PR #9, opened 2021-07-11
cbuild stage0: need also xz and patch

yopito:
if not using `bootstrap.sh`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ to start with. Also, you will need the following tools:
 * `flex`
 * `perl`
 * `m4`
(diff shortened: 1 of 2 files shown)


q66:
what do you need xz for?

yopito:
xz is needed to extract kernel headers tarball

q66:
i don't see why it would be, during bootstrap the distfiles are all extracted with python's builtin functionality

yopito:
you're right. xz is uneeded anymore. Has changed since last week ;)

PR #22, opened 2021-08-03
cbuild: package dependency must be calculated for target arch

### LINKED ISSUE #21: crossbuild: can't find a requirement if not in base-chroot
while building for a crossbuild arch (let's say aarch64), shlibs fails to find requirement of the being built package: It does not find a lib provided by a package that is installed for **target** arch.

this is because apk is launched against host `paths.masterdir()` and not against the target's equivalent path here: https://github.com/chimera-linux/cports/blob/d9449a27ee256261624beb78574fe8fb796aece0/cbuild/hooks/pre_pkg/04_generate_runtime_deps.py#L62

practical example:
* package nvi rely on libdb-18.1.so provided by a ~~db~~ libdb package
* crossbuild package nvi for aarch64
* errror: `SONAME: libdb-18.1.so <-> UNKNOWN PACKAGE!`
* wrong call `apkg info --root /build/chimera/masterdir --allow-untrusted --installed so:libdb-18.1.so`
* right call should be `apkg info --root /build/chimera/masterdir/usr/aarch64-unknown-linux-musl  --allow-untrusted --installed so:libdb-18.1.so`

Edit: package providing .so lib is named libdb

yopito:
Closes #21

### DIFF
--- a/cbuild/hooks/pre_pkg/04_generate_runtime_deps.py
+++ b/cbuild/hooks/pre_pkg/04_generate_runtime_deps.py
@@ -58,8 +58,10 @@ def invoke(pkg):
             continue
         # otherwise, check if it came from an installed dependency
         if not pkg.bootstrapping or not (dep in bootstrap_map):
+            # dependency must be calculated for target arch
             info = subprocess.run([
-                "apk", "info", "--root", paths.masterdir(),
+                "apk", "info",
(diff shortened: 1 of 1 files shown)


yopito:
need some rework: fail with subpackage:
```
AttributeError: 'Subpackage' object has no attribute 'build_profile'
```

PR #26, opened 2021-08-29
bootstrap.sh: fully support custom BOOTSTRAP_ROOT

yopito:


q66:
why?

yopito:
This helps to script bootstrap tests, since a not-random BOOTSTRAP_ROOT can be controlled, instead of creating a new one each time.
Since BOOTSTRAP_ROOT already can be reused (once created), this is just a small enhancement.

q66:
it's not meant to be reused, the sole purpose of the variable is for testing (so that failed bootstraps can be restarted)

yopito:
"reuse" is probably not the right term, sorry. I mean be able to fully choose BOOTSTRAP_ROOT, it's more convinient sometimes (especially on boostrap-fail-bootstrap-fail-... sequences)

Issue #27, opened 2021-08-29
main/ca-certificates: build fails on stage0


yopito:
`main/ca-certificates` is building fine on on regular workflow (with binpkgs from stage2) but currently fails on bootstraping stage0 (with rev e31ae19).

~~looks like that `post_patch` phase in this context is not within a bubblewrap chroot, so clang fails (see `-isystem` argument):~~

bootstraping stage0 output:
```
$ rm -rf masterdir-stage0/
$ BOOTSTRAP_ROOT=/build/chimera/SCRATCH_cports.git/bootstrap.ROOTDIR ./bootstrap.sh
[...]
=> cbuild: bootstrapping stage 0
=> base-chroot-0.66-r0: building for x86_64...
   [runtime] musl-devel>=0: found (1.2.2-r0)
   [runtime] base-files>=0: found (0.142-r0)
   [runtime] elftoolchain>=0: found (0.7.1_svn20210623-r0)
   [runtime] llvm>=0: found (12.0.0-r0)
   [runtime] clang>=0: found (12.0.0-r0)
   [runtime] lld>=0: found (12.0.0-r0)
   [runtime] bmake>=0: found (20210420-r0)
   [runtime] bsdutils>=0: found (0.0.1-r0)
   [runtime] dash>=0: found (0.5.11.3-r0)
   [runtime] file>=0: found (5.40-r0)
   [runtime] apk-tools>=0: not found
   [runtime] awk>=0: not found
   [runtime] ncurses>=0: found (6.2-r0)
   [runtime] bsdgrep>=0: not found
   [runtime] bsdgzip>=0: not found
   [runtime] bsdpatch>=0: not found
   [runtime] bsdsed>=0: not found
   [runtime] bsdtar>=0: found (3.5.1-r0)
   [runtime] bsddiff>=0: not found
   [runtime] chroot-util-linux>=0: not found
=> apk-tools-2.12.5-r0: building [meson] (dependency of base-chroot) for x86_64...
   [target] zlib-devel: found (1.2.11-r0)
   [target] openssl-devel: found (1.1.1k-r0)
   [runtime] ca-certificates>=0: not found
   [runtime] apk-tools=2.12.5-r0: subpackage (ignored)
=> ca-certificates-20210119-r0: building (dependency of apk-tools) for x86_64...
   [host] openssl: found (1.1.1k-r0)
   [runtime] openssl<=2.0: found (1.1.1k-r0)
[...]

PR #28, opened 2021-08-29
main/ca-certificates: fix generation on stage0

yopito:
Closes: #27

is fine on regular packaging too.


q66:
I don't see by what mechanism this is supposed to fix the problem?

q66:
yeah, the template itself does not need fixing

PR #58, opened 2021-11-25
[RFC] main/dinit: crossbuild support, enhance checks.

yopito:
Details:
* I was able to crossbuild `dinit` from `x86_64` to all other architectures with this fix.
* building flags are the same than before for `x86_64`, package content is the same.
* not tested on runtime
* target arch package depends on `so:libunwind.so.1`, that is not the case for native one ?
* since init system, also build it statically ?


q66:
i don't see any particular reason for the init system to be static

PR #64, opened 2021-12-06
contrib/psmisc: new package (23.4)

RoastVeg:
- Tests disabled: `testsuite` subdirectory patched out because of a dependency on dejagnu
- Tested working on x86_64
- Build-tested working on aarch64, ppc64, ppc64le, and riscv64.


q66:
is there any reason to patch out the testsuite dir if `!check` is used?

also this is more of a thing for `contrib/`

q66:
(you can also try packaging `dejagnu` and have the tests run instead)

RoastVeg:
> is there any reason to patch out the testsuite dir if `!check` is used?

Yes, there is an immediate error under bmake with the call to "export". If patching the makefile (or switching to gmake, less nice) in that directory, and packaging dejagnu is a requirement to get this package merged then I can take a look.

> also this is more of a thing for `contrib/` (i would like to keep `main/` more or less like "one of each curated thing", and `psmisc` does not provide anything strictly important, the primary tools are all in `procps-ng`)

I trust your judgement and am happy to move it to contrib, though it is perhaps worth considering that several of these commands exist in busybox, which I personally would consider an "expanded coreutils", and might be expected by users familiar with distros that provide them in main. I'd broadly consider bsdutils(-extra), util-linux, procps-ng, psmisc etc all part of a greater whole.

q66:
that's a part of the point, i'd like people to try and move to better/more consistent alternatives to the `psmisc` commands (e.g. `pkill` instead of `killall`)

there is no particular reason to have this in `main`, things in `contrib` are treated more or less the same, but it provides sort of a logical distinction for what is actually endorsed by the distribution

packaging `dejagnu` would be nice because there was another thing in `main` that needed it to run tests, and the goal is to run tests where possible, so it's a two birds with one stone situation

RoastVeg:
> that's a part of the point, i'd like people to try and move to better/more consistent alternatives to the `psmisc` commands (e.g. `pkill` instead of `killall`)

I fear you may be fighting a losing battle, but I will not attempt to persuade you.

> there is no particular reason to have this in `main`, things in `contrib` are treated more or less the same, but it provides sort of a logical distinction for what is actually endorsed by the distribution

As a reflection of your personal endorsement (or lack thereof), I will move it to the appropriate folder.

> packaging `dejagnu` would be nice because there was another thing in `main` that needed it to run tests, and the goal is to run tests where possible, so it's a two birds with one stone situation

I'll package it, and enable tests in those packages that can be.

RoastVeg:
Thanks for merging dejagnu, you beat me to it. Unfortunately the tests fail because of https://gitlab.com/psmisc/psmisc/-/issues/18

I can do one of two things:
- leave the package as-is (`["!check"]`)

q66:
usually where most tests pass but some fail and they fail for harmless reasons, disabling the specific tests is preferrable (makes it easy to track too)

RoastVeg:
Agreed in principle. Two things to consider:
- [In this case](https://gitlab.com/psmisc/psmisc/-/blob/master/testsuite/killall.test/killall.exp) some, but not all, tests in the loop over all signal names are failing. I can comment out the loop altogether.
- The test is actually picking up a bug in usability on musl. It might be worth fixing the issue with a gnu-style getopt, which would also fix other packages which don't use getopt correctly.

q66:
if you want to package gnu-getopt, i see no harm in it (feel free to put it in main/). Package name should be `gnu-getopt`, it should be compiled with `!lto` in `options`.

RoastVeg:
This should be good to go now

q66:
1) squash your spurious merge commits
2) you are still patching psmisc tests
3) you don't really need need the new psmisc patch, just `-Dgetopt_long_only=gnu_getopt_long_only` in `tool_flags["CFLAGS"]` should be fine
4) don't build any shared library for `gnu-getopt`
5) use the makefile for `gnu-getopt` build (even if you didn't, this build process is incorrect and does not generate proper libraries, and also there is `cbuild.util.compiler`)

PR #65, opened 2021-12-07
main/os-prober: handle os-probes/mounted/{x86,arm64}/efi/ subdir

yopito:


yopito:
generated package on x86_64 has the same layout than VoidLinux grub package.

PR #80, opened 2022-01-10
main/libpulse: default_progs() already handles usr/share/ content.

yopito:
side effect: new automatic subpackage `libpulse-progs-bashcomp`.
zsh completion stuff remain in libpulse-progs package.


q66:
not really a side effect, it's fully intended

PR #102, opened 2022-08-08
[RFC] go lang support as a build sytle

yopito:
golang support as a build style, and some packaging to illustrate its usage.
apart from various WIP stuff in packages, I'll be glad with some informal review of this build style.

### DIFF
--- a/src/cbuild/util/golang.py
+++ b/src/cbuild/util/golang.py
@@ -1,0 +1,121 @@
+from pathlib import Path
+
+def get_go_env(pkg):
+    if not pkg.profile().goarch:
+        pkg.error("Unknown architecture for golang")
+
+    # * GOBIN is not suitable for crossbuild:
+    #   "go: cannot install cross-compiled binaries when GOBIN is set"
+    # * GOPATH must NOT contain a go.mod file, so cant' be '.'
+    env = {
+        "GOMODCACHE" : "/cbuild_cache/golang/pkg/mod",
+        "GOARCH"     : pkg.profile().goarch,
+        "GOPATH"     : f"{pkg.chroot_cwd}/{pkg.make_dir}"
+    }
+    return env
+
+class Golang:
+    def __init__(self, tmpl, jobs = None, env = {}, wrksrc = None):
+        self.template = tmpl
+        self.wrksrc = wrksrc
+        self.env = env
+        self.jobs = jobs
+
+    def _invoke(
+        self, command = None, args = [], jobs = None, offline = True,
+        base_env = {}, env = {}, wrksrc = None
+    ):
+
+        if not command:
+            self.template.error("Missing go command argument")
+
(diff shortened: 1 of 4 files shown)


q66:
thanks, i don't think we need this right now but once things are a bit more stable we can work out some review

q66:
i wasn't happy with the whole gcompat nonsense so i experimented with generating toolchains and at that point i was far enough to have a working template, so that's in the repo now, suggest dropping the toolchain from here and also drop gcompat because it makes no sense for chimera (any binaries you'd want to run with that would be linked with libgcc, and we don't have that and there is a good chance we never will)

we can try getting in the build style eventually (it looks like it needs a lot of work though)

the random software should be standalone pull requests

yopito:
OK, i'll rework this PR, only with the go build style.

src/cbuild/util/golang.py
```diff
@@ -0,0 +1,123 @@
...
+        if ldflags:
+            myargs += ["-ldflags", f"{(' ').join(ldflags)}"]
+        if self.template.make_build_args:
+            myargs += self.template.make_build_args
+        myargs += args
+
+        return self._invoke("install", myargs, jobs, True, None, env, wrksrc)
+
+    def check(self):
+        myargs = ["-v"]
```
vazub, line 102: I would rather suggest to be silent by default, not verbose.

nekopsykose: iirc go test verbose doesn't even have anything useful except being loud- failures still print the failure and source line, so -v isn't useful, just spammy

vazub:
I've tried the style and it works rather well overall. One thing that requires further investigation is that the app I've built did get debug symbols stripped, but a separate `-dbg` subpackage was not created.

src/cbuild/util/golang.py
yopito: Thanks for your inputs. OK, I will remove it

src/cbuild/util/golang.py
```diff
@@ -0,0 +1,123 @@
...
+    return env
+
+class Golang:
+    def __init__(self, tmpl, jobs = None, env = {}, wrksrc = None):
+        self.template = tmpl
+        self.wrksrc = wrksrc
+        self.env = env
+        self.jobs = jobs
+        # -dbg packages fail to be generated
+        tmpl.options["debug"] = False
```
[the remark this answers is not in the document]
vazub: Removed the line and tested locally - works well, -dbg gets created. So just removing the line here helps.

vazub (changes requested):

vazub (approved):
Should be fine to merge, unless @q66 has any additional concerns/requirements.

q66:
i did a preliminary merge of this after some cleanup/adjustments, but without the documentation; only stable interfaces can be documented and i plan to not make this stable for a while (just like cargo)

PR #125, opened 2022-10-07
contrib/neovim: new package (0.9.0)

yopito:
build and usage fine on ` x86_64`, even within `cbuild chroot` environnement.  
crossbuild need some investigations.


q66:
something i noticed in all your stuff, you're separating things like options from the main block while you shouldn't

yopito:
OK, i'll fix that

q66:
also there's a lot of pointless comments scattered all over the place and we should probably build neovim against standard lua because luajit is a mess

yopito:
> also there's a lot of pointless comments scattered all over the place and we should probably build neovim against standard lua because luajit is a mess

I missed that neovim can live without luajit, have to review and test.

yopito:
reworked to remove luaJIT usage. ready for review

yopito:
ah, bad commit message. quick git push within a minute, sorry

q66:
can you please squash the update commits, update where appropriate and so on? basically do whatever you may think is necessary for merging, so i can review it

yopito:
here it is. thanks for rewieving

q66:
alright thanks, i'll probably merge it by pieces (i'll do adjustments to the templates as i think is necessary, probably faster than having you iterate on it; and it's a fair amount of commits)

PR #142, opened 2023-02-09
chimerautils: fix sha256 sum for commit a8b03cd4e0d9

bjornstromberg:
build from source fails on sha256 sum for chimera utils commit a8b03cd4e0d955f462da8c902e57884e17f2fe4f  downloaded 3 times  and got sha256 sum 6688d5acdba49622da7f42d7ea3c1c1c8fabab5dadcd20186530fbe302d995f3 three times. 

updating the template, builds fine.


JamiKettunen:
I think the commit message should just be `main/chimerautils: fix checksum`, otherwise +1 from me

q66:
i'm planning to tag a chimerautils release probably tomorrow anyway, so no need to fix this for now

bjornstromberg:
@JamiKettunen yea thought i would get to create a new squash commit message in the pull req.  

anyway since @q66 will tag a new release i agree its kind of useless to merge this.

PR #148, opened 2023-03-01
cbuild: show the source of a symlink when creating auto-deps

the-eater:
more of a proposal than "this is how it should be"

but I just spend hunting for a bit where a auto-dep came from, so display the path to the symlink itself


q66:
yup, this is fine, i have no problem with this (the hook messages generally could use improvements)

PR #150, opened 2023-03-02
[wip] add flatpak

the-eater:
still some policy issues needing to be fixed, and tests to be looked at

this mr depends !149


contrib/flatpak/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "flatpak"
+pkgver = "1.14.3"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--with-system-bubblewrap", "--with-system-dbus-proxy", "--with-system-helper-user=_flatpak",
```
JamiKettunen, line 5: So there's `--with-system-helper-user=_flatpak` and `/usr/lib/sysusers.d/flatpak.conf` installed (containing `u _flatpak - "FLatpak system helper" -`), however afaik that file is a systemd specific thing and isn't used which means you'd instead need to set `system_users = ["_flatpak"]`

JamiKettunen:
Btw there's also some missing deps from the packages, namely `xdg-desktop-portal` from `xdg-desktop-portal-gnome` etc which would be required for open file dialogs and such

q66:
what do you mean? i see them in the PR

JamiKettunen:
Missing `depends = ["xdg-desktop-portal"]` on `xdg-desktop-portal-gnome` for example, I had to install it manually after installing `xdg-desktop-portal-gnome`.

Additionally I wonder if `xdg-desktop-portal-gnome` could be `install_if`'d so it gets installed when `flatpak` and `gnome-shell` already are, or is that too much?

the-eater:
does xdg-desktop-portal-gnome actually depend on xdg-desktop-portal tho? I thought it was only like framework definitions and gir stuff but no actual moving parts?

JamiKettunen:
@the-eater you can try installing just `flatpak` and `xdg-desktop-portal-gnome` without installing/running `xdg-desktop-portal`, you'll notice the open file dialogs don't work in installed apps

q66:
i've added gpg, so please rebase them out; also i think that flatpak and portals are fine for `main` (this way we'll have an easier job setting the dependency relationships)

q66:
I merged the stuff after reworking some of it. Some observations for later:

* I had to redo formatting a lot, you should stick to 80 chars per line max, and instead of stuff like
```
foo = [ bla, bla, bla
        ble, ble ]
```
do
```
foo = [
    .... contents here....
]
```
* Use proper long SPDX license names instead of legacy (e.g. `LGPL-2.1-only` or `LGPL-2.1-or-later` instead of `LGPL-2.1`)
* The `gobject-introspection` is always a `hostmakedepends`, not `makedepends` (and `!cross` should be added)
* Usually things should not be building gtk-docs.
* I noticed a lot of double-empty-lines, usually there should be just one empty line

But otherwise it's mostly good, thanks.

the-eater:
cool, thanks!

yea about the lines, I use pep-8 which defines that there should be 2 empty lines around each function,  I'll update my IDE settings lol

PR #153, opened 2023-03-08
[pkgbulk] intel gpu stack (opencl, libva)

the-eater:


nekopsykose:
the libexecinfo from here is never going to actually work without also patching musl (to be useful, that is, not as purely a makedep) so you should patch the things using instead (usually just a few lines to remove backtrace() )

q66:
stale PR, feel free to reopen if you end up having any interest in making this mergeable (though ideally it should be multiple PRs)

PR #156, opened 2023-03-21
main/xwayland: allow glamor_composite_clipped_region to do overflows

the-eater:


the-eater:
Confirmed works, and builds, assembly doesn't contain any of the UBSan checks either

PR #160, opened 2023-03-27
contrib/unudhcpd: new package (0.2.1)

JamiKettunen:
There wasn't (yet) any other DHCP server in Chimera and I like having one for my USB networking -> SSH shell on phone use case :)
image


JamiKettunen:
Drafting until I check out which hardening options can be applied

PR #161, opened 2023-03-27
contrib/htop: new package (3.2.2)

JamiKettunen:
It's better than `top` :p
image

I considered instead packaging [`btop`](https://github.com/aristocratos/btop) but that had hundreds of various random compile errors which seemed specific to `libc++` so I didn't bother, who knows perhaps LLVM 16 will change the situation

### DIFF
--- a/contrib/htop/template.py
+++ b/contrib/htop/template.py
@@ -1,0 +1,12 @@
+pkgname = "htop"
+pkgver = "3.2.2"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--enable-unicode", "--enable-sensors"]
+makedepends = ["ncurses-devel", "libsensors-devel"]
+pkgdesc = "Interactive process viewer"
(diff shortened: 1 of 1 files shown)


JamiKettunen:
Drafting until I check out which hardening options can be applied

nekopsykose:
> who knows perhaps LLVM 16 will change the situation  

yes, btop builds with llvm16 toolchain fine (only one tiny makefile patch, next release has it fixed without patches)

q66:
i see no reason why it wouldn't build with 15 either

nekopsykose:
16 finalized ranges without experimental iirc, or maybe that was 15

nekopsykose:
ah, no, 16 fixed using ranges with libstdc++, so that's not relevant

JamiKettunen:
Seems  https://github.com/aristocratos/btop/issues/490 is relevant, I may come back to that later today and close this if I get it running, we'll see.

PR #164, opened 2023-04-11
contrib/minicom: new package (2.8)

JamiKettunen:
Seems to work for UART TX/RX with my FTDI FT232R USB adapter; `update-check` is broken but #163 addresses that.
image

### DIFF
--- a/contrib/minicom/template.py
+++ b/contrib/minicom/template.py
@@ -1,0 +1,12 @@
+pkgname = "minicom"
+pkgver = "2.8"
+pkgrel = 0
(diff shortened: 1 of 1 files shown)


JamiKettunen:
There's now a [`cu(1)`](https://man.freebsd.org/cgi/man.cgi?cu(1)) port in [`chimerautils`](https://github.com/chimera-linux/chimerautils/commits/ae67339) which should later come to `cports` too, thus I'm closing this since it seems to be enough for what I need. Anyone else feel free to pick this up :)

PR #165, opened 2023-04-18
Android tools + dependencies

JamiKettunen:
Brings in `adb`, `fastboot` and friends normally required for dealing with development and flashing of Android devices.

### DIFF
--- a/contrib/android-tools/template.py
+++ b/contrib/android-tools/template.py
@@ -1,0 +1,32 @@
+pkgname = "android-tools"
+pkgver = "34.0.4"
+pkgrel = 0
+build_style = "cmake"
+hostmakedepends = ["cmake", "ninja", "perl", "go", "pkgconf", "protobuf"]
+makedepends = [
+    "brotli-devel",
+    "lz4-devel",
+    "pcre2-devel",
+    "libusb-devel",
+    "zstd-devel",
+    "protobuf-devel",
+    "linux-headers",
+    "gtest-devel",
+]
+depends = ["python", "android-udev-rules"]
+pkgdesc = "Android platform tools (e.g. adb and fastboot)"
+maintainer = "Jami Kettunen <jami.kettunen@protonmail.com>"
+license = "Apache-2.0 AND ISC AND GPL-2.0-only AND MIT"
+url = "https://github.com/nmeum/android-tools"
+source = f"{url}/releases/download/{pkgver}/android-tools-{pkgver}.tar.xz"
+sha256 = "7a22ff9cea81ff4f38f560687858e8f8fb733624412597e3cc1ab0262f8da3a1"
+hardening = ["vis", "cfi"]
+# no tests
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_license("vendor/boringssl/LICENSE", name="boringssl.LICENSE")
+    self.install_license(
+        "vendor/boringssl/third_party/fiat/LICENSE", name="fiat.LICENSE"
+    )
--- a/contrib/android-udev-rules/template.py
+++ b/contrib/android-udev-rules/template.py
@@ -1,0 +1,17 @@
+pkgname = "android-udev-rules"
+pkgver = "20230614"
+pkgrel = 0
+pkgdesc = (
+    "Android udev rules list aimed to be the most comprehensive on the net"
+)
+maintainer = "Jami Kettunen <jami.kettunen@protonmail.com>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/M0Rf30/android-udev-rules"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "563b8f3194110b7173dfaad61e74da2842b28c4f2409dedce04d5572ef2f056f"
+options = ["!splitudev"]
+system_groups = ["adbusers"]
+
+
+def do_install(self):
+    self.install_file("51-android.rules", "usr/lib/udev/rules.d")


JamiKettunen:
Thoughts on having `android-udev-rules` as a dep of `android-tools` (like currently here), should it be done somehow differently? technically `adb` for example can work without the udev rules and you'd just `sudo adb start-server` or whatever

q66:
i see no harm in it (it does not break anything, does it), in general the simplest way is fine (especially if it's a small dependency, hard deps that are in practice optional are only questionable if they are big)

contrib/protobuf/template.py
```diff
@@ -0,0 +1,59 @@
...
+    return [
+        "usr/lib/libprotoc.so",
+        "usr/include/google/protobuf/compiler",
+    ]
+
+@subpackage("protobuf-devel")
+def _devel(self):
+    self.depends = [
+        f"libprotobuf~{pkgver}",
+        f"libprotobuf-lite~{pkgver}",
```
nekopsykose, line 56: technically this isn't needed, as the devel packages depend on the base libs already

contrib/protobuf/template.py
```diff
@@ -0,0 +1,59 @@
...
+    self.pkgdesc = f"Protocol buffers C++ library (lite version)"
+    return ["usr/lib/libprotobuf-lite.so.*"]
+
+@subpackage("libprotoc")
+def _libprotoc(self):
+    self.pkgdesc = f"Protocol buffers compiler library"
+    return ["usr/lib/libprotoc.so.*"]
+
+@subpackage("libprotoc-devel")
+def _libprotoc_devel(self):
```
nekopsykose, line 44: tbh, there's no real point to separately split -devel's, since it's more convenient to consume them all as a single `protobuf-devel`

q66:
ping

nekopsykose:
>  (new protobuf needs abseil-cpp)   

yeah, though 23.x works with the current 202301 version (iirc), so that should already work. 24 probably does need the new one tho

JamiKettunen:
`android-tools` still doesn't cross-build currently due to https://paste.c-net.org/4pldor7eeftm; on Void this was worked around by [removing `usr/lib/cmake` of  `protobuf`](https://github.com/void-linux/void-packages/blob/a62d54d/srcpkgs/protobuf/template#L32) but that on the other hand causes issues even for native arch `android-tools` build when linking similar to:
```
undefined symbol: absl::lts_20230802::log_internal::LogMessageFatal::LogMessageFatal(char const*, int, std::__1::basic_string_view<char, std::__1::char_traits<char>>)
```

q66:
do that, i wanted to package mosh but it needs protobuf, so i'd like to merge this soon

contrib/android-tools/template.py
```diff
@@ -0,0 +1,32 @@
...
+    "gtest-devel",
+]
+depends = ["python", "android-udev-rules"]
+pkgdesc = "Android platform tools (e.g. adb and fastboot)"
+maintainer = "Jami Kettunen <jami.kettunen@protonmail.com>"
+license = "Apache-2.0 AND ISC AND GPL-2.0-only AND MIT"
+url = "https://github.com/nmeum/android-tools"
+source = f"{url}/releases/download/{pkgver}/android-tools-{pkgver}.tar.xz"
+sha256 = "7a22ff9cea81ff4f38f560687858e8f8fb733624412597e3cc1ab0262f8da3a1"
+hardening = ["vis", "cfi"]
```
nekopsykose, line 23: i'd be a little surprised if this actually worked with cfi (not doubting you; just a cautious double check of all the tooling)

nekopsykose (approved):
looks good to me now ^_^

JamiKettunen:
Hm think I'll drop the `libprotoc` subpkg as it's 1.3M while the main `protobuf` one with only `protoc` alone is just ~6K, or any other opinions how it should be done?

contrib/android-tools/template.py
JamiKettunen: `adb` and `fastboot` for my typical usage at least didn't crash and burn, but they have so much functionality that I'm not sure it's even possible to properly test it all with a single device (as some have certain fastboot commands not implemented at all on bootloader level etc)

Maybe it's just better to forget about CFI to avoid potentially bricking devices during critical operations with a CFI-related crash even though so far it "appears to be fine"

nekopsykose: that sounds like a realistic enough test if you want to keep it on :)  

if you think it has the potential to brick devices and whatnot then it would be safer to disable it, yes. the hardening itself is probably of 'lesser' value because like most hardening approaches, it has most value for longrun services, things with privileges, ... . these commands are basically user tools, so it's always just a small nice to have i guess, when it's safe

main/abseil-cpp/template.py
```diff
@@ -1,14 +1,16 @@
...
 configure_args = [
     "-DCMAKE_CXX_STANDARD=17",
     "-DBUILD_SHARED_LIBS=ON",
     "-DABSL_PROPAGATE_CXX_STD=ON",
+    "-DABSL_USE_EXTERNAL_GOOGLETEST=ON",
+    "-DABSL_BUILD_TESTING=ON",
 ]
 hostmakedepends = ["cmake", "ninja", "pkgconf"]
-makedepends = ["linux-headers"]
+makedepends = ["linux-headers", "gtest-devel"]
```
nekopsykose, line 13: the added absl test libraries from build_testing link to gtest which means gtest gets added as a runtime dependency even though nothing actually uses it after build-time check. would be nice to have a subpackage for those specific extra libs that link to it (they'll still be deps of -devel)

contrib/android-tools/template.py
q66: it's c++, it's most likely safe

q66:
i incorporated the abseil changes and protobuf separately, so rebase them out here

q66:
ok, this is good enough for now, let's try it

PR #166, opened 2023-04-20
Allow for console switching with AltGr in shortcuts

vazub:
Allow using AltGr in system shortcuts, primarily for console switching. Caters to situations when for any reason LAlt is unavailable or otherwise can not be used easily (modern DIY keyboards/keymaps etc.). Only extends aforementioned shortcuts, does not affect any other use cases for AltGr. Follows the lead of OpenBSD, that allows for similar feature.

### DIFF
--- a/main/kbd/patches/altgr-shortcuts.patch
+++ b/main/kbd/patches/altgr-shortcuts.patch
@@ -1,0 +1,23 @@
+diff --git a/data/keymaps/i386/include/linux-keys-bare.inc b/data/keymaps/i386/include/linux-keys-bare.inc
+index 64a4ee9..a2acf8c 100644
+--- a/data/keymaps/i386/include/linux-keys-bare.inc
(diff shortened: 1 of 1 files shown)


q66:
i'm still not quite sure if this is a good idea; the main thing that comes to mind is that this will only work while already on a tty, as within X/wayland environments binds are subject to xkb definitions instead, so say, somebody with a fancy keyboard will now have an inconsistency where they can switch VTs via altgr+Fx from a VT but not from a graphical environment

vazub:
Checked it inside Gnome - works fine with Ctrl+AltGr+F<n> - no clash of overlapping with any default shortcuts.
Other combinations did not actually work, so I trimmed the PR scope to this case only.

vazub:
My contribution interests shifted elsewhere for the time being. Retracting for now.

PR #168, opened 2023-04-23
contrib/limine: new package (4.20230422.0)

vazub:
Limine is a modern, advanced, portable, multiprotocol bootloader, also used as the reference implementation for the [Limine boot protocol](https://github.com/limine-bootloader/limine/blob/trunk/PROTOCOL.md).

### DIFF
--- a/contrib/limine/template.py
+++ b/contrib/limine/template.py
@@ -1,0 +1,24 @@
+pkgname = "limine"
+pkgver = "4.20230422.0"
+pkgrel = 0
+archs = ["x86_64", "aarch64"]
+build_style = "gnu_configure"
+configure_args = [
+    "--enable-all",
+    "TOOLCHAIN_FOR_TARGET=llvm"
+]
+make_cmd = "gmake"
+hostmakedepends = ["gmake", "mtools", "nasm"]
+pkgdesc = "Advanced, portable, multiprotocol bootloader"
+maintainer = "vazub <chimera@zubko.cc>"
+license = "BSD-2-Clause"
+url = "https://limine-bootloader.org"
+source = f"https://github.com/limine-bootloader/{pkgname}/releases/download/v{pkgver}/{pkgname}-{pkgver}.tar.xz"
+sha256 = "d42fe5488c221e2f69e1f7c920e5231841dd5ff65993b70104d7a6be00c93fcc"
+hardening = ["vis", "cfi", "sst"]
+# no test suite
+options = ["!check"]
+
+def post_install(self):
+    self.rm(self.destdir / f"usr/share/doc/{pkgname}/LICENSE")
+    self.install_license("LICENSE")


contrib/limine/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "limine"
+pkgver = "4.20230422.0"
+pkgrel = 0
+archs = ["x86_64", "aarch64", "!ppc*", "!riscv*"]
```
q66, line 4: should be just `["x86_64", "aarch64"]`

contrib/limine/template.py
```diff
@@ -0,0 +1,25 @@
...
+]
+make_cmd = "gmake"
+hostmakedepends = ["gmake", "mtools", "nasm"]
+pkgdesc = "Advanced, portable, multiprotocol bootloader"
+maintainer = "vazub <chimera@zubko.cc>"
+license = "BSD-2-Clause"
+url = "https://limine-bootloader.org"
+source = f"https://github.com/limine-bootloader/{pkgname}/releases/download/v{pkgver}/{pkgname}-{pkgver}.tar.xz"
+sha256 = "d42fe5488c221e2f69e1f7c920e5231841dd5ff65993b70104d7a6be00c93fcc"
+hardening = ["vis", "cfi", "sst", "cfi-genptr", "cfi-icall"]
```
q66, line 19: have you verified this all actually works? (i.e. tested setup and booting)

`cfi-icall` is a default so it's unnecessary, and `cfi-genptr` is mostly a workaround and if it's actually needed, it's a better idea to fix the problem in the code instead

contrib/limine/template.py
```diff
@@ -0,0 +1,25 @@
...
+# no test suite
+options = ["lto", "!check"]
```
q66, line 21: lto is default so drop the option

contrib/limine/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "limine"
+pkgver = "4.20230422.0"
+pkgrel = 0
+archs = ["x86_64", "aarch64", "!ppc*", "!riscv*"]
+build_style = "gnu_configure"
+configure_args = [
+    "--prefix=/usr",
```
q66, line 7: no need to specify this (`gnu_configure` does it by default)

q66 (changes requested):

contrib/limine/template.py
vazub: fixed

contrib/limine/template.py
vazub: fixed

contrib/limine/template.py
vazub: fixed

vazub:
My contribution interests shifted elsewhere for the time being. Retracting for now.

PR #169, opened 2023-04-23
contrib/libmd: new package (1.0.4)

vazub:
This library provides message digest functions found on BSD systems either on their libc (NetBSD, OpenBSD) or libmd (FreeBSD, DragonflyBSD, macOS, Solaris) libraries and lacking on others like GNU systems.

### DIFF
--- a/contrib/libmd/template.py
+++ b/contrib/libmd/template.py
@@ -1,0 +1,22 @@
+pkgname = "libmd"
+pkgver = "1.0.4"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--prefix=/usr"
+]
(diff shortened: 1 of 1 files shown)


q66:
i assume this is for libbsd? this is of no use on its own so i suggest you submit this with something else that will need it (but the same goes for libbsd itself, not quite sure what you need that for)

also stuff like the lto option (or i noticed a lot of spurious hardening fields in your limine PR) does not have to be specified, it's the default

vazub:
Bith libmd and libbsd are dependencies for wlroots, which, in turn, is the basis for the majority of Wayland compositors out there.

vazub:
Retracting for now.

PR #170, opened 2023-05-06
contrib/xfce4-dev-tools: new package (4.18.0)

Sachin-Bhat:
There are some dependencies that haven't been packaged yet namely `libtool`, `gtk-doc` and `intltool`.


q66:
stale + will not pass CI/does not follow rules/etc, closing for now

PR #172, opened 2023-05-14
contrib/budgie-desktop: new package (and all the dependencies it needs)

toukoAMG:
I'm going to create this as a draft because I haven't been able to build it as it keeps giving me a shim error after one of the packages start building, and this seems to be an issue for most contrib packages, but I think they all should work

AFAIK budgie doesn't support gnome-bluetooth 4, so I disabled it for now. Also I have no idea if the sound would work or not since budgie doesn't support pipewire yet, but it seems like they are working on it ([albeit with no progress since April 9th](https://github.com/BuddiesOfBudgie/budgie-desktop/commits/pipewire)), and I included libpulse as a dependency, which might work.

There are also libwnck, libgnomekbd, libxklavier since those weren't available, and were needed for budgie-screensaver and budgie-desktop.

I'd like to get around to adding budgie-desktop-view and budgie-control-center in the future as well.


q66:
i have no idea what "shim error" is supposed to be

does budgie still take ages to update every time a new mutter version releases? because unless they can guarantee that their support is ready by the time of a new gnome major release, this would block gnome updates in main, which is not really an option

toukoAMG:
> i have no idea what "shim error" is supposed to be
somehow it doesn't happen anymore, but now I found out that libgnomekbd doesn't build

as far as i can see, i'm pretty sure the recent mutter update didn't take that long as budgie 10.7.1 (update in which mutter-12 support was added) seems to have had support for it around the time mutter 44 was in beta

q66:
stale PR; feel free to reopen if you want to continue working on it

as is it's not in a reviewable shape - at least the following should be met:

1) one commit per new package, good commit messages, no fixup commits
2) templates formatted with `black` and passing `flake8`
3) the CI should pass

PR #175, opened 2023-05-21
contrib/qt6-* : enable check and some minor fixes

yopito:
some initial work on testing for qt6 packages, probably too verbose and not enough accurate ?
non-check changes on `contrib/qt6-qtbase` should be reorganized separately, ie mode dedicated commits ?

### DIFF
--- a/contrib/qt6-qtbase/template.py
+++ b/contrib/qt6-qtbase/template.py
@@ -255,7 +255,7 @@ def _devel(self):
     return self.default_devel(
         extra=[
             "usr/lib/qt6/metatypes",
-            "usr/lib/qt6/mkspecs/modules",
+            "usr/lib/qt6/mkspecs",
             "usr/lib/qt6/modules",
             "usr/lib/*.prl",
         ]
--- a/contrib/qt6-qtbase/template.py
+++ b/contrib/qt6-qtbase/template.py
@@ -71,7 +71,7 @@ debug_level = 1  # defatten, especially with LTO
 # FIXME
 hardening = ["!int"]
 # TODO
-options = ["!check", "!cross"]
+options = ["!cross"]
 
 if self.profile().arch == "riscv64":
     tool_flags = {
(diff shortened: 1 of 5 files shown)


JamiKettunen:
There's also at least this currently:
```
$ apk info -L qt6-qtbase
qt6-qtbase-6.5.0-r0 contains:
usr/lib/objects-None/moctestplugin_resources_1/.rcc/qrc_qmake_plugin_resource.cpp.o
```

nekopsykose:
the ones with `Could not launch app` you probably have to make a fake screen wrapper for (like xvfb-run does, but maybe someone made one for wayland)

should probably also put `CTEST_OUTPUT_ON_FAILURE` into the build_style because you always want that for ctest testing (and also perhaps `CTEST_PARALLEL_LEVEL`)

yopito:
> There's also at least this currently:
> 
> ```
> $ apk info -L qt6-qtbase
> qt6-qtbase-6.5.0-r0 contains:
> usr/lib/objects-None/moctestplugin_resources_1/.rcc/qrc_qmake_plugin_resource.cpp.o
> ```
you mean remove it too ?

> the ones with `Could not launch app` you probably have to make a fake screen wrapper for (like xvfb-run does, but maybe someone made one for wayland)

I don't think so: `QT_QPA_PLATFORM` (that is `offscreen`) is a fake screen wrapper.
Most of these kind of tests want something that is not yet installed (like `qmlscene`) ...
 
> should probably also put `CTEST_OUTPUT_ON_FAILURE` into the build_style because you always want that for ctest testing (and also perhaps `CTEST_PARALLEL_LEVEL`)

Yes, it would be convenient.

nekopsykose:
> I don't think so: QT_QPA_PLATFORM (that is offscreen) is a fake screen wrapper.
Most of these kind of tests want something that is not yet installed (like qmlscene) ...  

ah, good point. didn't know about offscreen :)

q66:
i will get this in manually with 6.5.1 update and reformatting

PR #185, opened 2023-06-10
Mesa 23.1.2

JamiKettunen:
Also only disable RadeonSI glthread for `gnome-shell` instead of globally as further tests revealed this was the component affected, see https://gitlab.freedesktop.org/mesa/mesa/-/issues/9141#note_1938665.


JamiKettunen:
(this was closed by 58f6f6dad249e5aa0bf5fedaaf838a779715a416)

PR #187, opened 2023-06-12
Add Prism Launcher (7.0) and dependencies

anellie:
This PR adds PrismLauncher, along with 2 build dependencies it needed (qt6-5compat, extra-cmake-modules).
This is my first PR, so I hope I got the templates right, especially since I wasn't sure of a good way to specify the Java installation for PL.
It builds and runs fine on x86; it should too on ARM and PPC but I was unable to test it due to not having any hardware available.

### DIFF
--- a/contrib/extra-cmake-modules/template.py
+++ b/contrib/extra-cmake-modules/template.py
@@ -1,0 +1,23 @@
+pkgname = "extra-cmake-modules"
+pkgver = "5.107.0"
+pkgrel = 0
+build_style = "cmake"
+hostmakedepends = ["cmake", "ninja"]
+pkgdesc = "Extra modules and scripts for CMake"
+maintainer = "aurelia <git@elia.garden>"
+license = "BSD-3-Clause"
+url = "https://github.com/KDE/extra-cmake-modules"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "b59d1bd4723be532ada21c1e584e2e61940e066f7ace0131173095431700bec3"
+# 10 out of 83 tests fail? It still seemed to run fine when building with it.
+#  2 - ExecuteKDEModules
+#  3 - KDEFetchTranslations
+#  4 - KDEInstallDirsTest.vars_in_sync_no_args
+#  5 - KDEInstallDirsTest.not_cache_variable
+#  6 - KDEInstallDirsTest.vars_in_sync_kde_arg
+#  7 - KDEInstallDirsTest.vars_in_sync_cmake_arg
+# 16 - ecm_add_tests-single_tests
+# 17 - ecm_add_tests_did_run-single_tests
+# 18 - ecm_add_tests-multi_tests
+# 19 - ecm_add_tests_did_run-multi_tests
+options = ["!check"]
--- a/contrib/prismlauncher/template.py
+++ b/contrib/prismlauncher/template.py
@@ -1,0 +1,24 @@
+pkgname = "prismlauncher"
+pkgver = "7.0"
+pkgrel = 0
+build_style = "cmake"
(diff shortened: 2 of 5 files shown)


q66:
make sure to run your files through `black` and `flake8` ideally beforehand, right now it's not so important as the CI is not loaded significantly, but once there is more stuff it may result in unnecessary stalls

anellie:
Yeah, will do next time! I admittedly just forgot with these, will fix the formatting and redo the commits.

anellie:
Should be good to go now! `flake8 contrib` reports no errors. Sorry for the trouble and thanks for the review!

contrib/extra-cmake-modules/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "extra-cmake-modules"
+pkgver = "5.107.0"
+pkgrel = 0
+build_style = "cmake"
+hostmakedepends = ["cmake", "ninja"]
+makedepends = []
```
q66, line 6: you can remove this one

contrib/extra-cmake-modules/template.py
```diff
@@ -0,0 +1,25 @@
...
+pkgdesc = "Extra modules and scripts for CMake"
+maintainer = "aurelia <git@elia.garden>"
+license = "BSD-3-Clause"
+url = "https://github.com/KDE/extra-cmake-modules"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "b59d1bd4723be532ada21c1e584e2e61940e066f7ace0131173095431700bec3"
+
```
q66, line 13: no need for the empty line here

contrib/prismlauncher/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "prismlauncher"
+pkgver = "7.0"
+pkgrel = 0
+build_style = "cmake"
+# TODO is this really the ideal way to do this?
+configure_env = {"JAVA_HOME": "/usr/lib/jvm/java-17-openjdk"}
```
q66, line 6: i'll think about this one a bit more if you don't mind

contrib/qt6-5compat/template.py
```diff
@@ -0,0 +1,30 @@
...
+makedepends = ["qt6-qtbase-devel"]
+pkgdesc = "Module containing unsuppored Qt5 APIs"
+maintainer = "aurelia <git@elia.garden>"
+license = (
+    "LGPL-2.1-only AND LGPL-3.0-only AND GPL-3.0-only WITH Qt-GPL-exception-1.0"
+)
+url = "https://www.qt.io"
+source = f"https://download.qt.io/official_releases/qt/{pkgver[:-2]}/{pkgver}/submodules/qt5compat-everywhere-src-{pkgver}.tar.xz"
+sha256 = "a9e2f53a193fc2e131b01a2f6e7a1fbfe31309c2413fdc213e5a81c558c21261"
+
```
q66, line 15: ditto empty line shouldn't be here

q66 (changes requested):
looks good otherwise, just please wait a bit until i address the _JAVA_HOME stuff :)

anellie:
Redid the commits with the requested changes and removed the `JAVA_HOME`, which `java-common`  should handle now. Have not built it yet due to r1 of the JDK not being available as a binary package yet

anellie:
Weird, it still behaves the same as it did without specifying `JAVA_HOME`... `java_common` should have taken care of doing so, correct?

q66:
profile.d is for login shells, package builds don't spawn a login shell

q66:
well, just do the configure_env, there is not much point in setting up anything more elaborate

q66:
qt6-5compat should be called qt6-qt5compat to follow the existing local and upstream convention, otherwise lgtm

anellie:
Renamed it now

PR #189, opened 2023-06-12
Add zoxide (0.9.1)

anellie:
Adds a package for zoxide.
Since zoxide already had a `.cargo/config.toml` in the project dir, the code for configuring vendoring sources threw an error.
I modified it to append to the file instead, which felt like the right approach and fixes this issue.

### DIFF
--- a/contrib/zoxide/template.py
+++ b/contrib/zoxide/template.py
@@ -1,0 +1,29 @@
+pkgname = "zoxide"
+pkgver = "0.9.1"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo"]
+makedepends = ["rust"]
+pkgdesc = "Fuzzy cd command for interactive shells"
+maintainer = "aurelia <git@elia.garden>"
+license = "MIT"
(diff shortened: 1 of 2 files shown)


contrib/zoxide/template.py
```diff
@@ -0,0 +1,31 @@
...
+    self.install_file(
+        "contrib/completions/zoxide.bash",
+        "usr/share/bash-completion/completions",
+        name="zoxide",
+    )
+    self.install_file(
+        "contrib/completions/_zoxide",
+        "usr/share/zsh/site-functions",
+    )
+    # TODO this is not split off into a completions package, is that okay?
```
q66, line 26: yeah i've been thinking about this for other packages already, i might add `-fishcomp` splitting before getting back to you here

JamiKettunen: As of an hour ago `-fishcomp` splitting was added with 0744645a09d3b4aff93fda6665db9e248f8ef993

anellie:
Removed the TODO, should be good to go now

PR #190, opened 2023-06-14
contrib/bat: new package (0.23.0)

anellie:


q66:
this should be built with system curl, libgit2 and libssh2, the latter two need adding

q66:
note that both of these libraries should go in main/

anellie:
Added `http-parser` and `libgit2`, other 2 were already present. It seems like `bat` is refusing to use the system versions though; they don't show in `ldd`. Adding `LIBGIT2_NO_VENDOR` didn't help either

nekopsykose:
(i didn't check too deeply why that happens (same thing goes for zstd and most sys crates!) but i assume it's something to do with bindgen checking if the vendored crate headers match up to some system thing)

anellie:
Should be good now, `ldd` lists all the libraries it should I think. Thanks for the tip @nekopsykose, gonna keep that in mind for future packages

contrib/bat/template.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def post_patch(self):
+    from cbuild.util import cargo
+
+    self.cargo.vendor()
+    cargo.setup_vendor(self)
+
+
+def post_install(self):
```
q66, line 33: please make use of the new install_completion helper

contrib/bat/template.py
```diff
@@ -0,0 +1,39 @@
...
+    self.install_man(next(self.find("target/", "bat.1")))
+    self.install_license("LICENSE-APACHE")
```
q66, line 35: LICENSE-APACHE should not be installed

q66:
outside of the little license nit, i don't see anything bad anymore

PR #192, opened 2023-06-15
Fix bluetoothd service file

wezm:
There was an extra 'o' in the directory name preventing it from starting.


q66:
increment pkgrel

wezm:
Ahh right, of course. pkgrel bumped.

q66:
ah, also, follow correct commit message rules next time (missed it this time)

PR #193, opened 2023-06-15
contrib/swaylock: new package (1.7.2)

paulapatience:


q66:
swaylock is a part of https://github.com/chimera-linux/cports/pull/194, since that PR is more complete, i'll take that one if you don't mind

PR #194, opened 2023-06-15
Sway, wlroots, and sway tools

flukeeey:
I've packaged wlroots (needed by sway), sway, swaybg, swayidle, swaylock. Together, these provide a functional sway desktop experience.

I've enabled all features where possible, including:

- Vulkan, OpenGL renderers
- libinput, DRM backends
- Xwayland support
- Extra image file format support for swaybg
- PAM support for swaylock
- Systray support for sway

Vulkan doesn't work too well on my machine but its there anyways.

This is my first PR to Chimera so be gentle. :p


q66:
i'll let the CI run first, then i'll do the manual review

contrib/sway/template.py
```diff
@@ -0,0 +1,44 @@
...
+    "wlroots-devel",
+]
+pkgdesc = "i3-compatible Wayland compositor"
+maintainer = "flukey <flukey@vapourmail.eu>"
+license = "MIT"
+url = "https://swaywm.org"
+source = f"https://github.com/swaywm/sway/releases/download/{pkgver}/sway-{pkgver}.tar.gz"
+sha256 = "73f08fd2cf7948e8af900709efe44eae412ae11c5773960e25c9aa09f73bad41"
+hardening = ["vis", "cfi"]
+options = ["!lint"]  # disable checking of pkgdesc
```
q66, line 30: don't disable lints, do a compliant pkgdesc instead (e.g. `Wayland compositor compatible with i3`)

contrib/sway/template.py
```diff
@@ -0,0 +1,44 @@
+pkgname = "sway"
+pkgver = "1.8.1"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Dwerror=false"]
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+    "ninja",
```
q66, line 9: ninja doesn't need explicit hostdep (it's always installed with meson); ditto everywhere else

q66 (changes requested):
also the maintainer fields commit should be split up between the individual commits

PR #196, opened 2023-06-16
contrib/tig: new package (2.5.8)

wezm:


contrib/tig/template.py
```diff
@@ -0,0 +1,33 @@
...
+hostmakedepends = ["gmake", "asciidoc", "xmlto", "pkgconf"]
+makedepends = ["ncurses-devel"]
+depends = ["git"]
+pkgdesc = "Text-mode interface for git"
+maintainer = "Wesley Moore <wes@wezm.net>"
+license = "GPL-2.0-or-later"
+url = "https://github.com/jonas/tig"
+source = f"{url}/releases/download/{pkgname}-{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "b70e0a42aed74a4a3990ccfe35262305917175e3164330c0889bd70580406391"
+
```
q66, line 18: drop extra whitespace

contrib/tig/template.py
```diff
@@ -0,0 +1,33 @@
...
+# test suite tries to access /dev/tty which fails
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_file(
```
q66, line 24: use install_completion

contrib/tig/template.py
```diff
@@ -0,0 +1,33 @@
+pkgname = "tig"
+pkgver = "2.5.8"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_gen = []
```
q66, line 5: any reason the autotools cannot be regenerated?

q66 (changes requested):

contrib/tig/template.py
wezm: No reason, configure was already present so was making use of it but if the done thing is to regenerate it then I'm happy to make that change.

wezm:
I addressed the review comments

PR #197, opened 2023-06-17
contrib/fd: new package (8.7.0)

wezm:
Template for `fd`. I have disabled the default cargo features:

- `use-jemalloc`: fails to build
- `completions`: this is unused

Only zsh completions are included as generating bash and fish completions requires invoking the built binary, which I assume will not work when cross-compiling.

I was a little unsure of the best way to pass `--no-default-features` to cargo. Setting `make_*_args` works for build and check but setting `make_install_args` overwrites the default behavior (of passing `--path .`), so I instead invoke `install` manually to preserve that default behavior.

### DIFF
--- a/contrib/fd/template.py
+++ b/contrib/fd/template.py
@@ -1,0 +1,22 @@
+pkgname = "fd"
+pkgver = "8.7.0"
+pkgrel = 0
+build_style = "cargo"
+# disable the default use-jemalloc and completions features
+make_build_args = ["--no-default-features"]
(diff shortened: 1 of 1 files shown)


q66:
I altered the `make_install_args` behavior (overriding it probably doesn't make much sense), you should be able to just use `make_install_args` now

wezm:
Thanks for tweaking `make_install_args` I've updated it to use that now.

PR #199, opened 2023-06-17
Foot Wayland terminal emulator & its deps

flukeeey:
I use foot as my terminal emulator on wayland, here are my package templates.

I have enabled:
- Grapheme clustering
- Runtime shaping
- SVG support (bundled nanosvg)

Missing:
- PGO - TODO, but having trouble with
- Utempter support - missing libutempter package, not sure if this is needed tbh? Do other terminals enable this?
- dinit user service for the foot daemon - this has pros and cons, so I choose not to enable it personally, lmk if needed

Optional runtime deps (not sure how to specify in template):
- libnotify: desktop notifications by default uses notify-send
- xdg-utils: URLs are by default launched with xdg-open
- bash-completion: If you want completion for positional arguments


q66:
install licenses properly (in this case they are seemingly already being installed so move them to proper location instead)

flukeeey:
Thanks for the feedback!

I've made the changes as requested, but a few things:

1. The `tllist-1.1.0` package is empty, it's just a header and .pc file in `tllist-devel-1.1.0`, is it okay that the main package is empty like this?
2. I moved the license files using mkdir and mv, hope this is ok, there doesn't seem to be a cleaner or more idiomatic way to do this? I know `install_license` exists for copying from build dir to destdir, but not for moving...
3. xdg-open and libnotify are optional dependencies, am I reading right that you want me to make them hard dependencies instead by adding them to `depends`? I don't want/need these on my machine, so I'd rather keep them as optional as upstream recommends

q66:
`mv` is fine for moving, for creating the directory use `install_dir`; for `tllist`, just make it one package

also remove the PGO TODO comment, we're not going to do PGO for distribution packages, it makes no sense

flukeeey:
Done!

tllist header and .pc files are now in a single `tllist` package.

I also made `foot-themes` depend on `foot`, as it doesn't make sense to have the themes installed for a non-existent app (?).

PR #200, opened 2023-06-18
go build style: documentation

yopito:
following PR#102: documentation for the go build style.
As a reminder.

quoting https://github.com/chimera-linux/cports/pull/102#issuecomment-1556053524 :
> [...] without the documentation; only stable interfaces can be documented and i plan to not make this stable for a while (just like cargo)

### DIFF
--- a/Packaging.md
+++ b/Packaging.md
@@ -1196,6 +1196,49 @@ contents of the wheel. The `do_check` will run `pytest` or fail.
 
 The `make_install_target` is used as a glob pattern to match built wheels.
 
+#### go
+
+A build style for packages written in Go language (golang). Requires to have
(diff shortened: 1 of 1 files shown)


yoctozepto:
Could we consider `go` build style _stable_ now? @q66 @nekopsykose

I can update the docs to match the current state of go affairs.

nekopsykose:
it hasn't changed in forever so in that sense sure, though i have no opinion myself

q66:
i think it's still p sketch but i haven't gotten around to rewriting it yet

yoctozepto:
> i think it's still p sketch but i haven't gotten around to rewriting it yet

Could you elaborate?

PR #201, opened 2023-06-18
main/imagemagick: set keeplibtool on libmagick

wezm:
The module `.la` files were being removed so none of the imagemagic commands were working.


q66:
drop the keeplibtool from the main package, as well as the post_install hook .la file removal

wezm:
done

PR #204, opened 2023-06-20
Add alacritty (0.12.1) and its dependencies

nbfritch:
This PR adds the alacritty terminal emulator and cargo-auditable, which it requires as a build dependency.
These patches have been verified to build and work on x86_64.

The build successfully completed on aarch64, but there is some problem rendering fonts when running.

### DIFF
--- a/contrib/alacritty/template.py
+++ b/contrib/alacritty/template.py
@@ -1,0 +1,54 @@
+pkgname = "alacritty"
+pkgver = "0.12.1"
+pkgrel = 0
+build_style = "cargo"
+make_install_args = ["--path", "alacritty"]
+hostmakedepends = [
+    "cargo",
+    "cargo-auditable",
+    "cmake",
+    "fontconfig-devel",
+    "freetype-devel",
+    "gmake",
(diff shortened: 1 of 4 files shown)


contrib/alacritty/template.py
```diff
@@ -0,0 +1,54 @@
...
+url = "https://github.com/alacritty/alacritty"
+source = (
+    f"https://github.com/alacritty/alacritty/archive/refs/tags/v{pkgver}.tar.gz"
+)
+sha256 = "14bce0bfc538872c97e0e38b9233a9d1fa992dcf83a22b6035da5fe58a55bc6c"
+
+
+def post_install(self):
+    self.install_license("LICENSE-MIT")
+    self.install_file(
```
q66, line 32: use install_completion

nekopsykose:
> and cargo-auditable, which it requires as a build dependency.  

what makes you think so? it's only used when `cargo auditable ..` is invoked, which nothing does here

q66:
PR non-responsive, superseded by https://github.com/chimera-linux/cports/pull/284

PR #206, opened 2023-06-21
contrib/podman: new package (4.5.1)

RoastVeg:
Lots of things still wrong with this:

- [ ] configuration for zfs/btrfs storage backends
- [ ] shell completions
- [ ] netavark
- [ ] podman tests failing
- [ ] yajl is unmaintained and likely needs CVE patches (at a minimum)


JamiKettunen:
After rebasing and trying to build this for myself I found:
- `conmon`, `crun`, `yajl` & `slirp4netns` are missing `>` at the end of the `maintainer` string.
- tests don't pass currently on `slirp4netns` so I added `options = ["!check"]` (along with a comment)
- licenses for `go-md2man` & `yajl` need to be installed
- [isn't `yajl` ISC-licenced](https://github.com/lloyd/yajl/blob/2.1.0/COPYING)?
- missing `,` (and newline) between `man1` and `man5` items for `podman-docker` subpkg

Now with it built and installed I shall test out https://clickable-ut.dev/ and report back if I see something odd

JamiKettunen:
Sure enough I'm seeing `Error: open /etc/containers/policy.json: no such file or directory` when trying to actually build an app with [`clickable`](https://pypi.org/project/clickable-ut/), even following the `Getting Started` steps with everything defaults (by only pressing enter on each question) reproduces this.

Another thing I remember from Void was the packaging of https://github.com/containers/storage to install the `containers.storage` configuration file so that the graph driver defaults to overlay, not vfs (which uses too much storage)

Anyway great work on this so far, looking forward to seeing `podman` fully functional down the road ^^

the-eater:
In my short testing exposing ports doesn't seem to work either, which might be related to the slirp4netns? or maybe lack of iptables

RoastVeg:
> In my short testing exposing ports doesn't seem to work either, which might be related to the slirp4netns? or maybe lack of iptables

Was this as a user service, or as root?

q66:
got any plans to do further work on this? a bunch of it was meanwhile packaged by psykose, but podman itself and some deps still remain

RoastVeg:
Yes I'll pick this back up again

contrib/podman/files/registries.conf
```diff
@@ -0,0 +1,79 @@
+# For more information on this configuration file, see containers-registries.conf(5).
```
nekopsykose, line 1: this should be part of a containers-common instead, like in https://github.com/void-linux/void-packages/blob/314341f80a7eeb430d4b4efe01e733d8873cf22c/srcpkgs/containers-common/template or https://git.alpinelinux.org/aports/tree/community/containers-common/APKBUILD , since there's a bunch more configs too (and so you don't need to manually specify this config at all)

contrib/protobuf/template.py
```diff
@@ -0,0 +1,26 @@
+pkgname = "protobuf"
```
nekopsykose, line 1: this isn't really split correctly, https://github.com/chimera-linux/cports/pull/165 does it completely instead

contrib/podman/files/registries.conf
nekopsykose: preferably more like the latter tho (also how archlinux does it) since you have to match up the versions of a few repos for all the files and check them against the podman version

RoastVeg:
I no longer have bandwitdth for this PR. Please feel free to continue it in a new PR.

PR #208, opened 2023-06-24
contrib/tree: new package (2.1.1)

flukeeey:
Another screwy Makefile, please check if there's a better way of handling this one. :)

### DIFF
--- a/contrib/tree/template.py
+++ b/contrib/tree/template.py
@@ -1,0 +1,22 @@
+pkgname = "tree"
+pkgver = "2.1.1"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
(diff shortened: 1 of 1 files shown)


contrib/tree/template.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def init_install(self):
+    self.make_install_args = [
+        f"DESTDIR={self.chroot_destdir / 'usr/bin'}",
+        f"MANDIR={self.chroot_destdir / 'usr/share/man'}",
+    ]
+
+
+def post_install(self):
```
q66, line 25: GPL should not be installed

flukeeey: Fixed.

PR #209, opened 2023-06-24
contrib/wireguard-tools: new package (1.0.20210914)

flukeeey:
Some notes:

- Had to use `tool_flags` to fix a Makefile problem, hope this is correct
- Wasn't sure whether to use /run or /var/run as RUNSTATEDIR, upstream defaults to /var/run
- My wg-quick dinit service forces the user to use /etc/wireguard/wg0.conf as their config, systemd allows template units, e.g. the user could specify `systemd start wg-quick@myconfig`, but I don't see a way of doing that with dinit?
- wg-quick script can use either iptables or nftables to set its rules, it prefers nftables, but I'm not sure how to specify iptables OR nftables as a dependency

### DIFF
--- a/contrib/wireguard-tools-wg-quick
+++ b/contrib/wireguard-tools-wg-quick
@@ -1,0 +1,1 @@
+wireguard-tools
--- a/contrib/wireguard-tools/files/wg-quick-all
+++ b/contrib/wireguard-tools/files/wg-quick-all
@@ -1,0 +1,7 @@
+# wg-quick service
+
+type               = scripted
+command            = /usr/libexec/wg-quick-all up
+stop-command       = /usr/libexec/wg-quick-all down
+before             = pre-network.target
+depends-on         = init-done.target
--- a/contrib/wireguard-tools/files/wg-quick-all.sh
+++ b/contrib/wireguard-tools/files/wg-quick-all.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+export PATH=/usr/bin
+set -e
+
+for cfg in /etc/wireguard/*.conf; do
+    [ -r "$cfg" ] || continue
+    wg-quick "$1" "${cfg}"
+done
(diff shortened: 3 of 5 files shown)


q66:
for the service, how about making a wrapper script that will bring up every `.conf` file in `/etc/wireguard`? basically something like this:

```
#!/bin/sh

set -e

for cfg in /etc/wireguard/*.conf; do
    [ -r "$cfg" ] || continue
    wg-quick "$1" "$cfg"
done
```

and then have the dinit service invoke something like

```
command = /usr/libexec/wg-quick-all up
stop-command = /usr/libexec/wg-quick-all down
```

that seems a lot more useful to me, but i don't use wireguard, so maybe i'm missing something? e.g. a mechanism to have several config files and only enable some, etc

q66:
for iptables vs nftables, i just wouldn't specify any firewall dependency and have things fail gracefully at runtime when none is present (make the user install whatever they like)

flukeeey:
That all sounds golden to me. I used your script but just stripped the '.conf' suffix for the wg-quick command (it doesn't want the .conf at the end). Changed /var/run to /run to keep it correct.

q66:
the manpage of wg-quick implies that the .conf should be kept

flukeeey:
Oops, you're right, when specifying an absolute path, the .conf is indeed required.

One other thing, wg-quick has a problem with `stat`:

> stat: unrecognized option: c
> usage: stat [-FLnq] [-f format | -l | -r | -s | -x] [-t timefmt] [file|handle ...]
> stat: unrecognized option: c
> usage: stat [-FLnq] [-f format | -l | -r | -s | -x] [-t timefmt] [file|handle ...]
> /usr/bin/wg-quick: line 47: ((: ( &  & 0007) == 0: syntax error: operand expected (error token is "&  & 0007) == 0")
> Warning: `/etc/wireguard/wg0.conf' is world accessible

Looking at the script, stat is only used to check if the config file is world-accessible or not, and it continues regardless. But maybe we should try to patch this?

q66:
stat is not a standardized command and the bsd and gnu variants have different syntax, so you'll have to patch it

flukeeey:
Turned out to be easy enough to patch. `wg-quick` looks good now. I tested it with a world-readable folder/config and it warned as necessary.

q66:
ping

q66:
stale PR, fails CI, feel free to reopen and fix (or somebody else)

PR #210, opened 2023-06-24
contrib/acpid: new package (2.0.34)

flukeeey:
A standalone ACPI event daemon.

Notes:

- All events are sent to handler.sh script; power button shuts down, lid close suspends
- acpid creates a socket (for other tools to listen to events it seems), but I haven't set an socket-* options in the dinit service file


q66:
I don't really like this handler, I'll think about it a bit first

q66:
just drop the default handler, acpid is used by various things on its own in scenarios where having it handle stuff is undesirable, and people who want theirs handled through it can configure it themselves without the distro making any promises about it (besides, i have a more elaborate power management framework in mind for later)

for the service, drop the `restart` (it's implicitly that way) and also make the dependency `init-done.target`, it's not worth it to start daemons like this so early, it doesn't get you anything and only means potential races

also make it `before = login.target`

flukeeey:
Sounds good. Should I move the current handler script and event file to an examples folder or something first? Or just comment out the `everything` file contents?

For someone using a more minimal WM/DE setup, it might be nice to give them something to get started with easily. It could require a lot of groking for someone to figure out how it works, which event IDs to use, write a shell script, etc.

q66:
just drop it

as i see it if somebody wants a setup where this is done manually they should be capable of using internet search (there are countless resources) and having a halfassed example script isn't of much benefit

flukeeey:
Done. Just FYI, having no `/etc/acpi` folder results in an error:

> starting up with netlink and the input layer
> opendir(/etc/acpi/events): No such file or directory

Not sure if we should create an empty folder in the package to allow acpid to at least launch?

q66:
acpid should be able to launch, the service should work, it should just not handle any events ootb

but otherwise it should be useful (as i said many things use it e.g. by listening on its socket for events, without using the handlers)

contrib/acpid/files/acpid
```diff
@@ -0,0 +1,6 @@
+type            = process
+command         = /usr/bin/acpid -f
+logfile         = /var/log/acpid.log
```
q66, line 3: the acpid manpage says it logs everything to syslog, is this needed?

nekopsykose: with -f it logs to stderr

```c
	/* open the log */
	if (foreground)
		log_to_stderr = 1;
	else
		open_log();
```

generally speaking though, it's nicer to not put it in syslog even if it wasn't stderr

PR #212, opened 2023-06-24
xcb-util-wm: fix empty xcb_ewmh.h by using gmake, add missing dependency

the-eater:
configure says it's missing xcb-macros

and also xcb_ewmh.h is currently empty because of what looks like some redirecting shenanigans bmake doesn't like


q66:
`xorg-util-macros` is `hostmakedepends`

PR #214, opened 2023-06-25
contrib/fdupes: new package (2.2.1)

autumnontape:
Part 2 of 3 of tools for squishing Btrfs filesystems, though this one can be used for other purposes, too. I tried the package, and it works for me.

The license text is part of the README... I checked what other packages in cports do in that situation, and it looks like they just install the whole README to /usr/share/licenses, so I did that too.

### DIFF
--- a/contrib/fdupes/template.py
+++ b/contrib/fdupes/template.py
@@ -1,0 +1,17 @@
+pkgname = "fdupes"
+pkgver = "2.2.1"
+pkgrel = 0
+build_style = "gnu_configure"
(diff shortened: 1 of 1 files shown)


autumnontape:
I'm wondering if pkgconf should be in hostmakedepends even though the configure script is able to find the libraries it needs (ncursesw and pcre2) without it. I'm not familiar with autotools, and I don't know if its non-pkg-config dependency discovery sucks in some situation like cross compiling that I haven't tried or what.

PR #218, opened 2023-06-26
contrib/spice-vdagent: new package (0.22.1)

JamiKettunen:
Tools to make interacting with Chimera Linux QEMU/KVM guests more integrated such as a shared bi-directional clipboard, drag & drop from host to guest etc.

GDM integration was dropped due to breaking launch of `spice-vdagent` user side daemon (via XDG autostart); should be looked into but imo nothing too major. A `spice-vdagent` user service may be added later after missing graphical support for user services is addressed.

Configuring `--with-session-info=systemd` allows for multi-user `spice-vdagent` to work at the cost of `elogind` dependency; without it (which is the state of this PR currently) only a single instance of the user side daemon can run / work with `spice-vdagentd` system service.

### DIFF
--- a/contrib/spice-vdagent/files/spice-vdagentd
+++ b/contrib/spice-vdagent/files/spice-vdagentd
@@ -1,0 +1,7 @@
+# spice-vdagentd service
+
+type               = process
+command            = /usr/bin/spice-vdagentd -x
+depends-on         = dbus
+restart            = true
+smooth-recovery    = true
--- a/contrib/spice-vdagent/template.py
+++ b/contrib/spice-vdagent/template.py
@@ -1,0 +1,36 @@
+pkgname = "spice-vdagent"
+pkgver = "0.22.1"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--with-session-info=none"]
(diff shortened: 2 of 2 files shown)


grigio:
> Marking as a draft due to not having found a proper way to launch spice-vdagent user side daemon yet

Did you find a way to fix that ? I think that issue is common with other sw like `syncthing` and other user side daemons

q66:
wë'll need the graphical support for user services done first before doing any user service stuff that relies on graphical things

JamiKettunen:
I've made kind of a big discovery, so it was the GDM integration all along which broke this which I realized after now making it a subpackage like:
```py
@subpackage("spice-vdagent-gdm")
def _gdm(self):
    self.pkgdesc = f"{pkgdesc} (GNOME display manager integration)"
    self.install_if = [f"{pkgname}={pkgver}-r{pkgrel}", "gdm"]

    return ["usr/share/gdm"]
```
I'll probably just drop that entirely for now along with the files until it's figured out; not that copy/paste in gdm is that useful anyway

grigio:
> apk add '!spice-vdagent-gdm'

I tried it, but it seems `spice-vdagent` do not work, are other steps required?

PR #222, opened 2023-07-17
main/base-cbuild-host: add python to dependencies

winterqt:
Without Python as a dependency, the procedure documented in the documentation to install this package to pull in all of cbuild's dependencies is incorrect.


winterqt:
(Force-pushed because I forgot to bump the revision.)

PR #223, opened 2023-07-20
contrib/xstow: new package (1.1.1)

wezm:


contrib/xstow/template.py
```diff
@@ -0,0 +1,13 @@
+pkgname = "xstow"
+pkgver = "1.1.0"
+pkgrel = 0
+build_style = "gnu_configure"
+make_dir = "."
+hostmakedepends = ["automake"]
+makedepends = ["ncurses-devel"]
+pkgdesc = "Extended replacement of GNU Stow"
```
q66, line 8: this description is not very descriptive, it should say what the software does, not what it's a replacement of

q66:
you should also test it with `hardening = ["vis", "cfi"]`

wezm:
Ok I tweaked the description and turned on those hardening options. Still seems to build and run with them enabled

wezm:
I nudged upstream to publish a new release with the clang patches included, which they did. So I've updated the template and dropped the patches.

PR #225, opened 2023-07-23
cbuild: print arch which pkg can't be cross-compiled for

JamiKettunen:
The existing error message below doesn't make much sense:
```
=> qemu-8.0.3-r0: ERROR: cannot be cross-compiled for True
```
Make it print the architecture of the target instead:
```
=> qemu-8.0.3-r0: ERROR: cannot be cross-compiled for aarch64
```


q66:
yes, this was actually a typo on my part (i always intended to use arch)

PR #226, opened 2023-07-23
contrib/wl-clipboard: new package (2.2.0)

wezm:
I marked cfi as not enabled as the tool crashes with Illegal instruction when enabled.

Is there a way to capture optional dependencies? Looking through the docs and templates it appears not. `wl-clipboard` is an optional dependency of Neovim. When present it enables yank/paste from the system clipboard. `wl-clipboard` itself has [optional runtime dependencies](https://github.com/bugaevc/wl-clipboard/blob/master/BUILDING.md) on `xdg-utils` and `/etc/mime.types` (doesn't look like we have a package providing that so far).

### DIFF
--- a/contrib/wl-clipboard/template.py
+++ b/contrib/wl-clipboard/template.py
@@ -1,0 +1,18 @@
+pkgname = "wl-clipboard"
+pkgver = "2.2.0"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+    "wayland-progs",
+]
+makedepends = ["wayland-devel", "wayland-protocols"]
+depends = ["xdg-utils"]
+pkgdesc = "Command-line copy/paste utilities for Wayland"
+maintainer = "Wesley Moore <wes@wezm.net>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/bugaevc/wl-clipboard"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "414005cfa229d0e54f89a0d9a8473938e4c29adc21a9e556847a4d44ad508874"
+hardening = ["vis", "!cfi"]


contrib/wl-clipboard/template.py
```diff
@@ -0,0 +1,16 @@
+pkgname = "wl-clipboard"
+pkgver = "2.2.0"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+]
+makedepends = ["wayland-devel", "wayland-progs", "wayland-protocols"]
```
bugaevc, line 9: Perhaps the build instructions aren't quite clear on this (something for me to improve!), but wl-clipboard needs the `wayland-scanner` binary on the machine it's being built _on_, not the machine it's being built _for_ (these are called the _build_ and _host_ machines respectively in GNU and Meson terminology) — `wayland-scanner` is a code generator, it runs at build time. So if I'm reading this right, you should move `wayland-progs` up to `hostmakedepends` (and perhaps try actually cross-compiling it 🙂).

nekopsykose: cbuild does it backwards- makedepends is target-sysroot when cross, hostmakedepends is current machine when cross

nekopsykose: (so yes, it should be moved up, but by autoconf terminology the hostdeps would be the wrong place)

bugaevc: (Yes, I understood that.)

wezm: > Perhaps the build instructions aren't quite clear on this (something for me to improve!), but wl-clipboard needs the `wayland-scanner` binary

No that's just my inexperience and unfamiliarity with with the dependencies.

I've updated the template and tested that it cross-compiles.

q66:
`/etc/mime.types` is not a package, it's a debian-specific file, `xdg-utils` is already present in the repos (for a long time) and i'd just make it a hard dep

q66:
there is no way to do optional deps per se (outside of via `install_if`, but that pattern does not apply everywhere), in general using hard deps for optional stuff is fine unless the optional thing itself has a large dependency tree that may be unwanted or is itself very large; considering we're not aiming to be a small system, there is no harm in installing an extra package or two

bugaevc:
Please note that xdg-utils is exactly a package that has a huge dependency tree, at least in Debian: https://floss.social/@bugaevc/109655649574874908
That being said, you really want to have it by default with wl-clipboard, and most likely if you're using wl-clipboard you already have some desktop environment installed and that has dragged xdg-utils in already, so I guess that's fine if you don't have a mechanism for soft dependencies.

q66:
`xdg-utils` in chimera has no dependencies, therefore it cannot have a dependency tree (it used to depend on `xset`, which would have added a modest amount of dependencies, but that is now dealt with in revision 96006b3acad83111b7d6c85bb1a32e6519f88112)

wezm:
Ok I added `xdg-utils` as a dependency.

PR #249, opened 2023-08-03
contrib/aerc: new package (0.15.2)

nekopsykose:


contrib/aerc/template.py
```diff
@@ -0,0 +1,39 @@
+pkgname = "aerc"
+pkgver = "0.15.2"
+pkgrel = 0
+build_style = "makefile"
+make_check_target = "tests"
+hostmakedepends = [
+    "scdoc",
+]
+makedepends = [
+    "go",
```
q66, line 10: should probably be hostdep

contrib/gmime/template.py
```diff
@@ -0,0 +1,37 @@
+pkgname = "gmime"
+pkgver = "3.2.13"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--disable-dependency-tracking",
```
q66, line 6: this will probably become unneeded if you switch it to gmake

nekopsykose: true, sure

i wonder if it's possible to universally pass it? afaik gnu_conf always takes it, and it only speeds up configure since we don't care

contrib/aerc/template.py
```diff
@@ -0,0 +1,39 @@
...
+    "notmuch-devel",
+]
+checkdepends = [
+    "gnupg",
+    "gpgme",
+]
+go_build_tags = ["notmuch"]
```
q66, line 17: this probably doesn't do anything

contrib/aerc/template.py
nekopsykose: fixed

contrib/aerc/template.py
nekopsykose: the script it sources by default checks if [this thing](https://git.sr.ht/~rjarry/aerc/tree/master/item/contrib/goflags.sh) compiles, then passes it. passing it like this forces it so it doesn't get silently disabled if that fails for some reason

wdyt?

contrib/gmime/template.py
q66: it's probably better not to, when the dependency tracking setup fails it's better to figure out why than to silence it because it often indicates problems which may or may not result in miscompilation

contrib/aerc/template.py
nekopsykose: ah, though it's ignored for that anyway, sure

q66: not sure what you mean, but the tags only get passed when using go build via the util module, which does not happen here (as it uses a makefile)

nekopsykose: yeah, i assumed the tags were always added to GOFLAGS (`GOFLAGS=-tags $go_build_tags`). maybe they should? sometimes that would be useful to put `libsqlite3` in there even when not using go buildstyle

PR #256, opened 2023-08-04
main/mandoc: put the apropos trigger in a separate subpackage

nekopsykose:
this allows not running makewhatis every time a man page is touched


main/mandoc/template.py
```diff
@@ -72,3 +71,12 @@ def _base(self):
     self.build_style = "meta"
 
     return []
+
+
+@subpackage("mandoc-apropos")
+def _apropos(self):
+    self.pkgdesc = f"{pkgdesc} (apropos trigger)"
+    self.triggers = ["/usr/share/man"]
+    self.build_style = "meta"
```
nekopsykose, line 80: this is probably wrong, but to make it not so would have to move the makewhatis/apropos symlinks to this

main/mandoc/template.py
```diff
@@ -72,3 +71,12 @@ def _base(self):
     self.build_style = "meta"
 
     return []
+
+
+@subpackage("mandoc-apropos")
+def _apropos(self):
+    self.pkgdesc = f"{pkgdesc} (apropos trigger)"
+    self.triggers = ["/usr/share/man"]
```
nekopsykose, line 79: this could also include usr/local/share/man, technically

PR #259, opened 2023-08-05
contrib/rust-analyzer: new package (2023.07.31)

nekopsykose:


contrib/rust-analyzer/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "rust-analyzer"
+_pver = "2023-08-14"
+pkgver = _pver.replace("-", ".")
```
q66, line 3: use the version string for pkgver (i.e. `pkgver = "2023.08.14"`) and then use `{pkgver.replace('.', '-')}` in `source`

nekopsykose: done

PR #261, opened 2023-08-05
contrib/helix: new package (23.05)

wezm:
Helix [looks for system runtime files](https://docs.helix-editor.com/install.html#multiple-runtime-directories) alongside the binary. Runtime files include themes and 'queries' for syntax highlighting but also a bunch of dynamic libraries for treesitter parsers. For this reason I've installed everything into `/usr/libexec/helix` with `/usr/bin/hx` as a symlink to `/usr/libexec/helix/hx`. However, if there's a better option happy to move things around.

There a couple of patches for treesitter parsers that don't build with clang. I opened PR upstream as well.


contrib/helix/template.py
```diff
@@ -0,0 +1,37 @@
...
+pkgdesc = "Fast modal terminal-based text editor"
+maintainer = "Wesley Moore <wes@wezm.net>"
+license = "MPL-2.0"
+url = "https://github.com/helix-editor/helix"
+source = f"{url}/releases/download/{pkgver}/helix-{pkgver}-source.tar.xz"
+sha256 = "c1ca69facde99d708175c686ce5bf3585e119e372c83e1c3dc1d562c7a8e3d87"
+
+
+def do_install(self):
+    self.cargo.install(wrksrc="helix-term")
```
wezm, line 16: If alice's change in #259 is merged this can be tweaked to use `cargo_build_path`.

nekopsykose: didn't know this was even possible, i guess this also technically works for the original issue. not sure which is better :)

wezm:
I've converted this to a draft PR. I'm running into crashes (on x86_64) with the binary I built. Sometimes it panics with a Rust index out of bounds error, other times it just segfaults but it does so reliably as soon as I append line below (`o`) when editing `contrib/alacritty/template.py`. I can edit the same file in the same version of Helix on Arch.

nekopsykose:
i use the master branch of helix that i build myself and don't get any crashes. it could be:

- some difference between this release and master that they fixed later
- the vendored tree-sitter grammars we both use (which are also older revisions at this release) could be affected by different env cflags between building in cbuild compared to manually in cli

can you reproduce the same failure just running `cargo --install helix_term` on the repo manually (at the 23.05 version)?

wezm:
Thanks for the suggestions @nekopsykose. I was able to reproduce the issue with a `cargo install` from the 23.05 tarball but it was fixed when installing from master. Some reading through the commits since 23.05 revealed https://github.com/helix-editor/helix/pull/7227 and when applied as a patch to 23.05 it seems to fix the issue.

nekopsykose:
nice :D

wezm:
@q66 anything needed from me to get this one over the line?

contrib/helix/template.py
q66: let's just get this in as is for now

q66:
no, i just didn't get to it, sorry; it generally looks good, so i'll merge it as is

PR #263, opened 2023-08-05
experimental/zig: new package (0.11.0)

nekopsykose:


nekopsykose:
gotta fix

```console
$ zig version
0.11.0
error: Zig was built/linked incorrectly: LLVM and Clang have separate copies of libc++
       If you are dynamically linking LLVM, make sure you dynamically link libc++ too
```

which is

```cc
// Get a pointer to a static variable in libc++ from LLVM and make sure that
// it matches our own.
//
// This check is needed because if static/dynamic linking is mixed incorrectly,
// it's possible for Clang and LLVM to end up with duplicate "copies" of libc++.
//
// This is not benign: Static variables are not shared, so equality comparisons
// that depend on pointers to static variables will fail. One such failure is
// std::generic_category(), which causes POSIX error codes to compare as unequal
// when passed between LLVM and Clang.
//
// See also: https://github.com/ziglang/zig/issues/11168
bool ZigClangIsLLVMUsingSeparateLibcxx() {

    // Temporarily create an InMemoryFileSystem, so that we can perform a file 
    // lookup that is guaranteed to fail.
    auto FS = new llvm::vfs::InMemoryFileSystem(true);
    auto StatusOrErr = FS->status("foo.txt");
    delete FS;

    // This should return a POSIX (generic_category) error code, but if LLVM has
    // its own copy of libc++ this will actually be a separate category instance.
    assert(!StatusOrErr);
    auto EC = StatusOrErr.getError();
    return EC.category() != std::generic_category();
}

```

nekopsykose:
@andrewrk do you know why this check fails? i imagine "llvm has its own libc++ copy" is true just because the entire base c++ std is libcxx (for the whole system, there's no gcc components) (in which case i'm not sure how to handle that), but maybe there's actually something else that's wrong

nekopsykose:
for light reference  

```console
$ ldd /usr/bin/zig
	/lib/ld-musl-x86_64.so.1 (0x7ff981200000)
	libclang-cpp.so.16 => /lib/libclang-cpp.so.16 (0x7ff97bc00000)
	libLLVM-16.so => /lib/libLLVM-16.so (0x7ff974400000)
	libz.so.1 => /lib/libz.so.1 (0x7ff9897b1000)
	libzstd.so.1 => /lib/libzstd.so.1 (0x7ff98112c000)
	libc.so => /lib/ld-musl-x86_64.so.1 (0x7ff981200000)
	libc++.so.1 => /lib/../lib/libc++.so.1 (0x7ff97fcb4000)
	libffi.so.8 => /lib/../lib/libffi.so.8 (0x7ff9897a5000)
	libedit.so.0 => /lib/../lib/libedit.so.0 (0x7ff989767000)
	libncursesw.so.6 => /lib/../lib/libncursesw.so.6 (0x7ff97b989000)
$ ldd /usr/lib/libLLVM-16.0.6.so
	ldd (0x7fddaea00000)
	libffi.so.8 => /usr/lib/../lib/libffi.so.8 (0x7fddb6f90000)
	libedit.so.0 => /usr/lib/../lib/libedit.so.0 (0x7fddb6f52000)
	libz.so.1 => /usr/lib/../lib/libz.so.1 (0x7fddb6f34000)
	libzstd.so.1 => /usr/lib/../lib/libzstd.so.1 (0x7fdda6f2c000)
	libncursesw.so.6 => /usr/lib/../lib/libncursesw.so.6 (0x7fddae989000)
	libc++.so.1 => /usr/lib/../lib/libc++.so.1 (0x7fdda6de0000)
	libc.so => ldd (0x7fddaea00000)
$ ldd /usr/lib/libclang-cpp.so.16
	ldd (0x7f6d44200000)
	libLLVM-16.so => /usr/lib/../lib/libLLVM-16.so (0x7f6d38800000)
	libc++.so.1 => /usr/lib/../lib/libc++.so.1 (0x7f6d440b4000)
	libc.so => ldd (0x7f6d44200000)
	libffi.so.8 => /usr/lib/../lib/../lib/libffi.so.8 (0x7f6d4c7dd000)
	libedit.so.0 => /usr/lib/../lib/../lib/libedit.so.0 (0x7f6d4c79f000)
	libz.so.1 => /usr/lib/../lib/../lib/libz.so.1 (0x7f6d4c781000)
	libzstd.so.1 => /usr/lib/../lib/../lib/libzstd.so.1 (0x7f6d3852c000)
	libncursesw.so.6 => /usr/lib/../lib/../lib/libncursesw.so.6 (0x7f6d4c70a000)
```

experimental/zig/patches/nostrip.patch
```diff
@@ -0,0 +1,12 @@
+we do this ourselves later
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -835,7 +835,7 @@
+ elseif("${CMAKE_BUILD_TYPE}" STREQUAL "RelWithDebInfo")
+   set(ZIG_RELEASE_ARG -Doptimize=ReleaseFast)
+ else()
+-  set(ZIG_RELEASE_ARG -Doptimize=ReleaseFast -Dstrip)
```
andrewrk, line 8: This also affects conditional compilation & code generation - stack trace printing code will be omitted when strip is true. If you strip later, this code will remain, bloating the binary, but unable to be useful.

andrewrk (commented):
I'm sorry, I don't have any clue why this check is failing. Based on your cmake configuration and ldd output, it seems like it should be OK.

However, I don't think it's safe to ignore. Maybe there is someone more familiar with C++ you could ask?

In the long term, this issue will be resolved by https://github.com/ziglang/zig/issues/16270.

experimental/zig/patches/nostrip.patch
nekopsykose: ah, good to know. i assume the fact it's not actually strip(1) but rather objcopy'd with debuglink (-dbg subpackaged) still makes it useless? (i.e. zig itself won't be able to load it, so may as well just strip it fully)

andrewrk: Hmm well it may be the case that zig could be enhanced to find -dbg files being separately packaged. But let me switch topics for a moment. You shouldn't need that patch because if you want debug info you should use `-DCMAKE_BUILD_TYPE=RelWithDebInfo` as you can see in the if statement above. I would expect this to already be the default in chimera's package manager. Not sure why you would need a patch here.

nekopsykose: the default is [=None](https://github.com/chimera-linux/cports/blob/46fe927aff7a3e98a52f8f2b83f4dbf1a9f36a65/src/cbuild/util/cmake.py#L92), aside from this conditional the only difference for cmake is `RelWithDebInfo` appends `-O2 -g` to c*flags (None does nothing), and in this case everything is identical regardless of build type up until this conditional

nekopsykose: completely offtopic, the general philosophical reasoning distros use =None instead of other types is because aside from setting some default flags appended to env flags, they don't actually do anything by themselves. i.e.:

```
None:
Release: -O3 -DNDEBUG
RelWithDebInfo: -O2 -g
MinSizeRel: -Os -DNDEBUG
```

and they are appended to the env (`CFLAGS="stuff" cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo` gives `CFLAGS="stuff -O2 -g"`), which is sometimes unwanted. this basic difference aside, it's project-dependent what anyone does (sometimes a non-release type adds a bunch of stuff one doesn't want, sometimes the *release* type adds a bunch of stuff one doesn't want (in contrast to None), ...)

it's really a case by case basis so i read through every cmake project because people end up doing anything imaginable with conditionals :D

andrewrk: Thanks for the explanation! That is a useful perspective for me to be aware of.

mikdusan:
@nekopsykose, a couple of next-steps that may help:

We need to see the `zig build-exe` command for stage3 artifact; apologies we don't have an easier way to do it:

1. apply this tmp patch:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 2c812f2a7..0395383d4 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -877,6 +877,7 @@ set(ZIG_BUILD_ARGS
   "-Dtarget=${ZIG_TARGET_TRIPLE}"
   "-Dcpu=${ZIG_TARGET_MCPU}"
   "-Dversion-string=${RESOLVED_ZIG_VERSION}"
+  --verbose
 )
 
 add_custom_target(stage3 ALL
```

2. re-run the build from scratch and capture the output. I'm interested in the `zig2 build-exe` command line.

3. after the build configures or completes, verify `$BUILD/DIR/config.h` has this definition:
```
#define ZIG_LLVM_LINK_MODE "dynamic"
```

My current speculation is that a `libc++.a` is somehow linked.

mikdusan:
oh... I think this is the cause:

https://github.com/ziglang/zig/blob/d34201c8491007c5a24b4175a5170a5b6cc3d55e/build.zig#L632-L641

Looks like we should enhance the linux case, if `libstdc++` is not found, then try `libc++` before fallback to `exe.linkLibCpp();` (which iirc means link zig's bundled libc++.a).

Additionally there is a slight complication. We can no longer assume `THING.so` is to be found, because chimera follows OpenBSD's .so style and does not have a version-less pathname. We'd need some logic to handle that search. I can't recall if we have any code that already filters a dir for patterns { .so.X, .so.X.Y } and sorts them to use the highest version.

nekopsykose:
> 3. after the build configures or completes, verify $BUILD/DIR/config.h has this definition:

shared, but i assume the same thing:
```
/*
 * Copyright (c) 2016 Andrew Kelley
 *
 * This file is part of zig, which is MIT licensed.
 * See http://opensource.org/licenses/MIT
 */

#ifndef ZIG_CONFIG_H
#define ZIG_CONFIG_H

// Used by zig0.cpp
#define ZIG_VERSION_MAJOR 0
#define ZIG_VERSION_MINOR 11
#define ZIG_VERSION_PATCH 0
#define ZIG_VERSION_STRING ""

// Used by build.zig for communicating build information to self hosted build.
#define ZIG_CLANG_LIBRARIES "/usr/lib/libclang-cpp.so.16"
#define ZIG_CMAKE_BINARY_DIR "/builddir/zig-0.11.0/build"
#define ZIG_CMAKE_PREFIX_PATH ""
#define ZIG_CMAKE_STATIC_LIBRARY_PREFIX "lib"
#define ZIG_CMAKE_STATIC_LIBRARY_SUFFIX ".a"
#define ZIG_CXX_COMPILER "/usr/lib/ccache/bin/clang++"
#define ZIG_DIA_GUIDS_LIB ""
#define ZIG_LLD_INCLUDE_PATH "/usr/include"
#define ZIG_LLD_LIBRARIES "/usr/lib/liblldMinGW.a;/usr/lib/liblldELF.a;/usr/lib/liblldCOFF.a;/usr/lib/liblldWasm.a;/usr/lib/liblldMachO.a;/usr/lib/liblldCommon.a"
#define ZIG_LLVM_INCLUDE_PATH "/usr/include"
#define ZIG_LLVM_LIBRARIES "/usr/lib/libLLVM-16.so;-lrt;-ldl;-lm;-lz;-lzstd;-lcurses"
#define ZIG_LLVM_LIB_PATH "/usr/lib"
#define ZIG_LLVM_LINK_MODE "shared"

#endif
```

> 2. re-run the build from scratch and capture the output. I'm interested in the zig2 build-exe command line.

good point, not sure why i didn't get a verbose log in the first place since it's very useful:

```
root@cbuild: /builddir/zig-0.11.0/build$ /builddir/zig-0.11.0/build/zig2 build --prefix /builddir/zig-0.11.0/build/stage3 --zig-lib-dir /builddir/zig-0.11.0/lib -Dconfig_h=/builddir/zig-0.11.0/build/config.h -Denable-llvm -Doptimize=ReleaseFast -Dstrip -Dno-langref -Dno-autodocs -Dpie -Dtarget=native -Dcpu=baseline -Dversion-string=0.11.0 --verbose
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/absvdi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/absvsi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/absvti2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/addf3_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/addodi4_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/addosi4_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/addoti4_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/bswapdi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/bswapsi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/bswapti2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/clzdi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/clzsi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/clzti2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/cmpdi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/cmpsi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/cmpti2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/comparedf2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/comparesf2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/ctzdi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/ctzsi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/ctzti2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/divc3_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/divdf3_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/divsf3_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/divtf3_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/divti3_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/divxf3_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/extendf_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/ffsdi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/ffssi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/ffsti2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/fixint_test.zig
/builddir/zig-0.11.0/build/zig2 build-exe --stack 33554432 /builddir/zig-0.11.0/src/main.zig /builddir/zig-0.11.0/build/zigcpp/libzigcpp.a /usr/lib/libclang-cpp.so.16 /usr/lib/liblldMinGW.a /usr/lib/liblldELF.a /usr/lib/liblldCOFF.a /usr/lib/liblldWasm.a /usr/lib/liblldMachO.a /usr/lib/liblldCommon.a /usr/lib/libLLVM-16.so -lz -lzstd -lcurses -lunwind -lc++ -lc -fstrip -OReleaseFast --cache-dir /builddir/zig-0.11.0/zig-cache --global-cache-dir /tmp/.cache/zig --name zig -target native-native -mcpu x86_64 --mod build_options::/builddir/zig-0.11.0/zig-cache/c/8be3380d7aa95aa051fdfc613f919a48/options.zig --deps build_options -I /usr/include -I /usr/include -L /usr/lib --zig-lib-dir /builddir/zig-0.11.0/lib -fPIE --listen=- 
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/float_from_int_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/fmodq_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/fmodx_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/int_from_float_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/modti3_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/mulXi3_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/mulc3_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/mulf3_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/mulodi4_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/mulosi4_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/muloti4_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/negdi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/negsi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/negti2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/negvdi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/negvsi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/negvti2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/paritydi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/paritysi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/parityti2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/popcountdi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/popcountsi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/popcountti2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/powiXf2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/shift_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/subodi4_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/subosi4_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/suboti4_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/truncf_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/ucmpdi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/ucmpsi2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/ucmpti2_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/udivmoddi4_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/compiler_rt/udivmodti4_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/compress/deflate/compressor_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/compress/deflate/deflate_fast_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/compress/lzma/test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/compress/xz/test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/crypto/test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/fs/test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/hash/crc/catalog_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/io/test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/json/JSONTestSuite_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/json/dynamic_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/json/hashmap_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/json/scanner_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/json/static_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/json/stringify_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/json/test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/math/big/int_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/net/test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/os/linux/test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/os/test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/os/windows/test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/rand/test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/unicode/throughput_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/zig/parser_test.zig
info: truncate /builddir/zig-0.11.0/build/stage3/lib/zig/std/zig/perf_test.zig
steps [3/5] zig build-exe zig ReleaseFast native-native...
```

> My current speculation is that a libc++.a is somehow linked.  

to rule that out in another way too, slightly- the build root does not install `libcxx-devel-static`, so the only libcxx-related .a in there is libc++experimental.a (required to use that at all). the regular libc++.a isn't present, so there's no way to link it.  

> because chimera follows OpenBSD's .so style and does not have a version-less pathname.  

ehrm, no. libraries meant to be linked to have a bare .so:  

```
$ fd 'libc\+\+'                          
bldroot/usr/lib/libc++.so.1.0
bldroot/usr/lib/libc++experimental.a
bldroot/usr/lib/libc++.so
bldroot/usr/lib/libc++abi.so.1.0
bldroot/usr/lib/libc++abi.so.1
bldroot/usr/lib/libc++abi.so
bldroot/usr/lib/libc++.so.1
```

nekopsykose:
extracted from above:  

```
/builddir/zig-0.11.0/build/zig2 build-exe
--stack 33554432
/builddir/zig-0.11.0/src/main.zig
/builddir/zig-0.11.0/build/zigcpp/libzigcpp.a
/usr/lib/libclang-cpp.so.16
/usr/lib/liblldMinGW.a
/usr/lib/liblldELF.a
/usr/lib/liblldCOFF.a
/usr/lib/liblldWasm.a
/usr/lib/liblldMachO.a
/usr/lib/liblldCommon.a
/usr/lib/libLLVM-16.so
-lz
-lzstd
-lcurses
-lunwind
-lc++
-lc
-fstrip
-OReleaseFast
--cache-dir /builddir/zig-0.11.0/zig-cache
--global-cache-dir /tmp/.cache/zig
--name zig
-target native-native
-mcpu x86_64
--mod build_options::/builddir/zig-0.11.0/zig-cache/c/8be3380d7aa95aa051fdfc613f919a48/options.zig
--deps build_options
-I /usr/include
-I /usr/include
-L /usr/lib
--zig-lib-dir /builddir/zig-0.11.0/lib
-fPIE
--listen=-
```

it has a regular -lc++ which i assume would pick up the one from the path. i assume the real issue is just that `libLLVM` links a libc++ too but this has some semantics that fail a check (in contrast to a gcc system, where libLLVM will actually be linking libstdc++, and libc++ is purely statically linked for zig. i don't think this whole-system-is-libc++ is very common..)

nekopsykose:
but if i'm wrong, then:

```
Looks like we should enhance the linux case, if libstdc++ is not found, then try libc++ before fallback to exe.linkLibCpp(); (which iirc means link zig's bundled libc++.a).
```

if that is actually the case and there is a bundled libc++.a somewhere, then that would cause it. but one never exists in the buildroot from what i can tell, unless zig's compilation process makes one and deletes it after.

nekopsykose:
>> because chimera follows OpenBSD's .so style and does not have a version-less pathname.

> ehrm, no. libraries meant to be linked to have a bare .so:

to clarify- development (`-devel`) packages contain the unversioned links, so this is like any distro with devel packages (debian, ..).

mikdusan:
> if that is actually the case and there is a bundled libc++.a somewhere, then that would cause it. but one never exists in the buildroot from what i can tell, unless zig's compilation process makes one and deletes it after.

After a build, `find $BUILDROOT -name "libc++.a"` or similar... I'm expecting it to be found in a `zig-cache/` tree and that would be proof it is linked.

And if it is, here's a patch that just may solve issue...
```diff
diff --git a/build.zig b/build.zig
index 2dd71c5ec..e68c385e8 100644
--- a/build.zig
+++ b/build.zig
@@ -630,14 +630,19 @@ fn addCmakeCfgOptionsToExe(
         const lib_suffix = if (static) exe.target.staticLibSuffix()[1..] else exe.target.dynamicLibSuffix()[1..];
         switch (exe.target.getOsTag()) {
             .linux => {
-                // First we try to link against gcc libstdc++. If that doesn't work, we fall
-                // back to -lc++ and cross our fingers.
-                addCxxKnownPath(b, cfg, exe, b.fmt("libstdc++.{s}", .{lib_suffix}), "", need_cpp_includes) catch |err| switch (err) {
-                    error.RequiredLibraryNotFound => {
-                        exe.linkLibCpp();
-                    },
-                    else => |e| return e,
-                };
+                // First we try to link against system libstdc++.
+                // Next we try to link against system libc++.
+                // If that doesn't work, we fall to -lc++ and cross our fingers.
+                var found = false;
+                for ([_][]const u8{ "stdc++", "c++" }) |name| {
+                    addCxxKnownPath(b, cfg, exe, b.fmt("lib{s}.{s}", .{name, lib_suffix}), "", need_cpp_includes) catch |err| switch (err) {
+                        error.RequiredLibraryNotFound => continue,
+                        else => |e| return e,
+                    };
+                    found = true;
+                    break;
+                }
+                if (!found) exe.linkLibCpp();
                 exe.linkSystemLibrary("unwind");
             },
             .ios, .macos, .watchos, .tvos => {
```

nekopsykose:
> After a build, find $BUILDROOT -name "libc++.a" or similar... I'm expecting it to be found in a zig-cache/ tree and that would be proof it is linked.  

it indeed doesn't find anything, same output as the `fd` above. i guess it's not actually named the same..

but then i noticed, that the ldd above was misleading, because it prints the entire graph, flat. i.e. if anything links libc++, obviously ldd will output libc++. but zig itself does not contain a NEEDED for it- so it was actually statically linked!

and so:

> And if it is, here's a patch that just may solve issue...  

it works. thanks! the `exe.linkLibCpp();` does do that you expected and statically linked it.

mikdusan:
FYI, the fix landed in zig master, and andrew has flagged it to be part of `0.11.1`.

see commit https://github.com/ziglang/zig/commit/901457d173467c71b681a8c69f4b77c94d516da7

nekopsykose:
thanks!

q66:
ppc64le fail:

```
[13/19] Linking C executable zig1
[14/19] Running zig1.wasm to produce /builddir/zig-0.11.0/build/compiler_rt.c
FAILED: compiler_rt.c /builddir/zig-0.11.0/build/compiler_rt.c 
cd /builddir/zig-0.11.0 && /builddir/zig-0.11.0/build/zig1 /builddir/zig-0.11.0/lib build-obj lib/compiler_rt.zig -ofmt=c -OReleaseSmall --name compiler_rt -femit-bin="/builddir/zig-0.11.0/build/compiler_rt.c" --mod build_options::/builddir/zig-0.11.0/build/config.zig --deps build_options -target ppc64le-linux
error: UnknownArchitecture
[15/19] Running zig1.wasm to produce /builddir/zig-0.11.0/build/zig2.c
FAILED: zig2.c /builddir/zig-0.11.0/build/zig2.c 
cd /builddir/zig-0.11.0 && /builddir/zig-0.11.0/build/zig1 /builddir/zig-0.11.0/lib build-exe src/main.zig -ofmt=c -lc -OReleaseSmall --name zig2 -femit-bin="/builddir/zig-0.11.0/build/zig2.c" --mod build_options::/builddir/zig-0.11.0/build/config.zig --deps build_options -target ppc64le-linux
error: UnknownArchitecture
ninja: build stopped: subcommand failed.
=> A failure has occurred!
Traceback (most recent call last):
  File "/home/builder-ppc64le/worker-ppc64le/builder-ppc64le/build/src/runner.py", line 2211, in fire
    do_pkg(cmd)
  File "/home/builder-ppc64le/worker-ppc64le/builder-ppc64le/build/src/runner.py", line 1563, in do_pkg
    build.build(
  File "/home/builder-ppc64le/worker-ppc64le/builder-ppc64le/build/src/cbuild/core/build.py", line 120, in build
    buildm.invoke(pkg, step)
  File "/home/builder-ppc64le/worker-ppc64le/builder-ppc64le/build/src/cbuild/step/build.py", line 15, in invoke
    pkg.run_step("build", optional=True)
  File "/home/builder-ppc64le/worker-ppc64le/builder-ppc64le/build/src/cbuild/core/template.py", line 1257, in run_step
    if not run_pkg_func(self, "do_" + stepn) and not optional:
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/builder-ppc64le/worker-ppc64le/builder-ppc64le/build/src/cbuild/core/template.py", line 169, in run_pkg_func
    func(pkg)
  File "/home/builder-ppc64le/worker-ppc64le/builder-ppc64le/build/src/cbuild/build_style/cmake.py", line 9, in do_build
    self.make.build()
  File "/home/builder-ppc64le/worker-ppc64le/builder-ppc64le/build/src/cbuild/util/make.py", line 79, in build
    return self._invoke(
           ^^^^^^^^^^^^^
  File "/home/builder-ppc64le/worker-ppc64le/builder-ppc64le/build/src/cbuild/util/make.py", line 62, in _invoke
    return self.template.do(
           ^^^^^^^^^^^^^^^^^
  File "/home/builder-ppc64le/worker-ppc64le/builder-ppc64le/build/src/cbuild/core/template.py", line 1227, in do
    return chroot.enter(
           ^^^^^^^^^^^^^
  File "/home/builder-ppc64le/worker-ppc64le/builder-ppc64le/build/src/cbuild/core/chroot.py", line 767, in enter
    return subprocess.run(
           ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['bwrap', '--unshare-all', '--hostname', 'cbuild', '--ro-bind', PosixPath('/home/builder-ppc64le/cbuild/bldroot'), '/', '--bind', PosixPath('/home/builder-ppc64le/cbuild/bldroot/builddir'), '/builddir', '--ro-bind', PosixPath('/home/builder-ppc64le/cbuild/bldroot/destdir'), '/destdir', '--ro-bind', PosixPath('/home/builder-ppc64le/cbuild/sources'), '/sources', '--dev', '/dev', '--proc', '/proc', '--tmpfs', '/tmp', '--tmpfs', '/var/tmp', '--new-session', '--die-with-parent', '--bind', PosixPath('/home/builder-ppc64le/cbuild/cache'), '/cbuild_cache', '--uid', '1337', '--gid', '1337', '--chdir', PosixPath('/builddir/zig-0.11.0/build'), '--ro-bind-data', '5', '/tmp/cbuild-lld-args', 'linux64', '--', 'ninja', '-j72', 'all']' returned non-zero exit status 1.
```

q66:
aarch64 is pass (incl tests), but the test suite should probably print some progress or at least a summary, because being completely silent is pretty unhelpful

nekopsykose:
think that should address both, but if `all` isn't ideal then there aren't any other options (the default is just errors) short of a feature request for other output.

if the powerpc remap works then i'll send that upstream

q66:
the patch let it get past the original error, but now i'm getting this:

```
[19/19] Building stage3
FAILED: stage3/bin/zig /builddir/zig-0.11.0/build/stage3/bin/zig 
cd /builddir/zig-0.11.0 && /builddir/zig-0.11.0/build/zig2 build --prefix /builddir/zig-0.11.0/build/stage3 --zig-lib-dir /builddir/zig-0.11.0/lib -Dconfig_h=/builddir/zig-0.11.0/build/config.h -Denable-llvm -Doptimize=ReleaseFast -Dstrip -Dno-langref -Dno-autodocs -Dpie -Dtarget=native -Dcpu=baseline -Dversion-string=0.11.0
ninja: build stopped: subcommand failed.
```

there is no error message...

q66:
```
root@cbuild: /builddir/zig-0.11.0$ lldb ./build/zig2 -- build --prefix /builddir/zig-0.11.0/build/stage3 --zig-lib-dir /builddir/zig-0.11.0/lib -Dconfig_h=/builddir/zig-0.11.0/build/config.h -Denable-llvm -Doptimize=ReleaseFast -Dstrip -Dno-langref -Dno-autodocs -Dpie -Dtarget=native -Dcpu=bas
eline -Dversion-string=0.11.0
(lldb) target create "./build/zig2"
Current executable set to '/builddir/zig-0.11.0/build/zig2' (powerpc64le).
(lldb) settings set -- target.run-args  "build" "--prefix" "/builddir/zig-0.11.0/build/stage3" "--zig-lib-dir" "/builddir/zig-0.11.0/lib" "-Dconfig_h=/builddir/zig-0.11.0/build/config.h" "-Denable-llvm" "-Doptimize=ReleaseFast" "-Dstrip" "-Dno-langref" "-Dno-autodocs" "-Dpie" "-Dtarget=native" "-Dcpu=baseline" "-Dversion-string=0.11.0"
(lldb) run
Process 169 launched: '/builddir/zig-0.11.0/build/zig2' (powerpc64le)
Process 169 stopped
* thread #1, name = 'zig2', stop reason = signal SIGSEGV: invalid address (fault address: 0x10430cca0)
    frame #0: 0x000000010430cca0
error: memory read failed for 0x10430cc00
(lldb) bt
* thread #1, name = 'zig2', stop reason = signal SIGSEGV: invalid address (fault address: 0x10430cca0)
  * frame #0: 0x000000010430cca0
    frame #1: 0x000000010245374c zig2`fs_file_File_supportsAnsiEscapeCodes__7394(a0=(handle = 2)) at zig2.c:696753:10
    frame #2: 0x0000000102452e78 zig2`Progress_start__17472(a0=0x00003fffffffb088, a1=(ptr = "", len = 0), a2=0) at zig2.c:472212:7
    frame #3: 0x000000010038e5cc zig2`main_updateModule__217(a0=0x00003fffe1403970) at zig2.c:337717:7
    frame #4: 0x000000010033f494 zig2`main_cmdBuild__225(a0=mem_Allocator__3929 @ 0x00003ffffffff2b0, a1=mem_Allocator__3929 @ 0x00003ffffffff2a0, a2=anon__225_41 @ 0x00003ffffffff290) at zig2.c:353920:7
    frame #5: 0x00000001003053f8 zig2`main_mainArgs__195(a0=mem_Allocator__3929 @ 0x00003ffffffff6e0, a1=mem_Allocator__3929 @ 0x00003ffffffff6d0, a2=anon__195_41 @ 0x00003ffffffff6c0) at zig2.c:273031:9
    frame #6: 0x00000001002fbd40 zig2`main_main__193 at zig2.c:271961:7
    frame #7: 0x00000001002fb624 zig2`main(a0=16, a1=0x00003ffffffffa48, a2=0x00003ffffffffad0) at zig2.c:271572:8
    frame #8: 0x00003fffefa67430
    frame #9: 0x00003fffefa673a0
    frame #10: 0x00000001002f970c zig2`_start_c(p=<unavailable>) at crt1.c:18:2
    frame #11: 0x00000001002f96c8 zig2`_start + 40
```

q66:
the relevant function is

```
static bool fs_file_File_supportsAnsiEscapeCodes__7394(struct File__7377 const a0) {
 struct File__7377 const *t1;
 anon__7394_43 t6;
 anon__7394_43 t7;
 struct File__7377 t2;
 struct File__7377 t0;
 int32_t t4;
 bool t3;
 bool t5;
 t0 = a0;
 t1 = (struct File__7377 const *)&t0;
 t2 = (*t1);
 t3 = fs_file_File_isTty__7393(t2);
 if (t3) {
  t4 = a0.handle;
  t3 = t4 == INT32_C(1);
  if (t3) {
   t5 = true;
   goto zig_block_2;
  }
  t4 = a0.handle;
  t3 = t4 == INT32_C(2);
  t5 = t3;
  goto zig_block_2;

  zig_block_2:;
  if (t5) {
   t6 = os_getenvZ__2167(((uint8_t const *)&fs_file_File_supportsAnsiEscapeCodes__anon_111096__111096));
   t5 = t6.ptr != ((uint8_t const *)(uintptr_t)0x0ul);
   if (t5) {
    t6 = t6;
    memcpy(&t7, &t6, sizeof(anon__7394_43));
    t5 = mem_eql__anon_4688__4688(t7, (anon__7394_43){((uint8_t const *)&fs_file_File_supportsAnsiEscapeCodes__anon_111097__111097), (uintptr_t)4ul});
    if (t5) {
     return false;
    }
    goto zig_block_4;

    zig_block_4:;
    goto zig_block_3;
   }
   goto zig_block_3;

   zig_block_3:;
   goto zig_block_1;
  }
  goto zig_block_1;

  zig_block_1:;
  return true;
 }
 goto zig_block_0;

 zig_block_0:;
 return false;
}
```

the mentioned line is

```
t5 = mem_eql__anon_4688__4688(t7, (anon__7394_43){((uint8_t const *)&fs_file_File_supportsAnsiEscapeCodes__anon_111097__111097), (uintptr_t)4ul});
```

nekopsykose:
unfortunate, no idea how to fix that one :)  

> if the powerpc remap works then i'll send that upstream

that was helpfully also already fixed: https://github.com/ziglang/zig/commit/1c5c3f499a5144880feba9c24c51b0283a0e43c5

q66:
here is the listing of the local variables in frame 1:

```
(const File__7377) a0 = (handle = 2)
(const File__7377 *) t1 = 0x00003fffffffac18
(anon__7394_43) t6 = (ptr = "linux", len = 5)
(anon__7394_43) t7 = (ptr = "linux", len = 5)
(File__7377) t2 = (handle = 2)
(File__7377) t0 = (handle = 2)
(int32_t) t4 = 2
(bool) t3 = true
(bool) t5 = true
```

and the single relevant global:

```
(const uint8_t[5]) fs_file_File_supportsAnsiEscapeCodes__anon_111097__111097 = "dumb"
```



the segfault seems to be from `mem_eql__anon_4688__4688`, but i get no debug info for it; and the source file is massive... i wonder if i can extract and isolate the function and reproduce the issue on its own

but that's for later i guess

q66:
I have isolated a standalone case: https://gist.github.com/q66/c2ca25698a348c01b96809553893e610

however, compiling and running that does not segfault...

q66:
after dehardening the build:

```
FAILED: stage3/bin/zig /builddir/zig-0.11.0/build/stage3/bin/zig                                                                                                                                                                                                       
cd /builddir/zig-0.11.0 && /builddir/zig-0.11.0/build/zig2 build --prefix /builddir/zig-0.11.0/build/stage3 --zig-lib-dir /builddir/zig-0.11.0/lib -Dconfig_h=/builddir/zig-0.11.0/build/config.h -Denable-llvm -Doptimize=ReleaseFast -Dstrip -Dno-langref -Dno-autodo
cs -Dpie -Dtarget=native -Dcpu=baseline -Dversion-string=0.11.0                                                                                                                                                                                                        
thread 465 panic: reached unreachable code                                                                                                                                                                                                                             
/builddir/zig-0.11.0/lib/std/target.zig:289:33: 0x101349eb in default (build)                                                                                                                                                                                          
                        else => unreachable,                                                                                                                                                                                                                           
                                ^                                                                                                                                                                                                                                      
/builddir/zig-0.11.0/lib/std/target.zig:89:58: 0x100d3b5f in defaultVersionRange (build)                                                                                                                                                                               
                    .version_range = VersionRange.default(tag, arch),                                                                                                                                                                                                  
                                                         ^                                                                                                                                                                                                             
/builddir/zig-0.11.0/lib/std/zig/system/NativeTargetInfo.zig:38:57: 0x100ca437 in detect (build)                                                                                                                                                                       
    var os = cross_target.getOsTag().defaultVersionRange(cross_target.getCpuArch());                                                                                                                                                                                   
                                                        ^                                                                                                                                                                                                              
/builddir/zig-0.11.0/lib/std/Build/Step/Compile.zig:397:48: 0x1014c6eb in create (build)                                                                                                                                                                               
    const target_info = NativeTargetInfo.detect(options.target) catch @panic("unhandled error");                                                                                                                                                                       
                                               ^                                                                                                                                                                                                                       
/builddir/zig-0.11.0/lib/std/Build.zig:490:31: 0x100f0467 in addExecutable (build)                                                                                                                                                                                     
    return Step.Compile.create(b, .{                                                                                                                                                                                                                                   
                              ^                                                                                                                                                                                                                                        
/builddir/zig-0.11.0/test/link/macho/bugs/13457/build.zig:18:32: 0x102c3ceb in add (build)                                                                                                                                                                             
    const exe = b.addExecutable(.{                                                                                                                                                                                                                                     
                               ^                                                                                                                                                                                                                                       
/builddir/zig-0.11.0/test/link/macho/bugs/13457/build.zig:9:8: 0x1028016f in build (build)                                                                                                                                                                             
    add(b, test_step, .Debug);                                                                                                                                                                                                                                         
       ^                                                                                                                                                                                                                                                               
/builddir/zig-0.11.0/lib/std/Build.zig:1638:33: 0x10231897 in runBuild__anon_61581 (build)                                                                                                                                                                             
        .Void => build_zig.build(b),                                                                                                                                                                                                                                   
                                ^                                                                                                                                                                                                                                      
/builddir/zig-0.11.0/lib/std/Build.zig:1622:25: 0x101c8673 in dependencyInner__anon_57442 (build)                                                                                                                                                                      
    sub_builder.runBuild(build_zig) catch @panic("unhandled error");                                                                                                                                                                                                   
                        ^                                                                                                                                                                                                                                              
/builddir/zig-0.11.0/lib/std/Build.zig:1597:27: 0x1016ba7b in anonymousDependency__anon_27022 (build)                                                                                                                                                                  
    return dependencyInner(b, name, build_root, build_zig, args);                                                                                                                                                                                                      
                          ^                                                                                                                                                                                                                                            
/builddir/zig-0.11.0/test/tests.zig:656:46: 0x100fcf6f in addLinkTests (build)                                                                                                                                                                                         
            const dep = b.anonymousDependency(case.build_root, case.import, .{});                                                                                                                                                                                      
                                             ^                                                                                                                                                                                                                         
/builddir/zig-0.11.0/build.zig:492:42: 0x100ee4f3 in build (build)                                                                                                                                                                                                     
    test_step.dependOn(tests.addLinkTests(b, enable_macos_sdk, false, enable_symlinks_windows));                                                                                                                                                                       
                                         ^                                                                                                                                                                                                                             
/builddir/zig-0.11.0/lib/std/Build.zig:1639:43: 0x100ce0e7 in runBuild__anon_7105 (build)                                                                                                                                                                              
        .ErrorUnion => try build_zig.build(b),                                                                                                                                                                                                                         
                                          ^                                                                                                                                                                                                                            
/builddir/zig-0.11.0/lib/build_runner.zig:297:29: 0x100c981f in main (build)                                                                                                                                                                                           
        try builder.runBuild(root);                                                                                                                                                                                                                                    
                            ^                                                                                                      
/builddir/zig-0.11.0/lib/std/start.zig:574:37: 0x100b436b in posixCallMainAndExit (build)                                                                                                                                                                              
            const result = root.main() catch |err| {                                                                               
                                    ^                                                                                              
???:?:?: 0x0 in ??? (???)                                                                                                          
error: the following build command crashed:                                                                                        
/builddir/zig-0.11.0/zig-cache/o/12265b7ef1432cf89421f2821423a3d5/build /builddir/zig-0.11.0/build/zig2 /builddir/zig-0.11.0 /builddir/zig-0.11.0/zig-cache /tmp/.cache/zig --prefix /builddir/zig-0.11.0/build/stage3 --zig-lib-dir /builddir/zig-0.11.0/lib -Dconfig_
h=/builddir/zig-0.11.0/build/config.h -Denable-llvm -Doptimize=ReleaseFast -Dstrip -Dno-langref -Dno-autodocs -Dpie -Dtarget=native -Dcpu=baseline -Dversion-string=0.11.0
ninja: build stopped: subcommand failed.                                                                                           
```

Issue #245, opened 2023-08-03
cbuild checkapk implementation


nekopsykose:
basically a:

`./cbuild checkapk [--full] x/y` (path required) that does:  

- on new package (old doesn't exist):
  - full file list per (sub)package that looks like
    ```
    x:
    usr/lib/libx.so.1
    x-devel:
    usr/lib/libx.so
    ```
    (optionally sonames, etc)
    file size nice too
- on upgrade:
  - a diff of the old and new package (files, size)
- `--full` always does the 'new' behaviour and prints the full list as an override to not try find old/new

adbdump already does the first part, but you have to manually pass a path to the local repo and it's extremely verbose  

apk info -L/apk manifest works only for installed things too for a simple filelist

one of the issues i had with checkapk is that's it's really slow (the download/diff is serial to compare against), would be nice if it could automatically reuse the cache for the 'remote' file to diff against (checkapk doesn't) and also do all the downloads for the remote files in parallel if possible

PR #265, opened 2023-08-05
contrib/pipx: new package (1.2.0)

nekopsykose:
would be nice to have the SETUPTOOLS_SCM pretention be a global env when python

### DIFF
--- a/contrib/python-argcomplete/template.py
+++ b/contrib/python-argcomplete/template.py
@@ -1,0 +1,26 @@
+pkgname = "python-argcomplete"
+pkgver = "3.1.1"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+    "python-setuptools_scm",
+    "python-wheel",
+]
+checkdepends = [
+    "python-pytest",
+]
+pkgdesc = "Python tab completion plugin"
+maintainer = "psykose <alice@ayaya.dev>"
+license = "Apache-2.0"
+url = "https://github.com/kislyuk/argcomplete"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "8da1c3e591234df5ea4e741efae090fd34f1dd8cc220215ab31abd2b22b6aaa3"
+# missing pexpect
+options = ["!check"]
+
+
+def init_build(self):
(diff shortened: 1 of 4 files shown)


contrib/python-argcomplete/template.py
```diff
@@ -0,0 +1,26 @@
...
+license = "Apache-2.0"
+url = "https://github.com/kislyuk/argcomplete"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "8da1c3e591234df5ea4e741efae090fd34f1dd8cc220215ab31abd2b22b6aaa3"
+# missing pexpect
+options = ["!check"]
+
+
+def init_build(self):
+    self.env["SETUPTOOLS_SCM_PRETEND_VERSION"] = f"{pkgver}"
```
q66, line 26: what does this actually do in practice

nekopsykose: if you don't build from a blessed pypi archive the build fails because setuptools_scm doesn't know what to version the archive. or sometimes it doesn't fail but just outputs a `0.0.0` version.

this makes setuptools_scm ignore everything and just set the version in the var. which is always correct because the pkgver= is correct and we can skip all the useless ways of detecting it from magic files in the directory

nekopsykose: (one could say it's preferred to "use the pypi tarball" or similar, but in reality you can't always end up doing that (sdists miss tests pretty frequently, sometimes have bad files, ..), so it's more consistent to just override the version and use a normal git archive)

nekopsykose: similarly, there's a `python-pbr` env var for the same thing, and also some `poetry-dynamic-version` env var for the same thing, etc. hatch_vcs has the benefit of just being a shim for setuptools_scm so it reuses this

yeah, there's 500 layers of this in python that all write a 10 character version string to a file somewhere..

nekopsykose: the final output version is then visible in

```
site-packages/somepackage-$version.dist-info
```

and the `METADATA` in that folder:

```
Version: $version
```

having it be wrong just makes some things fail later that check the version at runtime for some reason (sometimes), and just looks ugly. it's part of the reason https://github.com/chimera-linux/cports/issues/245 is nice, because seeing a file diff shows the version in python site-packages after a build, so you can immediately see if it's completely wrong (or has a git/dev/whatever suffix by accident)

PR #267, opened 2023-08-05
contrib/yj: new package (5.1.0)

wezm:


contrib/yj/template.py
```diff
@@ -0,0 +1,11 @@
...
+pkgver = "5.1.0"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go"]
+pkgdesc = "Convert between YAML, TOML, JSON, and HCL"
+maintainer = "Wesley Moore <wes@wezm.net>"
+license = "Apache-2.0"
+url = "https://github.com/sclevine/yj"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "9a3e9895181d1cbd436a1b02ccf47579afacd181c73f341e697a8fe74f74f99d"
```
q66, line 11: add `options = ["!debug"]`

wezm: Done. Just for my own understanding, what's the reason for this? Something to do with `strip` and Go as discussed in the chat?

nekopsykose: llvm-objcopy fails to copy the debug info on ppc64\*, yeah. llvm-strip still works fine (iirc)

PR #268, opened 2023-08-06
contrib/qbittorrent: new package (4.5.4)

nekopsykose:
what doesn't work (daemon, not here anymore):

- daemon doesn't actually start:
  `alert : Could not create required directory '/.cache/qBittorrent'`
  the working-dir chroots to /var/lib/qbittorrent and i assume something goes wrong here with the mkdir from the daemon itself after (it uses cwd as home)..

- the group is on the wrong package
  the _qbittorrent user/group should only be on -nox, but it doesn't work to put it there with self.system_users, and you get `chown: _qbittorrent: illegal group name`

### DIFF
--- a/contrib/qbittorrent/template.py
+++ b/contrib/qbittorrent/template.py
@@ -1,0 +1,33 @@
+pkgname = "qbittorrent"
+pkgver = "4.5.4"
+pkgrel = 0
+build_style = "cmake"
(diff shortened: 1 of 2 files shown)


nekopsykose:
look bad commented but i guess i'll figure out the daemon another day since it's separate anyway

PR #278, opened 2023-08-06
contrib/doasedit: new package (1.0.6)

nekopsykose:


contrib/doasedit/template.py
```diff
@@ -0,0 +1,15 @@
+pkgname = "doasedit"
+pkgver = "1.0.6"
+pkgrel = 0
+depends = ["opendoas"]
+pkgdesc = "Shellscript to edit files with doas"
```
q66, line 5: Shell script

PR #280, opened 2023-08-06
contrib/emacs: new package (29.1)

nekopsykose:


q66:
i'd prefer not to bring in useless legacy junk like x athena widgets

nekopsykose:
i'm not sure what the other options are. there is:  

```
AC_ARG_WITH([x-toolkit],[AS_HELP_STRING([--with-x-toolkit=KIT],
 [use an X toolkit (KIT one of: yes or gtk, gtk2, gtk3, lucid or athena, motif, no)])],
[	  case "${withval}" in
	    y | ye | yes )	val=gtk ;;
	    n | no )		val=no  ;;
	    l | lu | luc | luci | lucid )	val=lucid ;;
	    a | at | ath | athe | athen | athena )	val=athena ;;
	    m | mo | mot | moti | motif )	val=motif ;;
	    g | gt | gtk  )	val=gtk ;;
	    gtk2  )	val=gtk2 ;;
	    gtk3  )	val=gtk3 ;;
	    * )
```

to sort these:  

- athena: current
- lucid: no idea what this is
- motif: no idea what this is
- gtk: this is some really old thing
- gtk2: lol
- gtk3: this is the non-pgtk legacy gtk3 thing, that works on x11 only (not even in xwayland, to be clear). this is perhaps the only alternative for an "-x11" packages (though i would name it -gtk3). and it's also buggy as hell and nobody used it in practice (pre-29.1, for the past few years). people always used athena through xwayland since that was the only option, and athena on x11 proper.


if lucid/motif are less garbage, then perhaps those, but aside from that this would have no x11 build since gtk3 is just worse (and this is merely one library for emacs)

nekopsykose:
(ah and to be specific, -pgtk is actually just for wayland and prints a warning if you start it in xwayland or x11, that it should not be used there)

nekopsykose:
apparently lucid uses the same thing as athena and nobody uses motif either, so..

nekopsykose:
so i guess the only actual option is having no x11 emacs at all or having a -gtk3 one (that i can't test given it cannot run on wayland because it is some really weird hacks that only works on real hacks)

pick an option and i'm fine with whatever :D

q66:
if gtk3 works fine these days maybe we can have pgtk + gtk3, at least that way we get the same toolkit on both, and a console version otherwise?

nekopsykose:
> we can have pgtk + gtk3

sure, though i can't test it. i guess it probably works

> though i still wonder wtf emacs upstream is smoking   

i want some too tbh

contrib/emacs-console/template.py
```diff
@@ -0,0 +1,56 @@
...
+# no tests
+options = ["!check"]
+
+system_groups = ["_games"]
+
+
+def post_install(self):
+    # remove suid from game exe
+    (
+        next(self.find(self.destdir / "usr/libexec/emacs", "update-game-score"))
```
q66, line 55: what's the find for? seems equivalent to `(self.destdir / "usr/libexec/emacs/update-game-score").chmod(0o755)`

nekopsykose: i guess `f"usr/libexec/emacs/{pkgver}/{self.profile.triplet()}/update-game-score"` could work to be direct too

PR #284, opened 2023-08-06
contrib/alacritty: new package (0.12.2)

wezm:
This is a continuation of #204 I've bumped it to the latest release, slimmed down the dependencies, added man pages and terminfo, and updated it to use `install_completion`.

I'm having trouble with cross-compilation though. I'm getting a lot of errors such as these but I'm not sure what's different with this codebase to other Rust ones I've packaged:

```
error[E0463]: can't find crate for `core`
  |
  = note: the `aarch64-chimera-linux-musl` target may not be installed
  = help: consider downloading the target with `rustup target add aarch64-chimera-linux-musl`

error[E0463]: can't find crate for `compiler_builtins`

For more information about this error, try `rustc --explain E0463`.
error[E0463]: can't find crate for `core`
 --> /builddir/alacritty-0.12.2/vendor/lazy_static/src/inline_lazy.rs:8:1
  |
8 | extern crate core;
  | ^^^^^^^^^^^^^^^^^^ can't find crate
  |
  = note: the `aarch64-chimera-linux-musl` target may not be installed
  = help: consider downloading the target with `rustup target add aarch64-chimera-linux-musl`

error[E0463]: can't find crate for `std`
 --> /builddir/alacritty-0.12.2/vendor/lazy_static/src/inline_lazy.rs:9:1
  |
9 | extern crate std;
  | ^^^^^^^^^^^^^^^^^ can't find crate
  |
  = note: the `aarch64-chimera-linux-musl` target may not be installed
  = help: consider downloading the target with `rustup target add aarch64-chimera-linux-musl`
```

### DIFF
--- a/contrib/alacritty-terminfo
+++ b/contrib/alacritty-terminfo
@@ -1,0 +1,1 @@
+alacritty
--- a/contrib/alacritty/template.py
+++ b/contrib/alacritty/template.py
@@ -1,0 +1,45 @@
+pkgname = "alacritty"
+pkgver = "0.12.2"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo", "cmake", "pkgconf", "python", "ncurses"]
+makedepends = ["fontconfig-devel", "freetype-devel", "libxcb-devel", "rust"]
+depends = [f"alacritty-terminfo={pkgver}-r{pkgrel}"]
+pkgdesc = "Cross-platform, GPU-accelerated terminal emulator"
+maintainer = "nbfritch <nbfritch@gmail.com>"
+license = "MIT OR Apache-2.0"
+url = "https://github.com/alacritty/alacritty"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "285b44d2d2e83118ab7fe61d575811bb85d5e21147808171bd0e583e9e794748"
+
+
+def do_install(self):
+    self.cargo.install(wrksrc="alacritty")
+    self.install_man("extra/alacritty.man", name="alacritty", cat=1)
+    self.install_man("extra/alacritty-msg.man", name="alacritty-msg", cat=1)
+    self.install_license("LICENSE-MIT")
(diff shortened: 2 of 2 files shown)


nekopsykose:
> I'm having trouble with cross-compilation though. I'm getting a lot of errors such as these but I'm not sure what's different with this codebase to other Rust ones I've packaged:  

there's no `makedepends = rust` so there is no tgt-sysroot stdlib? just a guess

contrib/alacritty-terminfo
```diff
@@ -0,0 +1 @@
+alacritty
```
nekopsykose, line 1: i forget if it's required, but shouldn't you depends=alacritty-terminfo=pkgver-rpkgrel on this from the base package? it's normally separate to allow standalone installation of the terminfo on another machine, but the terminal itself would need it

wezm:
You're right on both counts @nekopsykose. Thanks for the help.

contrib/alacritty/template.py
```diff
@@ -0,0 +1,45 @@
...
+pkgdesc = "Cross-platform, GPU-accelerated terminal emulator"
+maintainer = "nbfritch <nbfritch@gmail.com>"
+license = "MIT OR Apache-2.0"
+url = "https://github.com/alacritty/alacritty"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "285b44d2d2e83118ab7fe61d575811bb85d5e21147808171bd0e583e9e794748"
+
+
+def do_install(self):
+    self.cargo.install(wrksrc="alacritty")
```
q66, line 17: you can probably remove this and turn the `do_install` into a `post_install`

nekopsykose: install --path=. doesn't work in workspaces, you need either this wrksrc override or the proposed change in https://github.com/chimera-linux/cports/pull/259 .

q66: alright, let's go with this for now

PR #302, opened 2023-08-10
contrib/execline: new package (2.9.3.0)

nekopsykose:
funnily this squats things like /usr/bin/cd, but nothing actually puts anything in such names since they are shell builtins. probably safe?

### DIFF
--- a/contrib/execline-devel
+++ b/contrib/execline-devel
@@ -1,0 +1,1 @@
+execline
--- a/contrib/execline/template.py
+++ b/contrib/execline/template.py
@@ -1,0 +1,33 @@
+pkgname = "execline"
+pkgver = "2.9.3.0"
+pkgrel = 0
+build_style = "configure"
+configure_args = [
+    "--prefix=/usr",
+    "--disable-allstatic",
+    "--enable-multicall",
+    "--enable-shared",
+    "--libdir=/usr/lib",
+]
+make_cmd = "gmake"
+hostmakedepends = ["gmake"]
+makedepends = ["skalibs-devel"]
+pkgdesc = "Small non-interactive scripting language"
+maintainer = "psykose <alice@ayaya.dev>"
+license = "ISC"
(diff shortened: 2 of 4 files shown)


q66:
i don't really like the way this provides a shitload of core names, can something be done about it

q66:
so why is that anyway

nekopsykose:
all the components are just individual programs that execute other programs from path. it's not any different from typing `prog` in a shell script.

so for instance

```
#!/usr/bin/execlineb

foreground { echo "hello" }
echo "hello"
```

the `foreground` and `echo` come from `PATH`. the execline interpreter only interprets the language to then execute it (like how /bin/sh interprets a specific syntax in a shell script). the content is all regular path programs (no builtins). and all of them are `prog args.. prog` form, like one long chain that execs into eachother. just how the language was designed i guess

q66:
that's still not a reason it can't have its own private namespace (e.g. in libexec) for all this bs

nekopsykose:
if it was in libexec none of the programs would work because libexec is not in PATH

q66:
it's literally a separate interpreter, there is no rule it has to only ever consider PATH

nothing that is not execline will ever care about all the binaries it installs

nekopsykose:
the interpreter doesn't do anything except rewrite the script a little (to have a correct argv list, from the input string(file) given). the final program is a single long `argv` thing, by design. there's no interpreter sitting watching over the whole execution that can map the names to some magic private libexec dir- it's no different than if one types the final result directly into a execvp and exec()s it. it's somewhat possible to have it work by 'put magic dir in PATH before using any execline program ever', but that means you cannot have any script using it just work without first also writing a wrapper script or whatnot to modify the path first (or modifying it to append to path first). debian does that, and it's basically unusable as a result.

> nothing that is not execline will ever care about all the binaries it installs  

the `fd*` programs are expected to be in path for other things that just use them directly from other s6-* utilities (i.e. just `fdmove`, ..)

PR #305, opened 2023-08-10
contrib/wine: new package (8.13)

nekopsykose:
doesn't actually work yet, just fails on start


nekopsykose:
works, just had a stale prefix

contrib/wine/template.py
```diff
@@ -0,0 +1,120 @@
...
+        "regsvr32",
+        "wineboot",
+        "winecfg",
+        "wineconsole",
+        "winefile",
+        "winemine",
+        "winepath",
+    ]:
+        self.rm(dd / f"usr/bin/{link}")
+        self.ln_s("wineapploader", dd / f"usr/bin/{link}")
```
q66, line 93: ditto

PR #311, opened 2023-08-13
contrib/mold: update to 2.1.0

nekopsykose:
https://github.com/rui314/mold/releases/tag/v2.1.0


contrib/onetbb/template.py
```diff
@@ -0,0 +1,45 @@
...
+# a lot of the tests can deadlock and it's a pain to figure out which to disable
+options = ["!check"]
+
+
+@subpackage("onetbb-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage("libtbb")
```
q66, line 38: is there any reason to split these

nekopsykose: nothing generally ever links the others and it's 4 line of code

q66: it leaves the main package literally empty and these packages have 320k and 128k installed size respectively

just drop the subpackages

PR #326, opened 2023-08-17
contrib/obs-studio: new package (29.1.3)

nekopsykose:


contrib/libsrt/template.py
```diff
@@ -0,0 +1,38 @@
+pkgname = "libsrt"
```
q66, line 1: should be called just srt

nekopsykose: renamed

PR #336, opened 2023-08-17
main/cmake: update to 3.27.3

nekopsykose:
also uses ninja for cmake's self-bootstrap to make it more faster since it's free

### DIFF
--- a/main/cmake/template.py
+++ b/main/cmake/template.py
@@ -1,11 +1,12 @@
 pkgname = "cmake"
-pkgver = "3.27.1"
+pkgver = "3.27.3"
 pkgrel = 0
 build_style = "configure"
 configure_args = [
     "--prefix=/usr",
     "--mandir=/share/man",
     "--docdir=/share/doc/cmake",
+    "--generator=Ninja",
     "--system-libarchive",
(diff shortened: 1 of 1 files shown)


nekopsykose:
the sigill with cfi is unrelated to the upgrade and also just fails on 3.27.1. unsetting it works (and is fine, since this is neither a longrun service nor super meaningful for hardening), but perhaps you want to explore that

q66:
it built fine previously though? is this due to the ninja switch or something

main/cmake/template.py
```diff
@@ -1,30 +1,38 @@
...
     "--system-zstd",
     f"--parallel={self.conf_jobs}",
 ]
-makedepends = ["libarchive-devel", "ncurses-devel", "linux-headers"]
+make_cmd = "ninja"
+makedepends = [
+    "libarchive-devel",
+    "linux-headers",
+    "ncurses-devel",
+    "ninja",
```
q66, line 22: hostmakedepends

nekopsykose:
> is this due to the ninja switch or something  

nope, you can just uhh try a local build? just master and building current (no changes) does the same failure

PR #361, opened 2023-08-18
contrib/keyd: new package (2.4.3)

feurry:


q66:
please no merge commits (use rebase)

contrib/keyd/template.py
```diff
@@ -0,0 +1,20 @@
+pkgname = "keyd"
+pkgver = "2.4.3"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_use_env = True
+hostmakedepends = ["gmake"]
+makedepends = ["linux-headers"]
+pkgdesc = "Key remapping daemon for linux"
+maintainer = "feurry <=@gmail.com>"
```
nekopsykose, line 10: are you `=@gmail.com`? (it has to be a valid email that isn't someone elses)

contrib/keyd/patches/no-systemd.patch
```diff
@@ -0,0 +1,15 @@
+--- a/Makefile
```
nekopsykose, line 1: this patch doesn't do anything? `$(DESTDIR)$(PREFIX)/lib/systemd/` doesn't exist so it doesn't actually install a service file and goes to else, so then it wouldn't have to be patched

contrib/keyd/patches/no-systemd.patch
```diff
@@ -0,0 +1,15 @@
...
++++ b/Makefile
```
nekopsykose, line 2: you should patch out the -O3 in https://github.com/rvaiya/keyd/blob/eeea96aa4dda2b94a1431362c781064543625246/Makefile#L37

contrib/keyd/patches/no-systemd.patch
feurry: Installation was failing because my system does have a directory /usr/lib/systemd with 
../systemd/user/dconf.service and 
../systemd/user/org.gnome.Evince.service

nekopsykose: it also works fine just removing this, i just checked

contrib/keyd/template.py
```diff
@@ -0,0 +1,20 @@
...
+license = "MIT"
+url = "https://github.com/rvaiya/keyd"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "d4075f673879f4950c1c0d43985797603200e993596a5d7fcec1597c2350c380"
+# no tests available
+options = ["!check"]
+
+def post_install(self):
+    self.install_license("LICENSE")
+    self.install_file(self.files_path / "keyd", "etc/dinit.d")
```
nekopsykose, line 20:
```suggestion
    self.install_service(self.files_path / "keyd")
```

contrib/keyd/patches/no-systemd.patch
feurry: no that's not the error message. What I mean is that line 50 of the makefile
@if [ -e $(DESTDIR)$(PREFIX)/lib/systemd/ ]; then \
fails on my system because I have a directory /usr/lib/systemd/

This is the error message.
$ doas gmake install
install: /usr/lib/systemd/system/keyd.service: No such file or directory
gmake: *** [Makefile:50: install] Error 71

contrib/keyd/template.py
feurry: thanks. fixed.

contrib/keyd/patches/no-systemd.patch
feurry: thanks. changed this.

contrib/keyd/patches/no-systemd.patch
nekopsykose: that happens because it uses `install -D`, but that never happens in actual build so it doesn't matter

contrib/keyd/template.py
feurry: Thanks. Made the change.

contrib/keyd/patches/no-systemd.patch
feurry: removed and package builds successfully.

feurry:
I think I finally have this in the proper state to commit

q66:
mainly 3 issues here:

1) fix your lint issues (in a chimera env, you will want to install `python-black` and `python-flake8` and ensure to run `black` on your code as well as check lint errors with `flake8`)
2) the package should create a `keyd` group (`system_groups = ["keyd"]`)
3) try if the package works with `hardening = ["vis", "cfi"]`; if it doesn't work, it's not a blocker (in which case set to just `hardening = ["vis"]` and put a comment like `# FIXME cfi` above), but if it does, let's have it there

feurry:
1. passed flake8 and black
2. added keyd group
3. 'vis' and 'cfi' hardnening enable and compiling

PR #362, opened 2023-08-18
contrib/dmidecode: new package (3.5)

nekopsykose:


contrib/dmidecode/template.py
```diff
@@ -0,0 +1,14 @@
+pkgname = "dmidecode"
+pkgver = "3.5"
+pkgrel = 0
```
q66, line 3: i believe this should be `archs = ["x86_64"]`, no?

nekopsykose: aarch64 also has smbios/dmi afaik. so perhaps just those 2, not sure if riscv will ever support that either..

nekopsykose: hm, think riscv also has it

nekopsykose: from what i can tell any platform that implements EFI can optionally then provide the smbios tables for this to end up working, i guess

q66: these 3 then

PR #366, opened 2023-08-20
contrib/pigz: new package (2.8)

nekopsykose:


q66:
there should probably be ani unpigz.1 symlink

nekopsykose:
good idea

PR #372, opened 2023-08-20
contrib/aerc: actually enable notmuch support

nekopsykose:


q66:
should use tool_flags

PR #385, opened 2023-08-23
contrib/qemu: update to 8.1.0

nekopsykose:
https://wiki.qemu.org/ChangeLog/8.1

### DIFF
--- a/contrib/qemu/template.py
+++ b/contrib/qemu/template.py
@@ -1,5 +1,5 @@
 pkgname = "qemu"
-pkgver = "8.0.4"
+pkgver = "8.1.0"
 pkgrel = 0
 build_style = "gnu_configure"
 # TODO vde liburing libssh capstone
@@ -36,7 +36,7 @@ configure_args = [
     "--disable-bsd-user",
     "--disable-werror",
     "--disable-xen",
-    "--audio-drv-list=pa,jack,sdl",
+    "--audio-drv-list=pa,pipewire,jack,sdl",
 ]
 make_cmd = "gmake"
 hostmakedepends = [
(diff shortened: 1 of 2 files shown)


contrib/qemu/template.py
```diff
@@ -246,7 +247,7 @@ def _system(self):
                     "usr/lib/qemu/s390-ccw.img",
                     "usr/lib/qemu/s390-netboot.img",
                 ]
-                self.options += ["execstack"]
+                self.options += ["execstack", "textrels"]
```
nekopsykose, line 250: not sure if this is ok or not for

```
=> qemu-system-s390x-8.1.0-r0: moving: /home/demon/src/cports/bldroot/destdir/qemu-8.1.0/usr/lib/qemu/s390-netboot.img -> /home/demon/src/cports/bldroot/destdir/qemu-system-s390x-8.1.0
   usr/lib/qemu/s390-ccw.img
   usr/lib/qemu/s390-netboot.img
=> qemu-system-s390x-8.1.0-r0: ERROR: found textrels:
```

q66: should be fine yeah

contrib/qemu/template.py
```diff
@@ -36,7 +36,7 @@
     "--disable-bsd-user",
     "--disable-werror",
     "--disable-xen",
-    "--audio-drv-list=pa,jack,sdl",
+    "--audio-drv-list=pa,pipewire,jack,sdl",
```
q66, line 39: hm, maybe we should drop pa from the list now that pipewire is in

nekopsykose: i can't get it to work lol `pipewire: Failed to initialize PW contextaudio: Could not init `pipewire' audio driver`

that said, the reason i didn't drop pa is because it's not automatic; you manually do -audio/-audiodev with an explicit driver, so removing pa means everyone has to update what they pick (in contrast to a random application that just automatically uses whatever you compiled it with and just connects to audio server)

q66: ok, fair

PR #393, opened 2023-08-23
nuke typing_extensions from 2 packages

nekopsykose:
annoyin ass compat packages

### DIFF
--- a/contrib/python-twisted/patches/no-typing-ext.patch
+++ b/contrib/python-twisted/patches/no-typing-ext.patch
@@ -1,0 +1,142 @@
+we have a new enough python version and don't need this
+--
+--- a/setup.cfg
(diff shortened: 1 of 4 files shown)


q66:
y tho

nekopsykose:
useless empty dep

that said twisted is p big anyway and setuptools_scm is just fixed in next version anyway

q66:
i'll merge this if you also remove the `python-typing_extensions` package since it's now a leaf

nekopsykose:
it should be kept because there are things that would actually use it beyond backported types in the future (mypy, ..)

q66:
can't we just nuke typing_extensions from these individually

doesn't seem super hard to do

nekopsykose:
that wouldn't make sense in every case

PR #403, opened 2023-08-24
contrib/linux-tools: new package (6.4.12)

nekopsykose:
starting with cpupower only because it's at least easy to build

of course, linux-tools itself will remain empty since everything should be neatly split

### DIFF
--- a/contrib/linux-tools/template.py
+++ b/contrib/linux-tools/template.py
@@ -1,0 +1,77 @@
+pkgname = "linux-tools"
+pkgver = "6.4.12"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+_make_args = [
+    "-C",
+    "tools",
+    # FIXME: cpufreq-bench is completely broken with optimisations because of
+    # int UB that gets optimised out and then breaks in div-by-zero
+    "CPUFREQ_BENCH=0",
+    "LLVM=1",
+    "NLS=false",
+    "WERROR=0",
+    "libdir=/usr/lib",
+    "mandir=/usr/share/man",
+    "prefix=/usr",
+    "sbindir=/usr/bin",
+]
+make_build_target = ""
+make_build_args = _make_args + [
+    "cpupower",
+]
+make_install_target = ""
+make_install_args = _make_args + [
+    "cpupower_install",
+]
+hostmakedepends = [
+    "gmake",
+]
+makedepends = [
+    "linux-headers",
+    "pciutils-devel",
+]
+pkgdesc = "Linux Kernel userspace tools"
+maintainer = "psykose <alice@ayaya.dev>"
+license = "GPL-2.0-only"
+url = "https://www.kernel.org"
+source = f"https://cdn.kernel.org/pub/linux/kernel/v{pkgver[:pkgver.find('.')]}.x/linux-{pkgver}.tar.xz"
+sha256 = "cca91be956fe081f8f6da72034cded96fe35a50be4bfb7e103e354aa2159a674"
+# vis breaks everything
+hardening = []
+# nah 💀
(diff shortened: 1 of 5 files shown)


q66:
hm, perhaps we should build the tools as individual templates?

contrib/linux-tools/template.py
```diff
@@ -0,0 +1,81 @@
...
+# nah 💀
+options = ["!check"]
+
+
+def init_build(self):
+    self.make_build_args += [f"CC={self.get_tool('CC')}"]
+
+
+def do_build(self):
+    self.make.build(["cpupower"])
```
q66, line 48: this probably shouldn't be needed, just set `make_build_target` to `cpupower`

contrib/linux-tools/template.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def do_install(self):
+    self.make.install(["cpupower_install"])
```
q66, line 52: ditto

contrib/linux-tools/template.py
```diff
@@ -0,0 +1,81 @@
...
+source = f"https://cdn.kernel.org/pub/linux/kernel/v{pkgver[:pkgver.find('.')]}.x/linux-{pkgver}.tar.xz"
+sha256 = "cca91be956fe081f8f6da72034cded96fe35a50be4bfb7e103e354aa2159a674"
+# vis breaks everything
+hardening = []
+# nah 💀
+options = ["!check"]
+
+
+def init_build(self):
+    self.make_build_args += [f"CC={self.get_tool('CC')}"]
```
q66, line 44: this shouldn't be needed, it's already in there implicitly

contrib/linux-tools/template.py
```diff
@@ -0,0 +1,81 @@
...
+def _cpupower(self):
+    self.pkgdesc = "Linux standard cpu power management tool"
+    return [
+        "usr/bin/cpupower",
+        "usr/share/bash-completion/completions/cpupower",
+        "usr/share/man/man1/cpupower*",
+    ]
+
+
+@subpackage("libcpupower")
```
q66, line 65: is there a point in splitting the lib rather than just having `cpupower` and `cpupower-devel`? the binary should be tiny and depends on the lib anyway

nekopsykose: yeah it's neater this way

contrib/linux-tools/template.py
nekopsykose: it's not, make.build adds it but we don't call it here

contrib/linux-tools/template.py
q66: it's not

contrib/linux-tools/template.py
q66: it would be if you apply the suggestions above

contrib/linux-tools/template.py
nekopsykose: there's multiple targets for both so i left it in args

nekopsykose:
> hm, perhaps we should build the tools as individual templates?  

i don't see why, it's just calling `make x` and `make x_install`

q66:
because then one can drop all the custom `do_build` crap

contrib/linux-tools/template.py
```diff
@@ -0,0 +1,77 @@
...
+    "NLS=false",
+    "WERROR=0",
+    "libdir=/usr/lib",
+    "mandir=/usr/share/man",
+    "prefix=/usr",
+    "sbindir=/usr/bin",
+]
+make_build_target = ""
+make_build_args = _make_args + [
+    "cpupower",
```
q66, line 22: just put that in `make_build_target` lol

PR #435, opened 2023-08-30
spirv/vulkan updates

nekopsykose:


q66:
the only things which need glslang rebuilds are libplacebo, shaderc and gamescope

PR #445, opened 2023-08-31
contrib/qemu-user: disable pie for riscv64

JamiKettunen:
This breaks running binaries for any target arch.

Still to-be-tested tomorrow on my VisionFive 2 board tomorrow in full, so far I only narrowed it down to this `hardening` opt in a minimal template, so marking as a draft for now.

### DIFF
--- a/contrib/qemu-user/template.py
+++ b/contrib/qemu-user/template.py
@@ -1,6 +1,6 @@
 pkgname = "qemu-user"
 pkgver = "8.1.0"
-pkgrel = 0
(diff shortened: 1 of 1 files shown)


JamiKettunen:
This still segfaults unfortunately unlike the [minimal template](https://github.com/JamiKettunen/cports/commit/fd7c208e) I have, time to keep digging I guess...

q66:
this is obviously not the way to go and not due to pie so i'll close this one

PR #446, opened 2023-09-01
main/procps: update to 4.0.4

nekopsykose:
https://gitlab.com/procps-ng/procps/-/blob/4ddcef2fd843170c8e2d59a83042978f41037a2b/NEWS

### DIFF
--- a/main/procps/template.py
+++ b/main/procps/template.py
@@ -1,5 +1,5 @@
 pkgname = "procps"
-pkgver = "4.0.3"
+pkgver = "4.0.4"
 pkgrel = 0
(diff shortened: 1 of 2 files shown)


main/procps/template.py
```diff
@@ -8,6 +8,7 @@
     "--disable-modern-top",
     "--without-systemd",
 ]
+configure_gen = []
```
q66, line 11: usually when dealing with trailing configure_gen, it should be made to regen, unless that is somehow impossible (e.g. cyclic or broken)

PR #450, opened 2023-09-01
contrib/wget2: update to 2.1.0

nekopsykose:
https://gitlab.com/gnuwget/wget2/-/blob/e3f039a4eb2a1de781d51120b5eb6c298a01e1a9/NEWS


contrib/wget2/template.py
```diff
@@ -1,11 +1,12 @@
 pkgname = "wget2"
-pkgver = "2.0.1"
+pkgver = "2.1.0"
 pkgrel = 0
 build_style = "gnu_configure"
 configure_args = [
     "--with-lzma",
     "--with-bzip2",
 ]
+configure_gen = []
```
q66, line 9: ditto

PR #465, opened 2023-09-03
main/samba: update to 4.18.6

nekopsykose:


q66:
4.19.0 is out (probably also gonna need an ldb update)

nekopsykose:
the whole build is effectively -j1 looking at it, i'd guess WAF_MAKE=1 is kinda scuffed

nekopsykose:
could also add `vfs_io_uring` to shared-modules with liburing moved to main

q66:
it's pretty useless and i don't really want to move it right now

PR #480, opened 2023-09-05
contrib/gmic: update to 3.3.0

nekopsykose:


contrib/gmic/template.py
```diff
@@ -33,19 +30,31 @@
...
+sha256 = "9c00a2b4c33ff4879860eadb89f6c836062433a5debdaffbf41f98bfa83081f3"
 # vis broken
 # FIXME int: gmic_qt PreviewWidget::updateOriginalImagePosition
 hardening = ["!int"]
 # no tests
-# breaks in parallel if lib isn't built first by itself
-options = ["!check", "!parallel"]
+options = ["!check"]
+
+tool_flags = {
```
q66, line 40: have you actually checked this 

because unlike in gcc, -O2 and -O3 are practically the same in clang

nekopsykose: tbh mostly just what i remembered from some old gcc tests. goood point and now that i measure it pretty much all the filters are identical in some quick benches.

i also remember that e.g. with gcc13 flac -8 was 30% faster or so with -O3 over -O2, but here it's identical with both (just checked). good to keep in mind :)

PR #482, opened 2023-09-05
main/nghttp2: update to 1.56.0

nekopsykose:
https://nghttp2.org/blog/2023/09/04/nghttp2-v1-56-0/


main/nghttp2/template.py
```diff
@@ -31,3 +31,8 @@ def post_install(self):
 @subpackage("nghttp2-devel")
 def _devel(self):
     return self.default_devel()
+
```
q66, line 34: unless these take a meaningful amount of space, don't

nekopsykose: they're 90% of the package size. we also split progs from most libs in general

nekopsykose: for a similar example, xz and liblzma are split, and there the former is 50% of the latter (100kb to 200kb). the same goes in pretty much almost every case. this is one of the bigger ones where the progs are 1.9MB of the 2MB..

if you want to have a "lol it's 1MB" policy you should probably remove 90% of the default_progs()/default_libs() from the repo because they are almost all like this :D

q66: it's only split where it actually makes a difference (if it adds 200k to the size or something it's generally not split)

but in this case the installed size with progs is 2M and without progs 200k, so it's fine to split

q66:
(one commit is fine for next time)

PR #486, opened 2023-09-05
main/libarchive: also rename bsdcpio/bsdtar mans

nekopsykose:
since we rename (not symlink) the bins

could also symlink these i guess


q66:
yeah fair enough

nekopsykose:
got jumpscared by tar(5) when typing man tar lmao

PR #489, opened 2023-09-05
contrib/whois: new package (5.5.18)

nekopsykose:


contrib/whois/template.py
```diff
@@ -0,0 +1,47 @@
...
+}
+
+
+def init_install(self):
+    self.make_install_args.append(f"INSTALL_ROOT={self.chroot_destdir}")
+
+
+@subpackage("mkpasswd")
+def _mkpasswd(self):
+    # conflicting /usr/bin/mkpasswd
```
nekopsykose, line 42: the manpage also conflicts. not sure what to do about that

the actual mkpasswd (and manpage) are completely different, just happen to have the same name..

q66: i'll probably change expect to do the same as debian does, i.e. call the scripts expect_foo

you can drop the subpackage

contrib/whois/template.py
```diff
@@ -0,0 +1,38 @@
...
+hostmakedepends = [
+    "gettext",
+    "gmake",
+    "perl",
+    "pkgconf",
+]
+makedepends = [
+    "libidn2-devel",
+]
+depends = ["!expect"]
```
q66, line 21: don't need this (even if i didn't resolve the conflict, you get an implicit conflict through two versioned providers of the `cmd:`)

PR #495, opened 2023-09-07
contrib/go: update to 1.21.1

nekopsykose:
https://go.dev/doc/devel/release#go1.21.minor


nekopsykose:
should bump every go revdep along with these security ones

PR #551, opened 2023-09-14
main/linux-rpi: disable evbug

JamiKettunen:
This just spams `dmesg` with every input event from keyboard, mice etc and was already disabled for `linux-lts` and `linux-stable`; excluded `pkgrel` bump for now since it just was built, can add it if preferred already.


q66:
just bump it i guess

PR #554, opened 2023-09-14
main/shadow: enable bcrypt hash support

nekopsykose:
this now shows up in e.g. chpasswd --crypt-method.

unlike glibc, musl supports bcrypt in standard crypt interface, without requiring libxcrypt. it's also generally better to pick than sha512.  

---

unrelatedly, i'd say it's better to pick bcrypt as the default hash for /etc/shadow creations, but i'm not sure how good the compatibility with glibc would be (it doesn't support it, but a libxcrypt-enabled glibc would afaik..), which could be troublesome when using a shadow from chimera on a glibc system (?)

### DIFF
--- a/main/shadow/template.py
+++ b/main/shadow/template.py
@@ -1,6 +1,6 @@
 pkgname = "shadow"
 pkgver = "4.14.0"
-pkgrel = 0
+pkgrel = 1
 build_style = "gnu_configure"
 configure_args = [
     "--enable-shared",
@@ -9,6 +9,7 @@ configure_args = [
     "--with-libpam",
     "--with-acl",
     "--with-attr",
+    "--with-bcrypt",
     "--without-libbsd",
     "--without-selinux",
     "--without-nscd",


nekopsykose:
(it also does work, changing your password with it will give you a $2a in shadow etc. it takes like 1 second for a passwd authentication to happen with that because it is much more hardened than the default (very low) sha hash rounds.)

q66:
i don't think i really care much about using shadow from chimera from a glibc system...

nekopsykose:
that would make sense :)  

i looked into it, there is a ENCRYPT_METHOD in login.defs, but `"This only affect the generation of group passwords. The generation of user passwords is done by PAM"`

for which, [pam_unix.so](https://man7.org/linux/man-pages/man8/pam_unix.8.html) has e.g. `blowfish` et al to pick the actual algorithm. the same goes for `rounds=`, ..

nekopsykose:
even if we don't change to bcrypt/blowfish, i think we should set `rounds=` in pam.d/passwd(? or wherever) because the musl crypt default is very low for sha512 (and cannot be changed for compat reasons)

nekopsykose:
of course, any of this only affects new account creation. and is also orthogonal to this configure toggle

q66:
just patch the login.defs default (which is in shadow patches) and edit files/passwd.pam and that should be enough

q66:
looking at it, you should probably also add a new patch patching `libmisc/salt.c` as it currently does:

```
        if (NULL != meth)
                method = meth;
        else {
                method = getdef_str ("ENCRYPT_METHOD");
                if (NULL == method) {
                        method = getdef_bool ("MD5_CRYPT_ENAB") ? "MD5" : "DES";
                }
        }
```

so that should default to bcrypt i guess?

and also the login.defs comment is wrong, according to changelog ENCRYPTMETHOD_SELECT was removed in 2007 and this has been unconditionally supported since

nekopsykose:
tl;dr from irc- running `passwd` uses `/etc/pam.d/passwd` to read the pam_unix.so blowfish/rounds by linux-pam to then generate the crypt string, but what it generates:

- has a too short 'salt' (the `s` in `$2b$10$s$`) which has to be 22+ but is 16. the bcrypt algorithm uses 16 salt bytes but chars are narrower than bytes, and so base64(16-byte-salt) gives 21.something chars and rounds up to 22 text chars for representation (which is why the musl salt must be 22+, for security reasons..)
  - an aside- the [linux-pam code for the salt is a disaster](https://github.com/linux-pam/linux-pam/blob/77bd338125cde583ecdfb9fd69619bcd2baf15c2/modules/pam_unix/passverify.c#L385), and seemingly says to not raise the output above 16.. it works to just put 22, but we should probably rewrite this to getrandom in the process.
- passes rounds=5 by default (but respects rounds= from the pam_unix.so arg; maybe should raise default anyway?). musl doesn't support rounds= for bcrypt (unlike sha512 where that is the common way to pass it), and this must be passed as a `$X$` value, 0-padded (`08`, `11`, etc, for bcrypt round count).
- in the error path where this fails, it segfaults-
  - [this path](https://github.com/linux-pam/linux-pam/blob/77bd338125cde583ecdfb9fd69619bcd2baf15c2/modules/pam_unix/passverify.c#L473C3-L473C8) appends to the pointer,
  - then it [calls explicit_bzero](https://github.com/linux-pam/linux-pam/blob/77bd338125cde583ecdfb9fd69619bcd2baf15c2/modules/pam_unix/passverify.c#L497) with this offset pointer (UB), musl crashes in the memset

todo:

- [ ] change `pam.d` entries that use pam_unix.so with a password arg to have `blowfish` and `rounds`
- [ ] fix the above (ugh)  
- [ ] patch shadow defaults

richfelker:
> * passes rounds=5 by default (but respects rounds= from the pam_unix.so arg; maybe should raise default anyway?). musl doesn't support rounds= for bcrypt (unlike sha512 where that is the common way to pass it), and this must be passed as a `$X$` value, 0-padded (`08`, `11`, etc, for bcrypt round count).

This is not a "musl does not support". It's that the bcrypt data format does not use the rounds=n form that shacrypt does. This is a standard data format we have to adhere to not something we can redefine at will.

PR #584, opened 2023-09-20
cbuild: use `meson test`/`meson install` directly

nekopsykose:
this lets us directly pass args to meson for test/install that aren't possible otherwise via the `ninja` call
- for install, this is useful to skip a layer of indirection as ninja calls back out to `meson install` and lets perhaps --skip-subprojects be used, etc
- for check, this allows passing extra args later such as -t and friends

### DIFF
--- a/src/cbuild/build_style/meson.py
+++ b/src/cbuild/build_style/meson.py
@@ -10,11 +10,11 @@ def do_build(self):
 
 
 def do_check(self):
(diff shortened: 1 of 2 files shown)


nekopsykose:
theoretically could also replace build with `meson compile`, but that one seems useless since it does not have any args that could get meaningfully used

PR #591, opened 2023-09-20
contrib/gas: new package (2.41)

### LINKED ISSUE #582: [Question] Is there gas or similar assembler available ?
I'm trying to compile a program on chimera and it complains that it needs an assembler `as`. I'm guessing it needs something like a gnu assembler which I don't think is available in chimera. Is there something like gas or equivalent provided by llvm toolchain available in chimera ?

nekopsykose:
closes #582 

nothing else seems to provide /usr/bin/as in general, so even though almost nothing will use this, it doesn't really hurt to have it where it's really required for something particular

### DIFF
--- a/contrib/gas/template.py
+++ b/contrib/gas/template.py
@@ -1,0 +1,44 @@
+pkgname = "gas"
+pkgver = "2.41"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--disable-gold",
+    "--disable-gprof",
+    "--disable-gprofng",
+    "--disable-ld",
+    "--disable-multilib",
+    "--disable-nls",
+    "--disable-shared",
+    "--disable-werror",
+    "--with-system-zlib",
+    "--with-zstd",
+]
+# require a very specific autotools
+configure_gen = []
+make_cmd = "gmake"
+hostmakedepends = [
+    "gmake",
+    "pkgconf",
+    "texinfo",
+]
+makedepends = [
+    "zlib-devel",
+    "zstd-devel",
+]
+pkgdesc = "GNU Assembler"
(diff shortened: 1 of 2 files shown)


q66:
wat? what's the point of this instead of providing well, a binutils package

nekopsykose:
nothing else from the binutils is useful, since ar/nm/ranlib already exist (in 2 flavors, no less), the linker is useless, and gprof doesn't work anyway

nekopsykose:
the only thing of interest is really that:  

- `elfedit` doesn't exist in other tooling
- `llvm-objdump` doesn't have an `objdump` symlink technically

but all the other tools exist  

and, there is also `libbfd`, which is used for.. perf. but technically that is not redistributable (debian refuses to because of gpl3+2 mix; although i think this is stupid as nobody cares + it's way better than shelling to addr2line and is like 100x faster for address lookup), and we don't have perf yet

q66:
the binutils tools have a large amount of custom extensions that don't always exist or don't exist in the same form in the llvm tools; same applies to the linker

this already used to be packaged: https://github.com/chimera-linux/cports/commit/0cc2f0b795363fd8547203cfb51d1c6f49ed735d

nekopsykose:
> the binutils tools have a large amount of custom extensions that don't always exist or don't exist in the same form in the llvm tools; same applies to the linker  

and in practice, the only thing i found that is actually required for something with no replacement or easy workaround is gas itself, like in the requested issue.

q66:
there is also e.g. the win32 stuff needed for gnu-efi things, u-boot in practice requires it (which also needs gcc and is currently worked around with the baremetal toolchain, but i don't want it to remain that way forever), etc

PR #609, opened 2023-09-21
contrib/swww: new package (0.8.1)

Froggo8311:


contrib/swww/template.py
```diff
@@ -0,0 +1,12 @@
+pkgname = "swww"
+pkgver = "0.8.1"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = [ "cargo" ]
+makedepends = [ "rust" ]
```
wezm, line 6: CI is waiting on approval but the [`black`](https://pkgs.chimera-linux.org/package/current/main/x86_64/python-black) job will fail these lines. You'll need to format the template with it.

JamiKettunen:
Please squash the format commit, you can e.g. `git rebase -i origin/master` and mark it as `f` for `fixup` (instead of default `pick`) which will merge it to the commit above (`contrib/swww: add package`)

q66:
as @JamiKettunen said, commits need squashing (ditto for every other PR)

Froggo8311:
@q66 Why not do it through github?

q66:
because 1) you're the one who wants it merged, so you should be the one to put in the effort to make sure things are in a correct format 2) the commits are expected to have correct commit messages 3) it may not always be merged through github 4) ...

Froggo8311:
Squashed & rebased

q66:
fix commit message

Froggo8311:
fixed

PR #613, opened 2023-09-21
contrib/hyprland: new package (0.29.1)

Froggo8311:


contrib/hyprland/template.py
```diff
@@ -0,0 +1,49 @@
...
+makedepends = [
+    "cairo-devel",
+    "hwdata-devel",
+    "libavcodec",
+    "libavformat",
+    "libavutil",
+    "libdisplay-info-devel",
+    "libdrm-devel",
+    "libgbm-devel",
+    "libgles2",
```
nekopsykose, line 22:
```suggestion
    "ffmpeg-devel",
    "libdisplay-info-devel",
    "libdrm-devel",
    "libgbm-devel",
```

Froggo8311: That also works, thanks. I forgot to remove gles after debugging and didn't know there was an `ffmpeg-devel` package that provided the others

q66:
no

jiripospisil:
Why?

jiripospisil:
Huh? You're the one who opened the PR in the first place. And yeah, if you're going to exclude one of the most popular Wayland compositor, something more than "no" would have been great.

q66:
hyprland's case is a very particular combo of "made by very loudly bigoted clowns" and "is a terrible codebase whose author displays a total lack of fundamentals"

neither thing is particularly uncommon in foss/tech (sadly) and it's not possible to check/verify everything, but since we're dealing with a particular combo of these where it's really visible and obvious, i see no reason to let it in

i think you're also overestimating its actual popularity (not that it would have any effect on the decision)

so there you go

PR #614, opened 2023-09-22
contrib/solaar: update to 1.1.10, properly install udev rules

JamiKettunen:
To use the program as non-root user it needs udev rules which were already installed but in a "useless" location from distro perspective.
image

### DIFF
--- a/contrib/solaar/template.py
+++ b/contrib/solaar/template.py
@@ -27,3 +27,12 @@ source = (
 sha256 = "40887b508e4cfc753e5c2e82caa4af7f057cadad4a32f238f9aef898b8ccfb2c"
 # no tests
 options = ["!check"]
+
+
+def post_install(self):
+    self.install_dir("usr/lib/udev/rules.d")
+    self.mv(
+        self.destdir
+        / "usr/share/solaar/udev-rules.d/42-logitech-unify-permissions.rules",
+        self.destdir / "usr/lib/udev/rules.d",
+    )
--- a/contrib/solaar/template.py
+++ b/contrib/solaar/template.py
@@ -1,8 +1,13 @@
 pkgname = "solaar"
-pkgver = "1.1.9"
+pkgver = "1.1.10"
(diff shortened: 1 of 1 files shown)


contrib/solaar/template.py
```diff
@@ -20,3 +20,12 @@
     f"https://github.com/pwr-Solaar/Solaar/archive/refs/tags/{pkgver}.tar.gz"
 )
 sha256 = "3cb0cb4c79ef2ea0ac7c9adc9c4a6f4f15d28e99ff2df33850de0520ced9f116"
+
+
+def post_install(self):
+    self.mkdir(self.destdir / "usr/lib/udev/rules.d", parents=True)
```
q66, line 26: use install_dir

JamiKettunen: right forgot about that

contrib/solaar/template.py
```diff
@@ -1,22 +1,34 @@
...
     "python-evdev",
     "python-pyudev",
     "python-psutil",
     "python-pyyaml",
     "python-xlib",
     "python-gobject",
+    "python-dbus",
 ]
-checkdepends = list(depends)
+checkdepends = ["python-pytest"] + depends
```
q66, line 15: if it has no tests, what's the deal with this stuff

JamiKettunen: well I saw
```
=> solaar-1.1.10-r0: running do_check...
=> solaar-1.1.10-r0: ERROR: pytest not found
```
but I'll drop it since `!check` alone avoids it anyway

q66: just remove the checkdepends line entirely

q66: also removing setuptools from hostdeps is wrong

JamiKettunen: just noticed setuptools in hostdeps wasn't needed anymore to even "cross-build" which is why I tried that

PR #631, opened 2023-09-27
contrib/thinkfan: new package (1.3.1) ← dep: contrib/yaml-cpp: new package (0.8.0)

valpackett:
My first packages :wave: 

https://github.com/vmatare/thinkfan and its dependency https://github.com/jbeder/yaml-cpp

### DIFF
--- a/contrib/thinkfan/files/thinkfan
+++ b/contrib/thinkfan/files/thinkfan
@@ -1,0 +1,8 @@
+# thinkfan service
+
+type            = process
+command         = /usr/bin/thinkfan -n -q
+before          = login.target
+depends-on      = local.target
+restart         = true
+smooth-recovery = true
--- a/contrib/thinkfan/template.py
+++ b/contrib/thinkfan/template.py
@@ -1,0 +1,20 @@
+pkgname = "thinkfan"
+pkgver = "1.3.1"
+pkgrel = 0
+build_style = "cmake"
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+makedepends = ["yaml-cpp-devel"]
+pkgdesc = "Simple fan control program"
+maintainer = "Val Packett <val@packett.cool>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/vmatare/thinkfan"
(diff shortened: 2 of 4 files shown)


contrib/thinkfan/files/thinkfan
```diff
@@ -0,0 +1,9 @@
+# thinkfan service
+
+type            = bgprocess
```
q66, line 3: can we make this a process service instead?

contrib/thinkfan/template.py
```diff
@@ -0,0 +1,19 @@
+pkgname = "thinkfan"
+pkgver = "1.3.1"
+pkgrel = 0
+build_style = "cmake"
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+makedepends = ["yaml-cpp-devel"]
+pkgdesc = "Simple, lightweight fan control program"
```
q66, line 7: just "Lightweight fan control program" should be descriptive enough

contrib/yaml-cpp/template.py
```diff
@@ -0,0 +1,22 @@
+pkgname = "yaml-cpp"
+pkgver = "0.8.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DYAML_BUILD_SHARED_LIBS=ON"]
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+pkgdesc = "YAML parser and emitter in C++ matching the YAML 1.2 spec"
```
q66, line 7: i'd make this terser, like "C++ YAML parser and emitter"

q66:
outside of these, please fix the lint errors

contrib/thinkfan/files/thinkfan
valpackett: kinda, it wouldn't write to syslog in that case but I guess that's fine

q66: can always just log to a file

valpackett:
done (just silenced the logging as it's not useful)

q66:
lint fails

PR #665, opened 2023-10-02
cbuild: always call apk with --no-interactive

JamiKettunen:
Otherwise existance of host `/etc/apk/interactive` could affect `cbuild`.

This might be too naïve of a fix, not sure yet and needs proper testing.


JamiKettunen:
Oops #663 already existed, closing as a duplicate

PR #701, opened 2023-10-06
main/linux-headers: update to 6.5.5

nekopsykose:
this allows using new defines for software that makes use of them. one example is https://codeberg.org/dnkl/foot/commit/61eb56dfda605427156b52d019be4bf5ce84c3aa ; compiling with 6.3+ headers would allow it to be set, but is still backwards compatible because unknown flags are just ignored.

in general, the headers are backwards compatible. having older headers doesn't improve compatibility with older kernels; it only prevents applications from using newer features/defines, without benefit to people using the older LTS series (it works either way), and with a negative to people using latest stable (loss of potential feature).

### DIFF
--- a/main/linux-headers/template.py
+++ b/main/linux-headers/template.py
@@ -1,14 +1,14 @@
 pkgname = "linux-headers"
-pkgver = "6.1.19"
+pkgver = "6.5.5"
 pkgrel = 0
 make_cmd = "gmake"
 hostmakedepends = ["gmake", "perl"]
 pkgdesc = "Linux API headers for userland development"
 maintainer = "q66 <q66@chimera-linux.org>"
 license = "GPL-2.0-only"
-url = "http://www.kernel.org"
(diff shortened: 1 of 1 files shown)


main/linux-headers/template.py
```diff
@@ -8,7 +8,7 @@
 license = "GPL-2.0-only"
 url = "http://www.kernel.org"
 source = f"$(KERNEL_SITE)/kernel/v{pkgver[0]}.x/linux-{pkgver}.tar.xz"
```
sidkshatriya, line 10: Why not use secure protocol i.e. https://www.kernel.org vs http:// ? Yes, there is a sha256 afterwards which is great but probably nice to use HTTPS always if available ?

nekopsykose: (see `sites` in `src/cbuild/core/template.py` for those values)

PR #705, opened 2023-10-06
main/libxml2: update to 2.11.5

nekopsykose:


nekopsykose:
not sure if it was intentional or not to not update this ever since adding it and if i missed some secret reason, aside from the update-check not functioning

q66:
the only reason was uniquely broken update-check (returning versions i.e. not complaining but not returning current versions) so it went under the radar

q66:
is there any reason for the subpackage shuffle

nekopsykose:
devel gets the remainder of usr/share/doc/libxml2 after python gets libxml2/python (before they were separate dirs)

PR #760, opened 2023-10-14
main/shaderc: update to 2023.7

nekopsykose:
https://github.com/google/shaderc/blob/0d6f72f3ec57fe68377363d9e810385e6b6e37e1/CHANGES


q66:
there is no need to rebuild anything

PR #768, opened 2023-10-14
contrib/tailscale: new package (1.50.1)

valpackett:
and more golang flag stuff, similar to [what other distros do](https://shibumi.dev/posts/hardening-executables/) — actually ideally we'd also handle annoying go ldflags thing to be able to pass both `linkmode=external` and per-package stuff like those `-X` defines; not sure what the right way to do it would be though. or maybe I shouldn't touch any of that stuff…

### DIFF
--- a/contrib/tailscale/template.py
+++ b/contrib/tailscale/template.py
@@ -1,0 +1,26 @@
+pkgname = "tailscale"
+pkgver = "1.50.1"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    f"-ldflags=-X tailscale.com/version.longStamp={pkgver} -X tailscale.com/version.shortStamp={pkgver}"
+]
+hostmakedepends = ["go"]
+depends = ["iptables", "ca-certificates"]
+pkgdesc = "Mesh VPN daemon based on WireGuard"
+maintainer = "Val Packett <val@packett.cool>"
+license = "BSD-3-Clause"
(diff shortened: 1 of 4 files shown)


contrib/tailscale/template.py
```diff
@@ -0,0 +1,26 @@
...
+]
+hostmakedepends = ["go"]
+depends = ["iptables", "ca-certificates"]
+pkgdesc = "Mesh VPN daemon based on WireGuard"
+maintainer = "Val Packett <val@packett.cool>"
+license = "BSD-3-Clause"
+url = "https://github.com/tailscale/tailscale"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "183a7d559590a759dd77aa9c2b65486ab6e13c26f3c07fad0b536e318ad5e233"
+options = ["!strip"]
```
nekopsykose, line 16: !debug instead

nekopsykose:
trimpath looks ok, not sure why it's gated on debug==0 though. also for some projects it breaks `go test`, so ideally it's not set for do_check since it wouldn't matter there anyway

valpackett:
Not trimming for debug builds was suggested in various places but yeah I can't see why the full paths would ever help (?)

Would it be better to set it in `build()` next to `"-v"`?

q66:
all packages are for all intents and purposes release builds, debug just controls the presence of debug info for otherwise release packages

PR #773, opened 2023-10-15
contrib/protobuf: make -devel subpkg depend on protobuf (protoc)

JamiKettunen:
The protobuf compiler itself is required for development so this just makes sense, it was not picked up previously due to `libprotoc` being split into its own subpackage and main pkg left with the tiny `protoc` executable.

### DIFF
--- a/contrib/protobuf/template.py
+++ b/contrib/protobuf/template.py
@@ -1,6 +1,6 @@
 pkgname = "protobuf"
 pkgver = "24.4"
-pkgrel = 0
+pkgrel = 1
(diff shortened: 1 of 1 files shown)


q66:
the compiler is a host library dependency, the library is a target dependency; i don't particularly see a reason to link them together like that

JamiKettunen:
I was mostly thinking you build somthing locally outside cports, in which case you'd have to `apk add protobuf-devel protobuf` to get everything required for development, but whichever you prefer, this isn't really critical by any means

PR #788, opened 2023-10-17
contrib/greetd: new package (0.9.0)

Natthias:
This is my first time attempting to upstream a package, hopefully I get this right.

This MR also provides a contrib/tuigreet package, which can be used as a greeter for greetd. I also don't know if I got the dinit service right, but it works locally.

Both packages have been tested, and built on x86_64, as it is the only hardware I have. All tier 1 and tier 3 architectures cross compile successfully.

### DIFF
--- a/contrib/greetd/greetd.post-install
+++ b/contrib/greetd/greetd.post-install
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+# install greetd home dir
+install -d -m 755 /var/lib/greetd || :
+chown _greetd:_greetd /var/lib/greetd > /dev/null 2>&1 || :
--- a/contrib/greetd/patches/config.patch
+++ b/contrib/greetd/patches/config.patch
@@ -1,0 +1,17 @@
+--- a/config.toml
++++ b/config.toml
+@@ -1,7 +1,7 @@
+ [terminal]
+ # The VT to run the greeter on. Can be "next", "current" or a number
+ # designating the VT.
+-vt = 1
++vt = 7
+ 
+ # The default session, also known as the greeter.
+ [default_session]
+@@ -13,4 +13,4 @@
+ # The user to run the command as. The privileges this user must have depends
+ # on the greeter. A graphical greeter may for example require the user to be
+ # in the `video` group.
+-user = "greeter"
++user = "_greetd"
--- a/contrib/greetd/template.py
+++ b/contrib/greetd/template.py
@@ -1,0 +1,48 @@
+pkgname = "greetd"
+pkgver = "0.9.0"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = [
+    "bmake",
+    "cargo",
+    "pkgconf",
+    "scdoc",
+]
+makedepends = [
+    "linux-pam-devel",
+    "rust-std",
+]
+pkgdesc = "Minimal and flexible login manager daemon"
+maintainer = "natthias <natthias@proton.me>"
+license = "GPL-3.0-or-later"
+url = "https://git.sr.ht/~kennylevinsen/greetd"
+source = f"https://git.sr.ht/~kennylevinsen/greetd/archive/{pkgver}.tar.gz"
+sha256 = "a0cec141dea7fd7838b60a52237692d0fd5a0169cf748b8f8379d8409a3768eb"
+
+system_users = [
+    {
+        "name": "_greetd",
+        "id": None,
+        "home": "/var/lib/greetd",
+    }
+]
+
+
+def post_build(self):
+    self.do("make", "-C", "man", "all")
+
+
+def do_install(self):
+    self.install_bin(f"target/{self.profile().triplet}/release/agreety")
+    self.install_bin(f"target/{self.profile().triplet}/release/fakegreet")
+    self.install_bin(f"target/{self.profile().triplet}/release/greetd")
+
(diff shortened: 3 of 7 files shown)


contrib/greetd/greetd.post-install
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+# install greeter home dir
+install -d -m 755 /var/lib/greeter || :
+chown _greeter:_greeter /var/lib/greeter > /dev/null 2>&1 || :
```
nekopsykose, line 5: everything here is already done automatically by having system_users with this set, no?

Natthias: Afaict, no. I initially didn't have this, but the home dir was not created, so I did what gdm does

contrib/greetd/patches/config.patch
```diff
@@ -0,0 +1,17 @@
+--- a/config.toml	2023-08-29 22:25:37.253885597 +0200
++++ b/config.toml	2023-08-29 22:24:23.032078385 +0200
+@@ -1,7 +1,7 @@
+ [terminal]
+ # The VT to run the greeter on. Can be "next", "current" or a number
+ # designating the VT.
+-vt = 1
++vt = 7
```
q66, line 8: is there a reason not to use the default?

contrib/greetd/template.py
```diff
@@ -0,0 +1,53 @@
...
+    "rust-std",
+]
+pkgdesc = "Minimal and flexible login manager daemon"
+maintainer = "natthias <natthias@proton.me>"
+license = "GPL-3.0-or-later"
+url = "https://git.sr.ht/~kennylevinsen/greetd"
+source = f"https://git.sr.ht/~kennylevinsen/greetd/archive/{pkgver}.tar.gz"
+sha256 = "a0cec141dea7fd7838b60a52237692d0fd5a0169cf748b8f8379d8409a3768eb"
+
+system_users = [
```
q66, line 22: i'm not sure i like the choice of user name, i'd call it `_greetd` instead

contrib/greetd/template.py
```diff
@@ -0,0 +1,53 @@
...
+    {
+        "name": "_greeter",
+        "id": None,
+        "groups": ["video"],
```
q66, line 26: is there a reason to do that? shouldn't a graphical greeter use elogind/libseat to request the device access? using the video group is a big hammer (it gives access to the gpu device node)

contrib/greetd/template.py
```diff
@@ -0,0 +1,53 @@
...
+    }
+]
+
+
+def post_build(self):
+    self.do("make", "-C", "man", "all")
+
+
+def init_install(self):
+    self.env["RUST_TARGET"] = self.profile().triplet
```
q66, line 37: this doesn't seem like it actually does anything

contrib/tuigreet/template.py
```diff
@@ -0,0 +1,16 @@
...
+pkgver = "0.8.0"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = [
+    "cargo",
+    "scdoc",
+]
+makedepends = ["rust-std"]
+depends = ["greetd"]
+pkgdesc = "Graphical console greeter for greetd"
```
q66, line 11: so is it graphical, or console? :) the name implies it's console and no graphical stuff is involved, but the description says otherwise?

contrib/greetd/greetd.post-install
q66: `useradd` does not create a homedir for system users (`-r`)

contrib/greetd/patches/config.patch
nekopsykose: if https://man.sr.ht/~kennylevinsen/greetd/#changing-target-vt is followed, i assume starting on 1 (or 1-6 rather) is already claimed by getty. i don't think it breaks, but '7' is kind of an unspecified default for spawned 'login' things?

that said, not sure why the default is left at 1 in that case

q66: in that case yeah (chimera runs gettys at 1-6 by default)

q66:
Besides the review comments, I wonder if there is some kind of mechanism we could use to implement readiness notification (that the session has come up) for the dinit service.

contrib/greetd/template.py
Natthias: Yea, it should, I just followed upstreams reccomendation, as well as what other distros do. Should I use elogind or libseat to manage this?

q66: greetd itself doesn't need it (it doesn't do anything with the gpu), this is only relevant to graphical greeters, and at least e.g. gtkgreet expects you to run it in a wayland compositor, which will use logind/seatd

Natthias: Yup, but with libseatd, the _greetd user would need the _seatd group, no?

q66: depends on the libseat backend; for logind you don't need anything (it uses dbus to communicate), for seatd you need to be in the group (it uses a unix domain socket to communicate, owned by `root:_seatd`)

q66: for now it's probably fine not to add it to any groups as elogind is the recommended stock solution, when it's time to kill elogind we'll need to figure something out

PR #794, opened 2023-10-18
*/*: consistentise +depends to +list(depends)

nekopsykose:


q66:
this doesn't really fix anything nor makes anything more consistent

nekopsykose:
it's what you said should be done with `checkdepends = list(depends)` instead of `checkdepends = depends`

q66:
if you're concatenating then it's pointless to make a copy since it'll just get discarded right away

PR #795, opened 2023-10-18
main/python-black: update to 23.10.0

nekopsykose:


q66:
drop the experimental/ stuff, it's just noise

PR #803, opened 2023-10-19
main/libbytesize: update to 2.10

nekopsykose:
https://github.com/storaged-project/libbytesize/blob/6e83cc6f6dff4f126fc79284e0c3c1c50123380d/NEWS.rst


q66:
python-six dependency was dropped according to the changelog above

PR #805, opened 2023-10-19
contrib/sysstat: new package (12.7.4)

nekopsykose:


q66:
these date invocations are trivial to fix, so just fix them (`-v -1d`, see `man 1 date`)

PR #815, opened 2023-10-20
contrib/gptfdisk: new package (1.0.9)

reocat:


main/gdisk/patches/makefile.patch
```diff
@@ -0,0 +1,10 @@
+--- Makefile	2023-10-21 00:13:47.131679080 +0300
++++ Makefile	2023-10-21 00:46:48.279265063 +0300
+@@ -162,6 +162,8 @@
+ lint:	#no pre-reqs
+ 	lint $(SRCS)
+ 
++check: 	#empty 
++
```
nekopsykose, line 8: if there's no check just do `options = ["!check"]` and you don't need this

main/gdisk/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "gdisk"
```
nekopsykose, line 1: the actual project name is `gptfdisk`.

nekopsykose:
you most likely want the 00 patches from https://git.alpinelinux.org/aports/tree/main/gptfdisk?id=84a227baf001b6e0208e3352b294e4d7a40e93de too since they fix crashes with newer libraries here

main/gdisk/patches/makefile.patch
nekopsykose: also it seems like the test is in `builddir/gdisk_test.sh`, so you can run that from `self.do` when overriding `do_check`.

nekopsykose: ah, there is a target for it. you can do `make_check_target = "test"`

nekopsykose: (but that runs it with bash, when you can manually call `sh` on it and it probably works, if you want to skip the dep)

reocat: > if there's no check just do `options = ["!check"]` and you don't need this

ERROR: lint failed: check disabled but no reason given

q66:
please keep your changes squashed to one commit per template

q66:
use interactive rebase to squash all the commits together

also format the template with `black` and make sure it passes `flake8`

q66:
perhaps you want to read https://github.com/chimera-linux/cports/blob/master/USING_GIT.md

q66:
also, this package should go to contrib/, not main/

reocat:
> also, this package should go to contrib/, not main/

done too

q66:
CI fails

reocat:
> CI fails

Check is disabled now, should work better

contrib/gptfdisk/template.py
```diff
@@ -0,0 +1,28 @@
...
+    "ncurses-devel",
+    "popt-devel",
+]
+pkgdesc = "Fdisk-like partitioning tool for GPT disks"
+maintainer = "reocat <ng.ct_ml@tuta.io>"
+license = "GPL-2.0-or-later"
+url = "https://sourceforge.net/projects/gptfdisk"
+source = f"$(SOURCEFORGE_SITE)/{pkgname}/{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "dafead2693faeb8e8b97832b23407f6ed5b3219bc1784f482dd855774e2d50c2"
+# no check provided
```
q66, line 20: this is a lazy way out

cosnidering gptfdisk provides a test suite and it's capable of mostly passing, you should patch out the part that is not relevant for the test environment and let the rest run

reocat: Should be better now

reocat: But I can't check it in GitHub CI

reocat:
Ah, forgot to remove doas while tested some stuff, should be better now

q66:
I looked into the failing tests here and it seems they're just flaky (i guess it's missing some syncs) so it's fine to keep the offending ones patched out

contrib/gptfdisk/template.py
```diff
@@ -0,0 +1,34 @@
+pkgname = "gptfdisk"
+pkgver = "1.0.9"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_dir = "."
+make_check_target = "test"
+hostmakedepends = [
```
q66, line 8: all of these except `gmake` should be `makedepends`

contrib/gptfdisk/template.py
```diff
@@ -0,0 +1,34 @@
...
+    "linux-headers",
+    "libuuid-devel",
+    "ncurses-devel",
+    "popt-devel",
+]
+pkgdesc = "Fdisk-like partitioning tool for GPT disks"
+maintainer = "reocat <ng.ct_ml@tuta.io>"
+license = "GPL-2.0-or-later"
+url = "https://sourceforge.net/projects/gptfdisk"
+source = f"$(SOURCEFORGE_SITE)/{pkgname}/{pkgver}/{pkgname}-{pkgver}.tar.gz"
```
q66, line 19: you don't have to substitute pkgname here

q66:
fix the remaining two comments i raised, and then it's lgtm

contrib/gptfdisk/template.py
```diff
@@ -0,0 +1,34 @@
...
+sha256 = "dafead2693faeb8e8b97832b23407f6ed5b3219bc1784f482dd855774e2d50c2"
```
q66, line 20: also add `hardening = ["vis", "cfi"]`

PR #825, opened 2023-10-22
main/imagemagick: update to 7.1.1.21

nekopsykose:
https://github.com/ImageMagick/ImageMagick/releases/tag/7.1.1-21


q66:
please update the update-check to skip the weird version

nekopsykose:
yeah, just forgot how to make it take `7.*` so it skips all the dual-numerical ones (they keep adding e.g. 77,78,79, ..., so will have to keep skipping those)

q66:
just extend the existing `6*` ignore like `[67]*`

nekopsykose:
that ignores all the actual 7.xyz versions too

q66:
`[67][0-9]` then

PR #833, opened 2023-10-25
contrib/sbctl: new package (0.12)

flukeeey:
I use this tool to manage my Secure Boot setup.

Default path for keys and database files is `/usr/share/secureboot`, some distros change this to `/etc/secureboot`.

### DIFF
--- a/contrib/sbctl/template.py
+++ b/contrib/sbctl/template.py
@@ -1,0 +1,58 @@
+pkgname = "sbctl"
+pkgver = "0.12"
+pkgrel = 0
+build_style = "go"
+make_build_args = ["./cmd/sbctl"]
+hostmakedepends = ["go", "asciidoc", "gmake"]
+depends = [
+    "llvm-binutils",  # required to generate EFI bundles
+]
+pkgdesc = "Secure Boot key manager"
+maintainer = "flukey <flukey@vapourmail.eu>"
+license = "MIT"
+url = "https://github.com/Foxboron/sbctl"
+source = f"{url}/releases/download/{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "72475519733385b0effb422ff058e95f0282fa93eccf03df07e3402736646b97"
+options = ["!cross"]
+
+
+def post_build(self):
+    # Generate man page, bmake doesn't work
+    self.do("gmake", "man")
+
+    # Generate bash completions
+    with open(self.cwd / "sbctl.bash", "w") as cf:
+        self.do(
+            self.make_dir + "/sbctl",
+            "completion",
+            "bash",
+            stdout=cf,
+        )
+
+    # Generate zsh completions
+    with open(self.cwd / "sbctl.zsh", "w") as cf:
+        self.do(
+            self.make_dir + "/sbctl",
+            "completion",
+            "zsh",
+            stdout=cf,
+        )
+
+    # Generate fish completions
+    with open(self.cwd / "sbctl.fish", "w") as cf:
+        self.do(
+            self.make_dir + "/sbctl",
+            "completion",
+            "fish",
+            stdout=cf,
+        )
+
+
+def post_install(self):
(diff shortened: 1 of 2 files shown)


q66:
doesn't this depend on systemd-boot efi stubs?

flukeeey:
> doesn't this depend on systemd-boot efi stubs?

Generating EFI bundles requires the systemd EFI stub, yeah, but none of the other functionality does afaik.

q66:
we have the stubs though, so the dependency should be there for full functionality (it only adds ~200k)

flukeeey:
Oops missed that!

I installed `systemd-boot-efi` and tried to generate a bundle, but was met with a `objcopy: file format not recognized` error. I also tried with the stub from my Arch install, same result.

Is objcopy from `elftoolchain` behaving doing something different to what is standard? Relevant code in sbctl is here: https://github.com/Foxboron/sbctl/blob/748bc59d24a4d14087d8d92431606c5fb57afeb0/bundles.go#L197

Command I'm using for reference:

```
doas sbctl bundle \
  --amducode /usr/lib/firmware/amd-ucode.img \
  --cmdline /etc/kernel/cmdline \
  --initramfs /boot/initrd.img-6.1.59-hardened1-4-custom \
  --kernel-img /boot/vmlinuz-6.1.59-hardened1-4-custom \
  --os-release /etc/os-release \
  --efi-stub /usr/lib/systemd/boot/efi/linuxx64.efi.stub \
  /boot/EFI/Linux/chimera.efi
```

q66:
to what is "standard"? there is no standard

i assume this is just it expecting support for the baremetal efi target in the utils, which is not present

you can try patching it to use llvm-objcopy

flukeeey:
OK, will try to take a look.

The args seem compatible except for `--change-section-vma`, which doesn't exist in llvm-objcopy. There's `--set-start`, but the format is different.

q66:
@flukeeey you may find this useful: https://github.com/systemd/systemd/commit/789a642738d28cf2a8ad3f65df9c0c136e83af09

flukeeey:
> @flukeeey you may find this useful: [systemd/systemd@789a642](https://github.com/systemd/systemd/commit/789a642738d28cf2a8ad3f65df9c0c136e83af09)

Nice! I'll test this out.

q66:
ping

flukeeey:
Created a patch using the diff you provided. Built and works fine for me, I can generate, sign, and boot from bundles.

Generating bundles requires pulling in llvm-binutils, which seems like a big dependency for optional functionality maybe?

Also, maybe it would be good to add a script to /etc/kernel.d that could sign new kernels for Secure Boot, generate bundles, etc. But how would the user configure this?

Bundles/files-to-sign can be persisted to the sbctl database using the `--save` option, and then regenerated on kernel addition/update using `sbctl generate-bundles` and `sbctl sign-all` subcommands, but the problem is that the filenames of the kernels/initrd images change each time, so this wouldn't work.

Maybe we could just include a sample script somewhere, and let the user customise and add it to /etc/kernel.d if they want to?

contrib/sbctl/template.py
```diff
@@ -0,0 +1,60 @@
...
+depends = [
+    "llvm-binutils",  # required to generate EFI bundles
+]
+pkgdesc = "Secure Boot key manager"
+maintainer = "flukey <flukey@vapourmail.eu>"
+license = "MIT"
+url = "https://github.com/Foxboron/sbctl"
+source = f"{url}/releases/download/{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "72475519733385b0effb422ff058e95f0282fa93eccf03df07e3402736646b97"
+
```
q66, line 15: add `options = ["'!cross"]` since you are invoking built artifacts

q66:
llvm-binutils dep is okay here (it's not that big on its own)

let's not do the kernel hook stuff for now (i want to figure out something for systemd-boot first)

PR #870, opened 2023-11-04
contrib/nerdctl: update to 1.7.0

nekopsykose:


wezm:
Does this need a dependency on `iptables`? Didn't work until I installed it:

Before `apk add iptables`:

```
FATA[0009] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: time="2023-10-19T20:36:06+10:00" level=fatal msg="failed to call cni.Setup: plugin type=\"bridge\" failed (add): failed to locate iptables: exec: \"iptables\": executable file not found in $PATH"
```

Or maybe it's `runc` that needs the dep?

nekopsykose:
> Or maybe it's runc that needs the dep?  

no, this does the networking

> Does this need a dependency on iptables? Didn't work until I installed it:

by default it does FirewallStuff:tm: but personally i do --network=host and then it doesn't need it

nekopsykose:
then again, i did put depends=runc in containerd even though i don't use that either (`--runtime=crun`), so sure, i guess it makes sense for it to be there

PR #876, opened 2023-11-05
contrib/audacity: update to 3.4.0

nekopsykose:
the new opus import/export doesn't actually work and just fails to load mod-opus with ENOTTY. strace doesn't say anything interesting because every single module load does ioctl TIOCGWINSZ on the cfg file fd which fails with that, but only mod-opus fails to load..


nekopsykose:
ah, that's because mod-opus nor nothing in the chain links libopusfile so the syms are missing

nekopsykose:
fixed

PR #903, opened 2023-11-12
contrib/patchutils: new package (0.4.2)

reocat:


contrib/patchutils/template.py
```diff
@@ -0,0 +1,23 @@
...
+source = (
+    f"http://cyberelk.net/tim/data/patchutils/stable/patchutils-{pkgver}.tar.xz"
+)
+sha256 = "8875b0965fe33de62b890f6cd793be7fafe41a4e552edbf641f1fed5ebbf45ed"
+# 6 tests fail
+options = ["!check"]
+
+
+def do_configure(self):
+    self.do(self.chroot_cwd / "configure", "--prefix=/usr")
```
nekopsykose, line 23: build_style = configure does this for you

contrib/patchutils/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "patchutils"
+pkgver = "0.4.2"
+pkgrel = 2
```
nekopsykose, line 3: pkgrel starts at 0

contrib/patchutils/template.py
```diff
@@ -0,0 +1,20 @@
+pkgname = "patchutils"
+pkgver = "0.4.2"
+pkgrel = 0
+build_style = "configure"
```
q66, line 4: is gnu_configure

contrib/patchutils/template.py
```diff
@@ -0,0 +1,20 @@
...
+configure_args = ["--prefix=/usr"]
```
q66, line 5: is unnecessary with gnu_configure

contrib/patchutils/template.py
```diff
@@ -0,0 +1,20 @@
...
+make_cmd = "gmake"
+make_dir = "."
+hostmakedepends = ["gmake"]
+makedepends = ["xmlto"]
+depends = ["perl"]
+pkgdesc = "Collection of programs for manipulating patch files"
+maintainer = "reocat <ng.ct_ml@tuta.io>"
+license = "GPL-2.0-or-later"
+url = "https://cyberelk.net/tim/patchutils"
```
q66, line 14: url doesn't work over https

contrib/patchutils/template.py
```diff
@@ -0,0 +1,20 @@
...
+source = (
+    f"http://cyberelk.net/tim/data/patchutils/stable/patchutils-{pkgver}.tar.xz"
+)
+sha256 = "8875b0965fe33de62b890f6cd793be7fafe41a4e552edbf641f1fed5ebbf45ed"
+# 6 tests fail
```
q66, line 19: tests should be patched not to fail

contrib/patchutils/template.py
```diff
@@ -0,0 +1,20 @@
+pkgname = "patchutils"
+pkgver = "0.4.2"
+pkgrel = 0
+build_style = "configure"
+configure_args = ["--prefix=/usr"]
+make_cmd = "gmake"
+make_dir = "."
+hostmakedepends = ["gmake"]
+makedepends = ["xmlto"]
```
q66, line 9: is a runnable program build-time dependency, thus hostmakedepends

q66: also likely want to build with pcre2 support, thus pcre2-devel in makedepends

q66:
a bunch of the bundled scripts also use a `#!/bin/bash` shebang despite seemingly using no bashisms

q66:
before this is ready, these things are pending:

1) the scripts i mentioned above need correcting and testing
2) while disabling failing tests is potentially fine, i need to know why they fail, so you need to post the detailed test suite failure logs

reocat:
>     2. while disabling failing tests is potentially fine, i need to know why they fail, so you need to post the detailed test suite failure logs (in fact i strongly suspect they fail because of missing bash for the uncorrected scripts)

It's not looking like the reason is bash, or at least not in every failed tests. I tried to add bash in dependencies and removing patch, but it still the same. So I guess there potentially could be some another reasons for that. Maybe I'm wrong. Anyways, log from test suite:

[test-suite.log](https://github.com/chimera-linux/cports/files/13401463/test-suite.log)

q66:
yes these failures are unrelated, let's keep them patched out for now

the scripts still need fixing though

reocat:
Screenshot from 2023-11-19 11-23-11
And this was in sources:
Screenshot from 2023-11-19 11-23-30
Perl shebang getting ignored, causing errors like this:
`zsh: exec format error: unwrapdiff`
Should i try to hardcode path to perl?

q66:
put perl in hostmakedepends and it may work (re-check the tests with it too)

reocat:
Okay, looks like problem with perl was from my side and not linked to package. I guess i just forgot to update package

PR #908, opened 2023-11-15
contrib/lf: new package (31)

subnut:


subnut:
I have a question...

Why doesn't the go build-style add `go` to `hostmakedepends` by itself?

PR #918, opened 2023-11-17
main/libxml2: add libxml2-devel-static

sidkshatriya:
These are not currently present.

### DIFF
--- a/main/libxml2/template.py
+++ b/main/libxml2/template.py
@@ -2,7 +2,13 @@ pkgname = "libxml2"
 pkgver = "2.11.5"
 pkgrel = 0
 build_style = "gnu_configure"
-configure_args = ["--with-threads", "--with-icu", "--with-history"]
+configure_args = [
+    "--with-threads",
+    "--with-icu",
+    "--with-history",
+    "--enable-shared",
+    "--enable-static",
+]
 make_cmd = "gmake"
 hostmakedepends = [
     "automake",
@@ -33,6 +39,8 @@ def post_extract(self):
 
 
 def post_install(self):
+    # Delete unwanted python static lib that gets built due to --enable-static
+    self.rm(self.destdir / "usr/lib/python*/site-packages/*.a", glob=True)
     self.install_license("Copyright")
 
 


q66:
why? also declaring these subpackages is unnecessary, and python should *not* come with static libs, it's completely useless and wrong

sidkshatriya:
You may want to statically link with libxml2. Currently chimera does not have `libxml2.a` available. I like the fact that chimera packages static libraries -- libxml2 seems to be a notable library that does not have a static subpackage.

The static python lib appears if you select `configure --enable-static [...]`. We could remove it.

q66:
unless you have a real reason to link libxml2 statically, "you may want to" is not enough

there are plenty of packages which don't come with static libraries, and since the static libraries consist of llvm bitcode, they're mostly unsupported (there's a reason they're completely separate)

sidkshatriya:
Using the various static libraries on chimera, I was able to statically compile a pretty big program the other day. This is useful when you want to run the binary on another system, want to tightly control the dependencies, want performance etc. Here chimera served as a useful "building ground" -- just like musl can for static builds.

As far as libxml2 is concerned, this was sometime ago -- again I was trying to compile another big program statically (sorry, I just can't remember which one now) and I required a statically compiled version of `libxml2`. So I thought it might be a good idea to add a sub-package in chimera.

>  static libraries consist of llvm bitcode

Aren't `.a` files an `ar` archive of (platform) compiled `.o` files ? (and not llvm bitcode which is platform independent)

On chimera
```bash
$ cat >hello.c <<EOT
> int main() { return 42; }
> EOT
$ clang -o hello.o -c hello.c
$ file hello.o
hello.o: ELF 64-bit LSB relocatable, x86-64, version 1 (SYSV), not stripped
```

q66:
since chimera packages are built with LTO, the output is bitcode, not object files:

```
q66@chimera: /home/q66$ clang -o hello.o -c -flto=thin hello.c
q66@chimera: /home/q66$ file hello.o
hello.o: LLVM IR bitcode
```

that means it can change across compiler versions (e.g. when llvm recently transitioned to opaque pointers, though that's rare and it's "mostly" stable) but also it means it cannot be stripped (so the output is unreasonably chunky)

q66:
in any case there is no particular harm in having the staticlib available for this but this is not the way to do it - for one you don't need to manually declare the subpackages, for two you'll need to make sure none of the leftovers are present

sidkshatriya:
I've tried to address your suggestions -- this is ready for another look again.

q66:
can you merge it with https://github.com/chimera-linux/cports/pull/911 and put `Closes: https://github.com/chimera-linux/cports/pull/911` into the commit message?

sidkshatriya:
> can you merge it with #911 and put `Closes: https://github.com/chimera-linux/cports/pull/911` into the commit message?

Done ! :-)

q66:
no merge commits please, 1 commit per template

sidkshatriya:
If you don't want a merge commit, there is no way I can add the commit message `Closes: https://github.com/chimera-linux/cports/pull/911` without amending @nekopsykose 's commit which I'd rather not do, if I can avoid it. These PRs can be easily merged successively since they don't create any conflicts. Maybe you had another process in mind ?

q66:
it's just a simple version bump, just do it yourself

q66:
alternatively you can have two commits, first the @nekopsykose one and then your amendment (without changing revision) but it does not really matter

sidkshatriya:
I have incorporated @nekopsykose 's ea53b6499ebc0c31ce7d53abedc433bbc620271f (as is) from #911 and force pushed here. 

See https://github.com/chimera-linux/cports/pull/918/commits

sidkshatriya:
(I basically built atop @nekopsykose 's commit )

main/libxml2/template.py
```diff
@@ -33,6 +39,10 @@ def post_extract(self):
 
 
 def post_install(self):
+    # A side effect of --enable-static is the production of
```
q66, line 42: no need for a long comment like that, just say something like `delete spurious files introduced by --enable-static` or something

sidkshatriya: Made comment a single line.

sidkshatriya:
(BTW @nekopsykose 's commit is not only a version bump it also removes a patch that is not longer required )

q66:
bad symlink left in the PR

sidkshatriya:
What does bad symlink here mean ? I have symlink (which i just check works ) from `main/libxml2` -> `main/libxml2-devel-static` . Is it not supposed to be this way ?

(Thanks for all the handholding btw on my first PR to cports ! )

q66:
symlinks should only exist for subpackages explicitly declared in template, the one you added makes the CI fail

q66:
also, your commit message is missing the `Closes: ...`

sidkshatriya:
I force pushed.

- Fixed the symlink issue. BTW `cbuild` is a surprisingly elegant package build system. I love its features as I learn it !
- Added the `Closes: https://github.com/chimera-linux/cports/pull/918` line to the commit message
- Removed, from this PR, the commit [ea53b6499ebc0c31ce7d53abe](https://github.com/chimera-linux/cports/pull/911/commits/ea53b6499ebc0c31ce7d53abedc433bbc620271f) which is in PR #911. The reason being that you would require a `Closes: ...` comment for that one also, I'm guessing. I think the original PR author should do that and I would not want to ideally amend that commit. In any case both the PRs don't have any merge conflicts with each other and can be applied independently. TL;DR now this PR only contains my commit.

q66:
i'm not sure what makes you think that you'd need to amend the other commit? all you really need to do is put the `Closes: ...` in your commit's message and leave the other one as is

not having merge conflicts doesn't matter, the point is that if i merged the other PR first, the buildbot would pick it up, and then to have it pick up your changes also, i'd have to bump revision on the library

PR #922, opened 2023-11-18
contrib/netavark: install to usr/libexec/podman instead

nekopsykose:
the project makefile install rule puts it in LIBEXECPODMAN; this is where other tools that utilise this like buildah expect to find it too.

### DIFF
--- a/contrib/netavark/template.py
+++ b/contrib/netavark/template.py
@@ -1,6 +1,6 @@
 pkgname = "netavark"
 pkgver = "1.8.0"
-pkgrel = 0
+pkgrel = 1
 build_style = "cargo"
 hostmakedepends = ["cargo", "go-md2man", "protoc"]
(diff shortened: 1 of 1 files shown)


contrib/netavark/template.py
```diff
@@ -10,3 +10,11 @@
 url = "https://github.com/containers/netavark"
 source = f"{url}/archive/v{pkgver}.tar.gz"
 sha256 = "b1422ef6927458e9f80f7d322b751e29ab5d04d8ed6cb065baa82fa4291af10f"
+
+
+def do_install(self):
+    self.install_file(
```
q66, line 16: install_bin

nekopsykose: install_bin only goes to usr/bin tho which is not where this is going

q66: oh right, q66 moment

PR #923, opened 2023-11-19
contrib/duperemove: update to 0.14

autumnontape:
It would also be possible to work around broken CFI by passing `-fsanitize-cfi-icall-generalize-pointers` instead of turning it off completely.


q66:
you can just bring in the pending patch

autumnontape:
@q66 Done.

PR #924, opened 2023-11-19
contrib/worker: new package (<4.12.1>)

logout128:
Template created and formatted.


q66:
>\>4 commits
> \>read it as 1
> \>merge it

q66 moment

PR #925, opened 2023-11-20
contrib/i2c-tools: new package (4.3)

subnut:
Umm.. is the commit message OK?

I mean... we are also adding the `python-smbus` package in this same commit.. but how should I mention it in the commit msg?

### DIFF
--- a/contrib/i2c-tools/template.py
+++ b/contrib/i2c-tools/template.py
@@ -1,0 +1,31 @@
+pkgname = "i2c-tools"
+pkgver = "4.3"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_build_args = ["EXTRA=py-smbus"]
+make_install_args = ["sbindir=/usr/bin"] + make_build_args
+hostmakedepends = ["gmake", "python", "python-setuptools"]
+makedepends = ["linux-headers", "python-devel"]
+depends = ["perl"]
+pkgdesc = "Heterogeneous set of I2C tools and I2C library for Linux"
+maintainer = "Subhaditya Nath <sn03.general@gmail.com>"
+license = "GPL-2.0-or-later AND LGPL-2.1-or-later AND GPL-2.0-only"
+url = "https://git.kernel.org/pub/scm/utils/i2c-tools/i2c-tools.git"
+source = f"https://mirrors.edge.kernel.org/pub/software/utils/i2c-tools/i2c-tools-{pkgver}.tar.xz"
+sha256 = "1f899e43603184fac32f34d72498fc737952dbc9c97a8dd9467fadfdf4600cf9"
+# no tests
+options = ["!check"]
+
+
+# LGPL-2.1-or-later
+@subpackage("i2c-tools-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+# GPL-2.0-only
+@subpackage("python-smbus")
+def _python(self):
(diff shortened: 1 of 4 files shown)


nekopsykose:
it's just a subpackage so there's nothing to mention

contrib/i2c-tools/template.py
```diff
@@ -0,0 +1,31 @@
...
+maintainer = "Subhaditya Nath <sn03.general@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://git.kernel.org/pub/scm/utils/i2c-tools/i2c-tools.git"
+source = f"https://mirrors.edge.kernel.org/pub/software/utils/i2c-tools/i2c-tools-{pkgver}.tar.xz"
+sha256 = "1f899e43603184fac32f34d72498fc737952dbc9c97a8dd9467fadfdf4600cf9"
+# no tests
+options = ["!check"]
+
+
+@subpackage(f"{pkgname}-devel")
```
q66, line 21: please no gratuitous f-strings

contrib/i2c-tools/template.py
```diff
@@ -0,0 +1,31 @@
...
+def _devel(self):
+    self.license = "LGPL-2.1-or-later"
```
q66, line 23: also nope

contrib/i2c-tools/template.py
```diff
@@ -0,0 +1,31 @@
...
+    return self.default_devel()
+
+
+@subpackage("python-smbus")
+def _python(self):
+    self.pkgdesc = "Python bindings for Linux SMBus access through i2c-dev"
+    self.license = "GPL-2.0-only"
```
q66, line 30: ditto

contrib/i2c-tools/template.py
```diff
@@ -0,0 +1,31 @@
...
+build_style = "makefile"
+make_cmd = "gmake"
+make_build_args = ["EXTRA=py-smbus"]
+make_install_args = ["sbindir=/usr/bin"] + make_build_args
+hostmakedepends = ["gmake", "python", "python-setuptools"]
+makedepends = ["linux-headers", "python-devel"]
+depends = ["perl"]
+pkgdesc = "Heterogeneous set of I2C tools and I2C library for Linux"
+maintainer = "Subhaditya Nath <sn03.general@gmail.com>"
+license = "GPL-2.0-or-later"
```
q66, line 13: this should be specified as an spdx license expression covering all the licenses of the project

subnut: so... should it be `AND` or `OR` ?

contrib/i2c-tools/template.py
subnut: why tho? :see_no_evil:

q66: because it's unnecessary, doesn't achieve anything, and is inconsistent with how every other template is written

contrib/i2c-tools/template.py
q66: AND is for mixed licenses, OR is for dual/triple/etc licensing

subnut:
Updated.

contrib/i2c-tools/template.py
subnut: so.... `AND`, right?

q66: yes

PR #936, opened 2023-11-20
contrib/mako: new package (1.8.0)

handlerug:
jq is used for `makoctl menu` and `makoctl mode`. However, makoctl checks for the presence of jq. Should I remove jq from dependencies?

### DIFF
--- a/contrib/mako/template.py
+++ b/contrib/mako/template.py
@@ -1,0 +1,35 @@
+pkgname = "mako"
+pkgver = "1.8.0"
+pkgrel = 0
+build_style = "meson"
+configure_args = [
+    "-Dbash-completions=true",
+    "-Dfish-completions=true",
+    "-Dzsh-completions=true",
+]
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+    "scdoc",
+]
+makedepends = [
+    "basu-devel",
+    "cairo-devel",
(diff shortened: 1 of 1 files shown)


q66:
dbus should not be a runtime dependency here, if it merely tries to connect to it

as for jq, if it's required for expected functionality I'd keep it as is, if it's viably optional then I'd either drop it (if it's not important) or make a subpackage like mako-jq depending on jq, and have it be install_if on `{pkgname}={pkgver}-r{pkgrel}` - this will effectively turn it into a soft dependency (installed by default but optional)

q66:
user service would also be nice but needs to wait till graphical environment can be checked reasonably

contrib/mako/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "cairo-devel",
+    "elogind-devel",
+    "gdk-pixbuf-devel",
+    "linux-headers",
+    "pango-devel",
+    "wayland-devel",
+    "wayland-protocols",
+]
+depends = [
+    "dbus",
```
q66, line 25: there is no need to depend on dbus, nor it is a good idea (presumably the user will have a bus through other means, and it does not necessarily mean that it's *this* dbus)

q66: since basu does not split the progs, you get it regardless, due to having a library dependency

PR #937, opened 2023-11-20
contrib/swayidle: add elogind support

handlerug:


q66:
the commit message refers to swaylock but the PR modifies swayidle? additionally, elogind will be removed eventually, so i dunno if to enable this

PR #938, opened 2023-11-20
contrib/poweralertd: new package (0.2.0)

handlerug:


contrib/poweralertd/template.py
```diff
@@ -0,0 +1,22 @@
...
+maintainer = "Umar Getagazov <umar@handlerug.me>"
+license = "GPL-3.0-only"
+url = "https://sr.ht/~kennylevinsen/poweralertd"
+source = f"https://git.sr.ht/~kennylevinsen/poweralertd/archive/{pkgver}.tar.gz"
+sha256 = "f70076dad452c592e2fcdeba4fd533c11394c254f42c21522aa7b56f92a0bd69"
+hardening = ["vis", "cfi"]
+
+
+def post_install(self):
+    self.install_license("LICENSE")
```
nekopsykose, line 22: don't have to install gpl

nekopsykose: based on if licence needs distribution, yes for mit

q66:
this one lgtm; we could use basu here for the dbus access, but for now it's ok

q66:
we should have a dinit user service for this but that will need to wait until user services have a graphical.target sentinel (which is coming Real Soon now)

PR #939, opened 2023-11-20
contrib/brightnessctl: add elogind support

handlerug:


q66:
i'm not entirely sure about this one considering elogind will be removed eventually

q66:
it's going to be a very gradual process and it's sort of a catch-22 situation so i'm not entirely sure how long it'll actually take since the transition has to be done without creating regressions (that means porting over everything that uses it, while ensuring it remains working enough to support what hasn't been ported yet), but turnstile will eventually take over all session tracking functionality, and auxiliary daemons built around it will deal with the rest (uaccess and the likes)

it's not possible to keep elogind in the long term as it's too inflexible and geared towards just making stuff work somewhat in a legacy-style environment, but in a system that actually implements stuff equivalently to systemd, it's more often than not an obstacle

q66:
there will be initial usable API in turnstile soon so the process will commence, though setting brightness is not something that's actually within the scope at the moment (sounds like something best suited for an auxiliary service)

PR #948, opened 2023-11-22
contrib/fonts-noto-emoji-ttf: new package (2.042)

GeopJr:
The fontconfig is taken from [Arch](https://gitlab.archlinux.org/archlinux/packaging/packages/noto-fonts-emoji/-/blob/main/66-noto-color-emoji.conf?ref_type=heads)


q66:
this will require formatting with black

GeopJr:
Done!

q66:
there's an update available

JamiKettunen:
Hate to be that guy but https://github.com/googlefonts/noto-emoji/releases/tag/v2.042 :p

GeopJr:
Updated, thanks!

PR #949, opened 2023-11-22
main/libspelling: new package (0.2.0)

GeopJr:
Not sure if this should be in main instead

### DIFF
--- a/main/libspelling/template.py
+++ b/main/libspelling/template.py
@@ -1,0 +1,30 @@
+pkgname = "libspelling"
+pkgver = "0.2.0"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Denchant=enabled", "-Dvapi=true", "-Ddocs=false"]
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+    "gobject-introspection",
+    "vala",
+    "glib-devel",
+]
+makedepends = [
+    "enchant-devel",
+    "glib-devel",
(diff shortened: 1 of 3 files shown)


q66:
does anything use it?

GeopJr:
I believe [gtranslator](https://gitlab.gnome.org/GNOME/gtranslator/) does and I use it for Tuba. It's the only spelling checker for GTK4 so I assume more apps will do too :shrug:

nekopsykose:
can probably just move it to main when that happens since that takes a few seconds anyway

GeopJr:
FWIW, gnome-text-editor also plans to switch to it eventually [issue](https://gitlab.gnome.org/GNOME/gnome-text-editor/-/issues/594) / [MR](https://gitlab.gnome.org/GNOME/gnome-text-editor/-/merge_requests/117). Moving it to main then sounds good to me!

q66:
in that case, why not just put it in main straight away?

GeopJr:
Moved to main

main/libspelling/template.py
```diff
@@ -0,0 +1,31 @@
...
+    "vala",
+    "glib-devel",
+]
+makedepends = [
+    "enchant-devel",
+    "glib-devel",
+    "gtk4-devel",
+    "gtksourceview-devel",
+    "icu-devel",
+    "vala",
```
q66, line 19: shouldn't need vala in makedepends

PR #950, opened 2023-11-23
vulkan-validation-layers: new package (1.3.270)

flukeeey:
Required by wlroots's experimental and incomplete Vulkan renderer backend. Enabled via the `WLR_RENDERER=vulkan` environment variable when launching Sway.

Could someone check the `vulkan-utility-libraries-devel-static` subpackage please, it contains a single `.a` static lib, required by vulkan-validation-layers. I wasn't really sure how to package this, so I checked existing cports packages for inspiration, hope I got it right.

robin-hood-hashing is no longer maintained but vulkan-validation-layers seems to use it still for faster hashing (though is optional).

robin-hood-hashing and vulkan-utility-libraries are both header-only packages.

### DIFF
--- a/contrib/robin-hood-hashing/template.py
+++ b/contrib/robin-hood-hashing/template.py
@@ -1,0 +1,16 @@
+pkgname = "robin-hood-hashing"
+pkgver = "3.11.5"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DRH_STANDALONE_PROJECT=OFF"]
+hostmakedepends = ["cmake", "ninja"]
+pkgdesc = "Fast & memory efficient hashtable based on robin hood hashing"
+maintainer = "flukey <flukey@vapourmail.eu>"
+license = "MIT"
+url = "https://github.com/martinus/robin-hood-hashing"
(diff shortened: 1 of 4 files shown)


q66:
this already used to be packaged and it was a constant source of pain for various reasons so i don't think I'm interested in re-adding it without any users to justify it

nekopsykose:
> Required by wlroots's experimental and incomplete Vulkan renderer backend.  

that's not the case anymore, wlroots 0.17 dropped that requirement :)

flukeeey:
> > Required by wlroots's experimental and incomplete Vulkan renderer backend.
> 
> that's not the case anymore, wlroots 0.17 dropped that requirement :)

Ah perfect, thanks for the heads up!

> this already used to be packaged and it was a constant source of pain for various reasons so i don't think I'm interested in re-adding it without any users to justify it

I guess I can close this, and submit a PR to update wlroots, along with incrementing its dependents' pkgrels?

q66:
considering it was just released and every wlroots release breaks api, i doubt everything already builds against it, it's best to just wait

nekopsykose:
yes, need to wait for sway+wayfire to update too

PR #966, opened 2023-11-23
main/pipewire: update to 0.3.85

nekopsykose:
https://gitlab.freedesktop.org/pipewire/pipewire/-/releases/0.3.85


flukeeey:
1.0.0 was released FYI.

PR #985, opened 2023-11-24
contrib/rclone: new package (1.65.0)

flukeeey:


contrib/rclone/template.py
```diff
@@ -0,0 +1,51 @@
...
+build_style = "go"
+hostmakedepends = ["go"]
+pkgdesc = "Rsync for cloud storage"
+maintainer = "flukey <flukey@vapourmail.eu>"
+license = "MIT"
+url = "https://rclone.org"
+source = f"https://github.com/rclone/rclone/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "85feffc2d60554bcc3c59140750dc4ccf008e109b52c451956a1f52387af1bd6"
+# tests require network
+options = ["!check"]
```
q66, line 13: since you're executing built artifacts, it should also be !cross

q66: alternatively, you can do something like @wezm did in https://github.com/chimera-linux/cports/pull/992

PR #991, opened 2023-11-26
contrib/basu: new package (0.2.1)

handlerug:
See: https://github.com/chimera-linux/cports/pull/936#issuecomment-1820005931


q66:
since you're adding vis/cfi here, are you sure the library is properly tagged for hidden visibility to work? can you compile with and without and compare `nm -D` outputs on both?

q66:
it's fine if internal stuff disappears (in fact that is a part of the point), as long as all public api is visible

PR #995, opened 2023-11-29
rollup rpi kernel to 6.1.63

gompa:
i think this updates the kernel to 6.1.63
but iam not familiar with the build system so i could be missing something


q66:
i do kernel updates myself as needed, this alone is not enough anyway

PR #998, opened 2023-11-29
contrib/bat: clean up template

wezm:
Not sure whether this is worth the rebuild or not, but it removes some bits left over from when there was a patch.


q66:
no rebuild please (it does not change anything)

wezm:
no problem, reverted pkgrel

PR #1006, opened 2023-12-02
contrib/fonts-noto: new package (23.12.1)

GeopJr:
Not sure about the subpackage, I followed Void's but alpine seems to split it into much smaller ones

fontconfig from Arch

### DIFF
--- a/contrib/fonts-noto-extra
+++ b/contrib/fonts-noto-extra
@@ -1,0 +1,1 @@
+fonts-noto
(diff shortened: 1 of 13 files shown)


nekopsykose:
do you know what the difference between hinted/ttf and unhinted/otf and unhinted/ttf is? at a glance the ttf ones are double the size, which vaguely reminds me of the 'ttf' format being an older standard (which funnily isnt based on the file extension but the actual file data), but i'm not sure how the hinted/ plays into it or what other font features are present/not present between those 3

q66:
fonts where otf is possible should provide both ttf and otf packages and default to otf (see fonts-dejavu)

GeopJr:
Does this packaging make sense?

PR #1007, opened 2023-12-02
contrib/fonts-noto-sans-cjk: new package (2.004)

GeopJr:
Sans version

- Same subpackaging as #1006 
- fontconfig from Arch

### DIFF
--- a/contrib/fonts-noto-sans-cjk-extra
+++ b/contrib/fonts-noto-sans-cjk-extra
@@ -1,0 +1,1 @@
+fonts-noto-sans-cjk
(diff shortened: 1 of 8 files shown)


q66:
i'd prefer to have this split into two templates

q66:
i've been thinking about it and the suffix for these should definitely be -ttf, not -ttc; ttc is just a collection of ttfs, and is a detail that is meaningless in this context; otf and ttf suffix in this case mean "OpenType Fonts" and "TrueType Fonts"

PR #1011, opened 2023-12-05
contrib/conspy: new package (1.16)

JamiKettunen:


RoastVeg:
Isn't AGPL code generally forbidden?

JamiKettunen:
`ghostscript` and `jbig2dec` both are in binary repo already and have a `AGPL: forbidden from being a dependency except in special unambiguous cases` comment at the top of their `template.py`'s; considering this is a standalone program without any library part or development files I believe it should be fine?

PR #1065, opened 2023-12-11
main/perl: update to 5.38.2

nekopsykose:
https://metacpan.org/release/PEVANS/perl-5.38.2/changes


nekopsykose:
seems like a chimerautils update broke this  

```
=> perl-5.38.2-r0: running do_configure...
Guessing build platform ... x86_64-pc-linux-musl
Configuring for a native build
Checking perl version ... tail: illegal offset -- -1
perl5-5.38.2
No patchset found for perl5-5.38.2 in cnf/diffs
This perl version is probably not supported by perl-cross
```

nekopsykose:
ah no, just a side effect of old -cross

PR #1080, opened 2023-12-14
main/*: embrace tmpfiles.d

JamiKettunen:
Gets rid of `*-prepare` and `*-dir` dinit services that were just creating some dirs on boot in favor of using `tmpfiles.d`; this should be fine but definitely needs to be properly tested to be sure (I only tested a GNOME VM still booted to desktop but not every individual service file modified here).

Also converted `/var/db/nscd` to an "empty" packaged dir as it didn't sound like something that needs to be created every time on boot/service start, perhaps I'm wrong though or there was some other reason to do it like this previously.

Additionally: https://github.com/util-linux/util-linux/blob/stable/v2.39/Documentation/releases/v2.39.3-ReleaseNotes

### DIFF
--- a/main/chrony/files/chrony-dir
+++ b/main/chrony/files/chrony-dir
@@ -1,6 +1,0 @@
-# prepare chrony environment
-
-type       = scripted
-command    = /usr/bin/install -d -m 750 -o _chrony -g _chrony /run/chrony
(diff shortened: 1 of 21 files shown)


q66:
i incorporated this as a part of the sysusers work

PR #1099, opened 2023-12-18
main/libseccomp: update to 2.5.5

nekopsykose:


nekopsykose:
failures happen with old version too, so probably new toolchain

q66:
appears to break with -O2 and higher

PR #1104, opened 2023-12-18
main/python: update to 3.12.1

nekopsykose:
[ci skip]


nekopsykose:
rold boost into it too

nekopsykose:
it's almost probably missing a few python-* things in the rebuild list however

nekopsykose:
ok now it should be all, basically `apk list -d so:libpython* so:libboost* base-python3.11 python3.11` + `boost-devel` makedepends

PR #1112, opened 2023-12-19
contrib/restic: new package (0.16.2)

mhmdanas:


contrib/restic/template.py
```diff
@@ -0,0 +1,26 @@
...
+license = "BSD-2-Clause"
+url = "https://restic.net"
+source = f"https://github.com/restic/restic/releases/download/v{pkgver}/restic-{pkgver}.tar.gz"
+sha256 = "88165b5b89b6064df37a9964d660f40ac62db51d6536e459db9aaea6f2b2fc11"
+
+
+def do_build(self):
+    from cbuild.util import golang
+
+    self.golang.build(wrksrc="cmd/restic")
```
nekopsykose, line 17: you can probably use `make_build_args = ["./cmd/..."]` instead (so ./cmd/restic ?)

contrib/restic/template.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+    self.install_man("doc/man/*.1", glob=True)
```
nekopsykose, line 22: is it only .1 or are there others?

mhmdanas: It's only .1 at the moment.

PR #1114, opened 2023-12-19
contrib/wlr-randr: new package (0.3.1)

ifreund:


contrib/wlr-randr/template.py
```diff
@@ -0,0 +1,22 @@
+pkgname = "wlr-randr"
+pkgver = "0.3.1"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Dwerror=false"]
```
nekopsykose, line 5: this is passed by default via cbuild/util/meson

ifreund: Indeed, I fixed this for wlr-randr and also for the wlroots template which is where I had copied it from.

ifreund: Fixed for the other 2 templates that redundantly pass this as well.

ifreund:
CI seems to be failing because wlroots 0.17.0 is available. It does not make sense to upgrade wlroots until the compositors packaged by chimera see releases though.

Is there some workaround I should do to fix the CI? I could just drop the wlroots template cleanup or add a comment to fix it with the next upgrade I suppose.

nekopsykose:
nah, all good

PR #1116, opened 2023-12-19
contrib/shotman: new package (0.4.5)

mhmdanas:


contrib/shotman/template.py
```diff
@@ -0,0 +1,27 @@
...
+options = ["!check"]
+
+
+def post_build(self):
+    self.do("bmake", "shotman.1")
+
+
+def post_install(self):
+    self.install_license("LICENCE.md")
+    self.install_man("shotman.1")
```
mhmdanas, line 27: The package also has completions, but they're generated by running a binary, which won't work in cross. Not sure what to do about that.

nekopsykose: you could make it !cross like sbctl

mhmdanas: Is it worth making it !cross just for completions? We could generate them only on native builds, but idk

q66: !cross is fine

contrib/shotman/template.py
```diff
@@ -0,0 +1,53 @@
+pkgname = "shotman"
+pkgver = "0.4.5"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo", "pkgconf", "scdoc"]
+makedepends = ["libxkbcommon-devel", "rust-std"]
+depends = ["slurp"]
+pkgdesc = "Uncompromising screenshot GUI for Wayland"
```
q66, line 8: Screenshot GUI for Wayland

PR #1117, opened 2023-12-19
contrib/senpai: new package (0.3.0)

mhmdanas:


contrib/senpai/template.py
```diff
@@ -0,0 +1,21 @@
...
+pkgdesc = "IRC client that works best with bouncers"
+maintainer = "triallax <triallax@tutanota.com>"
+license = "ISC"
+url = "https://git.sr.ht/~taiite/senpai"
+source = f"https://git.sr.ht/~taiite/senpai/archive/v{pkgver}.tar.gz"
+sha256 = "c02f63a7d76ae13ed888fc0de17fa9fd5117dcb3c9edc5670341bf2bf3b88718"
+options = ["!debug"]
+
+def post_build(self):
+    self.do("gmake", "doc")
```
mhmdanas, line 16: Or I could just run `scdoc` directly to remove the `gmake` dependency if you want (the command fails with `bmake`).

mhmdanas:
I seem to be having connectivity issues with senpai, I'm marking this as draft until I determine whether they're an issue with senpai or just my internet connection.

contrib/senpai/template.py
nekopsykose: this is fine as is

PR #1124, opened 2023-12-20
contrib/regreet: new package (1.1.1)

Natthias:
Adds an actual graphical greeter for greetd. Upstream provides a systemd-tmpfiles config, but it requires patching, so using our own seems more sane. regreet.toml is for elogind.

This would require wlroots compositor as well as some manual config. I didn't provide any of the upstream example configs nor did I add a compositor to depends. Should I?

### DIFF
--- a/contrib/regreet/template.py
+++ b/contrib/regreet/template.py
@@ -1,0 +1,27 @@
+pkgname = "regreet"
+pkgver = "0.1.1"
+pkgrel = 0
+build_style = "cargo"
+make_build_args = [
+    "--features=gtk4_8",
+]
+hostmakedepends = [
+    "cargo",
+    "pkgconf",
+]
+makedepends = [
+    "gtk4-devel",
+    "rust-std",
+]
+depends = ["greetd"]
+pkgdesc = "Clean and customizable greeter for greetd"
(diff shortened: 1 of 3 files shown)


q66:
how does this compare to gtkgreet?

Natthias:
More configuration of the greeter through a config file rather than only being able to specify css style, regreet reads xsesssions and wayland-sessions rather than having to write an environments file, and regreet is gtk4 while gtkgreet is gtk3.

contrib/regreet/template.py
```diff
@@ -0,0 +1,34 @@
...
+    "gtk4-devel",
+    "rust-std",
+]
+depends = ["greetd"]
+pkgdesc = "Clean and customizable greeter for greetd"
+maintainer = "natthias <natthias@proton.me>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/rharish101/ReGreet"
+source = (
+    f"https://github.com/rharish101/ReGreet/archive/refs/tags/{pkgver}.tar.gz"
```
nekopsykose, line 22:
```suggestion
    f"{url}/archive/refs/tags/{pkgver}.tar.gz"
```

PR #1125, opened 2023-12-20
contrib/yambar: new package (1.10.0)

mhmdanas:


contrib/yambar/template.py
```diff
@@ -0,0 +1,33 @@
+pkgname = "yambar"
+pkgver = "1.10.0"
+pkgrel = 0
+build_style = "meson"
+make_check_wrapper = ["xvfb-run"]
```
nekopsykose, line 5: weston-headless-run instead

mhmdanas:
Not using this anymore.

PR #1126, opened 2023-12-20
contrib/vectorscan: new package (5.4.11)

a16bitsysop:


contrib/vectorscan/template.py
```diff
@@ -0,0 +1,40 @@
...
+    "-DCMAKE_SKIP_INSTALL_RPATH=ON",
+    "-DBUILD_SHARED_LIBS=ON",
+    "-DFAT_RUNTIME=OFF",
+]
+hostmakedepends = ["cmake", "ninja", "pkgconf", "ragel"]
+makedepends = ["boost-devel", "sqlite-devel"]
+pkgdesc = "High-performance regular expression matching library"
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "BSD-3-Clause"
+url = "https://www.vectorcamp.gr/vectorscan/index.html"
```
wezm, line 17: Link redirects to https://www.vectorcamp.gr/ which doesn't seem to describe the library.

a16bitsysop: Okay will find the correct one

contrib/vectorscan/template.py
```diff
@@ -0,0 +1,40 @@
+pkgname = "vectorscan"
+pkgver = "5.4.11"
+pkgrel = 0
+archs = ["aarch64", "ppc64", "x86_64"]
```
q66, line 4: should probably be ppc64le

(does it not build at all on other platforms?)

contrib/vectorscan/template.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@subpackage("vectorscan-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage("vectorscan-progs")
+def _subpkg(self):
+    self.pkgdesc = f"{pkgdesc} (command line tools)"
```
q66, line 39: any reason not to use default_progs?

contrib/vectorscan/template.py
a16bitsysop: without `index.html` it works, tried `home` and `default` with html and htm extensions as well and all redirected

contrib/vectorscan/template.py
a16bitsysop: at the minute no, but I think there will be a software backend in the next release for arch’s with unsupported simd extensions

contrib/vectorscan/template.py
a16bitsysop: it works without a `/` on the end of vectorscan as well

contrib/vectorscan/template.py
a16bitsysop: no, will try that

q66:
i'll need to manually test this on the other archs

a16bitsysop:
okay

q66:
tests are ok on all archs, however the PR is missing subpackage symlinks (the CI should catch this, but apparently doesn't?)

a16bitsysop:
ah, must have missed adding them.  It fails locally without them.

PR #1127, opened 2023-12-21
contrib/xonotic: move icon to hicolor

Froggo8311:


q66:
nah, the icon is supposed to be generic and there is no guarantee everything will be capable of rendering svg

the glx/sdl impls are split for a reason too, sometimes the sdl version has issues glx doesn't depending on setup

nekopsykose:
> nah, the icon is supposed to be generic and there is no guarantee everything will be capable of rendering svg  

you probably want to install both in that case, also the 'pixmaps' dir is technically deprecated (correct place is `/usr/share/icons/hicolor/SIZExSIZE/apps/`)

q66:
it's technically deprecated, but this does not do what the PR description says (i.e. allow icon themes to override it) since icon themes already take priority over this path

as for installing both, that's fine but there isn't actually much value in it (what'd be more helpful is multiple sizes of the bitmap icon, particularly the small sizes where neither a scalable icon nor the default bitmap render well)

Froggo8311:
wtf okay why did it close the thing

PR #1129, opened 2023-12-21
Add lint for 'written in' in pkgdesc

wezm:
This has come up a few times as something q66 would prefer not to have in `pkgdesc` so I added a lint for it.


q66:
i dunno about this one, i don't like gratuitously adding it but e.g. for libraries it may make sense (as it describes that language the API is available in)

i'll think about it

q66:
hm, nah probably not worth it for this one

PR #1131, opened 2023-12-21
main/libcamera: new package (0.1.0)

subnut:
IPA module signing has been patched out... because I couldn't decide whether keep it or patch it out :sweat_smile: 

Tested on x86_64... seems to be working (i.e. flatpak camera app works)..

### DIFF
--- a/main/libcamera/template.py
+++ b/main/libcamera/template.py
@@ -1,0 +1,55 @@
+pkgname = "libcamera"
+pkgver = "0.1.0"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Dtest=true"]
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+    "python-jinja2",
+    "python-ply",
+    "python-pyyaml",
+]
+makedepends = [
+    "gst-plugins-base-devel",
+    "gstreamer-devel",
+    "gtest-devel",
+    "libevent-devel",
+    "libunwind-devel",
+    "libyaml-devel",
+    "openssl-devel",
+    "udev-devel",
+]
+pkgdesc = "Open source camera stack and framework"
+maintainer = "Subhaditya Nath <sn03.general@gmail.com>"
+license = "LGPL-2.1-or-later AND GPL-2.0-or-later"
+url = "https://libcamera.org"
+source = f"https://github.com/libcamera-org/libcamera/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "6d9b0f33939de29a2d8950924f82713981b890f9b5dc1d68a65db76a1222b15a"
+nostrip_files = ["usr/lib/libcamera/ipa*.so"]
+
+
+def post_install(self):
+    from cbuild.util import strip
+
+    for f in (self.destdir / "usr/lib/libcamera").glob("ipa*.so"):
+        print(f"   Stripping and signing: {f.name}")
+        strip.strip_attach(self, f)
+        self.do(
+            "src/ipa/ipa-sign.sh",
+            "build/src/ipa-priv-key.pem",
+            self.chroot_destdir / f.relative_to(self.destdir),
+            f"{self.chroot_destdir / f.relative_to(self.destdir)}.sign",
+        )
+
+
+@subpackage("gstreamer-libcamera")
+def _gst(self):
+    self.pkgdesc = f"{pkgdesc} (GStreamer support)"
+    self.install_if = [f"{pkgname}={pkgver}-r{pkgrel}", "gstreamer"]
+    return ["usr/lib/gstreamer-1.0"]
+
+
+@subpackage("libcamera-devel")
+def _devel(self):
+    return self.default_devel()
--- a/main/gstreamer-libcamera
+++ b/main/gstreamer-libcamera
@@ -1,0 +1,1 @@
+libcamera
--- a/main/libcamera-devel
+++ b/main/libcamera-devel
@@ -1,0 +1,1 @@
+libcamera
--- a/main/libcamera/patches/00-meson_build.patch
+++ b/main/libcamera/patches/00-meson_build.patch
@@ -1,0 +1,11 @@
+--- a/meson.build
++++ b/meson.build
+@@ -248,7 +248,7 @@
+ # Create a symlink from the build root to the source root. This is used when
+ # running libcamera from the build directory to locate resources in the source
+ # directory (such as IPA configuration files).
+-run_command('ln', '-fsT', meson.project_source_root(), meson.project_build_root() / 'source',
++run_command('ln', '-fsF', meson.project_source_root(), meson.project_build_root() / 'source',
+             check : true)
+ 
+ configure_file(output : 'config.h', configuration : config_h)
--- a/main/libcamera/patches/01-ipu3_basename.patch
+++ b/main/libcamera/patches/01-ipu3_basename.patch
@@ -1,0 +1,22 @@
+Musl doesn't define basename in string.h, even if _GNU_SOURCE is defined
+
+--- a/utils/ipu3/ipu3-pack.c
++++ b/utils/ipu3/ipu3-pack.c
+@@ -8,6 +8,7 @@
+ 
+ #include <errno.h>
+ #include <fcntl.h>
++#include <libgen.h>
+ #include <stdint.h>
+ #include <stdio.h>
+ #include <string.h>
+--- a/utils/ipu3/ipu3-unpack.c
++++ b/utils/ipu3/ipu3-unpack.c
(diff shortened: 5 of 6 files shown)


main/libcamera/template.py
```diff
@@ -0,0 +1,47 @@
...
+hostmakedepends = [
+    "meson",
+    "ninja",
+    "pkgconf",
+    "python-pyyaml",
+    "python-ply",
+    "python-jinja2",
+]
+makedepends = [
+    # See: https://libcamera.org/getting-started.html#dependencies
```
nekopsykose, line 15: don't need this

main/libcamera/template.py
```diff
@@ -0,0 +1,47 @@
...
+    "libyaml-devel",
+    "libunwind-devel",
+    "udev-devel",
+    "gstreamer-devel",
+    "gst-plugins-base-devel",
+    "libevent-devel",
+    "gtest-devel",
```
nekopsykose, line 22: sort the list

main/libcamera/template.py
```diff
@@ -0,0 +1,47 @@
...
+
+@subpackage("libcamera-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage("libcamera-tools")
+def _tools(self):
+    self.pkgdesc = f"{pkgdesc} (tools)"
+    return ["usr/bin"]
```
nekopsykose, line 40: name this -progs and return default_progs unless it really has to be named tools for some reason

main/libcamera/template.py
```diff
@@ -0,0 +1,47 @@
+pkgname = "libcamera"
+pkgver = "0.1.0"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Dtest=true"]
+hostmakedepends = [
+    "meson",
+    "ninja",
```
nekopsykose, line 8:
```suggestion
```

meson already pulls this

also sort the list

q66:
module signing should be handled

main/libcamera/patches/02-disable-ipa-signing.patch
```diff
@@ -0,0 +1,11 @@
+--- a/src/meson.build
```
nekopsykose, line 1: you shouldn't disable the signing

i'm not sure how to make it run post-debugsplit though (stripping the binary after signing breaks the signature etc)

q66: we could expose stripping as a utility module and then it could be done in post_install

subnut: or we could include affected files under `nostrip_files`, ig..

main/libcamera/template.py
```diff
@@ -0,0 +1,45 @@
...
+    "libyaml-devel",
+    "udev-devel",
+]
+pkgdesc = "Open source camera stack and framework"
+maintainer = "Subhaditya Nath <sn03.general@gmail.com>"
+license = "LGPL-2.1-or-later AND GPL-2.0-or-later"
+url = "https://libcamera.org"
+source = f"https://github.com/libcamera-org/libcamera/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "6d9b0f33939de29a2d8950924f82713981b890f9b5dc1d68a65db76a1222b15a"
+nostrip_files = ["usr/lib/libcamera/*.so"]
```
nekopsykose, line 28: `usr/lib/libcamera/ipa*.so` are the signed files

main/libcamera/patches/02-disable-ipa-signing.patch
q66: no, i don't want that

main/libcamera/template.py
q66: so the approach should be as follows:

1) put the `ipa*.so` in `nostrip_files` to inhibit automatic stripping
2) create a post_install that does something like the following:

```
def post_install(self):
    from cbuild.util import strip
    for f in (self.destdir / "usr/lib/libcamera").glob("ipa*.so"):
        strip.strip_attach(self, f)
        self.do(resign_this_module)
```

main/libcamera/template.py
```diff
@@ -0,0 +1,45 @@
...
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage("libcamera-progs")
+def _progs(self):
+    return self.default_progs()
+
+
+@subpackage("libcamera-gstreamer")
```
q66, line 41: should be called `gstreamer-libcamera`

main/libcamera/template.py
```diff
@@ -0,0 +1,45 @@
...
+def _gst(self):
+    self.pkgdesc = f"{pkgdesc} (GStreamer support)"
+    self.install_if = [f"{pkgname}={pkgver}-r{pkgrel}", "gstreamer"]
+    return ["usr/lib/gstreamer-*"]
```
q66, line 45: also no need for a wildcard here

main/libcamera/template.py
```diff
@@ -0,0 +1,45 @@
...
+sha256 = "6d9b0f33939de29a2d8950924f82713981b890f9b5dc1d68a65db76a1222b15a"
+nostrip_files = ["usr/lib/libcamera/*.so"]
+
+
+@subpackage("libcamera-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage("libcamera-progs")
```
q66, line 36: probably no need to split the progs (they make up a few hundred kilobytes and the library package is already ~4M)

subnut: but the debug-progs are like 6M in total...

main/libcamera/template.py
```diff
@@ -0,0 +1,56 @@
...
+url = "https://libcamera.org"
+source = f"https://github.com/libcamera-org/libcamera/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "6d9b0f33939de29a2d8950924f82713981b890f9b5dc1d68a65db76a1222b15a"
+nostrip_files = ["usr/lib/libcamera/ipa*.so"]
+
+
+def post_install(self):
+    from cbuild.util import strip
+
+    for f in (self.chroot_destdir / "usr/lib/libcamera").glob("ipa*.so"):
```
q66, line 35: this won't match any files, because chroot_destdir only works from within the sandbox

main/libcamera/template.py
```diff
@@ -0,0 +1,58 @@
...
+nostrip_files = ["usr/lib/libcamera/ipa*.so"]
+
+
+def post_install(self):
+    from cbuild.util import strip
+
+    for f in (self.destdir / "usr/lib/libcamera").glob("ipa*.so"):
+        strip.strip_attach(self, f)
+
+    for f in (self.chroot_destdir / "usr/lib/libcamera").glob("ipa*.so"):
```
q66, line 38: i don't see how this is supposed to fix anything, considering globbing chroot_destdir literally won't match anything ever

subnut: Sorry... it's just that I'm terribly inexperienced at python.. 

Fixed.

main/libcamera/template.py
q66: ?

q66:
```
q66@banana: /home/q66/cports/packages/main/x86_64$ apk adbdump libcamera-progs-0.1.0-r0.apk|grep installed
  installed-size: 484 KiB
```

main/libcamera/template.py
```diff
@@ -0,0 +1,59 @@
...
+source = f"https://github.com/libcamera-org/libcamera/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "6d9b0f33939de29a2d8950924f82713981b890f9b5dc1d68a65db76a1222b15a"
+nostrip_files = ["usr/lib/libcamera/ipa*.so"]
+
+
+def post_install(self):
+    from cbuild.util import strip
+
+    for f in (self.destdir / "usr/lib/libcamera").glob("ipa*.so"):
+        strip.strip_attach(self, f)
```
q66, line 36: also add a log so the signing is not silent (e.g. `self.log(f"stripping and signing {f.name}")`

subnut: Yeah I was thinking about this too...

Instead of having the messages in `007_strip_debug.py`, could they be moved to `cbuild.util.strip` instead?

q66: no, that's not necessary and makes the API less useful

subnut: ok.

I did this (looks more cohesive with the other stripping messages IMO) -
```diff
      for f in (self.destdir / "usr/lib/libcamera").glob("ipa*.so"):
+         print(f"   Stripping and signing: {f.name}")    
```

PR #1132, opened 2023-12-21
contrib/epson-inkjet-printer-escpr: new package (1.8.2)

mhmdanas:


contrib/epson-inkjet-printer-escpr/template.py
```diff
@@ -0,0 +1,21 @@
...
+make_cmd = "gmake"
+make_dir = "."
+hostmakedepends = ["gmake"]
+makedepends = ["cups-devel"]
+depends = ["cups-filters"]
+pkgdesc = "Epson Inkjet Printer Driver (ESC/P-R)"
+maintainer = "triallax <triallax@tutanota.com>"
+license = "GPL-2.0-or-later"
+url = "https://download.ebz.epson.net/dsc/search/01/search?OSC=LX&productName=B700"
+source = f"https://download3.ebz.epson.net/dsc/f/03/00/15/37/36/9209d0195bbe0743c8f9f427bdc5d09186d72ba2/epson-inkjet-printer-escpr-{pkgver}-1.tar.gz"
```
q66, line 20: is this link stable

contrib/epson-inkjet-printer-escpr/update.py
```diff
@@ -0,0 +1 @@
+ignore = True
```
q66, line 1: no way to do update check?

contrib/epson-inkjet-printer-escpr/template.py
mhmdanas: Unfortunately not, it changes on every version. :/

q66: changing every version doesn't matter and wasn't the question, i'm asking if this link will remain the same for this version and if it'll provide a tarball of the same checksum

mhmdanas: Ah, if that's what what you mean, then yes, the link is stable.

PR #1136, opened 2023-12-22
contrib/swayidle: build with elogind support

mhmdanas:


contrib/swayidle/template.py
```diff
@@ -1,16 +1,13 @@
...
 hostmakedepends = [
     "meson",
     "pkgconf",
     "scdoc",
 ]
-makedepends = [
-    "wayland-devel",
-    "wayland-protocols",
-]
+makedepends = ["wayland-devel", "wayland-protocols", "elogind-devel"]
```
q66, line 10: sort

PR #1140, opened 2023-12-23
contrib/socat: fix BUILD_DATE, move socat-*.sh to examples

subnut:


q66:
unless upstream is willing to switch away from bash, i don't want to have downstream patches for this

also this PR has gratuitous changes like to pkgdesc and the formatting of makedepends, and the incorrect switch from gnu_configure to configure

contrib/socat/template.py
```diff
@@ -1,22 +1,31 @@
...
-source = f"{url}/download/socat-{pkgver}.tar.gz"
-sha256 = "6010f4f311e5ebe0e63c77f78613d264253680006ac8979f52b0711a9a231e82"
+source = f"{url}/download/socat-{pkgver}.tar.bz2"
+sha256 = "e1de683dd22ee0e3a6c6bbff269abe18ab0c9d7eb650204f125155b9005faca7"
 hardening = ["vis", "cfi"]
-# needs a smorgasbord of random network utilities
-options = ["!check"]
+
+
+def post_extract(self):
```
q66, line 18: also template-side fixups via sed are never allowed

q66: https://github.com/chimera-linux/cports/blob/master/Packaging.md#correct_templates

contrib/socat/patches/fix-date.patch
```diff
@@ -0,0 +1,11 @@
+--- a/configure
++++ b/configure
+@@ -22710,7 +22710,7 @@
+ # allow BUILD_DATE to be externally set for build reproducibility
+ if test "$SOURCE_DATE_EPOCH"; then
+   DATE_FMT="%d %b %Y %H:%M:%S"
+-  BUILD_DATE=$(LC_ALL=C date -u -d "@$SOURCE_DATE_EPOCH" "+$DATE_FMT")
++  BUILD_DATE=$(LC_ALL=C date -u -j -f %s "@$SOURCE_DATE_EPOCH" "+$DATE_FMT")
```
nekopsykose, line 8: pretty sure this doesn't work

```
$ LC_ALL=C date -u -j -f %s "@1234545" "+%d %b %Y %H:%M:%S"
Warning: Ignoring 9 extraneous characters in date string (@+1234545)
01 Jan 1970 00:00:00
```

contrib/socat/patches/fix-test.patch
```diff
@@ -0,0 +1,99 @@
...
+ 
+ cert:
+ 	# prepare critical files with correct permissions to avoid race cond
+--- a/test.sh
++++ b/test.sh
+@@ -8,12 +8,12 @@
+ # this script uses functions; you need a shell that supports them
+ 
+ # you can pass general options to socat: export OPTS="-d -d -d -d -lu"
+-# you can eg strace socat with: export TRACE="strace -v -tt -ff -D -x -s 1024 -o /tmp/$USER/socat.strace"
```
nekopsykose, line 18: rather than patch 50 uses of USER you could set it in the environment

contrib/socat/template.py
```diff
@@ -11,12 +11,12 @@
     "openssl-devel",
     "readline-devel",
 ]
+checkdepends = ["bash", "iproute2", "procps"]
+depends = ["bash"]
```
nekopsykose, line 15: socat itself doesn't need bash, the only thing that does are the *.sh scripts that i have never seen anyone use (they look more like examples than anything)

contrib/socat/patches/fix-date.patch
subnut: fixed

contrib/socat/template.py
subnut: moved to usr/share/examples

contrib/socat/patches/fix-test.patch
subnut: It was done like that because test 217 was failing whenever $USER was set in the environment.

Since then, i've discovered tons of heisenbugs in the testing code.  
So I've reverted back to disabling tests.

PR #1141, opened 2023-12-23
contrib/spice-vdagent: disable cfi

JamiKettunen:
Both the daemon and client can crash each other upon exit when built with CFI: https://paste.c-net.org/519lq1pmlhda


q66:
this is why i say cfi for glib things is futile

JamiKettunen:
I'll keep that in mind if something needs `glib-devel` to build now, in this case from my quick testing of the "expected" `dinitctl enable spice-vdagentd` and relogging to start the client worked just fine without issue, but relogging again broke the service for example (which without CFI keeps running if the client quits)

PR #1146, opened 2023-12-26
main/icu: update to 74.2

nekopsykose:


nekopsykose:
seems even older libreoffice is broken post rebuild

PR #1147, opened 2023-12-26
contrib/python-pybind11: new package (2.11.1)

a16bitsysop:


contrib/python-pybind11/template.py
```diff
@@ -0,0 +1,48 @@
...
+pkgdesc = "Seamless operability between C++11 and Python"
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "BSD-3-Clause"
+url = "https://pybind11.readthedocs.io/en/stable/index.html"
+source = f"https://github.com/pybind/pybind11/archive/refs/tags/v{pkgver}.zip"
+sha256 = "b011a730c8845bfc265f0f81ee4e5e9e1d354df390836d2a25880e123d021f89"
+
+
+def post_build(self):
+    from cbuild.build_style import python_module
```
q66, line 23: we don't import build-styles like this, also python_module is getting removed with 3.12 (all packages are being ported to pep517)

if you wish to execute setup.py, just do it directly

a16bitsysop: okay will change

a16bitsysop: made it a pep517 build and install symlinks for `/usr/share/cmake/pybind11` `/usr/share/pkgconfig/pybind11.pc` `/usr/include/pybind11` etc but dependency checker only finds the package for `/usr/include/pybind11`

nekopsykose: it's not meant to install those into sitepackages in the first place, but i'm not sure there is any way to fix that (since pep517 doesn't allow it)

i forget if there was a workaround or if the solution is to specialcase these and make them manual `self.do("python" "setup.py" ...` calls

a16bitsysop: All the other distro's I looked at have things installed installed in both locations, arch linux uses symlinks https://gitlab.archlinux.org/archlinux/packaging/packages/pybind11/-/blob/main/PKGBUILD?ref_type=heads

nekopsykose: yes, but that's a workaround because they are probably never used from there (nothing to my knowledge uses sitepackages/pybind11/{cmake,pkgconfig}). they're just linking it to the actual location- it could also be `mv` and work the same in the end. so if that's the case, might as well actually move it

a16bitsysop: okay will try moving

a16bitsysop: wouldn't this return the include location as inside `site-packages`
https://github.com/pybind/pybind11/blob/eeac2f45728633d7ee6fe792bea2890345697119/pybind11/commands.py#L6 so would need patching

nekopsykose: hm yes, in that case it's used from pybind itself, makes sense

link is fine then :)

contrib/python-pybind11/template.py
```diff
@@ -0,0 +1,69 @@
...
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "BSD-3-Clause"
+url = "https://pybind11.readthedocs.io/en/stable/index.html"
+source = f"https://github.com/pybind/pybind11/archive/refs/tags/v{pkgver}.zip"
+sha256 = "b011a730c8845bfc265f0f81ee4e5e9e1d354df390836d2a25880e123d021f89"
+options = ["!cross"]
+
+
+def post_configure(self):
+    self.do("cmake", "-B", "build-tests")
```
nekopsykose, line 27: i'm pretty sure you don't need an entire second build directory to run tests? only `-DPYBIND11_TEST=ON`

nekopsykose: oh right, it's python by default with this being an extra

nekopsykose: you should use util.cmake for these then, like qbittorrent

though the installing is optional since it installs the same files just outside site-packages (which we symlinked below)

a16bitsysop: Okay, yes I don't think cmake install is needed, the python install is what installs the python module as well.  So would just install things that are symlinked

contrib/python-pybind11/patches/010-pybind11-remove-unwanted-pip-build-deps.patch
```diff
@@ -0,0 +1,9 @@
+--- a/pyproject.toml
++++ b/pyproject.toml
+@@ -1,5 +1,5 @@
+ [build-system]
+-requires = ["setuptools>=42", "cmake>=3.18", "ninja"]
++requires = ["setuptools>=42"]
```
nekopsykose, line 6: does not removing this hurt anything since you install them anyway to build the tests?

a16bitsysop: it errored out without the patch, will try again

contrib/python-pybind11/template.py
```diff
@@ -0,0 +1,78 @@
...
+        "-c",
+        "cd build-tests/tests && python3 -m \
+        pytest --deselect test_exceptions.py::test_cross_module_exception_translator",
+    )
+
+
+def post_install(self):
+    import sys
+
+    _pypath = f"/usr/lib/python{sys.version_info[0]}.{sys.version_info[1]}/site-packages"
```
nekopsykose, line 56: can't use sys.version_info because that is using the host python and not the one actually being built against in cports

not sure what would work in this case though

a16bitsysop: I thought the cbuild was done in some kind of container or chroot?

could install to temp folder and “find” the path?

a16bitsysop: could also move the whole building tests into `init_check` so no longer !cross ?

nekopsykose: it is in a bwrap container but the `def:` hooks are evaluated on the host with the python used to run cbuild and not the python that runs `python -m build`, etc

nekopsykose: > could also move the whole building tests into init_check so no longer !cross ?
and yeah that sounds good

a16bitsysop: changed it, find is finding /usr/include first though?

nekopsykose: /usr/include sorts before /usr/lib lexicographically, you could find in self.destdir / usr/lib

a16bitsysop: that works, but stripping destdir still ends up stripping destdir / usr/lib/python

so stripped string starts with python version number

q66:
https://github.com/chimera-linux/cports/commit/f8f6636da530f58e158db6524359da36ba0725a0 should make it possible to avoid weird path-determining hacks

please simplify the symlinking, and also make it so that the symlinks are relative, not absolute (i.e. `/usr/include/pybind11` should point to `../lib/python3.12/site-packages/pybind11..`)

a16bitsysop:
okay will do

PR #1149, opened 2023-12-26
main/iputils: update to 20231222

nekopsykose:
https://github.com/iputils/iputils/releases/tag/20231222  

this also fixes EPERM on clockdiff because it needs +ep and not just +p (same as upstream script). the sys_nice isn't required it seems (works regardless) but it's what the upstream script does, so eh


nekopsykose:
i forgot if we need a new iso refresh to use apk caps, in which case i can undo that part tho

PR #1153, opened 2023-12-27
main/man-pages-posix: new package (2017a)

nekopsykose:


main/man-pages-posix/template.py
```diff
@@ -0,0 +1,17 @@
+pkgname = "man-pages-posix"
+_pver = "2017-a"
+pkgver = f"{_pver.replace('-', '')}"
+pkgrel = 0
+pkgdesc = "POSIX man pages"
+maintainer = "psykose <alice@ayaya.dev>"
+license = "custom:posix"
```
nekopsykose, line 7: as far as i can tell this says nothing about distribution so it's fine (?)

PR #1155, opened 2023-12-27
contrib/github-cli: new package (2.40.1)

mhmdanas:


contrib/github-cli/template.py
```diff
@@ -0,0 +1,43 @@
...
+make_check_args = ["./..."]
+hostmakedepends = ["go"]
+checkdepends = ["git", "openssh"]
+pkgdesc = "GitHub CLI tool"
+maintainer = "triallax <triallax@tutanota.com>"
+license = "MIT"
+url = "https://cli.github.com"
+source = f"https://github.com/cli/cli/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "0bb2af951b4716067747184b5b5bbd90c270edee5b45a84e62a5a803bf7ef467"
+options = ["!cross"]
```
nekopsykose, line 19: !debug too


though maybe we should finally figure out why that fails on ppc bleh

PR #1161, opened 2023-12-29
contrib/dtach: new package (0.9)

ifreund:


contrib/dtach/template.py
```diff
@@ -0,0 +1,18 @@
...
+pkgver = "0.9"
+pkgrel = 0
+build_style = "configure"
+pkgdesc = "Program that emulates the detach feature of screen"
+maintainer = "Isaac Freund <mail@isaacfreund.com>"
+license = "GPL-2.0-or-later"
+url = "https://dtach.sourceforge.net"
+source = f"$(SOURCEFORGE_SITE)/{pkgname}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "32e9fd6923c553c443fab4ec9c1f95d83fa47b771e6e1dafb018c567291492f3"
+
```
mhmdanas, line 11:
```suggestion
```

contrib/dtach/template.py
```diff
@@ -0,0 +1,18 @@
...
+# Makefile has no check target
+options = ["!check"]
+
+
+def do_install(self):
+    self.install_bin("dtach")
+    self.install_man("dtach.1")
```
mhmdanas, line 18: Does the makefile not have an install target?

mhmdanas (changes requested):

contrib/dtach/template.py
```diff
@@ -0,0 +1,18 @@
+pkgname = "dtach"
+pkgver = "0.9"
+pkgrel = 0
+build_style = "configure"
```
nekopsykose, line 4:
```suggestion
build_style = "gnu_configure"
```

nekopsykose: this also needs automake in hostmakedepends and then might fail to regen given the age, but it is autotools so this is the correct style

if you can't figure out the regen then put configure_gen = [] too

contrib/dtach/template.py
ifreund: nope

contrib/dtach/template.py
ifreund: looks like it does regen for now at least though a few warnings about deprecated stuff are logged.

PR #1162, opened 2023-12-30
contrib/python-pyproject-metadata: new package (0.7.1)

a16bitsysop:


contrib/python-pyproject-metadata/template.py
```diff
@@ -0,0 +1,17 @@
+pkgname = "python-pyproject-metadata"
+pkgver = "0.7.1"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = ["python-build", "python-installer", "python-wheel"]
```
nekopsykose, line 5: this is missing python-setuptools

a16bitsysop: added it

PR #1164, opened 2023-12-30
contrib/seogi: new package (1.1.4)

ifreund:
I need a wayland IME to test my compositor and this seemed like the easiest one to build/package :)

### DIFF
--- a/contrib/libhangul/template.py
+++ b/contrib/libhangul/template.py
@@ -1,0 +1,23 @@
+pkgname = "libhangul"
+pkgver = "0.1.0"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_gen = ["./autogen.sh"]
+hostmakedepends = ["automake", "gettext-devel", "libtool", "pkgconf"]
+pkgdesc = "Library to support hangul input and character classification"
+maintainer = "Isaac Freund <mail@isaacfreund.com>"
+license = "LGPL-2.1-or-later"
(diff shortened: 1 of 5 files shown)


contrib/libhangul/template.py
```diff
@@ -0,0 +1,22 @@
...
+pkgrel = 0
+build_style = "gnu_configure"
+configure_gen = ["./autogen.sh"]
+hostmakedepends = ["automake", "gettext-devel", "libtool", "pkgconf"]
+pkgdesc = "Library to support hangul input and character classification"
+maintainer = "Isaac Freund <mail@isaacfreund.com>"
+license = "LGPL-2.1-or-later"
+url = "https://github.com/libhangul/libhangul"
+source = f"{url}/archive/{pkgname}-{pkgver}.tar.gz"
+sha256 = "e2a81ef159ed098d3cc1a20377dba6204821b7ce2bc24cfb2f2543adf3bc5830"
```
nekopsykose, line 12: does it work or break with hardening = vis/cfi too? (same below)  

if not, leave a fixme, if at least vis works, add that

ifreund: libhangul fails to compile with vis unfortunately but seogi seems to work fine with vis and cfi enabled.

nekopsykose: thanks for testing!

PR #1165, opened 2023-12-30
usage: fix typos

paulapatience:


q66:
those are not even typos?

paulapatience:
Mentioning a `main/awk` package suggests that it exists. `main/gawk` actually does. I can change the commit message if you prefer.

q66:
it existed until recently, and besides, it wouldn't really matter much even if it doesn't

PR #1167, opened 2024-01-01
contrib/python-meson: new package (0.15.0)

a16bitsysop:


mhmdanas:
This seems like a library rather than an end-user application; is it needed for some other package? If not, I don't think this should be packaged right now.

a16bitsysop:
its needed for python-hypothesis, which is needed for checkdepends in python-pytest-asyncio, which is needed for checkdepends of python-pytest-mock

q66:
i'd prefer if you actually submitted that kind of stuff in a batch rather than individual PRs for individual templates without an obvious individual purpose

a16bitsysop:
okay will add some more commits

a16bitsysop:
added up to python-hypothesis, stuck in cannot import mock_xp.  Tried installing deps from requirements/test.txt with pip in chroot and the same

contrib/python-hypothesis/template.py
```diff
@@ -0,0 +1,39 @@
...
+pkgdesc = "Easy to use library for property-based testing"
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "MPL-2.0"
+url = "https://hypothesis.works/index.html"
+source = f"https://github.com/HypothesisWorks/hypothesis/archive/refs/tags/hypothesis-python-{pkgver}.zip"
+sha256 = "899688a4f7dcca9a6d32ce486a7765db9eef805ee4d84d3333fa6e06276e67b8"
+
+
+def init_build(self):
+    self.make_build_target = "hypothesis-python"
```
nekopsykose, line 28: this can be top level, doesn't have to be in init_build

contrib/python-hypothesis/template.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def init_check(self):
+    import sys
+    self.env["PYTHONPATH"] = "hypothesis-python/src"
+    self.env["PATH"] = "/usr/bin:" + "f{self.cwd}/testdir/usr/bin"
```
nekopsykose, line 34:
```suggestion
    self.env["PATH"] = "/usr/bin:" + f"{self.cwd}/testdir/usr/bin"
```

contrib/python-hypothesis/template.py
```diff
@@ -0,0 +1,39 @@
...
+    self.do("sh", "-c" ,"python3 -m installer --destdir=test_dir hypothesis-python/dist/*whl")
```
nekopsykose, line 35: the sh -c wrapper isn't needed at all, just call python -m installer itself normally

nekopsykose: that said it would be nice to make the default pep517 check do the equivalent of a new venv + installing the whl + running the tests, eh..

contrib/python-hypothesis/template.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def init_install(self):
+    self.make_install_target = "hypothesis-python/dist/*whl"
```
nekopsykose, line 39: can be toplevel

contrib/python-pexpect/template.py
```diff
@@ -0,0 +1,23 @@
...
+    "python-wheel",
+]
+checkdepends = ["bash", "mandoc", "python-ptyprocess", "python-pytest", "zsh"]
+pkgdesc = (
+    "Python module for controlling interactive programs in a pseudo-terminal"
+)
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "ISC"
+url = "https://pexpect.readthedocs.io/index.html"
+source = f"https://github.com/pexpect/pexpect/archive/refs/tags/{pkgver}.zip"
```
nekopsykose, line 18: .tar.gz

contrib/python-hypothesis/template.py
```diff
@@ -0,0 +1,39 @@
...
+    "python-pytest",
+    "python-pytest-xdist",
+    "python-sortedcontainers",
+]
+depends = ["python-attrs", "python-sortedcontainers"]
+pkgdesc = "Easy to use library for property-based testing"
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "MPL-2.0"
+url = "https://hypothesis.works/index.html"
+source = f"https://github.com/HypothesisWorks/hypothesis/archive/refs/tags/hypothesis-python-{pkgver}.zip"
```
nekopsykose, line 23: .tar.gz

contrib/python-ptyprocess/template.py
```diff
@@ -0,0 +1,22 @@
...
+    "python-installer",
+    "python-wheel",
+]
+checkdepends = ["bash", "python-pytest"]
+depends = ["python-flit_core"]
+pkgdesc = "Run a subprocess in a pseudo terminal"
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "ISC"
+url = "https://github.com/pexpect/ptyprocess/index.html"
+source = f"https://github.com/pexpect/ptyprocess/archive/refs/tags/{pkgver}.zip"
```
nekopsykose, line 17: .tar.gz (you get the picture)

contrib/python-ptyprocess/template.py
```diff
@@ -0,0 +1,22 @@
...
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-flit_core",
+    "python-installer",
+    "python-wheel",
+]
+checkdepends = ["bash", "python-pytest"]
+depends = ["python-flit_core"]
```
nekopsykose, line 12: this does not depend on flit_core at runtime

contrib/python-hypothesis/template.py
nekopsykose: looking at it, you might want to merely set `build_wrksrc = "hypothesis-python"` though, since everything is in that folder and the repo toplevel is not used

contrib/python-hypothesis/template.py
a16bitsysop: okay, yes would be easier.  Maybe activate venv as well?

a16bitsysop: without the shell it can’t find the whl, even with full path.  Will push

q66:
why close?

a16bitsysop:
accidentally deleted because github isn’t deleting branches when merged, I undeleted it.  Must have missed the checkbox to delete?

contrib/python-pytest-mock/template.py
```diff
@@ -0,0 +1,35 @@
+pkgname = "python-pytest-mock"
+pkgver = "3.12.0"
+pkgrel = 0
+build_style = "python_pep517"
+
```
nekopsykose, line 5:
```suggestion
```

contrib/python-pytest-mock/template.py
```diff
@@ -0,0 +1,35 @@
...
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
```
nekopsykose, line 9:
```suggestion
```

contrib/python-pytest-mock/template.py
```diff
@@ -0,0 +1,35 @@
...
+
+def init_build(self):
+    self.env["SETUPTOOLS_SCM_PRETEND_VERSION"] = f"{pkgver}"
+
+
+def init_check(self):
+    self.make_check_args = [
+        "--deselect",
+        "tests/test_pytest_mock.py::TestMockerStub::test_failure_message_with_",
+    ]
```
nekopsykose, line 31: make_check_args toplevel

contrib/python-pytest-mock/template.py
```diff
@@ -0,0 +1,35 @@
...
+pkgdesc = "Thin-wrapper around the mock package for easier use with pytest"
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "MIT"
+url = "https://pytest-mock.readthedocs.io/en/latest/index.html"
+source = f"https://github.com/pytest-dev/pytest-mock/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "aeedddfe2feefde936ba3d723e26db5757570537ab66bf4dd2cd17d4c263f022"
+
+
+def init_build(self):
+    self.env["SETUPTOOLS_SCM_PRETEND_VERSION"] = f"{pkgver}"
```
nekopsykose, line 24: if you use the pypi source you don't need this

the only reason to use git directly is for when the pypi source has the tests removed from it, but it has them here

contrib/python-pytest-mock/template.py
nekopsykose: also why does that fail?

contrib/python-meson/template.py
```diff
@@ -0,0 +1,31 @@
...
+    "python-wheel",
+]
+checkdepends = [
+    "python-cython",
+    "git",
+    "python-devel",
+    "python-pytest",
+    "python-pytest-mock",
+]
+depends = ["meson", "python-pyproject-metadata"]
```
nekopsykose, line 21: patchelf is a dependency at runtime too

contrib/python-meson/template.py
```diff
@@ -0,0 +1,31 @@
+pkgname = "python-meson"
+pkgver = "0.15.0"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "meson",
+    "ninja",
```
nekopsykose, line 7:
```suggestion
```

contrib/python-hypothesis/template.py
```diff
@@ -0,0 +1,67 @@
...
+# python-hypothesis-6.93.0-r0: forbidden directory 'builddir'
+# after building
+options = ["!check"]
+
+
+def init_check(self):
+    self.env["PATH"] = f"{self.chroot_cwd}/.cbuild-checkenv/bin:/usr/bin"
+    self.make_check_args = [
+        "--ignore",
+        "tests/array_api",
```
nekopsykose, line 42: you can do --ignore=.. to halve the linecount

it can also be toplevel and not in init_check

and the PATH seems not needed??

contrib/python-meson/template.py
```diff
@@ -0,0 +1,31 @@
...
+    "patchelf",
+    "python-build",
+    "python-installer",
+    "python-pyproject-metadata",
+    "python-wheel",
+]
+checkdepends = [
+    "python-cython",
+    "git",
```
nekopsykose, line 16:
```suggestion
    "git",
    "python-cython",
```

contrib/python-hypothesis/template.py
nekopsykose: also put -n self.make_jobs in here since you pulled xdist, since these tests take forever

nekopsykose: hm, i guess for some reason the path is indeed missing from some of the tests

contrib/python-hypothesis/template.py
```diff
@@ -0,0 +1,67 @@
...
+    "python-dateutil",
+    "python-pexpect",
+    "python-ptyprocess",
+    "python-pytest",
+    "python-pytz",
+    "python-pytest-xdist",
+    "python-sortedcontainers",
+    "python-typing_extensions",
+]
+depends = ["python-attrs", "python-sortedcontainers"]
```
nekopsykose, line 25: depends above checkdepends, then you can add it to the checkdepends list

contrib/python-hypothesis/template.py
```diff
@@ -0,0 +1,67 @@
...
+pkgdesc = "Easy to use library for property-based testing"
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "MPL-2.0"
+url = "https://hypothesis.works/index.html"
+source = f"https://github.com/HypothesisWorks/hypothesis/archive/refs/tags/hypothesis-python-{pkgver}.tar.gz"
+sha256 = "80ee73f86f0c48f65a762e03db240e7d4a4f41d6c15d06f4425e3a0bf38c0403"
+# with check link gives the error:
+# python-hypothesis-6.93.0-r0: forbidden directory 'builddir'
+# after building
```
nekopsykose, line 34: that's because it literally installs "builddir/python-hypothesis-6.96.2/hypothesis-python/" etc into the root, hm

nekopsykose: not sure why that happens, the .whl looks fine and it's passed to install fine..

nekopsykose: ah, it's because of the PATH= thing in init check. it runs the venv python, and running ./.venv/python -m installer ... with --destdir ends up installing to the same 'location' as the venv inside the destdir

guess you have to not override the path like that

nekopsykose: you can probably append the venv path after /usr/bin, then python still comes first, but `hypothesis` is found. but i really don't understand why the tests don't have it in path, the path should be updated when running with the venv python..

contrib/python-pytest-mock/template.py
a16bitsysop: okay

contrib/python-pytest-mock/template.py
a16bitsysop: this is the error https://github.com/chimera-linux/cports/actions/runs/7601459852/job/20700867634#step:8:725

nekopsykose: https://github.com/pytest-dev/pytest-mock/pull/403

a16bitsysop: yes thats it, guess its not worth adding 3 commits to fix the test

nekopsykose: it's one commit https://github.com/pytest-dev/pytest-mock/commit/3d48ff90257daf6b3c50100bd5dd36faf34e70dc

contrib/python-hypothesis/template.py
a16bitsysop: okay

contrib/python-hypothesis/template.py
a16bitsysop: I think they were different errors, so harder to bunch into fewer ignores

contrib/python-hypothesis/template.py
q66: i really don't like leaking the checkenv path into the template like this fwiw

q66: 8adeed69f785e5e6cc13892b4d342bfb199cb854 should probably be sufficient to cleanly get rid of this altogether?

a16bitsysop: without the path the tests can’t find hypothesis, added a commit with it commented out

nekopsykose: that commit automatically adds the path by default so it is there though

contrib/python-hypothesis/template.py
nekopsykose: you still gotta do --ignore=thing instead of making it two lines

q66: i assume `-n {self.make_jobs}` also does not work as a single argument

nekopsykose: eyah, could do `--numprocesses={}` instead

a16bitsysop: okay changed it, do you mean change to `-n{self.make_jobs}`

contrib/python-hypothesis/template.py
```diff
@@ -0,0 +1,48 @@
...
+    "python-click",
+    "python-dateutil",
+    "python-pexpect",
+    "python-ptyprocess",
+    "python-pytest",
+    "python-pytz",
+    "python-pytest-xdist",
+    "python-typing_extensions",
+] + depends
+pkgdesc = "Easy to use library for property-based testing"
```
nekopsykose, line 24:
```suggestion
pkgdesc = "Python library for property-based testing"
```

contrib/python-hypothesis/template.py
```diff
@@ -0,0 +1,48 @@
...
+depends = ["python-attrs", "python-sortedcontainers"]
+checkdepends = [
+    "python-black",
+    "python-click",
+    "python-dateutil",
+    "python-pexpect",
+    "python-ptyprocess",
+    "python-pytest",
+    "python-pytz",
+    "python-pytest-xdist",
```
nekopsykose, line 21:
```suggestion
    "python-pytest-xdist",
    "python-pytz",
```

contrib/python-pytest-mock/template.py
nekopsykose: you can use https://github.com/pytest-dev/pytest-mock/commit/6da5b0506d6378a8dbe5ae314d5134e6868aeabd.patch to shrink the patch to 1 commit too

a16bitsysop: okay

PR #1168, opened 2024-01-01
contrib/sbcl: new package (2.4.0)

paulapatience:


linfocito:
Don't know if it is relevant to the fail, but there is this comment about ECL bootstrap in Void's template for sbcl:

```
i686*|ppc*|*-musl)
	# ecl bootstrap is broken since 2.2.0:
	# https://build.voidlinux.org/builders/x86_64-musl_builder/builds/39694/steps/shell_3/logs/stdio
	makedepends+=" clisp"
	_bootstrap_lisp="clisp"
```

paulapatience:
Void's build log indicates that building SBCL failed, whereas here the build succeeds but the tests seem to take too long. On my machine the tests pass, and in much less time than 6 hours. I do not know what the problem could be.

paulapatience:
Added a patch to fix the riscv64 build, which I suspect was failing.

q66:
tests get stuck on aarch64: https://gist.github.com/q66/047bc00e4bccf5a8c2f7874984a294a6

q66:
The following patch is required on ppc64le to get it to 2nd-stage build:

```
Using PIE results in bad relocations for ppc64-assem.S:

ld: error: relocation R_PPC64_ADDR16_HI cannot be used against local symbol; recompile with -fPIC
>>> defined in ppc64-assem.o
>>> referenced by ppc64-assem.S:174 (./src/runtime/ppc64-assem.S:174)
>>>               ppc64-assem.o:(call_into_lisp)

ld: error: relocation R_PPC64_ADDR16_LO cannot be used against local symbol; recompile with -fPIC
>>> defined in ppc64-assem.o
>>> referenced by ppc64-assem.S:175 (./src/runtime/ppc64-assem.S:175)
>>>               ppc64-assem.o:(call_into_lisp)

The affected asm:

	lis reg_LRA,lra@h
	ori reg_LRA,reg_LRA,lra@l

The other changes are related to incomplete definitions of struct pt_regs.

--- a/src/runtime/Config.ppc64-linux
+++ b/src/runtime/Config.ppc64-linux
@@ -10,7 +10,7 @@
 # files for more information.
 
 CFLAGS += -m64 -std=gnu99
-LINKFLAGS += -m64 -Wl,--export-dynamic
+LINKFLAGS += -m64 -Wl,--export-dynamic -no-pie
 
 ASSEM_SRC = ppc64-assem.S
 ARCH_SRC = ppc-arch.c
--- a/src/runtime/os-common.c
+++ b/src/runtime/os-common.c
@@ -34,6 +34,10 @@
 #include "gc.h" // for find_page_index
 #endif
 
+#ifdef __powerpc__
+#include <asm/ptrace.h>
+#endif
+
 /*
  * historically, this used sysconf to select the runtime page size
  * per recent changes on other arches and discussion on sbcl-devel,
--- a/src/runtime/ppc-arch.c
+++ b/src/runtime/ppc-arch.c
@@ -10,6 +10,7 @@
  */
 
 #include <stdio.h>
+#include <asm/ptrace.h>
 
 #include "genesis/sbcl.h"
 #include "arch.h"
--- a/src/runtime/ppc-linux-os.c
+++ b/src/runtime/ppc-linux-os.c
@@ -32,6 +32,7 @@
 #include <sys/stat.h>
 #include <unistd.h>
 #include <sys/prctl.h>
+#include <asm/ptrace.h>
 
 #include "validate.h"
 #include "ppc-linux-mcontext.h"
```

It is currently building the rest, so no idea about tests and so on yet.

q66:
Additionally, `!lto` should definitely be applied here.

q66:
ppc64le test fail: https://gist.github.com/q66/cb141136000622aef1db450ebd3810cb

actually seems to be the same test aarch64 gets stuck on here; ppc64le just doesn't get stuck, and fails/crashes.

paulapatience:
Added !lto and your patch. I've updated the version to 2.4.2, which makes obsolete two of the patches I had previously applied.

Can you test with 2.4.2 to see if it fixes the tests? Otherwise one of the SBCL maintainers requested that we try with the latest commit.

PR #1169, opened 2024-01-02
Mark more non-meta packages as empty

JamiKettunen:
This should hopefully cover the rest of the empty packages in the repo (checked x86_64 apks size <=16K)

@psykose I'm not sure if the `volk` pkg should be empty?

also @q66 should `linux-headers-cross` be updated? I personally don't see a reason to have it lag behind `linux-headers` but maybe I'm missing something.

perhaps there could be a comment at the top of `linux-headers` template to remind about updating `linux-headers-cross` if they should be in sync

as for the rest OpenJDK patch was updated with latest version from https://github.com/openjdk/jdk17u-dev/pull/1906 and util-linux change was already untagged upstream as https://github.com/util-linux/util-linux/commit/77454e58d58f904cfdc02d3ca5bb65f1bd8739fc


nekopsykose:
> I'm not sure if the volk pkg should be empty?  

they only ship a static lib so it ends up that way

q66:
that's not the right fix then, there should be only one package in the first place

PR #1170, opened 2024-01-02
contrib/txr: new package (293)

paulapatience:


contrib/txr/template.py
```diff
@@ -0,0 +1,34 @@
...
+hostmakedepends = ["bash", "gmake"]
+makedepends = ["libffi-devel", "zlib-devel"]
+pkgdesc = "Data munging language"
+maintainer = "Paul A. Patience <paul@apatience.com>"
+license = "custom:txr"
+url = "https://www.nongnu.org/txr"
+source = f"https://www.kylheku.com/cgit/{pkgname}/snapshot/{pkgname}-{pkgver}.tar.bz2"
+sha256 = "6fc21ae7332f98f97af35ad3ca1808d0043c4c85384c4e7bebcfce967e36fa5c"
+hardening = ["vis"]
+options = ["!cross", "!lto"]
```
nekopsykose, line 17: is this actually !cross and !lto?

paulapatience: Yes. LTO causes issues with the GC; some tests fail. As for cross, TXR builds its stdlib by invoking the `txr` binary, so if I read `Packaging.md` correctly, it is not eligible for cross-compilation.

nekopsykose: sounds good :)

PR #1179, opened 2024-01-04
contrib/swww: install needed binary

Froggo8311:
The `swww-daemon` binary wasn't being installed when it needed to be


wezm:
Probably want to bump `pkgrel`

Froggo8311:
@wezm Thanks, forgot about that!

PR #1181, opened 2024-01-04
contrib/zzz: new package (20231124)

ifreund:
I realize that this is just a shell script and therefore perhaps not worth packaging. However, I wanted to allow running `doas zzz` with no password for use in scripts and realized that I shouldn't do that without having the zzz script in a root owned system directory. Therefore I made this package for my personal usage and thought it at the very least worth sharing as there is afaik currently no other user friendly option to suspend the system without logind installed.

Note that https://github.com/jirutka/zzz also exists but I personally would rather use a small shell script written by someone who really knows what they are doing rather than a significantly longer C program of middling quality.

### DIFF
--- a/contrib/zzz/template.py
+++ b/contrib/zzz/template.py
@@ -1,0 +1,14 @@
+pkgname = "zzz"
+pkgver = "20231124"
+pkgrel = 0
+pkgdesc = "Simple script to suspend or hibernate your computer"
+maintainer = "Isaac Freund <mail@isaacfreund.com>"
+license = "custom:none"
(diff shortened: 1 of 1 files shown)


q66:
i don't want this, it's not good enough (a proper implementation would need to be in a real language and presumably written around libturnstile for central inhibition support) and if added now i'd just remove it later anyway (considering it's a shell script <100LOC i see no reason why it needs to be packaged regardless)

ifreund:
Fair enough, I too would prefer a turnstile implementation long term of course :)

PR #1183, opened 2024-01-04
contrib/thermald: new package (2.5.4)

mikoxyz:


contrib/thermald/files/thermald
```diff
@@ -0,0 +1,5 @@
+# thermald service
+
+type		= process
+command		= /usr/bin/thermald --adaptive --dbus-enable --no-daemon
+depends-on	= dbus
```
nekopsykose, line 5:
```suggestion
type        = process
command     = /usr/bin/thermald --adaptive --dbus-enable --no-daemon
depends-on  = dbus
```
spaces not tabs

contrib/thermald/template.py
```diff
@@ -0,0 +1,44 @@
...
+    "pkgconf",
+]
+makedepends = [
+    "dbus-devel",
+    "dbus-glib-devel",
+    "glib-devel",
+    "libevdev-devel",
+    "xz-devel",
+    "libxml2-devel",
+    "upower-devel",
```
nekopsykose, line 26:
```suggestion
    "dbus-devel",
    "dbus-glib-devel",
    "glib-devel",
    "libevdev-devel",
    "xz-devel",
    "libxml2-devel",
    "upower-devel",
```

`|sort`

contrib/thermald/template.py
```diff
@@ -0,0 +1,44 @@
...
+pkgdesc = "Thermal daemon for x86_64-based Intel CPUs"
+maintainer = "miko <mikoxyzzz@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://github.com/intel/thermal_daemon"
+source = f"https://github.com/intel/thermal_daemon/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "2addedfc7974b30ed6a553a5690961eb48d6abacd4584e3b6d3fd292343d694e"
+hardening = ["vis"]
+
+
+# necessary since we don't want to run thermald's silly autogen
```
nekopsykose, line 37: the autogen looks fine though, you could do `configure_gen = ["NOCONFIGURE=1", "./autogen.sh"]` or whatnot

contrib/thermald/template.py
```diff
@@ -0,0 +1,44 @@
+pkgname = "thermald"
+pkgver = "2.5.4"
+pkgrel = 0
+archs = [
+    "x86_64",
+]
```
nekopsykose, line 6:
```suggestion
archs = ["x86_64"]
```

don't need a list for 1 entry

contrib/thermald/template.py
```diff
@@ -0,0 +1,44 @@
...
+    "libevdev-devel",
+    "xz-devel",
+    "libxml2-devel",
+    "upower-devel",
+]
+pkgdesc = "Thermal daemon for x86_64-based Intel CPUs"
+maintainer = "miko <mikoxyzzz@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://github.com/intel/thermal_daemon"
+source = f"https://github.com/intel/thermal_daemon/archive/refs/tags/v{pkgver}.tar.gz"
```
nekopsykose, line 32:
```suggestion
source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
```

contrib/thermald/template.py
mikoxyz: nope, doesn't work; it'll try to run `NOCONFIGURE=1` as a command instead of interpreting it as an env variable :<

i did try to add `NOCONFIGURE=1` to env but that seemed to screw some things up (specifically `MAKE` wasn't set in do_build for whatever reason) :<

nekopsykose: > nope, doesn't work; it'll try to run NOCONFIGURE=1 as a command instead of interpreting it as an env variable :<  

ah right. you want `configure_env = {"NOCONFIGURE": "1"}`

nekopsykose: you are right it funnily breaks though

mikoxyz: !

i'll fix that tomorrow. thanks!

nekopsykose: yeah idk why it's so broken, but you can use just

```
build_style = "gnu_configure"
configure_args = ["--disable-dependency-tracking"]
configure_env = {"NO_CONFIGURE": "1"}
configure_gen = ["./autogen.sh"]
make_cmd = "gmake"
```

and that works i guess

contrib/thermald/template.py
```diff
@@ -0,0 +1,42 @@
+pkgname = "thermald"
+pkgver = "2.5.4"
+pkgrel = 0
+archs = ["x86_64"]
+build_style = "gnu_configure"
+
```
q66, line 6: this extra line is wrong

PR #1185, opened 2024-01-06
contrib/zoxide: use install_completion

paulapatience:


q66:
for next time: simple refactorings that have no effect on resulting packages don't warrant pkgrel increase

PR #1186, opened 2024-01-06
contrib/frawk: new package (0.4.7)

paulapatience:
Although the package conditionally enables architecture-specific code, it always has a generic fallback, so I did not add the `archs` line.

### DIFF
--- a/contrib/frawk/template.py
+++ b/contrib/frawk/template.py
@@ -1,0 +1,31 @@
+pkgname = "frawk"
+pkgver = "0.4.7"
+pkgrel = 0
+build_style = "cargo"
+# disable llvm_backend (needs LLVM 12), unstable and use-jemalloc features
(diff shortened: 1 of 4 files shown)


contrib/frawk/template.py
```diff
@@ -0,0 +1,31 @@
+pkgname = "frawk"
+pkgver = "0.4.7"
+pkgrel = 0
+build_style = "cargo"
+# disable llvm_backend (needs LLVM 12), unstable and use-jemalloc features
+make_build_args = ["--no-default-features", "--features=allow_avx2"]
+make_install_args = make_build_args
+make_check_args = make_build_args
```
nekopsykose, line 8:
```suggestion
make_install_args = list(make_build_args)
make_check_args = list(make_build_args)
```

PR #1188, opened 2024-01-06
readme: link to other documents

firefly-cpp:
I recommend adding links in README for faster file navigation.


nekopsykose:
commit should be named something like `readme: link to other documents`

firefly-cpp:
@nekopsykose, done

PR #1190, opened 2024-01-07
contrib/clickclack: new package (0.2.3)

Froggo8311:


contrib/clickclack/update.py
```diff
@@ -0,0 +1 @@
+pattern = "(?!archive\\/)[0-9\\.]*(\\.tar\\.gz)"
```
nekopsykose, line 1: this returns a version that's just .tar.gz:  

```
Checking found version: .tar.gz
```

Froggo8311: tf?? that shouldnt happen lol

Froggo8311: OH

Froggo8311:
oops, my dumbass accidentally used a negative lookahead group instead of a non-capturing group

nekopsykose:
you need a group for the version part, you've done the opposite of how it works

contrib/clickclack/update.py
```diff
@@ -0,0 +1 @@
+pattern = "\\/archive\\/([\\d\\.]+)\\.tar\\.gz"
```
nekopsykose, line 1:
```suggestion
pattern = r"/archive/([\d\.]+)\.tar\.gz"
```

nekopsykose: ah, the page doesn't even have version urls at all

PR #1191, opened 2024-01-07
contrib/btop: new package (1.3.0)

nekopsykose:
seems this one doesn't crash


nekopsykose:
hm no, still happens after long enough, ehh...

nekopsykose:
hopefully will fix itself eventually with libc++18 or something, wontbebaitedagain

PR #1193, opened 2024-01-08
main/ncurses: delete terminfo provided by terminals directly

nekopsykose:
there's another two options too:  

- renaming the terminfo provided by the terminal to something like `foot-extra` as a choice for TERM later. that works, but i'm not sure there's much point in providing two..
- deleting the terminal one

generally terminals upstream their terminfo to ncurses. but the terminal one always updates first- and updates along with any 'new features'. a mismatch in either side being more current (terminal newer than ncurses terminfo or ncurses terminfo newer than the terminal) isn't good (and can lead to bugs due to feature mismatch), so i don't see a point to prefer the ncurses-term version instead of keeping the terminal-shipped one only (and personally i don't like this one-person centralised terminfo DB idea anyway). that said, other distros usually don't do this (picking one of the other two options), so maybe i'm missing some crucial info as to why it wouldn't be a good idea. the only negative i can think of is that installing `ncurses-term` won't give you "the terminfo for everything ever in one package"


q66:
hm i just wonder if we're gonna have to keep adding to the removals whenever a new terminal is added, which sounds kind of bad

q66:
i guess we could do both, i.e. make the terminfo packages replace ncurses-term, as well as remove the terminal from ncurses-term; that should address the primary issue with just removing it (which is apk conflicts when you have pre-removal ncurses-term installed and you run an upgrade)

can you think of any cases when version mismatch actually broke something?

nekopsykose:
> can you think of any cases when version mismatch actually broke something?

hrm.. the thing i had vaguely in memory was actually not a version mismatch at all (https://github.com/vim/vim/issues/11766) but rather just an application making use of an advertised cap (bracketed paste) misusing it upon it being advertised on terminfo update (with the terminal supporting it), and later being fixed in the application itself. this isn't a terminfo mismatch but rather an application bug..

i can't think of anything else offhand. i could just be wrong then :)  

with that i guess it's up to you, i don't mind adding replaces= to foot/alacritty terminfo, or removing it from them to keep ncurses.

nekopsykose:
from reading https://codeberg.org/dnkl/foot/issues/1367 this seems even more funny than either of those options:  

- for foot, upstream ncurses actually updates first and foot imports it later
- upstream ncurses is /wrong/ when making it up itself (and in more than one case; like in the original addition https://codeberg.org/dnkl/foot/issues/670)

sigh. i guess the best choice on our side then is:  

- for alacritty, delete the shipped one, depend on ncurses-term. these never had issues afaict and the shipped one is identical to upstream anyway
- for foot, install the shipped one as foot-extra (like arch/alpine etc) just in case it's needed due to tomfoolery like above, and depend on ncurses-term for the canonical default `foot` TERM

PR #1196, opened 2024-01-08
jsonschema

mikoxyz:


contrib/python-hatch-fancy-pypi-readme/template.py
```diff
@@ -0,0 +1,27 @@
...
+build_style = "python_pep517"
+hostmakedepends = [
+    "python",
+    "python-build",
+    "python-hatchling",
+    "python-installer",
+    "python-wheel",
+]
+depends = [
+    "python",
```
nekopsykose, line 13: don't need python also when another python dep is there, same for hostmakedepends

mikoxyz: mhm, oki

contrib/python-hatch-fancy-pypi-readme/template.py
```diff
@@ -0,0 +1,27 @@
+pkgname = "python-hatch-fancy-pypi-readme"
```
nekopsykose, line 1: we patched this out everywhere so far, but i guess we could just keep it, since it's a tiny module and saves some patches (with no downside to having it in build hostdeps)

contrib/python-jsonschema/template.py
```diff
@@ -0,0 +1,33 @@
...
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python",
+    "python-build",
+    "python-hatch-fancy-pypi-readme",
+    "python-hatch_vcs",
+    "python-hatchling",
+    "python-installer",
+    "python-pytest",
```
nekopsykose, line 12: pytest in checkdepends for all of them

contrib/python-maturin/template.py
```diff
@@ -0,0 +1,39 @@
...
+def post_patch(self):
+    from cbuild.util import cargo
+
+    self.cargo = cargo.Cargo(self)
+    self.cargo.vendor()
+    cargo.setup_vendor(self)
+
+
+def post_install(self):
+    self.install_license("license-apache")
```
nekopsykose, line 38: don't need to install apache

contrib/python-referencing/template.py
```diff
@@ -0,0 +1,30 @@
...
+pkgver = "0.32.1"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python",
+    "python-build",
+    "python-hatch_vcs",
+    "python-hatchling",
+    "python-installer",
+    "python-rpds",
```
nekopsykose, line 11: this doesn't need to be in hostdeps, you can pass --skip-dependency-check in make_build_args if it's build that complains

contrib/python-referencing/template.py
```diff
@@ -0,0 +1,30 @@
...
+    "python-attrs",
+    "python-rpds",
+]
+pkgdesc = "Cross-specification JSON referencing"
+maintainer = "miko <mikoxyzzz@gmail.com>"
+license = "MIT"
+url = "https://github.com/python-jsonschema/referencing"
+source = f"{url}/releases/download/v{pkgver}/referencing-{pkgver}.tar.gz"
+sha256 = "3c57da0513e9563eb7e203ebe9bb3a1b509b042016433bd1e45a2853466c3dd3"
+# test_sanity.py causes a circular dependency since it depends on jsonschema, which depends on this package
```
nekopsykose, line 25: you could --deselect this one test in that case

contrib/python-hatch-fancy-pypi-readme/template.py
q66: i'd prefer patching it out, since it's entirely useless, and it's very little work to patch it out

contrib/python-jsonschema-specifications/template.py
```diff
@@ -0,0 +1,50 @@
...
+# get rid of this once pep517's check stuff is fixed up
+def pre_check(self):
+    whl = list(
+        map(
+            lambda p: str(p.relative_to(self.cwd)),
+            self.cwd.glob(self.make_install_target),
+        )
+    )
+
+    self.do(
```
q66, line 33: do we really need this venv nonsense for every single one of these? because it hasn't been needed for literally anything until now

contrib/python-maturin/template.py
```diff
@@ -0,0 +1,35 @@
+pkgname = "python-maturin"
```
q66, line 1: on a second thought, split specifically this into its own PR and leave the others in this one

q66: while at it, this does not bother to properly set features (`--features=full,native-tls,password-storage`), which need to be set (because we don't want to build with rustls)

contrib/python-jsonschema-specifications/template.py
nekopsykose: no, just the native-module ones. but you could make it the default for do_check and it would be an improvement like i explained yesterday

nekopsykose: > no, just the native-module ones.  

(or to be specific- it depends on the build-backend, setuptools keeps it in build/lib.* too. but maturin does not persist a whole pythonpath import on disk and only gives you a wheel. with a venv though you can always 'just install the wheel' and not care about any of the specifics, like how in some templates we add PYTHONPATH to the build/lib folder, etc)

q66:
i've merged maturin and i've done the venv stuff for check earlier too, so you can remove all that and rebase this

contrib/python-jsonschema-specifications/template.py
```diff
@@ -0,0 +1,25 @@
...
+    "python-hatch_vcs",
+    "python-hatchling",
+    "python-installer",
+]
+checkdepends = [
+    "python-referencing",
+    "python-pytest",
+]
+depends = ["python-referencing"]
+pkgdesc = "Json Schema meta-schemas and vocabularioes, exposed as a Registry"
```
q66, line 16: vocabularioes

contrib/python-rpds/template.py
```diff
@@ -0,0 +1,38 @@
...
+url = "https://github.com/crate-py/rpds"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "10d04e7160e5973c97bc371c1bb26b10d91385d532d4b9d13d724ff4d4f23819"
+
+
+def post_patch(self):
+    from cbuild.util import cargo
+
+    self.cargo = cargo.Cargo(self)
+    self.cargo.vendor()
```
q66, line 33: cargo.Cargo(self).vendor()

q66:
also prefer $(PYPI_SITE) for source urls, like most of the other python packages

PR #1197, opened 2024-01-08
contrib/btrbk: new package (0.32.6)

z-erica:
also includes its dependencies contrib/mbuffer (20231216) and contrib/asciidoctor (2.0.20)

### DIFF
--- a/contrib/btrbk/patches/fix_illegal_time_format.patch
+++ b/contrib/btrbk/patches/fix_illegal_time_format.patch
@@ -1,0 +1,11 @@
+--- a/doc/Makefile.orig	2024-01-08 20:20:50.584116218 +0100
++++ b/doc/Makefile	2024-01-08 20:21:40.783645535 +0100
+@@ -23,7 +23,7 @@
+ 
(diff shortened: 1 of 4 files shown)


contrib/asciidoctor/template.py
```diff
@@ -0,0 +1,45 @@
...
+pkgdesc = "Ruby-based AsciiDoc converter"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "MIT"
+url = "https://asciidoctor.org"
+source = f"!https://rubygems.org/downloads/{pkgname}-{pkgver}.gem"
+sha256 = "835eabd445e4ae88f56a5f4e07593c3612b2be72eb661c612c3a8e1e17c57479"
+
+
+def do_extract(self):
+    self.cp(self.sources_path / f"{pkgname}-{pkgver}.gem", self.builddir)
```
nekopsykose, line 15: from what i can tell there's no point to copy this?

nekopsykose:
could you separate asciidoctor to a separate mr? there's something else i'd like to use it for too so we can merge it separately

PR #1200, opened 2024-01-09
contrib/shfmt: new package (3.7.0)

nekopsykose:


contrib/shfmt/template.py
```diff
@@ -0,0 +1,29 @@
...
+pkgver = "3.7.0"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    "-ldflags",
+    f"-X main.version={pkgver}",
+    "./cmd/shfmt",
+]
+make_check_args = ["./cmd/shfmt/..."]
+hostmakedepends = [
```
q66, line 11: since it's two entries only, make it one line

PR #1201, opened 2024-01-09
contrib/eza: new package (0.17.0)

nekopsykose:


contrib/eza/template.py
```diff
@@ -0,0 +1,28 @@
...
+hostmakedepends = [
+    "cargo",
+    "pkgconf",
+]
+makedepends = [
+    "libgit2-devel",
+    "rust-std",
+    "zlib-devel",
+]
+pkgdesc = "Modern ls replacement"
```
q66, line 14: pkgdesc bad

contrib/eza/template.py
```diff
@@ -0,0 +1,28 @@
+pkgname = "eza"
+pkgver = "0.17.0"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = [
```
q66, line 5: probably one line

contrib/eza/template.py
nekopsykose: what would you like it to be

q66: make something up, but Subjective Adjectives shouldn't go in pkgdescs; pkgdesc should be neutral and descriptive ("Directory listing utility" or whatever)

nekopsykose: when they're that small it kinda sounds really weird for some reason, but okie

q66: i mean it's just as small as the previous one

PR #1202, opened 2024-01-09
contrib/chicken: new package (5.3.0)

z-erica:


contrib/chicken/template.py
```diff
@@ -0,0 +1,32 @@
...
+make_cmd = "gmake"
+make_env = {"C_COMPILER": "clang", "CXX_COMPILER": "clang++"}
+hostmakedepends = ["gmake"]
+pkgdesc = "Practical and portable Scheme system"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "BSD-3-Clause"
+url = "https://call-cc.org"
+source = f"https://code.call-cc.org/releases/{pkgver}/chicken-{pkgver}.tar.gz"
+sha256 = "c3ad99d8f9e17ed810912ef981ac3b0c2e2f46fb0ecc033b5c3b6dca1bdb0d76"
+options = ["!parallel"]
```
q66, line 14: why?

contrib/chicken/template.py
```diff
@@ -0,0 +1,32 @@
...
+
+def post_install(self):
+    self.install_license("LICENSE")
+    self.mv(
+        self.destdir / "usr/share/chicken/doc",
+        self.destdir / "usr/share/doc/chicken",
+    )
+
+
+@subpackage("chicken-static")
```
q66, line 25: definitely not wanted; do you mean chicken-devel instead?

q66:
the build also respects neither cbuild-provided CFLAGS nor LDFLAGS

contrib/chicken/template.py
```diff
@@ -0,0 +1,41 @@
+pkgname = "chicken"
+pkgver = "5.3.0"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+hostmakedepends = ["gmake"]
+depends = ["chicken-devel"]
```
q66, line 7: `"chicken-devel={pkgver}-r{pkgrel}"`

contrib/chicken/template.py
```diff
@@ -0,0 +1,41 @@
...
+def post_install(self):
+    self.install_license("LICENSE")
+    self.mv(
+        self.destdir / "usr/share/chicken/doc",
+        self.destdir / "usr/share/doc/chicken",
+    )
+
+
+@subpackage("chicken-devel")
+def _static(self):
```
q66, line 35: rename function

q66:
should the stdlib (?) parts in /usr/lib/chicken be a part of the libs package?

q66:
from what i can find these are modules to be imported from programs, so they need to be available potentially at runtime so they can be used by programs; so they should probably go in -libs, so that programs can be used without installing the whole compiler

q66:
imo the documentation is not clear at all, and i don't understand how these are compile-time only if they are shared libs; could you perhaps join chicken's irc or equivalent (if they have it) and verify with somebody familiar with this?

PR #1206, opened 2024-01-10
contrib/ibus-skk: new package (1.4.3)

z-erica:


contrib/skkdict/template.py
```diff
@@ -0,0 +1,17 @@
+pkgname = "skkdict"
+pkgver = "1_git20231226"
```
nekopsykose, line 2:
```suggestion
pkgver = "0_git20231226"
```

stuff with no version starts at 0 for this (and if there's a 'latest' version, then it would be that version with _git appended, etc)

contrib/skkdict/template.py
```diff
@@ -0,0 +1,17 @@
...
+license = "Unicode-DFS-2016"
+url = "http://openlab.jp/skk/dic.html"
+source = f"https://github.com/skk-dev/dict/archive/{_commit}.tar.gz"
+sha256 = "c754fba1e7857151626216a837e4c83d254826f89bd4d73bc857e0dd2a9adda0"
+# no tests
+options = ["!check"]
+
+
+def do_install(self):
+    self.install_dir("usr/share/skk")
```
nekopsykose, line 16: the install_file below already makes the directory so you don't need this

contrib/libskk/template.py
```diff
@@ -0,0 +1,39 @@
...
+make_cmd = "gmake"
+hostmakedepends = [
+    "gettext-devel",
+    "pkgconf",
+    "gobject-introspection",
+    "vala",
+    "gmake",
+]
+makedepends = [
+    "libgee-devel",
```
nekopsykose, line 15: sort these lists

contrib/libskk/template.py
```diff
@@ -0,0 +1,39 @@
...
+    "json-glib-devel",
+    "libxkbcommon-devel",
+    "vala-devel",
+]
+depends = ["skkdict"]
+pkgdesc = "Japanese input method library"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/ueno/libskk"
+source = f"https://github.com/ueno/libskk/releases/download/{pkgver}/libskk-{pkgver}.tar.xz"
```
nekopsykose, line 25:
```suggestion
source = f"{url}/releases/download/{pkgver}/libskk-{pkgver}.tar.xz"
```

contrib/libskk/template.py
```diff
@@ -0,0 +1,39 @@
+pkgname = "libskk"
+pkgver = "1.0.5"
+pkgrel = 0
+build_style = "configure"
+configure_args = ["--prefix=/usr"]
```
nekopsykose, line 5: this is gnu_configure, and for it you won't need --prefix

it's quite old so it might be hard to reconf, in that case configure_gen = []. but it looks old enough to need a reconf to work correctly with new clang anyway..

contrib/ibus-skk/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "ibus-skk"
+pkgver = "1.4.3"
+pkgrel = 0
+build_style = "configure"
```
nekopsykose, line 4: same here, also gnu_configure

nekopsykose: probably also needs make_dir = "." then

contrib/libskk/template.py
```diff
@@ -0,0 +1,39 @@
...
+source = f"{url}/releases/download/{pkgver}/libskk-{pkgver}.tar.xz"
+sha256 = "b295eb9c248e353750c96474b2b55c4e627524fc6dc79439ff56f73ba09c3949"
+
+
+@subpackage("libskk-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+tool_flags = {
```
q66, line 34: wrong place, should be in the main metadata block or at very least not after any subpackage definitions

contrib/skkdict/template.py
```diff
@@ -0,0 +1,16 @@
...
+license = "Unicode-DFS-2016"
+url = "http://openlab.jp/skk/dic.html"
+source = f"https://github.com/skk-dev/dict/archive/{_commit}.tar.gz"
+sha256 = "c754fba1e7857151626216a837e4c83d254826f89bd4d73bc857e0dd2a9adda0"
+# no tests
+options = ["!check"]
+
+
+def do_install(self):
+    self.install_file("SKK-JISYO.*", "usr/share/skk/", glob=True)
```
q66, line 16: skip trailing slash

PR #1208, opened 2024-01-10
contrib/gopass: new package (1.15.11)

cal-andrew:


contrib/gopass/template.py
```diff
@@ -0,0 +1,22 @@
+pkgname = "gopass"
+pkgver = "1.15.11"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go"]
+pkgdesc = "Slightly more awesome standard UNIX password manager for teams"
```
q66, line 6: pkgdescs should be terse, descriptive, and not subjective; i.e. what is "awesome" is not decided by the author but by users; "standard" is highly debatable; etc.

cal-andrew: I changed it to 'standard unix password manager for teams'.
Password-store has the description 'standard unix password manager', so that should be good.

q66: that should be changed then

cal-andrew: is advanced console-based password manager good?

nekopsykose: presumably just "console-based password manager", but then they would both have the same pkgdesc. do we care about them being unique?

cal-andrew: pass could be "simple console-based password manager" cause gopass has more features

q66: i don't really care much about them being unique, maybe `Pass-compatible password manager with more features` or something

cal-andrew: that sounds good, i've updated the pkgdesc to that

nekopsykose:
separate commit for password-store

also it needs a pkgrel bump since the desc metadata changed

q66:
lgtm

cal-andrew:
thanks!

PR #1210, opened 2024-01-10
user/rspamd: new package (3.7.5)

a16bitsysop:


contrib/fasttext/patches/no-march-native.patch
```diff
@@ -0,0 +1,24 @@
...
+     src/args.h
+--- a/setup.py
++++ b/setup.py
+@@ -75,8 +75,8 @@
+             FASTTEXT_SRC,
+         ],
+         language='c++',
+-        extra_compile_args=["-O0 -fno-inline -fprofile-arcs -pthread -march=native" if coverage else
+-                            "-O3 -funroll-loops -pthread -march=native"],
++        extra_compile_args=["-O0 -fno-inline -fprofile-arcs -pthread" if coverage else
```
q66, line 20: these should *all* be patched out

contrib/fasttext/template.py
```diff
@@ -0,0 +1,22 @@
...
+build_style = "cmake"
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+pkgdesc = "Library for fast text representation and classification"
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "MIT"
+url = "https://fasttext.cc/index.html"
+source = f"https://github.com/facebookresearch/fasttext/archive/refs/tags/v{pkgver}.zip"
+sha256 = "93cc8a21633e6c6ecf4b5164925ad68a5eda332373d4316d2f56d05f4bfbe823"
+# Tests require 300+MB test data
+# options = ["!check"]
```
q66, line 13: unwanted comment

contrib/rspamd/files/rspamd
```diff
@@ -0,0 +1,9 @@
+# rspamd service
+
+type            = process
+command         = /usr/bin/rspamd -u _rspamd -g _rspamd --no-fork
+logfile         = /var/log/rspamd/rspamd.log
+waits-for       = network.target
```
q66, line 6: depends-on

contrib/rspamd/files/rspamd
```diff
@@ -0,0 +1,9 @@
...
+restart         = true
```
q66, line 7: this is implicit

contrib/rspamd/files/rspamd
```diff
@@ -0,0 +1,9 @@
...
+restart-delay   = 3
+smooth-recovery = true
```
q66, line 9: don't need to specify any of these other two either

contrib/rspamd/template.py
```diff
@@ -0,0 +1,62 @@
...
+    "openssl-devel",
+    "pcre2-devel",
+    "snowball-devel",
+    "sqlite-devel",
+    "xxhash-devel",
+    "zstd-devel",
+]
+pkgdesc = "Spam filtering system"
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "Apache-2.0 AND BSD-1-Clause AND BSD-2-Clause AND BSD-3-Clause AND CC0-1.0 AND LGPL-2.1-or-later AND LGPL-3.0-only AND MIT AND Zlib"
```
q66, line 39: where do all these come from?

contrib/snowball/template.py
```diff
@@ -0,0 +1,42 @@
+pkgname = "snowball"
+pkgver = "2.2.0"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+# check_utf8 check_iso_8859_1 check_iso_8859_2 check_koi8r
+# all have missing sub-targets
+make_check_target = "check_stemtest"
+hostmakedepends = ["gmake", "perl"]
+pkgdesc = "Snowball rule-based stemmin alorithms"
```
q66, line 10: likely a typo here

contrib/snowball/template.py
```diff
@@ -0,0 +1,42 @@
...
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "BSD-3-Clause"
+url = "https://snowballstem.org/index.html"
+source = f"https://github.com/snowballstem/snowball/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "425cdb5fba13a01db59a1713780f0662e984204f402d3dae1525bda9e6d30f1a"
+env = {"LD_LIBRARY_PATH": "./"}
```
q66, line 16: what is this one for? libstemmer?

contrib/rspamd/template.py
```diff
@@ -0,0 +1,62 @@
...
+    "-DENABLE_HYPERSCAN=OFF",
+    "-DENABLE_LUAJIT=OFF",
+    "-DENABLE_URI_INCLUDE=ON",
+    "-DRSPAMD_GROUP=_rspamd",
+    "-DRSPAMD_USER=_rspamd",
+    "-DSYSTEM_FMT=ON",
+    "-DSYSTEM_XXHASH=ON",
+    "-DSYSTEM_ZSTD=ON",
+    "-D_CAN_RUN=0",
+    "-DHAVE_ATOMIC_BUILTINS_EXITCODE=0",
```
nekopsykose, line 17: these 2 look weird, what are they for?

contrib/rspamd/template.py
```diff
@@ -0,0 +1,62 @@
...
+    "libsodium-devel",
+    "libunwind-devel",
+    "lua5.4-devel",
+    "openssl-devel",
+    "pcre2-devel",
+    "snowball-devel",
+    "sqlite-devel",
+    "xxhash-devel",
+    "zstd-devel",
+]
```
nekopsykose, line 36: didn't you intend to use vectorscan as a dep here

contrib/rspamd/template.py
a16bitsysop: the bundled dependencies https://github.com/rspamd/rspamd/blob/master/contrib/DEPENDENCY_INFO.md and the web interface https://github.com/rspamd/rspamd/tree/master/interface.

Looks like BSD-1-Clause and LGPL-2.1-or-later can be removed and boost needs adding

contrib/snowball/template.py
a16bitsysop: yes, the tests can’t find libstemmer otherwise.  Other distros have a libstemmer sub package

contrib/rspamd/template.py
a16bitsysop: Needed for cross compilation, cmake generates some .cmake files during cross compilation that need these

contrib/rspamd/template.py
a16bitsysop: ah yes, must have missed that!

contrib/fasttext/patches/no-march-native.patch
a16bitsysop: okay

a16bitsysop: do the flags in CMakelists.txt need patching out as well?

q66: yes, all but perhaps `-std=c++11`

contrib/fasttext/template.py
q66: any reason you dropped the whole thing? is the test data comment not correct?

contrib/snowball/template.py
q66: there is a second typo in there

contrib/fasttext/patches/no-march-native.patch
a16bitsysop: okay

contrib/fasttext/template.py
a16bitsysop: I thought that’s what you meant as without the comment it fails locally, comment is correct.  It has it’s own script to download test data: https://github.com/facebookresearch/fastText/blob/main/tests/fetch_test_data.sh

q66: what? the `!check` is commented out, so it doesn't do anything

a16bitsysop: ah, I see. Missed that

q66:
with the amount of vendored heavily patched mess in here, i'm thinking rspamd would potentially be better for the user/ repository

same with snowball, which has a bunch of patching to make it a sharedlib and nothing else needs it for the time being

we can always move things at a later point if needed

a16bitsysop:
okay, does that still get binary packages?

nekopsykose:
yes

q66:
`apk add chimera-repo-user` and you're good

q66:
fasttext can stay where it was

a16bitsysop:
okay, moved it back

a16bitsysop:
I added chimera-repo-user, and tried updating my cports with bootstrap-update and it is not in cports/etc/apk/repositories and chimera-repo-user cannot be installed in the chroot

q66:
you don't need to do anything in the chroot, that package is for your outside system to install things from it

PR #1211, opened 2024-01-10
contrib/i3status-rust: new package (0.32.3)

cal-andrew:
The man page can be generated locally, but that requires Pandoc, which requires Haskell.
Alpine Linux doesn't have Pandoc in their repos, so they have the man page on a server and download it during the build, which is what the template currently does.

### DIFF
--- a/contrib/i3status-rust/template.py
+++ b/contrib/i3status-rust/template.py
@@ -1,0 +1,28 @@
+pkgname = "i3status-rust"
+pkgver = "0.32.3"
+pkgrel = 0
+build_style = "cargo"
+# Enable the notmuch feature
+make_build_args = ["--features=notmuch"]
+make_install_args = ["--features=notmuch"]
+make_check_args = ["--features=notmuch"]
+hostmakedepends = ["cargo", "pkgconf"]
+makedepends = [
+    "libpulse-devel",
+    "libsensors-devel",
+    "notmuch-devel",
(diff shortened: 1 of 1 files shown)


q66:
i don't want to depend on something we can't generate ourselves using self-hosted tooling, i'd prefer not installing the manpage at the moment

contrib/i3status-rust/template.py
```diff
@@ -0,0 +1,38 @@
...
+make_check_args = ["--features=notmuch"]
+hostmakedepends = ["cargo", "pkgconf"]
+makedepends = [
+    "libpulse-devel",
+    "libsensors-devel",
+    "notmuch-devel",
+    "openssl-devel",
+    "rust-std",
+]
+pkgdesc = "Feature-rich replacement for i3status"
```
q66, line 17: pkgdesc should say what it does, not what it's a replacement of

cal-andrew: changed pkgdesc to say what it does

cal-andrew:
> i don't want to depend on something we can't generate ourselves using self-hosted tooling, i'd prefer not installing the manpage at the moment

Fair point, I've taken out the manpage

cal-andrew:
I have closed this PR as I've stopped using Chimera, but if anyone else wants to maintain the package, feel free to pick this up.

PR #1212, opened 2024-01-10
contrib/python-toml: new package (0.10.2)

firefly-cpp:
New package: https://github.com/uiri/toml

### DIFF
--- a/contrib/python-toml/template.py
+++ b/contrib/python-toml/template.py
@@ -1,0 +1,24 @@
+pkgname = "python-toml"
+pkgver = "0.10.2"
+pkgrel = 0
+build_style = "python_pep517"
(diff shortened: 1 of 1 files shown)


q66:
https://github.com/chimera-linux/cports/commit/dbee0138b6072c585da414d6d07fd4ad40ca44fa

nekopsykose:
i just checked for toml in the alpine packages and

- it's literally unused (incorrect dep) in most places, like the package does `import tomllib` normally or whatever  
- in the few places where it is, all you have to do is replace `toml.load` with `tomllib.load`

so yeah, nothing really needs this. and when adding stuff that might depend on it it's easy to handle it 1-by-1

PR #1214, opened 2024-01-11
main/iproute2: update to 6.7.0

nekopsykose:
this also fixes cross (for fun)- the build does HOSTCC?=CC which breaks since we never set HOSTCC by default


main/iproute2/template.py
```diff
@@ -17,9 +17,17 @@
...
-sha256 = "8738c804afd09f0bf756937f0c3de23117832a98d8cbbf50386cf5005cd613ce"
+sha256 = "ff942dd9828d7d1f867f61fe72ce433078c31e5d8e4a78e20f02cb5892e8841d"
 hardening = ["vis", "cfi"]
 
+if self.profile().cross:
+    hostmakedepends += ["linux-headers"]
+
+
+def init_build(self):
+    with self.profile("host") as _:
```
q66, line 28: you don't need the `as` if you're not using it

main/iproute2/template.py
```diff
@@ -17,9 +17,17 @@
 license = "GPL-2.0-only"
 url = "https://wiki.linuxfoundation.org/networking/iproute2"
 source = f"$(KERNEL_SITE)/utils/net/{pkgname}/{pkgname}-{pkgver}.tar.xz"
-sha256 = "8738c804afd09f0bf756937f0c3de23117832a98d8cbbf50386cf5005cd613ce"
+sha256 = "ff942dd9828d7d1f867f61fe72ce433078c31e5d8e4a78e20f02cb5892e8841d"
 hardening = ["vis", "cfi"]
 
+if self.profile().cross:
```
q66, line 23: this is kind of an antipattern, you can just directly add it to hostmakedepends and it won't be any different

nekopsykose: very tru

PR #1216, opened 2024-01-11
contrib/asciidoctor: new package (2.0.20)

z-erica:


q66:
broken update-check, fix that before getting to anything else

contrib/asciidoctor/template.py
```diff
@@ -0,0 +1,41 @@
...
+pkgdesc = "Ruby-based AsciiDoc converter"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "MIT"
+url = "https://asciidoctor.org"
+source = f"!https://rubygems.org/downloads/{pkgname}-{pkgver}.gem"
+sha256 = "835eabd445e4ae88f56a5f4e07593c3612b2be72eb661c612c3a8e1e17c57479"
+
+
+def do_install(self):
+    import os.path
```
q66, line 15: imports are forbidden in templates outside cbuild.util

contrib/asciidoctor/template.py
```diff
@@ -0,0 +1,41 @@
...
+
+    gemdir = os.path.relpath(
```
q66, line 17: don't do this at all (construct a path from `self.ruby_version` which is present if ruby is available)

contrib/asciidoctor/template.py
```diff
@@ -0,0 +1,41 @@
+pkgname = "asciidoctor"
+pkgver = "2.0.20"
+pkgrel = 0
+makedepends = ["ruby"]
```
q66, line 4: hostmakedepends

contrib/asciidoctor/template.py
```diff
@@ -0,0 +1,41 @@
...
+    import os.path
+
+    gemdir = os.path.relpath(
+        self.do("ruby", "-e", "puts Gem.default_dir", capture_output=True)
+        .stdout.decode()
+        .strip(),
+        "/",
+    )
+
+    self.do(
```
q66, line 24: where does the gem come from? nothing copies it

contrib/asciidoctor/template.py
```diff
@@ -0,0 +1,41 @@
...
+        "install",
+        "--local",
+        "--install-dir",
+        self.chroot_destdir / gemdir,
+        "--ignore-dependencies",
+        "--verbose",
+        "--no-document",
+        self.chroot_builddir / f"{pkgname}-{pkgver}.gem",
+    )
+    self.rm(self.destdir / gemdir / "cache", recursive=True)
```
q66, line 35: should the cache be deleted?

contrib/asciidoctor/template.py
nekopsykose: isn't it already in the builddir by default from download? still archived, but since it's the only file the cwd ends up having it in the right dir

contrib/asciidoctor/template.py
```diff
@@ -0,0 +1,43 @@
...
+        f"{pkgname}-{pkgver}.gem",
+    )
+    self.rm(
+        self.destdir / f"usr/lib/ruby/{self.ruby_version}/cache",
+        recursive=True,
+    )
+
+    self.install_dir("usr/bin")
+    self.install_link(
+        f"/usr/lib/ruby/{self.ruby_version}/bin/{pkgname}",
```
nekopsykose, line 37:
```suggestion
        f"../lib/ruby/{self.ruby_version}/bin/{pkgname}",
```

q66:
we should probably add a real build-style for this

q66:
it's okay, i'll do that myself and then you can just rebase this

PR #1217, opened 2024-01-11
contrib/helvum: new package (0.5.1)

z-erica:


contrib/helvum/template.py
```diff
@@ -0,0 +1,20 @@
...
+    "libadwaita-devel",
+    "pipewire-devel",
+    "rust-std",
+]
+pkgdesc = "GTK patchbay for PipeWire"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "GPL-3.0-only"
+url = "https://gitlab.freedesktop.org/pipewire/helvum"
+source = (
+    f"{url}/uploads/f523a2c71046fb21a7584fca80fc0ee9/{pkgname}-{pkgver}.tar.xz"
```
nekopsykose, line 18: is it possible to use the non-randomnumbers url instead?

if you do that the cargo vendor output won't be in the tarball, so then you'd have to patch the meson files to not override the CARGO_HOME, and manually call cargo.util vendor yourself, but at least then the url doesn't change every release

PR #1218, opened 2024-01-11
main/python-nftables: new package (1.0.9)

mikoxyz:


main/nftables/template.py
```diff
@@ -1,3 +1,4 @@
+# update contrib/python-nftables alongside this
 pkgname = "nftables"
 pkgver = "1.0.9"
 pkgrel = 0
```
q66, line 4: bump rel

main/nftables/template.py
```diff
@@ -7,7 +8,6 @@
     "--with-cli=editline",
 ]
 hostmakedepends = [
```
q66, line 10: while at it sort it

main/nftables/template.py
```diff
@@ -1,3 +1,4 @@
+# update contrib/python-nftables alongside this
```
q66, line 1: main

PR #1219, opened 2024-01-11
contrib/cmus: new package (2.10.0)

z-erica:


contrib/cmus/template.py
```diff
@@ -0,0 +1,39 @@
...
+    "flac-devel",
+    "libaom-devel",
+    "libcddb-devel",
+    "libmodplug-devel",
+    "libpulse-devel",
+    "libsamplerate-devel",
+    "libvorbis-devel",
+    "linux-headers",
+    "ncurses-devel",
+    "opus-devel",
```
nekopsykose, line 21: it only links opusfile so i think opus-devel isn't needed

also samplerate/aom/cddb aren't linked at all

also alsa-lib is probably not needed since it supports pulse and we expect everything to use pulse/pipewire

q66:
enable jack backend, disable alsa backend

contrib/cmus/template.py
nekopsykose: > also samplerate/aom/cddb aren't linked at all  

also the libav* ffmpeg libs either

q66: as far as i can tell there is ffmpeg support but it's not enabled in the build, which is why it's not linked

```
CONFIG_AAC = n
CONFIG_ALSA = y
CONFIG_AO = n
CONFIG_ARTS = n
CONFIG_CDIO = n
CONFIG_COREAUDIO = n
CONFIG_CUE = y
CONFIG_FFMPEG = n
CONFIG_FLAC = y
CONFIG_JACK = n
CONFIG_MAD = n
CONFIG_MIKMOD = n
CONFIG_MODPLUG = y
CONFIG_MP4 = n
CONFIG_MPC = n
CONFIG_MPRIS = y
CONFIG_OPUS = y
CONFIG_OSS = y
CONFIG_PULSE = y
CONFIG_ROAR = n
CONFIG_SAMPLERATE = y
CONFIG_SNDIO = n
CONFIG_SUN = n
CONFIG_VORBIS = y
CONFIG_VTX = n
CONFIG_WAV = y
CONFIG_WAVEOUT = n
CONFIG_WAVPACK = y
CONFIG_BASS = n
```

some of these should be on, some should be off

q66: samplerate seems to be enabled according to the log though? so that's weird

q66:
oss backend should also be disabled

contrib/cmus/template.py
nekopsykose: the configure step does detect the not-linked ones so not sure why they're not present after

q66: ah, samplerate is linked when jack is on, according to the buildsystem

q66:
```
op/jack.so: $(jack-objs) $(libcmus-y)
        $(call cmd,ld_dl,$(JACK_LIBS) $(SAMPLERATE_LIBS))
```

PR #1222, opened 2024-01-12
contrib/distrobox: new package (1.6.0.1)

cal-andrew:
During the install phase, there is an error with distrobox's custom install script:
`install: target 'distrobox' is not a directory`
And I also get another error:
`Do you have permission to write to cports/bldroot/destdir/distrobox-1.6.0.1/usr/bin?`

I've enabled verbose mode on the install script for debugging btw.

### DIFF
--- a/contrib/distrobox/template.py
+++ b/contrib/distrobox/template.py
@@ -1,0 +1,14 @@
+pkgname = "distrobox"
+pkgver = "1.6.0.1"
+pkgrel = 0
+pkgdesc = "Use any Linux distribution inside your terminal"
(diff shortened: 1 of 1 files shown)


nekopsykose:
it calls `install -D` but that means something else with bsd install

you can replace

```
install -D -m 0755 -t $destdir/$bindir distrobox
```

with

```
install -d $destdir/$bindir
install -m 0755 distrobox $destdir/$bindir
```

or somethin like that

cal-andrew:
I have closed this PR as I've stopped using Chimera, but if anyone else wants to maintain the package, feel free to pick this up.

PR #1223, opened 2024-01-12
contrib/gnome-shell-extension-gsconnect: new package (56)

mhmdanas:
`gsconnect:components / ClipboardComponent` test fails with this error:

```
  stderr:

(gjs:55): Gdk-CRITICAL **: 20:04:04.485: gdk_seat_get_keyboard: assertion 'GDK_IS_SEAT (seat)' failed

(test program exited with status code -5)

TAP parsing error: Too few tests run (expected 2, got 0)
```

Not quite sure how to fix it.

### DIFF
--- a/contrib/gnome-shell-extension-gsconnect/template.py
+++ b/contrib/gnome-shell-extension-gsconnect/template.py
@@ -1,0 +1,27 @@
+pkgname = "gnome-shell-extension-gsconnect"
+pkgver = "56"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Dinstalled_tests=false"]
+# Would've used weston-headless-run here instead of xvfb-run, but that runs
+# into a gtk3 bug in one of the tests:
+# https://github.com/chimera-linux/cports/pull/1223#issue-2079623168
+make_check_wrapper = ["dbus-run-session", "xvfb-run"]
+hostmakedepends = [
+    "bash",
+    "desktop-file-utils",
+    "gettext",
+    "glib-devel",
+    "gtk-update-icon-cache",
+    "libxml2-progs",
+    "meson",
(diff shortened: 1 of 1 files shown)


nekopsykose:
i don't think we have a wrapper that makes it look like there is an actual seat in the headless env, would be nice to fix for a few other testsuites too (if it's even possible)

contrib/gnome-shell-extension-gsconnect/template.py
```diff
@@ -0,0 +1,24 @@
...
+    "meson",
+    "pkgconf",
+]
+depends = ["gnome-shell", "evolution-data-server", "openssl", "gsound"]
+checkdepends = ["gnome-shell", "weston", "dbus"]
+pkgdesc = "KDE Connect implementation for GNOME"
+maintainer = "triallax <triallax@tutanota.com>"
+license = "GPL-2.0-or-later"
+url = "https://github.com/GSConnect/gnome-shell-extension-gsconnect"
+source = f"{url}/archive/refs/tags/v56.tar.gz"
```
nekopsykose, line 23:
```suggestion
source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
```

mhmdanas: good catch

q66:
> i don't think we have a wrapper that makes it look like there is an actual seat in the headless env, would be nice to fix for a few other testsuites too (if it's even possible)

a headless environment by definition does not have a seat, this is a gtk3 bug specifically though (one that nobody is going to fix because it involves a large amount of refactoring and one that is fixed in gtk4)

q66:
using xvfb-run instead of wayland will not run into that bug fwiw

PR #1228, opened 2024-01-13
Orphan packages

cal-andrew:
I have stopped using Chimera Linux as I'm not very comfortable with the project relying on Github for code hosting (especially given the recent sourcehut outage), and Github being proprietary and all.
No hard feelings though and I hope Chimera Linux keeps going well.
So this just orphans my packages.


nekopsykose:
you have to bump the pkgrel too since the maintainer: string is in the package metadata

PR #1229, opened 2024-01-13
contrib/copyq: new package (7.1.0)

ttyyls:


contrib/copyq/template.py
```diff
@@ -0,0 +1,39 @@
...
+    "qt6-qtbase-core",
+    "qt6-qtbase-widgets",
+    "qt6-qtbase-devel",
+    "qt6-qtsvg-devel",
+    "qt6-qtwayland-devel",
+    "qt6-qttools-devel",
+    "qt6-qtdeclarative-devel",
+    "libnotify-devel",
+    "libxtst-devel",
+    "libxfixes-devel",
```
nekopsykose, line 26: all of these should be makedepends

also sort the lists

contrib/copyq/template.py
```diff
@@ -0,0 +1,39 @@
...
+pkgver = "7.1.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DWITH_QT6=ON",
+    "-DCMAKE_BUILD_TYPE=None",
+    "-DCMAKE_INSTALL_PREFIX=/usr",
+]
+hostmakedepends = [
+    "git",
```
nekopsykose, line 11: as far as i can tell git is not needed and whatever data would be used to generate a version would be wrong anyway

contrib/copyq/template.py
```diff
@@ -0,0 +1,39 @@
+pkgname = "copyq"
+pkgver = "7.1.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DWITH_QT6=ON",
+    "-DCMAKE_BUILD_TYPE=None",
+    "-DCMAKE_INSTALL_PREFIX=/usr",
```
nekopsykose, line 8: these are both set by the cmake build style already

contrib/copyq/template.py
nekopsykose: (by all of those i mean the -devel ones i highlighted)

contrib/copyq/template.py
```diff
@@ -0,0 +1,38 @@
...
+]
+makedepends = [
+    "cmake",
```
q66, line 11: use hostmakedepends properly

contrib/copyq/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "wayland-devel",
+]
+pkgdesc = "Clipboard manager with advanced features"
+maintainer = "ttyyls <contact@behri.org>"
+license = "GPL-3.0-or-later"
+url = "https://hluk.github.io/CopyQ"
+source = f"https://github.com/hluk/CopyQ/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "ce0265b0a86350fda7bfa1a9d4b74d794a4077551b28186012683567d6fd8158"
+# FIXME: segfaults with cfi
+hardening = ["vis", "!cfi", "int"]
```
nekopsykose, line 32:
```suggestion
hardening = ["vis", "!cfi"]
```

int is the default

contrib/copyq/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "extra-cmake-modules",
+    "ninja",
+    "pkgconf",
+]
+makedepends = [
+    "libxfixes-devel",
+    "libxtst-devel",
+    "qt6-qt5compat-devel",
+    "qt6-qtbase-core",
+    "qt6-qtbase-widgets",
```
nekopsykose, line 17:
```suggestion
```

these come with -devel already

contrib/copyq/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "qt6-qtbase-devel",
+    "qt6-qtdeclarative-devel",
+    "qt6-qtsvg-devel",
+    "qt6-qtwayland-devel",
+    "qt6-qttools-devel",
```
nekopsykose, line 22:
```suggestion
    "qt6-qttools-devel",
    "qt6-qtwayland-devel",
```

contrib/copyq/template.py
```diff
@@ -0,0 +1,38 @@
...
+]
+pkgdesc = "Clipboard manager with advanced features"
+maintainer = "ttyyls <contact@behri.org>"
+license = "GPL-3.0-or-later"
+url = "https://hluk.github.io/CopyQ"
+source = f"https://github.com/hluk/CopyQ/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "ce0265b0a86350fda7bfa1a9d4b74d794a4077551b28186012683567d6fd8158"
+# FIXME: segfaults with cfi
+hardening = ["vis", "!cfi", "int"]
+# hard depends on git
```
nekopsykose, line 33: they don't depend on git, they build with `-DWITH_TESTS=ON` to cmake and then are ran with `./copyq tests` (inside headless). but it's basically a separate debug build (the codepaths are ifdef'd on that too), so you'd have to build it twice, ...

doesn't seem worth it to run but you should update the comment

PR #1230, opened 2024-01-13
user/plan9port: new package (0_git20240110)

z-erica:


q66:
the policy is in general not to package stuff that does not have tagged releases (outside of special exceptions), see packaging.md

contrib/plan9port/template.py
```diff
@@ -0,0 +1,64 @@
...
+license = "MIT"
+url = "https://9fans.github.io/plan9port"
+source = f"https://github.com/9fans/plan9port/archive/{_commit}.zip"
+sha256 = "931b96e814b31769bf62b2ef2189ac2f474207d079932987c468b1c86854fb39"
+options = ["!cross"]
+
+
+def do_configure(self):
+    self.cp(self.files_path / "configure", ".")
+    self.do("./configure")
```
nekopsykose, line 26: you can also write those contents from here instead of needing a script

JamiKettunen:
should be moved under [the newly added `user/`](https://github.com/chimera-linux/cports/compare/c307559...383095d) after rebase

user/plan9port/files/acme
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec 9 acme
```
nekopsykose, line 2: does this need "$@" too? (and i'm not sure if it accepts files passed, for perhaps %U or whatnot to the desktop entry Exec (see the spec))

user/plan9port/template.py
```diff
@@ -0,0 +1,54 @@
...
+url = "https://9fans.github.io/plan9port"
+source = f"https://github.com/9fans/plan9port/archive/{_commit}.zip"
+sha256 = "931b96e814b31769bf62b2ef2189ac2f474207d079932987c468b1c86854fb39"
+options = ["!cross", "!lintstatic"]
+
+
+def do_configure(self):
+    with open(self.cwd / "LOCAL.config", "w") as config:
+        config.write(f"CC9=" + self.get_tool("CC") + "\n")
+        config.write(f"CC9FLAGS='" + self.get_cflags(shell=True) + "'\n")
```
nekopsykose, line 26: ah missed this before- ldflags are ignored too, so you want to put them in this same cc9flags array i guess (gives a bunch of warnings but there's no other way that i see)

PR #1231, opened 2024-01-13
contrib/nnn: new package (4.9.0)

ttyyls:


contrib/nnn/template.py
```diff
@@ -0,0 +1,26 @@
+pkgname = "nnn"
+pkgver = "4.9"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+hostmakedepends = ["pkgconf", "gmake", "musl-fts-devel", "readline-devel"]
+depends = ["gmake"]
```
nekopsykose, line 7: it depends on gmake at runtime?

ttyyls: > it depends on gmake at runtime?
nope, just fixed that

contrib/nnn/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "nnn"
+pkgver = "4.9"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+hostmakedepends = ["pkgconf", "gmake", "musl-fts-devel", "readline-devel"]
```
q66, line 6: separate hostmakedepends and makedepends properly

contrib/nnn/template.py
```diff
@@ -0,0 +1,26 @@
+pkgname = "nnn"
+pkgver = "4.9"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+hostmakedepends = ["pkgconf", "gmake"]
+makedepends = ["musl-fts-devel", "readline-devel"]
```
q66, line 7: libedit-readline-devel should be preferred unless there's a reason for the otherwise

contrib/nnn/template.py
```diff
@@ -0,0 +1,26 @@
...
+pkgdesc = "Unorthodox terminal file manager"
+maintainer = "ttyyls <contact@behri.org>"
+license = "BSD-2-Clause"
+url = "https://github.com/jarun/nnn"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "9e25465a856d3ba626d6163046669c0d4010d520f2fb848b0d611e1ec6af1b22"
+tool_flags = {"LDFLAGS": ["-lfts"]}
+hardening = ["vis", "cfi"]
+# no tests :^)
```
q66, line 16: drop the :^)

contrib/nnn/template.py
ttyyls: it required a patch which I just added

contrib/nnn/template.py
ttyyls: done

PR #1232, opened 2024-01-13
contrib/python-filelock: new package (3.13.1)

firefly-cpp:
Package: https://github.com/tox-dev/filelock

### DIFF
--- a/contrib/python-filelock/template.py
+++ b/contrib/python-filelock/template.py
@@ -1,0 +1,20 @@
+pkgname = "python-filelock"
+pkgver = "3.13.1"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-hatch_vcs",
+    "python-hatchling",
+    "python-installer",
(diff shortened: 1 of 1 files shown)


contrib/python-filelock/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "python-filelock"
+pkgver = "3.13.1"
+pkgrel = 0
+build_style = "python_pep517"
+_deps = [
+    "python-hatchling",
+    "python-hatch_vcs",
+]
```
nekopsykose, line 8: these are neither checkdepends nor runtime depends so they should be in hostmakedepends only

firefly-cpp: fixed now

nekopsykose: (you can remove the whole _deps thing and put them in hostmakedepends since it's only used in that one place now)

firefly-cpp: of course; done

PR #1235, opened 2024-01-14
contrib/razor: new package (2.86)

a16bitsysop:


a16bitsysop:
This builds locally, I get the same error with 2.85 locally but not 2.86

a16bitsysop:
fixed after removing perl-getopt-long which is now bundled with perl

contrib/razor/template.py
```diff
@@ -0,0 +1,16 @@
...
+    "gmake",
+    "perl",
+]
+makedepends = ["perl"]
+depends = ["perl", "perl-uri"]
+pkgdesc = "Distributed, collaborative spam detection and filtering network"
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "Artistic-2.0"
+url = "https://razor.sourceforge.net/index.html"
+source = f"https://cpan.metacpan.org/authors/id/T/TO/TODDR/Razor2-Client-Agent-{pkgver}.tar.gz"
```
q66, line 15: this should use CPAN_SITE like other perl modules

contrib/razor/template.py
```diff
@@ -0,0 +1,16 @@
+pkgname = "razor"
```
q66, line 1: should likely be called perl-razor2-client-agent

q66: that is, unless the perl module is secondary and the primary thing is the frontend tools

contrib/razor/template.py
a16bitsysop: okay

contrib/razor/template.py
a16bitsysop: some distros call it razor and some call it that:
https://repology.org/project/perl:razor2-client-agent/versions

a16bitsysop: changing the name brings back the previous build failure?

a16bitsysop: fixed it with another upstream patch, was parallel make

q66: i still don't really understand how this is supposed to be used

so it is a perl module, is it a program, or what

contrib/perl-razor2-client-agent/template.py
```diff
@@ -0,0 +1,16 @@
...
+hostmakedepends = [
+    "gmake",
+    "perl",
+]
+makedepends = ["perl"]
+depends = ["perl", "perl-uri"]
+pkgdesc = "Distributed, collaborative spam detection and filtering network"
+maintainer = "Duncan Bellamy <dunk@denkimushi.com>"
+license = "Artistic-2.0"
+url = "https://razor.sourceforge.net/index.html"
```
q66, line 14: website is dead?

nekopsykose: works for me, redirects to https://sourceforge.net/projects/razor/

q66: gets me a white page

contrib/razor/template.py
a16bitsysop: it is a program written in perl, it installs in user bin:
razor-admin   razor-check   razor-client  razor-report  razor-revoke

but also installs perl modules in /usr/lib/perl5 that these use

contrib/perl-razor2-client-agent/template.py
a16bitsysop: I changed to the redirect address https://sourceforge.net/projects/razor

contrib/razor/template.py
q66: if it's primarily a program, calling it razor is fine

a16bitsysop: okay changed back, yes it a program for checking email hashes against a database for spam

PR #1236, opened 2024-01-14
contrib/fossil: new package (2.23)

z-erica:


contrib/fossil/template.py
```diff
@@ -0,0 +1,20 @@
...
+make_check_target = "test"
+# tests require $USER being set
+make_check_env = {"USER": "check"}
+makedepends = ["openssl-devel", "zlib-devel"]
+checkdepends = ["tcl"]
+pkgdesc = "Distributed software configuration management system"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "BSD-2-Clause"
+url = "https://fossil-scm.org"
+source = f"https://fossil-scm.org/home/tarball/47362306a7dd7c6fc3cab77cebe5d25469b0a9448479d9718eb5c49c8337b29/fossil-src-{pkgver}.tar.gz"
```
nekopsykose, line 15: i think there's a `"https://www.fossil-scm.org/home/tarball/version-{pkgver}/fossil-src-{pkgver}.tar.gz` to not have a hash in it

contrib/fossil/template.py
```diff
@@ -0,0 +1,20 @@
+pkgname = "fossil"
+pkgver = "2.23"
+pkgrel = 0
+build_style = "configure"
```
q66, line 4: anyhing preventing it from being gnu_configure?

q66: doesn't really matter, as long as it accepts the same configure arguments/works it should still be gnu_configure (sets up good defaults for cross-compiling, out of tree building and so on)

contrib/fossil/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "fossil"
+pkgver = "2.23"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_gen = []
+make_check_target = "test"
+# tests require $USER being set
+make_check_env = {"USER": "check"}
+makedepends = ["openssl-devel", "zlib-devel"]
```
nekopsykose, line 9: you probably want `--disable-internal-sqlite` too

contrib/fossil/template.py
nekopsykose: > as long as it accepts the same configure arguments/works

Error: Unknown option --with-libtool-sysroot

nekopsykose: fyi to test cross `./cbuild -a aarch64 pkg contrib/fossil`

nekopsykose: it needs the same --host/--build etc as regular autotools to work (since it uses --build to invoke the native cc for a buildtime exe), but in this case that seems like Effort so just mark it as !cross

contrib/fossil/template.py
nekopsykose: (otherwise it will have the sqlite+clang powerpc wrong endianness issue and then fail tests later on it, most likely anyway)

nekopsykose: from the test logs:  

```
Fossil was not compiled with JSON support.
Fossil was not compiled with TH1 docs support.
Fossil was not compiled with TH1 hooks support.
Fossil was not compiled with Tcl support.
```

in alpine config:  
```
		--json \
		--with-th1-docs \
		--with-th1-hooks \
		--with-tcl=1 \
		--with-tcl-private-stubs
```

might want to add those, probably. well, not sure if tcl is useful, but json seems so

contrib/tcllib/template.py
```diff
@@ -0,0 +1,12 @@
+pkgname = "tcllib"
+pkgver = "1.21"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = ["autoconf", "automake", "tcl"]
```
nekopsykose, line 5:
```suggestion
hostmakedepends = ["automake", "tcl"]
```

nekopsykose:
unrelatedly, do the tests fail on failure? there's a lot of various

```
test amend-tag-2.2 FAILED!
test json-login-c-c FAILED!
test th1-expr-2 FAILED!
RESULT: child killed: illegal instruction
```
etc

but it passes anyway. is that intentional? (the sigill looked a bit sus)

contrib/fossil/template.py
```diff
@@ -0,0 +1,30 @@
...
+    "--json",
+    "--with-th1-docs",
+    "--with-th1-hooks",
+    "--with-tcl=1",
+    "--with-tcl-private-stubs",
+]
+configure_gen = []
+make_check_target = "test"
+# tests require $USER being set
+make_check_env = {"USER": "check"}
```
q66, line 16: should no longer be necessary (ba2882c3df8e5d749c675fc3bdafc79bb0f6b055)

nekopsykose:
that would trigger the int hardening, yeah (which catches overflow). i'd guess they 'intentionally' overflow but without writing it in a non-UB way?

in any case i guess the tests don't actually exit nonzero then and that should be fixed

nekopsykose:
> im wondering just how intentional it is not to exit nonzero on failure  

it looks unintentional. that is, up to now the developers just manually looked at it. one of the latest replies:  

```
If merged into the current trunk (where test-fixes-2.24 is integrated), the improvements are:
[...]
    Exit code is 1 on errors; useful with e.g. make test.
```

you could probably look at that and see what line to change to make it exit 1 instead of 0

PR #1237, opened 2024-01-14
contrib/tealdeer: new package (1.6.1)

ttyyls:


contrib/tealdeer/template.py
```diff
@@ -0,0 +1,21 @@
...
+build_style = "cargo"
+hostmakedepends = ["cargo"]
+makedepends = ["rust-std"]
+pkgdesc = "Very fast implementation of tldr"
+maintainer = "ttyyls <contact@behri.org>"
+license = "MIT"
+url = "https://github.com/dbrgn/tealdeer"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "d42db25a56a72faec173c86192656c5381281dc197171f385fccffd518930430"
+# not working
```
q66, line 13: should say why

ttyyls: the last 8 tests required connectivity which is disabled at the check stage, I adjusted the comment accordingly

q66: is it possible (without involving something overly complicated) to exclude the network-using tests and let the others run?

sidkshatriya: How about omitting the network connectivity tests by simply removing the `#[test]` annotation from them ?

contrib/tealdeer/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "tealdeer"
+pkgver = "1.6.1"
+pkgrel = 0
+build_style = "cargo"
```
nekopsykose, line 4: should also use `--features native-tls` and use openssl-devel

nekopsykose: ah, that's only in master

q66: yeah that's a blocker, need to patch that then

contrib/tealdeer/template.py
q66: yes that looks good

contrib/tealdeer/template.py
nekopsykose: you'll need to add a

```
def do_prepare(self):
    # we patch the lockfile so vendor after patch
    pass


def post_patch(self):
    from cbuild.util import cargo

    self.cargo.vendor()
    cargo.setup_vendor(self)
```

and then generate a patch of the changed lock (`cargo update -p reqwest --precise=0.11.12`, or just update reqwest entirely) after adding the feature. but actually just that alone won't work, because by default it also uses `
native-roots = ["reqwest/rustls-tls-native-roots"]`, which will build ring anyway (and then not use it), so the build will fail. so you'd also have to remove those, and then also patch out the assertion in main.rs that checks for the feature... (it does work in the end)

nekopsykose: > and then generate a patch of the changed lock   

(and cargo.toml, and the changes in main.rs)

PR #1238, opened 2024-01-15
contrib/firewalld: new package (2.1.0)

mikoxyz:
making this as a draft since i'm not sure it's good enough to get merged

### DIFF
--- a/contrib/firewalld/patches/0001-don-t-build-tests.patch
+++ b/contrib/firewalld/patches/0001-don-t-build-tests.patch
@@ -1,0 +1,24 @@
+From 1136793b4e8c3b2b764bf75fef55a77893ccf698 Mon Sep 17 00:00:00 2001
+From: miko <mikoxyzzz@gmail.com>
+Date: Mon, 15 Jan 2024 15:26:43 +0100
+Subject: [PATCH] don't build tests
+
+build breaks if we don't do this. the tests don't even work in our
+environment anyway
+---
+ src/Makefile.am | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/src/Makefile.am b/src/Makefile.am
+index 58ada179..3ec1374c 100644
+--- a/src/Makefile.am
++++ b/src/Makefile.am
+@@ -1,4 +1,4 @@
+-SUBDIRS = icons tests
++SUBDIRS = icons
+ 
+ dist_bin_SCRIPTS_in = firewall-applet.in firewall-cmd.in firewall-offline-cmd.in firewall-config.in
+ dist_sbin_SCRIPTS_in = firewalld.in
+-- 
+2.43.0
+
--- a/contrib/firewalld/template.py
+++ b/contrib/firewalld/template.py
@@ -1,0 +1,61 @@
+pkgname = "firewalld"
+pkgver = "2.1.0"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--disable-systemd"]
+configure_gen = ["./autogen.sh"]
+make_cmd = "gmake"
+# icons won't install unless we do this...
+make_dir = "."
+hostmakedepends = [
+    "automake",
+    "gmake",
+    "gtk-doc-tools",
+    "intltool",
+    "libxml2-progs",
+    "pkgconf",
+    "python",
+]
+makedepends = [
+    "gettext-devel",
+    "glib-devel",
+]
+depends = [
+    "python-dbus",
+    "python-gobject",
+    "python-nftables",
+]
+pkgdesc = "Stateful zone-based firewall daemon with D-Bus interface"
+maintainer = "miko <mikoxyzzz@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://firewalld.org"
+source = f"https://github.com/firewalld/firewalld/releases/download/v{pkgver}/firewalld-{pkgver}.tar.bz2"
+sha256 = "22e3c700c2e4867796e9c22cf929cba4b2c143c8884240cfb9f3ef731366ff71"
+# tests don't work in our build env
+options = ["!check"]
+
+
+def do_prepare(self):
+    self.rm("src/icons/*/*/firewall-applet*", recursive=True, glob=True)
+
+
+def post_install(self):
+    from cbuild.util import python
+
+    python.precompile(self, "usr/lib")
+    self.install_service(self.files_path / "firewalld")
+
+
+@subpackage("firewall-config")
+def _config(self):
+    self.pkgdesc = "GTK-based configuration utility for firewalld"
+    self.depends = [f"{pkgname}={pkgver}-r{pkgrel}", "gtk+3"]
+    return [
+        "usr/bin/firewall-config",
+        "usr/share/applications",
+        "usr/share/firewalld/gtk3_chooserbutton.py",
+        "usr/share/firewalld/gtk3_niceexpander.py",
+        "usr/share/icons",
+        "usr/share/man/man1/firewall-config.1",
+        "usr/share/metainfo",
+    ]
--- a/contrib/firewall-config
+++ b/contrib/firewall-config
@@ -1,0 +1,1 @@
+firewalld
--- a/contrib/firewalld/files/firewalld
+++ b/contrib/firewalld/files/firewalld
@@ -1,0 +1,7 @@
+# firewalld service
+
+type        = process
+command     = /usr/bin/firewalld --nofork --nopid
+before      = pre-network.target
+depends-on  = dbus
+depends-on  = pre-local.target
--- a/contrib/firewalld/patches/0002-don-t-build-firewall-applet.patch
+++ b/contrib/firewalld/patches/0002-don-t-build-firewall-applet.patch
@@ -1,0 +1,64 @@
+From 20ab495f7602b0600496b2e18ca07ce8620d73fd Mon Sep 17 00:00:00 2001
+From: miko <mikoxyzzz@gmail.com>
+Date: Mon, 15 Jan 2024 14:58:18 +0100
+Subject: [PATCH 2/4] don't build firewall-applet
+
+requires pyqt5, which isn't packaged in chimera
+---
+ config/Makefile.am  | 6 ------
+ doc/xml/Makefile.am | 1 -
+ src/Makefile.am     | 2 +-
+ 3 files changed, 1 insertion(+), 8 deletions(-)
+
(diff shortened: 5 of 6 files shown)


contrib/firewalld/patches/0001-don-t-build-tests.patch
```diff
@@ -0,0 +1,188 @@
...
+-	$(MAKE) -C src/tests $@
+-
+-.PHONY: check-container check-integration installcheck-integration
+ .PHONY: dist-container
+ 
+ rpm test-rpm: dist
+diff --git a/src/tests/Makefile.am b/src/tests/Makefile.am
+deleted file mode 100644
+index 62b88d61..00000000
+--- a/src/tests/Makefile.am
```
nekopsykose, line 32: do you actually need to delete the whole file? normally you just have to remove some SUBDIR from a parent .am to ignore it instead

nekopsykose: same for below, rather than a 5k line patch file you probably need 2 lines

mikoxyz: > do you actually need to delete the whole file? normally you just have to remove some SUBDIR from a parent .am to ignore it instead

fuck im so silly

contrib/firewalld/template.py
```diff
@@ -0,0 +1,76 @@
+pkgname = "firewalld"
+pkgver = "2.1.0"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--disable-sysconfig", "--disable-systemd"]
+configure_gen = ["./autogen.sh"]
+make_cmd = "gmake"
+hostmakedepends = [
+    "autoconf",
```
nekopsykose, line 9:
```suggestion
```

redundant with automake

contrib/firewalld/template.py
```diff
@@ -0,0 +1,76 @@
...
+    "libxml2-progs",
+    "pkgconf",
+    "python",
+]
+makedepends = [
+    "gettext-devel",
+    "glib-devel",
+]
+depends = [
+    "dbus",
```
nekopsykose, line 23:
```suggestion
```

contrib/firewalld/template.py
```diff
@@ -0,0 +1,76 @@
...
+    from cbuild.util import python
+
+    python.precompile(self, "usr/lib")
+    self.mv(
+        self.destdir / "etc/modprobe.d", self.destdir / "usr/lib/modprobe.d"
+    )
+    self.mkdir(self.destdir / "usr/share/icons/hicolor", True)
+    self.cp("src/icons/48x48", self.destdir / "usr/share/icons/hicolor", True)
+    self.cp("src/icons/32x32", self.destdir / "usr/share/icons/hicolor", True)
+    self.cp("src/icons/24x24", self.destdir / "usr/share/icons/hicolor", True)
```
nekopsykose, line 52: the makefile already installs these from what i can tell?

contrib/firewalld/template.py
```diff
@@ -0,0 +1,76 @@
...
+    self.cp("src/icons/22x22", self.destdir / "usr/share/icons/hicolor", True)
+    self.cp("src/icons/16x16", self.destdir / "usr/share/icons/hicolor", True)
+    self.cp(
+        "src/icons/scalable", self.destdir / "usr/share/icons/hicolor", True
+    )
+    self.rm(self.destdir / "etc/logrotate.d", True)
+    self.rm(self.destdir / "etc/rc.d", True)
+    self.rm(self.destdir / "etc/sysconfig", True)
+    self.install_license("COPYING")
```
nekopsykose, line 61:
```suggestion
```

don't have to install gpl

contrib/firewalld/template.py
mikoxyz: hm, it didn't for me

nekopsykose: hm weird

in that case, you can remove the mkdir and do install_files

contrib/firewalld/template.py
```diff
@@ -0,0 +1,69 @@
...
+maintainer = "miko <mikoxyzzz@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://firewalld.org"
+source = f"https://github.com/{pkgname}/{pkgname}/releases/download/v{pkgver}/{pkgname}-{pkgver}.tar.bz2"
+sha256 = "22e3c700c2e4867796e9c22cf929cba4b2c143c8884240cfb9f3ef731366ff71"
+# tests don't work in our build env
+options = ["!check"]
+
+
+def do_prepare(self):
```
q66, line 36: do you need this at all? also use named arguments

q66: does the build system just glob everything from that place?

mikoxyz: build system doesn't even seem to want to install the icons for whatever reason. i have to manually install them, in which case it's just easier to nuke firewall-applet's icons so that i can just run install_files on the icon dirs

contrib/firewalld/template.py
```diff
@@ -0,0 +1,70 @@
...
+depends = [
+    "python-dbus",
+    "python-gobject",
+    "python-nftables",
+]
+pkgdesc = "Stateful zone based firewall daemon with D-Bus interface"
+maintainer = "miko <mikoxyzzz@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://firewalld.org"
+source = f"https://github.com/{pkgname}/{pkgname}/releases/download/v{pkgver}/{pkgname}-{pkgver}.tar.bz2"
```
q66, line 30: expanding pkgname in urls just makes things more annoying (because you can't copypaste it) without having any benefit

contrib/firewalld/template.py
q66: gotcha

contrib/firewalld/template.py
```diff
@@ -0,0 +1,70 @@
...
+    self.rm(self.destdir / "etc/logrotate.d", recursive=True)
+    self.rm(self.destdir / "etc/rc.d", recursive=True)
+    self.rm(self.destdir / "etc/sysconfig", recursive=True)
+    self.install_files("src/icons/48x48", "usr/share/icons/hicolor")
+    self.install_files("src/icons/32x32", "usr/share/icons/hicolor")
+    self.install_files("src/icons/24x24", "usr/share/icons/hicolor")
+    self.install_files("src/icons/22x22", "usr/share/icons/hicolor")
+    self.install_files("src/icons/16x16", "usr/share/icons/hicolor")
+    self.install_files("src/icons/scalable", "usr/share/icons/hicolor")
+    self.install_service(self.files_path / f"{pkgname}")
```
q66, line 56: don't expand pkgname (even if you wanted to, `f"{pkgname}"` is literally just `pkgname`)

contrib/firewalld/template.py
```diff
@@ -0,0 +1,70 @@
...
+makedepends = [
+    "gettext-devel",
+    "glib-devel",
+]
+depends = [
+    "python-dbus",
+    "python-gobject",
+    "python-nftables",
+]
+pkgdesc = "Stateful zone based firewall daemon with D-Bus interface"
```
q66, line 26: zone-based

contrib/firewalld/template.py
```diff
@@ -0,0 +1,70 @@
...
+    from cbuild.util import python
+
+    python.precompile(self, "usr/lib")
+    self.mv(
+        self.destdir / "etc/modprobe.d", self.destdir / "usr/lib/modprobe.d"
+    )
+    self.rm(self.destdir / "etc/logrotate.d", recursive=True)
+    self.rm(self.destdir / "etc/rc.d", recursive=True)
+    self.rm(self.destdir / "etc/sysconfig", recursive=True)
+    self.install_files("src/icons/48x48", "usr/share/icons/hicolor")
```
q66, line 50: any eason why you can't just

```
self.install_files("src/icons", "usr/share/icons", name="hicolor")
```

instead of installing every size separately

(ofc you'd have to expand the post_extract to exclude makefile.am)

contrib/firewalld/template.py
```diff
@@ -0,0 +1,70 @@
...
+def post_install(self):
+    from cbuild.util import python
+
+    python.precompile(self, "usr/lib")
+    self.mv(
+        self.destdir / "etc/modprobe.d", self.destdir / "usr/lib/modprobe.d"
+    )
+    self.rm(self.destdir / "etc/logrotate.d", recursive=True)
+    self.rm(self.destdir / "etc/rc.d", recursive=True)
+    self.rm(self.destdir / "etc/sysconfig", recursive=True)
```
q66, line 49: isn't this installed only if `--disable-sysconfig` is not passed

contrib/firewalld/template.py
```diff
@@ -0,0 +1,70 @@
...
+
+def do_prepare(self):
+    self.rm("src/icons/*/*/firewall-applet*", recursive=True, glob=True)
+
+
+def post_install(self):
+    from cbuild.util import python
+
+    python.precompile(self, "usr/lib")
+    self.mv(
```
q66, line 44: i'd prefer a patch in the build system over doing this (should be single-line)

contrib/firewalld/template.py
```diff
@@ -0,0 +1,70 @@
...
+        self.destdir / "etc/modprobe.d", self.destdir / "usr/lib/modprobe.d"
+    )
+    self.rm(self.destdir / "etc/logrotate.d", recursive=True)
```
q66, line 47: maybe actually also patch all these out of the makefile instead of removing them, more robust that way

contrib/firewalld/template.py
mikoxyz: `--disable-sysconfig` doesn't even seem to do anything. anyway, i just patched that stuff out

q66:
should be ok to undraft this now (i'll give it a final review later)

contrib/firewalld/template.py
```diff
@@ -0,0 +1,60 @@
...
+    from cbuild.util import python
+
+    python.precompile(self, "usr/lib")
+    self.install_service(self.files_path / "firewalld")
+
+
+@subpackage("firewall-config")
+def _config(self):
+    self.pkgdesc = "GTK-based configuration utility for firewalld"
+    self.depends = ["gtk+3"]
```
nekopsykose, line 52: doesn't this depend on the parent package too (+ `f"{pkgname}={pkgver}-r{pkgrel}"`)

mikoxyz: true!

PR #1243, opened 2024-01-15
contrib/python-werkzeug: new package (2.3.8)

firefly-cpp:
Package: https://werkzeug.palletsprojects.com


contrib/python-werkzeug/template.py
```diff
@@ -0,0 +1,21 @@
...
+    "python-installer",
+    "python-flit_core",
+]
+pkgdesc = "WSGI swiss-army knife"
+maintainer = "firefly-cpp <iztok@iztok.space>"
+license = "MIT"
+url = "https://werkzeug.palletsprojects.com"
+source = f"$(PYPI_SITE)/w/werkzeug/werkzeug-{pkgver}.tar.gz"
+sha256 = "554b257c74bbeb7a0d254160a4f8ffe185243f52a52035060b761ca62d977f03"
+# missing dependencies (pytest-timeout)
```
q66, line 16: package that then

firefly-cpp: This is a plan. However, it is a pretty long chain to cover all dependencies of dependencies.

nekopsykose: if it's only pytest-timeout then the chain ends there (unless you want to run the pytest-timeout tests too, then it's pexpect+ptyprocess as well)

firefly-cpp: Let's package pytest-timeout without running its tests for now. I am on it.

firefly-cpp: Tests are now enabled

contrib/python-werkzeug/template.py
```diff
@@ -0,0 +1,37 @@
+pkgname = "python-werkzeug"
+pkgver = "3.0.1"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-flit_core",
```
nekopsykose, line 8:
```suggestion
    "python-flit_core",
    "python-installer",
```

contrib/python-werkzeug/template.py
```diff
@@ -0,0 +1,37 @@
...
+]
+checkdepends = [
+    "python-markupsafe",
+    "python-pytest",
+    "python-pytest-timeout",
+    "python-requests",
+]
+depends = [
+    "python-markupsafe",
+]
```
nekopsykose, line 18:
```suggestion
depends = ["python-markupsafe"]
```

single line for 1 dep

contrib/python-werkzeug/template.py
```diff
@@ -0,0 +1,37 @@
...
+sha256 = "507e811ecea72b18a404947aded4b3390e1db8f826b494d76550ef45bb3b1dcc"
+
+
+def post_extract(self):
+    # tests require ephemeral-port-preserve
+    self.rm("tests/conftest.py")
+    self.rm("tests/test_debug.py")
+    # tests require watchdog
+    self.rm("tests/test_serving.py")
+    self.rm("tests/middleware/test_http_proxy.py")
```
nekopsykose, line 33: `make_check_args = ["--deselect", "tests/....` instead of removing files

firefly-cpp: @nekopsykose, skipping tests is not an option since scripts import the missing packages. Interestingly, --ignore flag in make_check_args is not working

firefly-cpp: see current version of template

nekopsykose: > skipping tests is not an option

> Interestingly, --ignore flag in make_check_args is not working

works fine though (though yes, need ignore instead of deselect). conftest.py is a special pytest import that defines fixtures, so you need

```
make_check_args = [
    "--noconftest",
    # need ephemeral-port-preserve + watchdog
    "--ignore=tests/middleware/test_http_proxy.py",
    "--ignore=tests/test_debug.py",
    "--ignore=tests/test_serving.py",
]
```

firefly-cpp: wonderful; thanks

PR #1244, opened 2024-01-15
python-referencing test enablement

mikoxyz:


contrib/python-referencing/template.py
```diff
@@ -1,6 +1,6 @@
 pkgname = "python-referencing"
 pkgver = "0.32.1"
-pkgrel = 0
+pkgrel = 1
```
q66, line 3: no need to bump again if it only enables tests

contrib/python-pytest-subtests/template.py
```diff
@@ -0,0 +1,29 @@
+pkgname = "python-pytest-subtests"
+pkgver = "0.11.0"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python",
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+    "python-setuptools_scm",
```
nekopsykose, line 10:
```suggestion
    "python-build",
    "python-installer",
    "python-setuptools_scm",
```

mikoxyz: omfg why did i include python as a hostmakedepends again

PR #1245, opened 2024-01-15
contrib/python-pytest-timeout: new package (2.2.0)

firefly-cpp:
Package: https://github.com/pytest-dev/pytest-timeout

### DIFF
--- a/contrib/python-pytest-timeout/template.py
+++ b/contrib/python-pytest-timeout/template.py
@@ -1,0 +1,24 @@
+pkgname = "python-pytest-timeout"
+pkgver = "2.2.0"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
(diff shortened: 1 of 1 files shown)


contrib/python-pytest-timeout/template.py
```diff
@@ -0,0 +1,23 @@
...
+pkgver = "2.2.0"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+    "python-wheel",
+]
+checkdepends = ["python-pytest"]
```
nekopsykose, line 11:
```suggestion
checkdepends = ["python-pytest"]
depends = ["python-pytest"]
```

firefly-cpp: corrected now

PR #1246, opened 2024-01-15
contrib/ledger: new package (3.3.2)

z-erica:


contrib/ledger/template.py
```diff
@@ -0,0 +1,18 @@
...
+build_style = "cmake"
+hostmakedepends = ["cmake", "ninja"]
+makedepends = ["boost-devel", "utfcpp-devel", "mpfr-devel", "libedit-devel"]
+pkgdesc = "Plain text accounting software"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "BSD-3-Clause"
+url = "https://ledger-cli.org"
+source = f"https://github.com/ledger/ledger/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "555296ee1e870ff04e2356676977dcf55ebab5ad79126667bc56464cb1142035"
+# some tests fail
```
q66, line 13: would be nice to list what tests fail and how in this PR, and whether the failures are harmless or not

contrib/utfcpp/template.py
```diff
@@ -0,0 +1,20 @@
...
+url = "https://github.com/nemtrif/utfcpp"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "ffc668a310e77607d393f3c18b32715f223da1eac4c4d6e0579a11df8e6b59cf"
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+
+
+@subpackage("utfcpp-devel")
```
nekopsykose, line 18: this is a header-only library so the -devel means the parent becomes empty. for these, just don't split them

contrib/ledger/template.py
q66: disabling check because of 2 tests out of 410 is bad, it's always a good idea to figure out why the tests are failing and then either address the failures or patch the tests out when the failure is harmless

nekopsykose: in case you didn't know already for future reference, for cmake skipping individual tests looks like `-E "(BaselineTest_cmd-pricedb|BaselineTest_cmd-prices)"`

contrib/utfcpp/template.py
nekopsykose: (the reason the parent didn't end up empty and ci didn't catch it is because it installs a /usr/share/utf8cpp/cmake)

nekopsykose:
i added utfcpp for taglib so you don't need that in here anymore

PR #1259, opened 2024-01-18
contrib/python-pytest-asyncio: new package (0.23.3)

firefly-cpp:
https://github.com/pytest-dev/pytest-asyncio


contrib/python-pytest-asyncio/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "python-pytest-asyncio"
+pkgver = "0.23.4a2"
```
nekopsykose, line 2: 0.23.3 is the latest stable version

firefly-cpp: The problem is that CI forces to use the latest. How to skip it?

nekopsykose: ignore = ["*a*"] in update.py

contrib/python-pytest-asyncio/template.py
```diff
@@ -0,0 +1,23 @@
...
+license = "Apache-2.0"
+url = "https://github.com/pytest-dev/pytest-asyncio"
+source = f"$(PYPI_SITE)/p/pytest-asyncio/pytest-asyncio-{pkgver}.tar.gz"
+sha256 = "af313ce900a62fbe2b1aed18e37ad757f1ef9940c6b6a88e2954de38d6b1fb9f"
+# missing deps
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_license("LICENSE")
```
nekopsykose, line 23: don't need to install apache

nekopsykose:
the commit message also still says `0.23.4a2`

firefly-cpp:
It is OK now

nekopsykose:
that changed the title and not the commit

firefly-cpp:
fixed

PR #1262, opened 2024-01-19
contrib/fennel: new package (1.4.0)

z-erica:


nekopsykose:
is there an advantage/disadvantage to building `fennel-bin` instead which seems to be a C shim that uses liblua?

nekopsykose:
that part would need fixing and whatnot yes, i just wonder how the usability/wahtever actually differs

nekopsykose:
sounds pointless then

PR #1265, opened 2024-01-19
newsraft and dependencies: gumbo

ttyyls:


contrib/gumbo/template.py
```diff
@@ -0,0 +1,37 @@
...
+    "gtest-devel",
+    "libtool",
+]
+pkgdesc = "HTML5 parsing library in pure C99"
+maintainer = "ttyyls <contact@behri.org>"
+license = "Apache-2.0"
+url = "https://codeberg.org/grisha/gumbo-parser"
+source = f"{url}/archive/{pkgver}.tar.gz"
+sha256 = "a0ff2e1b613403fe69ff6407f3f93221fdfa67da357be158bb1e6903b33c1c10"
+options = ["empty"]
```
nekopsykose, line 23: remove gumbo-libs and this

contrib/gumbo/template.py
```diff
@@ -0,0 +1,37 @@
...
+make_cmd = "gmake"
+hostmakedepends = [
+    "bash",
+    "gmake",
+    "gm4",
+    "pkgconf",
+    "automake",
+    "autoconf",
+    "gtest-devel",
+    "libtool",
```
nekopsykose, line 15: remove autoconf, put gtest-devel in makedepends  

it also seems bash is not needed

nekopsykose: also sort the list, same below

contrib/newsraft/template.py
```diff
@@ -0,0 +1,30 @@
...
+]
+pkgdesc = "Feed reader for terminal"
+maintainer = "ttyyls <contact@behri.org>"
+license = "ISC"
+url = "https://codeberg.org/newsraft/newsraft"
+source = f"{url}/archive/{pkgname}-{pkgver}.tar.gz"
+sha256 = "16dbda602a37a2c8052421e90dd00ca065bab4130c6c2c16d5ca2a8a184990b9"
+hardening = ["vis", "cfi"]
+# relies on symbols being exposed in the shared library
+options = ["!check"]
```
nekopsykose, line 25: what shared library? this is one executable

if symbols are missing in something else then it should not have =vis

nekopsykose: ah, i see. there is a separate libnewsraft.so that is used only for testing

sure, then remove vis

contrib/newsraft/template.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def post_install(self):
+    # self.install_license("doc/newsraft.1")
```
nekopsykose, line 29: the manpage is already installed by the makefile

contrib/newsraft/template.py
ttyyls: besides that, it still errors on parsing various date formats.

ttyyls: I think enabling vis and cfi is more valuable than running those tests anyway.

PR #1266, opened 2024-01-19
contrib/libaacs: new package (0.11.1)

mikoxyz:


contrib/libaacs/template.py
```diff
@@ -0,0 +1,35 @@
...
+license = "LGPL-2.1-or-later"
+url = "https://www.videolan.org/developers/libaacs.html"
+source = f"https://download.videolan.org/pub/videolan/libaacs/{pkgver}/libaacs-{pkgver}.tar.bz2"
+sha256 = "a88aa0ebe4c98a77f7aeffd92ab3ef64ac548c6b822e8248a8b926725bea0a39"
+hardening = ["cfi", "sst", "vis"]
+
+
+@subpackage("libaacs-static")
+def _static(self):
+    return self.default_static()
```
nekopsykose, line 25: you can remove this as the -devel creates a -devel-static already

mikoxyz: !

contrib/libaacs/template.py
```diff
@@ -0,0 +1,30 @@
...
+    "pkgconf",
+]
+makedepends = ["libgcrypt-devel"]
+pkgdesc = "Research implementation of the AACS specification"
+maintainer = "miko <mikoxyzzz@gmail.com>"
+license = "LGPL-2.1-or-later"
+url = "https://www.videolan.org/developers/libaacs.html"
+source = f"https://download.videolan.org/pub/videolan/libaacs/{pkgver}/libaacs-{pkgver}.tar.bz2"
+sha256 = "a88aa0ebe4c98a77f7aeffd92ab3ef64ac548c6b822e8248a8b926725bea0a39"
+hardening = ["cfi", "sst", "vis"]
```
nekopsykose, line 20: this is a library and sst does not even work for DSOs

nekopsykose: well, i guess there's progs, but we don't use this anywhere yet for reasons i forgot

mikoxyz: i just nuked the entire hardening thing if that's fine

nekopsykose: if vis/cfi works then you should leave it?

easy way to test: see if the library linked into libbluray actually rips stuff

mikoxyz: mmmm i'll test tomorrow

mikoxyz: vis works, cfi doesn't

nekopsykose:
this needs to be in main/ to use with libbluray

mikoxyz:
> this needs to be in main/ to use with libbluray

libbluray is a bit funny in that it doesn't actually link with libaacs, so i don't think we actually need to have this in main.

nekopsykose:
if you say so

PR #1267, opened 2024-01-19
contrib/lowdown: new package (1.1.0)

ttyyls:


contrib/lowdown/template.py
```diff
@@ -0,0 +1,21 @@
...
+]
+pkgdesc = "Simple markdown translator"
+maintainer = "ttyyls <contact@behri.org>"
+license = "ISC"
+url = "https://kristaps.bsd.lv/lowdown"
+source = f"{url}/snapshots/{pkgname}.tar.gz"
+sha256 = "f31e3950c4732b1e409174fa092eca40c55be77a448ee2818df987979d7b0879"
+hardening = ["vis", "cfi"]
+# there are no tests
+options = ["!check"]
```
nekopsykose, line 17: make_check_target = "regress"

ttyyls: ok

contrib/lowdown/template.py
```diff
@@ -0,0 +1,20 @@
+pkgname = "lowdown"
+pkgver = "1.1.0"
+pkgrel = 0
+build_style = "configure"
+configure_args = [
+    "PREFIX=/usr",
+    "MANDIR=/usr/share/man",
+]
```
nekopsykose, line 8: you have to put LDFLAGS= here too because those aren't taken from env. you can do that in init_configure and using self.get_ldflags(shell=True)

contrib/lowdown/template.py
```diff
@@ -0,0 +1,20 @@
...
+make_check_target = "regress"
+pkgdesc = "Simple markdown translator"
```
nekopsykose, line 10:
```suggestion
pkgdesc = "Markdown translator"
```

contrib/lowdown/template.py
ttyyls: ok.

contrib/lowdown/template.py
ttyyls: done

contrib/lowdown/template.py
```diff
@@ -0,0 +1,25 @@
...
+license = "ISC"
+url = "https://kristaps.bsd.lv/lowdown"
+source = f"{url}/snapshots/{pkgname}.tar.gz"
+sha256 = "f31e3950c4732b1e409174fa092eca40c55be77a448ee2818df987979d7b0879"
+hardening = ["vis", "cfi"]
+
+
+def init_configure(self):
+    ldfl = self.get_ldflags(shell=True)
+    self.env["LDLFAGS"] = ldfl
```
nekopsykose, line 21:
```suggestion
    configure_args += f"LDFLAGS={self.get_ldflags(shell=True)}"
```

they are in env already, the configure ignores the env

ttyyls: `cannot access local variable 'configure_args' where it is not associated with a value`

ttyyls: needed a self reference and I substituted the whole thing instead of mutating it. seems to work now.

nekopsykose: yeah i forgot the self. , that was the only change needed

PR #1271, opened 2024-01-19
contrib/python-pyqt6: new package (6.6.1)

nekopsykose:


contrib/python-pyqt6-sip/template.py
```diff
@@ -0,0 +1,24 @@
+pkgname = "python-pyqt6-sip"
```
q66, line 1: should probably be python-pyqt6_sip

nekopsykose: that's so darned ugly

but yeah, changed

PR #1273, opened 2024-01-20
contrib/schismtracker: new package (20231029)

z-erica:


contrib/schismtracker/template.py
```diff
@@ -0,0 +1,12 @@
+pkgname = "schismtracker"
+pkgver = "20231029"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = ["autoconf", "automake"]
```
nekopsykose, line 5:
```suggestion
hostmakedepends = ["automake"]
```

nekopsykose:
doesn't this need libxv/flac etc

nekopsykose:
former i guess idk but latter is "flac support"

contrib/schismtracker/template.py
```diff
@@ -0,0 +1,12 @@
...
+pkgver = "20231029"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = ["automake"]
+makedepends = ["sdl-devel", "flac-devel", "libxv-devel"]
+pkgdesc = "Reimplementation of Impulse Tracker"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "GPL-2.0-or-later"
+url = "https://schismtracker.org"
+source = f"https://github.com/schismtracker/schismtracker/releases/download/20231029/schismtracker-{pkgver}.source.tar.gz"
```
nekopsykose, line 11:
```suggestion
source = f"https://github.com/schismtracker/schismtracker/releases/download/{pkgver}/schismtracker-{pkgver}.source.tar.gz"
```

contrib/schismtracker/template.py
```diff
@@ -0,0 +1,12 @@
+pkgname = "schismtracker"
+pkgver = "20231029"
```
nekopsykose, line 2: looking at the log i saw a helpful

sh: git: not found

probably because

```
dnl PACKAGE_VERSION will be either "" if not using git, or date of the last git commit in the form YYYYMMDD
m4_define([last_git_commit], patsubst(m4_esyscmd([git log -n 1 --date=short --format=format:%cd]), [[^0-9]]))
AC_INIT([schismtracker],[last_git_commit])
```

which sets VERSION (to nothing) which then

```
	char ver[32] = VERSION;

...

	/* show build date if we don't know last commit date (no git) */
	if (ver[0]) {
		snprintf(top_banner_normal, sizeof(top_banner_normal) - 1,
			"Schism Tracker %s", ver);
	} else {
		snprintf(top_banner_normal, sizeof(top_banner_normal) - 1,
			"Schism Tracker built %s %s", __DATE__, __TIME__);
	}

```

annoyingly this is not the version. and also setting -DVERSION or anything doesn't work because the autotools part helpfully does VERSION="" in config.h

i can't think of a good solution that isn't renaming VERSION to VERSION2 in C and then passing it via tool_flags (patching it into autotools is easier but requires updating all the time). or just ignoring all of this, but it's a small patch..

PR #1276, opened 2024-01-20
contrib/htop: update to 3.3.0

JamiKettunen:
Also enable hardening. Didn't drop `configure_gen = []` as I couldn't figure out what caused:
```
/builddir/htop-3.3.0/configure: 11567: Syntax error: end of file unexpected (expecting "fi")
```

### DIFF
--- a/contrib/htop/template.py
+++ b/contrib/htop/template.py
@@ -1,14 +1,15 @@
 pkgname = "htop"
-pkgver = "3.2.2"
+pkgver = "3.3.0"
 pkgrel = 0
 build_style = "gnu_configure"
 configure_args = ["--enable-unicode", "--enable-sensors"]
+configure_gen = []
 makedepends = ["ncurses-devel", "libsensors-devel"]
 pkgdesc = "Interactive process viewer"
 maintainer = "Jami Kettunen <jami.kettunen@protonmail.com>"
 license = "GPL-2.0-only"
 url = "https://htop.dev"
 source = f"https://github.com/htop-dev/htop/releases/download/{pkgver}/htop-{pkgver}.tar.xz"
-sha256 = "bac9e9ab7198256b8802d2e3b327a54804dc2a19b77a5f103645b11c12473dc8"
-
-configure_gen = []
+sha256 = "a69acf9b42ff592c4861010fce7d8006805f0d6ef0e8ee647a6ee6e59b743d5c"
+# CFI cannot work with libsensors dlsym() stuff
+hardening = ["vis", "!cfi"]


q66:
have you actually tested this?

```
$ htop
zsh: illegal hardware instruction  htop
```

contrib/htop/template.py
```diff
@@ -1,14 +1,13 @@
 pkgname = "htop"
-pkgver = "3.2.2"
+pkgver = "3.3.0"
 pkgrel = 0
 build_style = "gnu_configure"
-configure_args = ["--enable-unicode", "--enable-sensors"]
```
q66, line 5: why remove that?

nekopsykose: they're both the default afaict (see log)

though i prefer leaving those in anyway

q66: it's still good to be explicit about default flags because it means things will actually properly fail instead of silently disabling the feature when something is wrong

nekopsykose: yep, that's why i like them :)

JamiKettunen: I copied that from void but just realized it's unneeded since it auto-detects both during configure:
```
  (Linux) sensors:           yes
...
  unicode:                   yes
```
sooo should I leave the flags in anyway or nah?

JamiKettunen:
@q66 I did, but under riscv64 which I just realized doesn't support CFI... the one time I decided why not try something on the VisionFive 2. maybe I'll try to see how broken it is and fix it, but then again it already failing on launch like that probably isn't a good sign

JamiKettunen:
As per (full log on https://paste.c-net.org/pf6f9qeyrzdg):
```
(lldb) bt
* thread #1, name = 'htop', stop reason = signal SIGILL: illegal operand
  * frame #0: 0x00005555555978cc htop`LibSensors_init at LibSensors.c:89:11
    frame #1: 0x00005555555a13c0 htop`Platform_init at Platform.c:1053:4
    frame #2: 0x000055555556e2c3 htop`CommandLine_run(argc=<unavailable>, argv=<unavailable>) at CommandLine.c:335:9
(lldb) di
htop`LibSensors_init:
...
    0x55555559789b <+363>: movq   0x1717e(%rip), %rdi       ; dlopenHandle.llvm.17717629786427404382
    0x5555555978a2 <+370>: leaq   -0x3db5d(%rip), %rsi      ; .str.8.llvm.17717629786427404382
    0x5555555978a9 <+377>: callq  0x5555555a5150            ; symbol stub for: dlsym
    0x5555555978ae <+382>: movq   %rax, 0x1719b(%rip)       ; sym_sensors_get_value.llvm.17717629786427404382
    0x5555555978b5 <+389>: testq  %rax, %rax
    0x5555555978b8 <+392>: je     0x5555555977c4            ; <+148> at LibSensors.c:93:8
    0x5555555978be <+398>: callq  0x5555555a5140            ; symbol stub for: dlerror
    0x5555555978c3 <+403>: testq  %rax, %rax
    0x5555555978c6 <+406>: jne    0x5555555977c4            ; <+148> at LibSensors.c:93:8
->  0x5555555978cc <+412>: ud1l   0x2(%eax), %eax
```
the problem is around https://github.com/htop-dev/htop/blob/3.3.0/linux/LibSensors.c#L41-L89, and from the looks of it there's no way for CFI to deal with it properly atm: https://discourse.llvm.org/t/rfc-accommodate-dynamically-resolved-function-pointers-with-cfi/75804

so I say let's forget about CFI for now and get this update in

PR #1278, opened 2024-01-20
contrib/python-flask: new package (3.0.1)

firefly-cpp:
https://github.com/pallets/flask

### DIFF
--- a/contrib/python-flask/template.py
+++ b/contrib/python-flask/template.py
@@ -1,0 +1,28 @@
+pkgname = "python-flask"
+pkgver = "3.0.1"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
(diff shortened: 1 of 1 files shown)


contrib/python-flask/template.py
```diff
@@ -0,0 +1,35 @@
...
+    "python-pytest",
+    "python-werkzeug",
+]
+depends = [
+    "python-blinker",
+    "python-click",
+    "python-itsdangerous",
+    "python-jinja2",
+    "python-werkzeug",
+]
```
nekopsykose, line 25: put depends above checkdepends then to `checkdepends = ["python-pytest"] + depends`

firefly-cpp: fixed

PR #1279, opened 2024-01-20
contrib/python-fasttext: new package (0.9.2)

a16bitsysop:


q66:
is it needed for anything? if yes, why not a separate template?

a16bitsysop:
not needed or anything, just for completeness because they are in the source.  It seems popular on pypi https://pypistats.org/packages/fasttext

in the packaging doc it says separate bindings, not put in own templare:
https://github.com/chimera-linux/cports/blob/master/Packaging.md#package-naming

but could be a separate template

q66:
it requires a separate buildsystem, so it qualifies for a separate template (called `python-fasttext`)

a16bitsysop:
okay, yes makes it easier to use built in functions.  Will change

contrib/fasttext-python/template.py
```diff
@@ -0,0 +1,24 @@
+pkgname = "fasttext-python"
```
nekopsykose, line 1: python-fasttext

nekopsykose:
is there a way to link it to the libfasttext.so.0 from fasttext or would that be too difficult

a16bitsysop:
I can have a go, I guess patch out FASTTEXT_SRC from https://github.com/facebookresearch/fastText/blob/b733943e84263f432fa47588643822194ee03dd1/setup.py#L75
and get it to link.  will have a look

nekopsykose:
sure, you could put /usr/include/whatever there and -lsomething in extra_link_args in the same place

a16bitsysop:
tried that and even deleting src dir and not linking but still seems to build and not depend on fasttext.so ?

nekopsykose:
yeah i don't remember how to do it and it seems to not support it

a16bitsysop:
okay, needs numpy as a runtime dep.  Which needs python-meson to build

nekopsykose:
numpy needs a bunch more stuff to work, like i'd have to build openblas first too, and the only fortran compiler pressent is flang-new so there's probably a bunch of subtle bugs in both openblas and numpy etc

if this needs numpy then you won't be able to use it for a long time

nekopsykose:
>  Which needs python-meson to build  

building numpy with meson doesn't even work yet because they use a forked version of meson copied inside the tree haha

q66:
if this needs numpy, convert to draft

a16bitsysop:
okay

q66:
we've got numpy now, maybe can do this now?

a16bitsysop:
Okay will edit this

q66:
hm, it says in the documentation that you need scipy too?

a16bitsysop:
I did a grep on the python directory and it looks like scipy is only used in one example?

https://github.com/facebookresearch/fastText/blob/1142dc4c4ecbc19cc16eee5cdd28472e689267e6/python/doc/examples/train_unsupervised.py#L18

q66:
is it? looks used here https://github.com/facebookresearch/fastText/blob/1142dc4c4ecbc19cc16eee5cdd28472e689267e6/eval.py#L16

q66:
hm but that is not installed, in installed files it does not seem used... okay

nekopsykose:
yeah, that stuff isn't installed for the python module, the repo is both the module and the actual lib and other stuff so it's confusing

PR #1282, opened 2024-01-21
main/nload: move from contrib

firefly-cpp:
The **nload** package is essential for analyzing network traffic and is included in almost all distributions. Since **nload** works well on Chimera, let's move to the main.


q66:
nah

PR #1283, opened 2024-01-21
contrib/python-asgiref: new package (3.7.2)

firefly-cpp:
https://github.com/django/asgiref

### DIFF
--- a/contrib/python-asgiref/template.py
+++ b/contrib/python-asgiref/template.py
@@ -1,0 +1,21 @@
+pkgname = "python-asgiref"
+pkgver = "3.7.2"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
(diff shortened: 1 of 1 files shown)


contrib/python-asgiref/template.py
```diff
@@ -0,0 +1,21 @@
...
+    "python-installer",
+    "python-setuptools",
+    "python-wheel",
+]
+checkdepends = ["python-pytest", "python-pytest-asyncio"]
+pkgdesc = "ASGI specs, helper code, and adapters"
+maintainer = "firefly-cpp <iztok@iztok.space>"
+license = "BSD-3-Clause"
+url = "https://github.com/django/asgiref"
+source = f"https://github.com/django/asgiref/archive/{pkgver}.tar.gz"
```
q66, line 16: use PYPI_SITE as is normal for python modules

PR #1284, opened 2024-01-21
contrib/oksh: new package (7.4)

ttyyls:


contrib/oksh/template.py
```diff
@@ -0,0 +1,28 @@
+pkgname = "oksh"
+pkgver = "7.4"
+pkgrel = 0
+build_style = "configure"
```
q66, line 4: gnu_configure (+ `configure_gen = []`) for this one (it technically isn't, but...)

contrib/oksh/template.py
```diff
@@ -0,0 +1,28 @@
...
+configure_args = [
```
q66, line 5: and ditch this stuff afterwards

PR #1285, opened 2024-01-21
Protobuf 25.2

JamiKettunen:
Also make `mosh` use `gmake` due to strange build errors even on current `master` of cports without any of this, not sure if it's a `bmake` bug or what as it used to build previously.


nekopsykose:
probably new incompatibility with bmake 20231210, in my experience every year or so bmake updates break random stuff :D  

in this case, it ended up with a wrong version.h written (which still has .sh in it) which then breaks etc

PR #1287, opened 2024-01-22
contrib/nc: new package (3.8.2)

subnut:


subnut:
netcat has a much simpler syntax than socat, and IMHO seems to be more popular than socat/ncat/etc.

Also, among netcat/gnu-netcat/openbsd-netcat, the openbsd one seems to be more popular IMHO.

I can't provide numbers, but that's the general trend I've seen in reddit/stackoverflow/etc.

subnut:
And libressl's nc is pretty much directly copied from the OpenBSD `src` tree (AFAIR)

subnut:
I did it this way instead of just packaging the whole libressl because... well we already have OpenSSL. And I presume q99 won't like the idea of having _both_ libressl _and_ openssl

q66:
the template doesn't meet packaging quality requirements, and i'm not interested in a hacky build like this anyway

it's also not guaranteed to work because the libretls ABI may not match the libressl ABI, so it's just overall a bad idea

i wouldn't be opposed to importing the freebsd variant of netcat in chimerautils, but definitely not this

nekopsykose:
> it's also not guaranteed to work because the libretls ABI may not match the libressl ABI, so it's just overall a bad idea  

the libretls abi does match the libtls abi. that's why it exists in the first place

the rest stands tho

q66:
afaict it's not required to, libressl can bump their soname and then it'll no longer match like this

nekopsykose:
yeah, every libressl major the soname goes up, and then one would have to wait for libretls to catch up with the changes. that's because libressl breaks every release

but aside from that they match :)

PR #1291, opened 2024-01-22
contrib/ctags: new package (6.1.0)

rbgarga:


rbgarga:
Maybe `contrib` would be better for this project, but I'm not used with the rules to define where to add new packages so I used `user` instead

nekopsykose:
contrib is fine for this

nekopsykose:
also `new package` not `New package`

rbgarga:
@nekopsykose done

contrib/universal-ctags/template.py
```diff
@@ -0,0 +1,37 @@
...
+    "pkgconf",
+    "python-docutils"
+]
+depends = [
+    "jansson",
+    "libxml2",
+    "libpcre2",
+    "libyaml"
+]
+checkdepends = ["python"] + depends
```
nekopsykose, line 31:
```suggestion
checkdepends = ["python"]
```

these are all traced by NEEDED from the elf file:  

```
=> universal-ctags-6.1.0-r0: running pre_pkg hook: 001_runtime_deps...
   SONAME: libxml2.so.2 <-> libxml2
   SONAME: libjansson.so.4 <-> jansson
   SONAME: libyaml-0.so.2 <-> libyaml
   SONAME: libpcre2-8.so.0 <-> libpcre2
```

rbgarga: oh, very nice!

nekopsykose:
also i wonder if it should merely be named 'ctags'? the repo and tool is ctags, and i don't think there would be a need to package exuberant- too

contrib/universal-ctags/template.py
```diff
@@ -0,0 +1,37 @@
...
+]
+make_cmd = "gmake"
+make_dir = "."
+hostmakedepends = [
+    "automake",
+    "gmake",
+    "jansson-devel",
+    "libtool",
+    "libxml2-devel",
+    "libyaml-devel",
```
nekopsykose, line 20: -devel ones should be moved to `makedepends=` instead

rbgarga: python-docutils is used to build manpages, if I get it right, it should also be moved to `makedepends`, right?

q66: no; `hostmakedepends` is stuff that needs to run in the host environment, while `makedepends` are target dependencies (typically what the project links to and what will become the target package's runtime dependencies)

the distinction is important for cross-compiling, otherwise they are the same

PR #1296, opened 2024-01-22
contrib/firefox: update to 122.0

nekopsykose:
would be nice to roll this with libvpx in the other mr but have to fix thunderbird, bleh  

dbus-glib was dropped in 120

[ci skip]


q66:
do they have any plans to actually fix that stuff in the ESR branch thunderbird is based on?

q66:
(i wonder how bad of an idea would be to roll tbirb to 122? :p)

nekopsykose:
> do they have any plans to actually fix that stuff in the ESR branch thunderbird is based on?  

i don't think so :'(  

> (i wonder how bad of an idea would be to roll tbirb to 122? :p)  

i was having the same thought. the next esr is like.. 128, which is in a long time..

i'm not sure how actually stable it is in practice though

PR #1299, opened 2024-01-22
user/git-extras: new package (7.1.0)

ttyyls:


user/git-extras/template.py
```diff
@@ -0,0 +1,100 @@
...
+        "git-undo",
+        "git-unlock",
+        "git-utimes",
+    ]:
+        self.install_bin("bin/" + utility)
+        self.install_man("man/" + utility + ".1")
+
+    # utilities without a dedicated manpage
+    self.install_bin("bin/git-scp")
+    self.install_link("git-scp", "usr/bin/git-rscp")
```
nekopsykose, line 93:
```suggestion
    self.install_bin("bin/*", glob=True)
    self.install_link("git-scp", "usr/bin/git-rscp")
    self.install_man("man/*.1", glob=True)
```

user/git-extras/template.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def post_install(self):
+    self.install_completion("etc/bash_completion.sh", "bash")
+    self.install_completion("etc/git-extras-completion.zsh", "zsh")
+    self.install_completion("etc/git-extras.fish", "fish")
+    self.install_license("LICENSE")
```
nekopsykose, line 100: put all of this in do_install

user/git-extras/template.py
nekopsykose: well- you'd have to `self.rm` git-rscp first to replace it with a link, since the first install will deref it and install a dupe, i guess

ttyyls: globbing causes git-rcsp to be selected anyway, which breaks the whole do_install step since hence the manual listing of utilities

user/git-extras/template.py
ttyyls: ok

user/git-extras/template.py
nekopsykose: that would make sense if i didn't just say how to work around it

ttyyls: mb I missed that

nekopsykose: or even better:  

```
    self.install_files("bin", "usr")
```

then it installs the link normally

nekopsykose: so the whole thing:  

```
def do_install(self):
    self.install_files("bin", "usr")
    self.install_man("man/*.1", glob=True)
    self.install_completion("etc/bash_completion.sh", "bash")
    self.install_completion("etc/git-extras-completion.zsh", "zsh")
    self.install_completion("etc/git-extras.fish", "fish")
```

ttyyls: most definitely

user/git-extras/template.py
```diff
@@ -0,0 +1,19 @@
+pkgname = "git-extras"
+pkgver = "7.1.0"
+pkgrel = 0
+depends = ["git", "bash"]
+pkgdesc = "Extra Git utilities"
+maintainer = "ttyyls <contact@behri.org>"
+license = "BSD-2-Clause"
```
nekopsykose, line 7:
```suggestion
license = "MIT"
```

PR #1300, opened 2024-01-23
contrib/rlwrap: new package (0.46.1)

ttyyls:


q66:
does this provide anything over just using socat? because this thing is dead

wezm:
Isn't socat a different thing? rlwrap lets you use readline line editing and history features on tools that don't use it.

q66:
`socat READLINE EXEC:<your command line>`

ttyyls:
fair enough

PR #1301, opened 2024-01-23
contrib/fish-shell: use correct check target

yanchan09:
fish has its own test target that should be used instead of just using ctest alone (see https://github.com/fish-shell/fish-shell/blob/3.7.0/cmake/Tests.cmake#L18)

Also removed `!int` hardening option as I wasn't able to reproduce the mentioned test failure.

### DIFF
--- a/contrib/fish-shell/template.py
+++ b/contrib/fish-shell/template.py
@@ -1,8 +1,8 @@
 pkgname = "fish-shell"
 pkgver = "3.7.0"
-pkgrel = 0
+pkgrel = 1
 build_style = "cmake"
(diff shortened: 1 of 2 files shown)


contrib/fish-shell/template.py
```diff
@@ -2,7 +2,7 @@
 pkgver = "3.7.0"
 pkgrel = 0
```
nekopsykose, line 3:
```suggestion
pkgrel = 1
```

nekopsykose:
well, let's see if that works..

PR #1304, opened 2024-01-23
contrib/libaacs: enable cfi

mikoxyz:


nekopsykose:
you could replace the patch with a regular file in files/

but i dunno how much i like this kind of thing

mikoxyz:
> you could replace the patch with a regular file in files/
> 
> but i dunno how much i like this kind of thing

can we even copy files into the src dir?

yanchan09:
honestly i feel like cbuild should have a better support e.g. helper functions for cfi (and ubsan?) exclusions in the first place

q66:
these shouldn't be exclusions but a patch

nekopsykose:
is this still a draft

mikoxyz:
please take a look at this again

PR #1307, opened 2024-01-23
main/evince: fix `SubOverflow` on some documents

mhmdanas:


nekopsykose:
though shouldn't it be 0 for ==

nekopsykose:
i forget if that breaks qsort or not

nekopsykose:
ah it is, since x>y will be 0 in that case

PR #1313, opened 2024-01-25
contrib/fractal: new package (6)

nekopsykose:
the meson+rust handling is a bit of a mess. should probably update helvum to do the same (currently it does not pass the envs and hence also not the lto env). i can't think of a better solution than just skipping the build and manually doing it i guess


subnut:
I'm curious about what's the benefit of running the natively-built app vs the flatpak one.

nekopsykose:
none except it's available outside aarch64/x86_64 too

PR #1316, opened 2024-01-25
contrib/haproxy: new package (2.9.6)

ttyyls:


contrib/haproxy/haproxy.pre-install
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+echo 'u haproxy - "haproxy user" /var/lib/haproxy /usr/bin/nologin' | \
+    /usr/bin/systemd-sysusers --replace /usr/lib/sysusers.d/haproxy.conf - || :
```
nekopsykose, line 4: you don't need this at all

contrib/haproxy/template.py
```diff
@@ -0,0 +1,56 @@
...
+    f"lua{_luaver}-devel",
+    "openssl-devel",
+    "pcre2-devel",
+    "zlib-devel",
+]
+pkgdesc = "TCP/HTTP reverse proxy for high availability environments"
+maintainer = "ttyyls <contact@behri.org>"
+license = "GPL-2.0-or-later AND LGPL-2.1-or-later WITH custom:OpenSSL-exception"
+url = "https://www.haproxy.org"
+source = f"{url}/download/{_majver}/src/{pkgname}-{pkgver}.tar.gz"
```
nekopsykose, line 40:
```suggestion
source = f"{url}/download/{pkgver[:-pkgver.rfind('.')]}/src/{pkgname}-{pkgver}.tar.gz"
```

contrib/haproxy/template.py
```diff
@@ -0,0 +1,55 @@
...
+]
+make_check_target = "reg-tests"
+make_use_env = True
+hostmakedepends = [
+    "gmake",
+    "pkgconf",
+]
+makedepends = [
+    "linux-headers",
+    f"lua{_luaver}-devel",
```
nekopsykose, line 30:
```suggestion
    "lua5.4-devel",
```

contrib/haproxy/files/haproxy.cfg
```diff
@@ -0,0 +1,43 @@
...
+
+defaults
+    mode http
+    timeout connect 5000ms
+    timeout client 50000ms
+    timeout server 50000ms
+
+listen http-in
+    bind *:80
+    server server1 127.0.0.1:8000 maxconn 32
```
nekopsykose, line 43: this is the same config twice though, why?

contrib/haproxy/template.py
```diff
@@ -0,0 +1,55 @@
...
+maintainer = "ttyyls <contact@behri.org>"
+license = "GPL-2.0-or-later AND LGPL-2.1-or-later WITH custom:OpenSSL-exception"
+url = "https://www.haproxy.org"
+source = f"{url}/download/{pkgver[:-pkgver.rfind('.')]}/src/{pkgname}-{pkgver}.tar.gz"
+sha256 = "ed517c65abd86945411f6bcb18c7ec657a706931cb781ea283063ba0a75858c0"
+hardening = ["vis", "cfi"]
+# hard depends on vtest which doesn't have releases
+options = ["!check"]
+
+system_users = ["haproxy"]
```
nekopsykose, line 45:
```suggestion
```

contrib/haproxy/files/haproxy.cfg
ttyyls: I took straight from the docuemntation, see https://docs.haproxy.org/2.9/configuration.html#2.7

contrib/haproxy/template.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def post_install(self):
+    self.install_file(self.files_path / "haproxy.cfg", "etc/haproxy")
+    self.install_dir("var/lib/haproxy", empty=True)
+    self.install_dir("var/log/haproxy", empty=True)
```
nekopsykose, line 51:
```suggestion
```
tmpfiles.d

contrib/haproxy/template.py
```diff
@@ -0,0 +1,55 @@
...
+
+    self.install_license("LICENSE")
+    self.install_license("doc/gpl.txt")
+    self.install_license("doc/lgpl.txt")
```
nekopsykose, line 55:
```suggestion
```

contrib/haproxy/files/haproxy
```diff
@@ -0,0 +1,6 @@
+# haproxy service
+
+type       = process
+command    = /usr/bin/haproxy
```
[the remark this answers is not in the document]
nekopsykose: looking at https://github.com/haproxy/haproxy/blob/96385f40b508a395fcd802a9dff13e968c12433d/admin/systemd/haproxy.service.in i don't think there's a way to map this, the best you can do is type = bgprocess and set pid-file to what it writes

contrib/haproxy/template.py
ttyyls: it fails CI ?

contrib/haproxy/template.py
```diff
@@ -0,0 +1,42 @@
+pkgname = "haproxy"
+pkgver = "2.9.3"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_build_args = [
+    "LUA_LIB=/usr/lib/lua5.4",
+    "LUA_LIB=/usr/include/lua5.4",
```
nekopsykose, line 8: both of these are autodetected from what i can tell

contrib/haproxy/template.py
nekopsykose: ah yes, keep the first one

contrib/haproxy/files/haproxy.cfg
nekopsykose:
```
Assuming haproxy is in $PATH, test these configurations in a shell with:

    $ sudo haproxy -f configuration.conf -c
```


```
/ # haproxy -c -f x.cfg 
[NOTICE]   (12) : haproxy version is 2.8.5-aaba8d0
[ALERT]    (12) : config : parsing [x.cfg:23] : 'maxconn' already specified. Continuing.
[ALERT]    (12) : config : Parsing [x.cfg:31]: proxy 'http-in' has the same name as frontend 'http-in' declared at x.cfg:11.
[ALERT]    (12) : config : Error(s) found in configuration file : x.cfg
[ALERT]    (12) : config : Fatal errors found in configuration.
```

contrib/haproxy/template.py
ttyyls: ok

contrib/haproxy/files/haproxy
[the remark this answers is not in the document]
nekopsykose: the -D doesn't do anything. it means the same thing as `daemon` in the config file (i.e. it's the same option), and it makes it fork to background (which breaks with type = process)

contrib/haproxy/template.py
nekopsykose: you still need a tmpfiles.d config for these

nekopsykose: see main/chrony/files/tmpfiles.conf for one example

contrib/haproxy/files/haproxy
```diff
@@ -0,0 +1,8 @@
+# haproxy service
+
+type       	    = bgprocess
+command    	    = /usr/bin/haproxy -f /etc/haproxy/haproxy.cfg -p /run/haproxy.pid
+pid-file	    = /run/haproxy.pid
```
nekopsykose, line 5: spaces not tabs

contrib/haproxy/template.py
```diff
@@ -0,0 +1,46 @@
...
+url = "https://www.haproxy.org"
+source = f"{url}/download/{pkgver[:-pkgver.rfind('.')]}/src/{pkgname}-{pkgver}.tar.gz"
+sha256 = "ed517c65abd86945411f6bcb18c7ec657a706931cb781ea283063ba0a75858c0"
+hardening = ["vis", "cfi"]
+# hard depends on vtest which doesn't have releases
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_file(self.files_path / "haproxy.cfg", "etc/haproxy")
```
nekopsykose, line 44: you have to install the sysusers.conf/tmpfiles.conf here too

contrib/haproxy/template.py
```diff
@@ -0,0 +1,58 @@
+pkgname = "haproxy"
+pkgver = "2.9.3"
+pkgrel = 0
+build_style = "makefile"
```
nekopsykose, line 4: it ignores cflags/ldflags, you'd probably want to change

```
CFLAGS = ...
```

to

```
CFLAGS := ... $(CFLAGS)
```

same for ldflags

ttyyls: isn't that what `make_use_env = True` is for ?

nekopsykose: make_use_env doesn't make makefiles that ignore cflags not ignore them, it replaces

````
make CFLAGS="default cflags" ...
```

with

```
CFLAGS="default cflags" ... make
```

which as the name implies uses the env instead of passing args directly. in this case, though, it seems all of the makefile LDFLAGS/CFLAGS are useless, so you can fix this by removing make_use_env

it will then fail because that passes LD=ld.lld and the makefile does $(LD) -o.. (links with LD instead of CC), which you can fix by changing the linker invocation to CC instead

ttyyls: oh my bad I got confused, I will quickly patch it

nekopsykose: or well, it does complain if you do that, so patch the cflags instead :)

nekopsykose: it also needs !int and doesn't work with cfi

contrib/haproxy/files/haproxy
```diff
@@ -0,0 +1,8 @@
+# haproxy service
+
+type           = bgprocess
+command        = /usr/bin/haproxy -f /etc/haproxy/haproxy.cfg -p /run/haproxy.pid
+pid-file       = /run/haproxy.pid
+logfile        = /var/log/haproxy.log
+run-as         = haproxy
```
nekopsykose, line 7: this can't work because dinit changes to this user before running the daemon. it then can't bind any (<1024) ports

normally instead you set

```
user        haproxy
group       haproxy
```

in the haproxy config

you can also name it `_haproxy` but not sure how much that matter

contrib/haproxy/patches/use_env_flags.patch
```diff
@@ -0,0 +1,29 @@
...
++LD = cc
+ 
+ #### Debug flags (typically "-g").
+ # Those flags only feed CFLAGS so it is not mandatory to use this form.
+@@ -289,13 +289,13 @@ ARCH_FLAGS        = $(ARCH_FLAGS.$(ARCH))
+ # These CFLAGS contain general optimization options, CPU-specific optimizations
+ # and debug flags. They may be overridden by some distributions which prefer to
+ # set all of them at once instead of playing with the CPU and DEBUG variables.
+-CFLAGS = $(ARCH_FLAGS) $(CPU_CFLAGS) $(DEBUG_CFLAGS) $(SPEC_CFLAGS)
++CFLAGS := $(ARCH_FLAGS) $(CPU_CFLAGS) $(DEBUG_CFLAGS) $(SPEC_CFLAGS) $(CFLAGS)
```
nekopsykose, line 19: while here, remove ARCH_FLAGS from both and CPU_CFLAGS since they're just noise

will have to check aarch64 later

contrib/haproxy/patches/use_env_flags.patch
```diff
@@ -0,0 +1,29 @@
+diff --git Makefile Makefile
+index a66c9b69c..64c44b85f 100644
+--- Makefile
++++ Makefile
+@@ -168,7 +168,7 @@ ARCH =
+ 
+ #### Toolchain options.
+ CC = cc
+-LD = $(CC)
++LD = cc
```
nekopsykose, line 10: you don't need to change this with make_use_env

nekopsykose:
looks good :)  

the only suggestion i have left is to perhaps patch https://github.com/haproxy/haproxy/blob/96385f40b508a395fcd802a9dff13e968c12433d/src/haproxy.c#L845 to do something, like in

main/pipewire/patches/readiness.patch

i have no idea if that works correctly with bgprocess or not though, especially since a reexec follows, etc

contrib/haproxy/files/haproxy
```diff
@@ -0,0 +1,7 @@
+# haproxy service
+
+type           = bgprocess
```
q66, line 3: this should be a process service, and it should be executed on the foreground; as far as i can tell the `-Ws` argument can do this, it just needs some minor patching to switch the systemd sd_notify logic to use dinit's readiness notification

q66: probably something like this should work (untested, test it yourself)

```
From af3e702b16d7ab50eddcc65ccd70ebcce5e41f23 Mon Sep 17 00:00:00 2001
From: q66 <q66@chimera-linux.org>
Date: Sun, 28 Jan 2024 04:50:21 +0100
Subject: [PATCH] allow for readiness notification with -Ws and dinit

---
 src/haproxy.c | 18 +++++++++++++++---
 1 file changed, 15 insertions(+), 3 deletions(-)

diff --git a/src/haproxy.c b/src/haproxy.c
index 5e13ae6..2ab0f1e 100644
--- a/src/haproxy.c
+++ b/src/haproxy.c
@@ -618,6 +618,8 @@ static void usage(char *name)
 		"        -W master-worker mode.\n"
 #if defined(USE_SYSTEMD)
 		"        -Ws master-worker mode with systemd notify support.\n"
+#else
+		"        -Ws master-worker mode with fd notify support.\n"
 #endif
 		"        -q quiet mode : don't display messages\n"
 		"        -c check mode : only check config files and exit\n"
@@ -914,6 +916,10 @@ void reexec_on_failure()
 	 * the READY=1 signal still need to be sent */
 	if (global.tune.options & GTUNE_USE_SYSTEMD)
 		sd_notify(0, "READY=1\nSTATUS=Reload failed!\n");
+#else
+	/* the reexec is transparent to the supervisor and the fd that was
+	 * used for the notification is already gone, so do nothing here
+	 */
 #endif
 
 	mworker_reexec_waitmode();
@@ -1761,9 +1767,6 @@ static void init_args(int argc, char **argv)
 				arg_mode |= MODE_MWORKER | MODE_FOREGROUND;
 #if defined(USE_SYSTEMD)
 				global.tune.options |= GTUNE_USE_SYSTEMD;
-#else
-				ha_alert("master-worker mode with systemd support (-Ws) requested, but not compiled. Use master-worker mode (-W) if you are not using Type=notify in your unit file or recompile with USE_SYSTEMD=1.\n\n");
-				usage(progname);
 #endif
 			}
 			else if (*flag == 'W')
@@ -3764,6 +3767,15 @@ int main(int argc, char **argv)
 #if defined(USE_SYSTEMD)
 					if (global.tune.options & GTUNE_USE_SYSTEMD)
 						sd_notifyf(0, "READY=1\nMAINPID=%lu\nSTATUS=Ready.\n", (unsigned long)getpid());
+#else
+					const char *rvar = getenv("HAPROXY_READY_FD");
+					if (rvar && (global.mode & MODE_FOREGROUND)) {
+						int fdn = atoi(rvar);
+						if (fdn && !fcntl(fdn, F_GETFD)) {
+							write(fdn, "READY=1\n", sizeof("READY=1"));
+							close(fdn);
+						}
+					}
 #endif
 					/* if not in wait mode, reload in wait mode to free the memory */
 					setenv("HAPROXY_LOAD_SUCCESS", "1", 1);
-- 
2.43.0
```

the service should be `type = process`, and have `ready-notification = pipevar:HAPROXY_READY_FD`, and the `-Ws` argument should be passed

contrib/haproxy/files/sysusers.conf
```diff
@@ -0,0 +1,3 @@
+# create haproxy user 
+
+u haproxy - "haproxy user" /var/lib/haproxy /usr/bin/nologin
```
yanchan09, line 3: Shouldn't system users in Chimera Linux usually be prefixed with an underscore?

contrib/haproxy/template.py
```diff
@@ -0,0 +1,60 @@
...
+pkgver = "2.9.4"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_build_args = [
+    "TARGET=linux-musl",
+    "USE_LUA=1",
+    "USE_NS=1",
+    "USE_OPENSSL=1",
+    "USE_PCRE2=1",
```
nekopsykose, line 11: you can also add

```
    "USE_QUIC=1",
    "USE_QUIC_OPENSSL_COMPAT=1",
```

nekopsykose: ah, USE_PROMEX too

contrib/haproxy/files/haproxy
```diff
@@ -0,0 +1,8 @@
+# haproxy service
+
+type                = process
+command             = /usr/bin/haproxy -Ws -f /etc/haproxy/haproxy.cfg -p /run/haproxy.pid
+ready-notification  = pipevar:HAPROXY_READY_FD
+pid-file            = /run/haproxy.pid
```
q66, line 6: pidfile should not be specified for process services (nor in the command)

contrib/haproxy/files/sysusers.conf
```diff
@@ -0,0 +1,3 @@
+# create haproxy user 
+
+u _haproxy - "haproxy user" /var/empty /usr/bin/nologin
```
q66, line 3: since introduction of sd-sysusers, it should no longer be necessary to explicitly specify nologin path (it's the same as the default now)

contrib/haproxy/files/tmpfiles.conf
```diff
@@ -0,0 +1,5 @@
+# Create haproxy state directories
+
+d /var/log/haproxy 0755 _haproxy _haproxy -
+
```
q66, line 4: drop empty line

contrib/haproxy/patches/use_env_flags.patch
```diff
@@ -0,0 +1,20 @@
...
+ # set all of them at once instead of playing with the CPU and DEBUG variables.
+-CFLAGS = $(ARCH_FLAGS) $(CPU_CFLAGS) $(DEBUG_CFLAGS) $(SPEC_CFLAGS)
++CFLAGS := $(DEBUG_CFLAGS) $(SPEC_CFLAGS) $(CFLAGS)
+ 
+ #### Common LDFLAGS
+ # These LDFLAGS are used as the first "ld" options, regardless of any library
+ # path or any other option. They may be changed to add any linker-specific
+ # option at the beginning of the ld command line.
+-LDFLAGS = $(ARCH_FLAGS) -g
++LDFLAGS := -g $(LDFLAGS)
```
q66, line 17: this is wrong; -g should not be added (instead, CFLAGS should be passed together with LDFLAGS)

contrib/haproxy/patches/0001-use-env-flags.patch
```diff
@@ -0,0 +1,32 @@
...
+ # set all of them at once instead of playing with the CPU and DEBUG variables.
+-CFLAGS = $(ARCH_FLAGS) $(CPU_CFLAGS) $(DEBUG_CFLAGS) $(SPEC_CFLAGS)
++CFLAGS := $(DEBUG_CFLAGS) $(SPEC_CFLAGS) $(CFLAGS)
+ 
+ #### Common LDFLAGS
+ # These LDFLAGS are used as the first "ld" options, regardless of any library
+ # path or any other option. They may be changed to add any linker-specific
+ # option at the beginning of the ld command line.
+-LDFLAGS = $(ARCH_FLAGS) -g
++LDFLAGS := $(CFLAGS) $(LDFLAGS)
```
q66, line 26: as far as i can tell, it's not using that (add `V=1` to `make_build_args` and see, while at it leave it there as it's much more useful in logs)

ttyyls:
afaict this last patch passes the env flags correctly, I had to hardcode cmd_{LD,CC} since there is no obvious way to override it from init_build when make_use_env = False.

ttyyls:
apparently it depends on some integer overflow behavior so I reverted make_use_env = True and let it combine with the internal SPEC_FLAGS logic which enables -fwrapv and -fno-strict-overflow otherwise it produces dysfunctional binaries.

Afaik it's ready for merging

q66:
yep this looks good now, thanks

PR #1317, opened 2024-01-25
main/cryptsetup: update to 2.7.0

nekopsykose:
https://mirrors.edge.kernel.org/pub/linux/utils/cryptsetup/v2.7/v2.7.0-ReleaseNotes  

notably we don't have `CONFIG_BLK_SED_OPAL` enabled

### DIFF
--- a/main/cryptsetup/template.py
+++ b/main/cryptsetup/template.py
@@ -1,16 +1,21 @@
 pkgname = "cryptsetup"
-pkgver = "2.6.1"
+pkgver = "2.7.0"
 pkgrel = 0
 build_style = "gnu_configure"
 configure_args = [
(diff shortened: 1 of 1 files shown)


q66:
don't the OPAL drives require proprietary userland software to interact with?

q66:
i guess perhaps not with this version of cryptsetup?

nekopsykose:
apparently not, cryptsetup can just do it now. well, i don't have any to test with..

nekopsykose:
also the Compatibility notes: spells out all the million cases where it won't work too. but aside from that, when it would work it would work i guess

mhmdanas:
Working fine for me FWIW

nekopsykose:
same for me

PR #1318, opened 2024-01-25
contrib/nginx: new package (1.24.0)

rbgarga:


contrib/nginx/template.py
```diff
@@ -0,0 +1,72 @@
...
+hostmakedepends = ["bmake"]
+makedepends = ["openssl-devel", "pcre2-devel", "zlib-devel"]
+pkgdesc = "Robust and small WWW server"
+maintainer = "Renato Botelho do Couto <renato@netgate.com>"
+license = "BSD-2-Clause"
+url = "https://nginx.org"
+source = f"https://nginx.org/download/{pkgname}-{pkgver}.tar.gz"
+sha256 = "77a2541637b92a621e3ee76776c8b7b40cf6d707e69ba53a940283e30ff2f55d"
+# Not available on nginx build tools
+options = ["!check"]
```
nekopsykose, line 59: it is, you can get it from https://hg.nginx.org/nginx-tests/archive/22f45bf99a9e etc

contrib/nginx/template.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+    self.install_dir("var/log/nginx", empty=True)
+    self.install_dir("var/tmp/nginx", empty=True)
+    self.install_file("man/nginx.8", "usr/share/man/man8")
```
nekopsykose, line 66:
```suggestion
    self.install_man("man/nginx.8")
```

contrib/nginx/template.py
```diff
@@ -0,0 +1,72 @@
...
+    "--without-mail_smtp_module",
+    "--with-mail_ssl_module",
+    "--with-stream_realip_module",
+    "--with-stream_ssl_module",
+    "--with-stream_ssl_preread_module",
+    "--with-threads",
+    "--with-mail=dynamic",
+    "--with-stream=dynamic",
+]
+make_cmd = "bmake"
```
nekopsykose, line 48:
```suggestion
```

it's the default already

contrib/nginx/template.py
```diff
@@ -0,0 +1,72 @@
...
+make_dir = "."
+hostmakedepends = ["bmake"]
```
nekopsykose, line 50:
```suggestion
```

ditto

nekopsykose:
you'll want to ignore the odd-numbered versions in update.py

contrib/nginx/template.py
```diff
@@ -0,0 +1,72 @@
...
+pkgrel = 0
+build_style = "configure"
+configure_args = [
+    "--prefix=/etc/nginx",
+    "--sbin-path=/usr/bin/nginx",
+    "--modules-path=/usr/libexec/nginx",
+    "--conf-path=/etc/nginx/nginx.conf",
+    "--http-log-path=/var/log/nginx/access.log",
+    "--error-log-path=/var/log/nginx/error.log",
+    "--pid-path=/run/nginx.pid",
```
nekopsykose, line 12: --lock-path too (it defaults to logs/nginx.lock apparently)

nekopsykose:
probably want `--with-file-aio` too, probably more stuff i'm forgetting..

wish the nginx config was more obvious in general

contrib/nginx/template.py
```diff
@@ -0,0 +1,72 @@
...
+    "--with-compat",
+    "--with-pcre",
+    "--user=_www",
+    "--group=_www",
+    "--http-client-body-temp-path=/var/tmp/nginx/client_body_temp",
+    "--http-fastcgi-temp-path=/var/tmp/nginx/fastcgi_temp",
+    "--http-proxy-temp-path=/var/tmp/nginx/proxy_temp",
+    "--http-scgi-temp-path=/var/tmp/nginx/scgi_temp",
+    "--http-uwsgi-temp-path=/var/tmp/nginx/uwsgi_temp",
+    "--with-http_v2_module",
```
nekopsykose, line 22: sort these afterward

contrib/nginx/files/nginx
```diff
@@ -0,0 +1,7 @@
+# nginx service
+
+type            = process
+command         = /usr/bin/nginx -g 'daemon off;'
+depends-on      = local.target
+restart         = true
```
nekopsykose, line 6:
```suggestion
```

default already

also, we could probably patch in readiness (like in pipewire readiness.patch)..

nekopsykose:
and `New`->`new`. looks good overall :)

contrib/nginx/template.py
nekopsykose: from what i can tell it would also need perl-fcgi perl-protocol-websocket and uwsgi. the first two should be easy to package

contrib/nginx/template.py
rbgarga: right, where do you suggest to save that?

nekopsykose: also probably make these into /run/nginx/ and put /run/nginx/ as 750 into the tmpfiles.d entry as well

nekopsykose: >  probably make these   

lock/pid specifically into /run/nginx

contrib/nginx/template.py
```diff
@@ -0,0 +1,72 @@
...
+    "--sbin-path=/usr/bin/nginx",
+    "--modules-path=/usr/libexec/nginx",
+    "--conf-path=/etc/nginx/nginx.conf",
+    "--http-log-path=/var/log/nginx/access.log",
+    "--error-log-path=/var/log/nginx/error.log",
+    "--pid-path=/run/nginx.pid",
+    "--with-compat",
+    "--with-pcre",
+    "--user=_www",
+    "--group=_www",
```
nekopsykose, line 16: not the biggest unix group separator person myself but maybe --user should be _nginx? _www is probably intended to be shared with a bunch of stuff

contrib/nginx/template.py
```diff
@@ -0,0 +1,72 @@
...
+source = f"https://nginx.org/download/{pkgname}-{pkgver}.tar.gz"
+sha256 = "77a2541637b92a621e3ee76776c8b7b40cf6d707e69ba53a940283e30ff2f55d"
+# Not available on nginx build tools
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+    self.install_dir("var/log/nginx", empty=True)
+    self.install_dir("var/tmp/nginx", empty=True)
```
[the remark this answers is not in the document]
rbgarga: Do you mean `/usr/lib/tmpfiles.d` ?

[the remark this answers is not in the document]
nekopsykose: yeah, like all the other instances of a tmpfiles config in cports, for these folders

contrib/nginx/template.py
rbgarga: It was discussed with q66 on IRC and he suggested to add _www user first on nginx and then if we add another webserver, then we move its creation to another place

rbgarga:
@nekopsykose wrt nginx version, I picked the one considered stable, 1.24.0, ten build CI failed because there are newer versions available.  Should I move to most recent one (1.25.3) or add something to bypass this check?

contrib/nginx/template.py
nekopsykose: that makes sense :)

nekopsykose:
put `ignore = ["1.2[13579].*"]` or similar in update.py

nekopsykose:
the intent of that being the odd-versions are all excluded, since they're preview releases

rbgarga:
This new version still doesn't have `nginx-test` but I believe all other changes are addressed

rbgarga:
Added `--with-file-aio`

nekopsykose:
nginx test is a separate repo you have to download. see the alpine apkbuild for example, and the hg repo i linked

it's just adding a _tests_hgrev = "sha" and an extra source with source_paths to put it in the right place. but then you'll need a few more deps too

nekopsykose:
also an open question (we can decide later): what to do about extra third-party modules? put them in the same template (like alpine)? build them all separate (if possible) (probably more sane, but needs to remember rebuilds on update)? never package any? i'd go with option 2 probably

rbgarga:
> also an open question (we can decide later): what to do about extra third-party modules? put them in the same template (like alpine)? build them all separate (if possible) (probably more sane, but needs to remember rebuilds on update)? never package any? i'd go with option 2 probably

Good question.  I'm not experient with cports yet so I'll need to look around and think a bit but definitely option 2 makes  sense to me

rbgarga:
> nginx test is a separate repo you have to download. see the alpine apkbuild for example, and the hg repo i linked
> 
> it's just adding a _tests_hgrev = "sha" and an extra source with source_paths to put it in the right place. but then you'll need a few more deps too

I didn't add missing deps yet but so far there are some ssl tests failing, if you have any idea about what could be missing please let me know

https://pastebin.com/XWLSdZxP

rbgarga:
nginx test suite is not going to be that simple to add.  It requires perl-fcgi, which requires perl-fcgi-client and this one requires `Module::Build::Tiny` to build, which is not supported by `perl_module` build style in cports

nekopsykose:
it looks like it just skips tests that the deps are missing for so you're actually already done :)

for the failures, probably caused by openssl 3.2 changes, maybe some other distro has a patch, or a newer test revision shows what fixed them

nekopsykose:
though add things that do exist- even vanilla perl in checkdepends adds some tests i think

rbgarga:
@nekopsykose I've got some tests passing, and SSL tests fixed for version 3.2.0, I believe it's a good start and we can add more perl modules to enable more tests in the future

Now we need to define a strategy for `_www` shared user creation and work with @yanchan09 on it.  IIRC @q66 suggested on IRC that we have a small cport that only creates this user and then both nginx and lighttpd could depend on it.  Any thoughts?

rbgarga:
I removed openssl 3.2.0 patch since it's now committed upstream

q66:
I finally pushed the new tmpfiles/sysusers implementations, so after I had revamped the passwd/group files in base-files (probably still today), we can proceed with this PR.

q66:
I have done the base-files rework. We now have a shared group www-data that is installed as a part of the base files. So the idea is now to have a dedicated user+group for each webserver (i.e. `_nginx` for nginx), and the user is expected to be present in `www-data` as a supplementary group.

Then we can have a default root directory to be shared by web servers, which will be writable by the www-data group.

contrib/nginx/template.py
```diff
@@ -0,0 +1,101 @@
...
+    self.install_service(self.files_path / "nginx")
+
+
+def do_check(self):
+    with self.pushd("nginx-tests"):
+        self.do(
+            "env",
+            "TEST_NGINX_BINARY=../objs/nginx",
+            "prove",
+            ".",
```
nekopsykose, line 100:
```suggestion
        self.do(
            "prove",
            ".",
            env={"TEST_NGINX_BINARY": "../objs/nginx"},
```

contrib/nginx/template.py
```diff
@@ -0,0 +1,101 @@
...
+    "ca-certificates",
+]
+pkgdesc = "Robust and small WWW server"
+maintainer = "Renato Botelho do Couto <renato@netgate.com>"
+license = "BSD-2-Clause"
+url = "https://nginx.org"
+_test_hash = "953461781bc0cd780bb5c4b3a3d727d7842e8d2e"
+source = [
+    f"https://nginx.org/download/{pkgname}-{pkgver}.tar.gz",
+    f"https://github.com/nginx/nginx-tests/archive/{_test_hash}.zip",
```
nekopsykose, line 69:
```suggestion
    f"https://hg.nginx.org/nginx-tests/archive/{_test_hash}.tar.gz",
```
0b5ec15c62ed

contrib/nginx/template.py
```diff
@@ -0,0 +1,101 @@
...
+    "--without-mail_smtp_module",
+]
+make_dir = "."
+hostmakedepends = ["linux-headers"]
+makedepends = ["openssl-devel", "pcre2-devel", "zlib-devel"]
+checkdepends = [
+    "ffmpeg",
+    "libgd-devel",
+    "libgd-progs",
+    "libxslt-devel",
```
nekopsykose, line 57: -devel in makedepends

also sort the list

and, i think ffmpeg specifically can be left out since that doesn't matter very much  

also, libgd/libxslt have modules in nginx (disabled). perhaps maybe want to enable those too?

contrib/nginx/template.py
nekopsykose: also want `f"--jobs={self.make_jobs}"`

rbgarga:
@nekopsykose I believe I addressed all the points from your last review

q66:
i think this on its own looks alright, though i wonder if we should have it come with a default configuration file? also perhaps importing relevant patches from aports would be good: https://gitlab.alpinelinux.org/alpine/aports/-/tree/master/main/nginx

PR #1321, opened 2024-01-25
contrib/pam_wrapper: new package (1.1.5)

mhmdanas:
Packaged this for `fprintd` tests, but couldn't get those working. `libsecret` 0.21.3 will use this package though for testing the PAM module added in the release, so might as well submit a PR.


nekopsykose:
pam_wrapper i think

nekopsykose:
(also fixes the versioncheck)

PR #1325, opened 2024-01-26
contrib/zathura: new package (0.5.4)

ttyyls:


contrib/zathura/template.py
```diff
@@ -0,0 +1,41 @@
...
+    "glib-devel",
+    "gtk+3-devel",
+    "json-glib-devel",
+    "libmagic",
+    "libnotify-devel",
+    "libseccomp-devel",
+    "sqlite-devel",
+]
+checkdepends = ["check-devel"]
+pkgdesc = "Highly customizable and functional document viewer"
```
nekopsykose, line 27:
```suggestion
pkgdesc = "Document viewer"
```

contrib/zathura/template.py
```diff
@@ -0,0 +1,41 @@
...
+    "python-sphinx",
+    "xserver-xorg-xvfb",
+]
+makedepends = [
+    "file-devel",
+    "girara-devel",
+    "glib-devel",
+    "gtk+3-devel",
+    "json-glib-devel",
+    "libmagic",
```
nekopsykose, line 21:
```suggestion
```

contrib/zathura/template.py
```diff
@@ -0,0 +1,41 @@
...
+build_style = "meson"
+hostmakedepends = [
+    "appstream-glib",
+    "intltool",
+    "librsvg-progs",
+    "meson",
+    "ninja",
+    "pkgconf",
+    "python-sphinx",
+    "xserver-xorg-xvfb",
```
nekopsykose, line 13: checkdepends

contrib/zathura/template.py
```diff
@@ -0,0 +1,41 @@
+pkgname = "zathura"
+pkgver = "0.5.4"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "appstream-glib",
+    "intltool",
```
nekopsykose, line 7:
```suggestion
    "gettext",
```

ttyyls:
vis/cfi breaks the modules and the default build for some reason fails to load with sandboxing set to strict ( still works with normal).

nekopsykose:
then write

```
# vis breaks symbols
hardening = ["!vis"]
```

nekopsykose:
> sandboxing set to strict  

probably because musl/one of the libraries uses a syscall not specified in seccomp_enable_strict_filter, which seems likely as seccomp is lol, and it's also fine because strict is not the default

contrib/mujs/patches/read_env_cflags.patch
```diff
@@ -0,0 +1,69 @@
...
+-	install -m 644 mujs.h $(DESTDIR)$(incdir)
+-	install -m 644 $(OUT)/mujs.pc $(DESTDIR)$(libdir)/pkgconfig
+-	install -m 755 $(OUT)/mujs $(DESTDIR)$(bindir)
++	install -d $(DESTDIR)$(INCDIR)
++	install -d $(DESTDIR)$(LIBDIR)
++	install -d $(DESTDIR)$(LIBDIR)/pkgconfig
++	install -d $(DESTDIR)$(BINDIR)
++	install -m 644 mujs.h $(DESTDIR)$(INCDIR)
++	install -m 644 $(OUT)/mujs.pc $(DESTDIR)$(LIBDIR)/pkgconfig
++	install -m 755 $(OUT)/mujs $(DESTDIR)$(BINDIR)
```
nekopsykose, line 44: all of these changes are useless, all you need is to pass prefix=/usr and then nothing has to be changed

contrib/mupdf/template.py
```diff
@@ -0,0 +1,49 @@
...
+pkgver = "1.23.9"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_build_args = [
+    "build=release",
+    "shared=yes",
+    "USE_SYSTEM_LIBS=yes",
+    "USE_SYSTEM_MUJS=yes",
+    "USE_SYSTEM_GLUT=yes",
```
nekopsykose, line 11: the forked version it uses is quite different so i'm not sure if this is good or not

contrib/mupdf/template.py
```diff
@@ -0,0 +1,49 @@
...
+]
+hostmakedepends = [
+    "gmake",
+    "pkgconf",
+]
+makedepends = [
+    "desktop-file-utils",
```
nekopsykose, line 18: hostmakedepends

subnut:
Just a suggestion

It'd be nice if there was a metapackage that installed `zathura` and all of its available backends.  
(eg. `zathura-all` being a metapackage that depends on zathura, zathura-cb, zathura-pdf-*, etc.)

Or we could make `zathura` an empty metapackage, and add `install-if = f"zathura-{pkgver}"` to the rest of the packages (and the main zathura shall have pkgname = zathura-core)


- The former method has the advantage of being similar to how most other distros do it.  
- The latter method has the advantage of having a simpler mental model  
  (You want zathura? You install `zathura`. That's it. Done.)

@q66 @nekopsykose What do you guys think?

nekopsykose:
that doesn't sound very useful, there's like 4 backends total (cb, djvu, ps, and poppler, i don't count mupdf because it's just poppler but worse) and it doesn't take any time to install what you want

nekopsykose:
looks good, though i assume you tested cfi on cb/djvu documents

ttyyls:
> looks good, though i assume you tested cfi on cb/djvu documents

I was about to comment about how things worked with the profile set to normal instead of strict but the truth is I just can't tell if it's a seccomp linking issue or a problem with loading symbols from the module but vis/cfi break in non deterministic ways. So I guess it's best to disable it as baseline.

also removed xvfb-run from the checkdepends since it's supposed to fail

nekopsykose:
> I was about to comment about how things worked with the profile set to normal instead of strict but the truth is I just can't tell if it's a seccomp linking issue or a problem with loading symbols from the module but vis/cfi break in non deterministic ways.  

cfi and seccomp=strict have nothing to do with eachother at all though so i'm not sure what that has to do with it working with cfi ootb

nekopsykose:
ah, ok. sounds fine :)

contrib/girara/template.py
```diff
@@ -0,0 +1,33 @@
...
+]
+makedepends = [
+    "check-devel",
+    "glib-devel",
+    "gtk+3-devel",
+    "json-glib-devel",
+    "libnotify-devel",
+]
+checkdepends = ["xserver-xorg-xvfb"]
+pkgdesc = "GTK+ user interface with a focus on simplicity"
```
q66, line 19: s/user interface/user interface library/

q66:
for the backends, i'd add a `zathura-backends` subpackage into `zathura` like this:

```
@subpackage("zathura-backends")
def _backends(self):
    self.pkgdesc = "{pkgdesc} (backends)"
    self.install_if = [f"{pkgname}={pkgver}-r{pkgrel}"]
    self.depends = [
        "virtual:zathura-whatever!zathura",
        ...
    ]
    self.options = ["empty"]
```

have it depend on all the backends available

the only drawback is that adding a new backend will require bumping the main package, but that's probably not a big deal

PR #1326, opened 2024-01-26
 contrib/networkmanager-openconnect: new package (1.2.10)

z-erica:
patching the gcr 3 dep to 4 requires yanking a widget that was entirely removed, as far as i can tell with no replacement. i assume its fine because its just displaying a cert with no interactivity. some tests are also completely broken, i may investigate later

### DIFF
--- a/contrib/networkmanager-openconnect/files/sysusers.conf
+++ b/contrib/networkmanager-openconnect/files/sysusers.conf
@@ -1,0 +1,3 @@
+# Create nm-openconnect system user
+
+u _nm-openconnect - "nm-openconnect user" /var/lib/NetworkManager /usr/bin/nologin
--- a/contrib/networkmanager-openconnect/template.py
+++ b/contrib/networkmanager-openconnect/template.py
@@ -1,0 +1,39 @@
+pkgname = "networkmanager-openconnect"
+pkgver = "1.2.10"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--with-gtk4=yes", "--disable-static"]
+make_cmd = "gmake"
+make_dir = "."
+hostmakedepends = [
+    "automake",
+    "file",
+    "gcr-progs",
+    "gettext-devel",
+    "glib-devel",
+    "gmake",
+    "intltool",
+    "libtool",
+    "pkgconf",
+]
+makedepends = [
+    "gcr3-devel",
+    "gtk+3-devel",
+    "gtk4-devel",
+    "libnma-devel",
+    "libsecret-devel",
+    "libxml2-devel",
+    "networkmanager-devel",
+    "openconnect-devel",
+    "webkitgtk-devel",
+]
+pkgdesc = "OpenConnect support for NetworkManager"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "GPL-2.0-or-later"
+url = "https://gitlab.gnome.org/GNOME/NetworkManager-openconnect"
+source = f"{url}/-/archive/{pkgver}/NetworkManager-openconnect-{pkgver}.tar.bz2"
+sha256 = "df21a8730438b1614de390ecf1f73d379536d388a8e464c9a802dab14dd23c8f"
+
+
+def post_install(self):
+    self.install_sysusers(self.files_path / "sysusers.conf")
--- a/contrib/networkmanager-openconnect/update.py
+++ b/contrib/networkmanager-openconnect/update.py
@@ -1,0 +1,1 @@
+pattern = r"\"/GNOME/NetworkManager-openconnect/-/tags/([0-9.]+)\""
--- a/contrib/vpnc-scripts/template.py
+++ b/contrib/vpnc-scripts/template.py
@@ -1,0 +1,16 @@
+pkgname = "vpnc-scripts"
+_commit = "4ed41c21e3857f96ab935b45092bbb07c3ccd5be"
+pkgver = "0_git20240308"
+pkgrel = 0
+pkgdesc = "OpenConnect network routing script"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "GPL-2.0-or-later"
+url = "https://www.infradead.org/openconnect/vpnc-script.html"
+source = f"https://gitlab.com/openconnect/vpnc-scripts/-/archive/{_commit}/vpnc-scripts-{_commit}.tar.bz2"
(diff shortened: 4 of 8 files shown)


contrib/gi-docgen/template.py
```diff
@@ -0,0 +1,28 @@
+pkgname = "gi-docgen"
+pkgver = "2023.3"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+    "python-packaging",
```
nekopsykose, line 8: this is depends=

nekopsykose: also put all these python things in _deps = and then make/depends = list(_deps) since it's the same list

nekopsykose: also hostmakedepends i think instead of makdepends

contrib/networkmanager-openconnect/template.py
```diff
@@ -0,0 +1,44 @@
...
+    "pkgconf",
+    "file",
+    "intltool",
+    "gettext-devel",
+    "gcr-progs",
+    "gmake",
+]
+makedepends = [
+    "libxml2-devel",
+    "libsecret-devel",
```
nekopsykose, line 21: sort 'em

contrib/networkmanager-openconnect/template.py
```diff
@@ -0,0 +1,44 @@
...
+    "networkmanager-devel",
+    "gtk4-devel",
+]
+pkgdesc = "OpenConnect support for NetworkManager"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "GPL-2.0-or-later"
+url = "https://gitlab.gnome.org/GNOME/NetworkManager-openconnect"
+source = f"{url}/-/archive/{pkgver}/NetworkManager-openconnect-{pkgver}.tar.bz2"
+sha256 = "df21a8730438b1614de390ecf1f73d379536d388a8e464c9a802dab14dd23c8f"
+options = ["!lintstatic"]
```
nekopsykose, line 36: this is wrong, it probably makes a /usr/lib/networkmanagert plugin with .a because of defaultlibs, put --default-library=shared in configure_args so it doesn't

contrib/networkmanager-openconnect/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://gitlab.gnome.org/GNOME/NetworkManager-openconnect/-/tags"
+pattern = r"href=\"/GNOME/NetworkManager-openconnect/-/tags/([0-9.]+)\""
```
nekopsykose, line 2:
```suggestion
pattern = r"NetworkManager-openconnect-([0-9.]+).tar.gz"
```

you can do something similar in the other mr too, never need the whole href thing

contrib/python-smartypants/template.py
```diff
@@ -0,0 +1,25 @@
...
+checkdepends = ["python-pytest", "python-docutils"]
+pkgdesc = (
+    "Translate ASCII punctuation characters into typographic HTML entities"
+)
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "BSD-3-Clause"
+url = "https://github.com/leohemsted/smartypants.py"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "b98191911ff3b4144ef8ad53e776a2d0ad24bd508a905c6ce523597c40022773"
+# tests fail
```
nekopsykose, line 20: what with

contrib/vpnc-scripts/template.py
```diff
@@ -0,0 +1,16 @@
...
+license = "GPL-2.0-or-later"
+url = "https://www.infradead.org/openconnect/vpnc-script.html"
+source = f"https://gitlab.com/openconnect/vpnc-scripts/-/archive/{_commit}/vpnc-scripts-{_commit}.tar.bz2"
+sha256 = "128f14953966dfcfedf2c4e236ca17ed3cb4e9aa79a3205af97e25d253f087e8"
+
+
+def do_install(self):
+    self.install_file("vpnc-script", "etc/vpnc", 0o755)
+    self.install_file("vpnc-script-ptrtd", "etc/vpnc", 0o755)
+    self.install_file("vpnc-script-sshd", "etc/vpnc", 0o755)
```
nekopsykose, line 16: /etc seems wrong. if not /usr/bin, there is /usr/libexec , or perhaps some other /usr path (but /etc seems bad unless it's user-editable)

nekopsykose:
it's usually always a toggle to build api docs, but they're always kinda meh to build

contrib/networkmanager-openconnect/template.py
nekopsykose: o rite yea, misread as meson initially

q66:
this should now be unblocked, so feel free to clean up and switch to gcr3-devel

contrib/networkmanager-openconnect/files/sysusers.conf
```diff
@@ -0,0 +1,3 @@
+# Create nm-openconnect system user
+
+u nm-openconnect - "nm-openconnect user" /var/lib/NetworkManager /usr/bin/nologin
```
q66, line 3: is there a way to pick the user for this? we generally aim for having underscore-prefixed system users/groups if we can help it

q66: looks like we'd need to patch `NM_OPENCONNECT_USER` in the code, and then the polkit rulesets

q66:
other than the user thing lgtm, so let me know what your thoughts are on that (if it makes sense at all to patch) and then we can get this in

contrib/networkmanager-openconnect/template.py
```diff
@@ -0,0 +1,39 @@
...
+pkgdesc = "OpenConnect support for NetworkManager"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "GPL-2.0-or-later"
+url = "https://gitlab.gnome.org/GNOME/NetworkManager-openconnect"
+source = f"{url}/-/archive/{pkgver}/NetworkManager-openconnect-{pkgver}.tar.bz2"
+sha256 = "df21a8730438b1614de390ecf1f73d379536d388a8e464c9a802dab14dd23c8f"
+
+
+def post_install(self):
+    self.install_sysuser(self.files_path / "sysusers.conf")
```
q66, line 39: install_sysusers

PR #1327, opened 2024-01-26
user/racket: new package (8.11.1)

z-erica:


Froggo8311:
each time i try building it seems the checksum is different resulting in a failed build

q66:
this is `user/` and `archs = ["aarch64", "x86_64"]`

PR #1329, opened 2024-01-27
main/sushi: add missing gtksourceview4 runtime dep

mhmdanas:
Fixes broken spacebar preview in Nautilus.


nekopsykose:
webkitgtk should also be there (it's pulled indirectly in most setups but it's also gjs-only and not detected)

PR #1331, opened 2024-01-27
contrib/furnace: new package (0.6)

z-erica:


contrib/furnace/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://github.com/tildearrow/furnace/releases"
+pattern = r"href=\"/tildearrow/furnace/tree/v([0-9.]+)\""
```
nekopsykose, line 2: don't need the url and just /v .tar.gz like in the other etc

nekopsykose: ah, very good point

PR #1333, opened 2024-01-27
contrib/acpica: new package (20230628)

mikoxyz:


contrib/acpica/update.py
```diff
@@ -0,0 +1,2 @@
+# the tag versioning scheme over at https://github.com/acpica/acpica is bogus
+ignore = True
```
nekopsykose, line 2: it actually isn't, you can map the RMM_DD_YY values back to YYYYMMDD

..whether you should is another story

mikoxyz: that solution would break in 2100 though

nekopsykose: that's a problem for future generations

subnut: IMO defining `fetch_versions` to remap RMM_DD_YY to YYYYMMDD is better than outright ignoring, because that way `update-check` still works.

PR #1335, opened 2024-01-27
contrib/neovide: new package (0.12.1)

subnut:
TODO
- [X] find source for gn with stable tarball SHA
- [ ] cross?
- [ ] hardening
- [ ] move to user/

### DIFF
--- a/contrib/neovide/patches/fix-skia-bindings.patch
+++ b/contrib/neovide/patches/fix-skia-bindings.patch
@@ -1,0 +1,22 @@
+--- a/vendor/skia-bindings/build_support/skia_bindgen.rs
++++ b/vendor/skia-bindings/build_support/skia_bindgen.rs
+@@ -805,15 +805,18 @@ pub(crate) mod definitions {
+     }
+ 
+     fn ninja_files_for_features(features: &features::Features) -> Vec<PathBuf> {
++        use crate::skia::env;
+         let mut files = vec!["obj/skia.ninja".into()];
+         if features.text_layout {
+             files.extend(vec![
+                 "obj/modules/skshaper/skshaper.ninja".into(),
+                 "obj/modules/skparagraph/skparagraph.ninja".into(),
+                 // shaper.cpp includes SkLoadICU.h
+-                "obj/third_party/icu/icu.ninja".into(),
+                 "obj/modules/skunicode/skunicode.ninja".into(),
+             ]);
++            if !env::use_system_libraries() {
++                files.push("obj/third_party/icu/icu.ninja".into());
++            }
+         }
+         if features.svg {
+             files.push("obj/modules/svg/svg.ninja".into());
--- a/contrib/neovide/template.py
+++ b/contrib/neovide/template.py
@@ -1,0 +1,117 @@
+pkgname = "neovide"
+pkgver = "0.12.1"
+pkgrel = 0
+archs = ["aarch64", "ppc64le", "riscv64", "x86_64"]
+build_style = "cargo"
+hostmakedepends = [
+    "cargo",
+    "ninja",
+    "libcxx-devel-static",
+]
+makedepends = [
+    "fontconfig-devel",
+    "freetype-devel",
+    "libexpat-devel",
+    "libjpeg-turbo-devel",
+    "libpng-devel",
+    "zlib-devel",
+]
+checkdepends = ["neovim"]
+pkgdesc = "Graphical client for neovim"
+maintainer = "Subhaditya Nath <sn03.general@gmail.com>"
+license = "MIT"
+url = "https://neovide.dev"
+_gn = "2139%5E20240120gitb5adfe5f574d-1.fc40"
+_skia = "2196c5b01b984e30bf28038ab861d0a350d4caf4"
+_wuffs = "e3f919ccfe3ef542cfc983a82146070258fb57f8"
+_spcross = "6e1fb9b09efadee38748e0fd0e6210d329087e89"
+_vmalloc = "a6bfc237255a6bac1513f7c1ebde6d8aed6b5191"
+_gn0, _gn1 = _gn.split("-")
+source = [
+    f"https://github.com/neovide/neovide/archive/refs/tags/{pkgver}.tar.gz>{pkgname}-{pkgver}.tar.gz",
+    f"https://kojipkgs.fedoraproject.org/packages/gn/{_gn0}/{_gn1}/src/gn-{_gn}.src.rpm>gn-{_gn}.tar.gz",
+    f"https://github.com/rust-skia/skia/archive/{_skia}.tar.gz",
+    f"https://github.com/google/wuffs-mirror-release-c/archive/{_wuffs}.tar.gz",
+    f"https://github.com/KhronosGroup/SPIRV-Cross/archive/{_spcross}.tar.gz",
+    f"https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/archive/{_vmalloc}.tar.gz",
+]
+source_paths = [
+    ".",
+    "_gn",
+    "skia",
+    "skia/third_party/externals/wuffs",
+    "skia/third_party/externals/spirv-cross",
+    "skia/third_party/externals/vulkanmemoryallocator",
+]
+sha256 = [
+    "95df61d51635a981697a6c4b3e63a9feb29bb90bfc69940c02c73273a61cfd10",
+    "fff8d73970df98b786270de0ff5e35adc7904037beade6b8f2952ac552c55a4e",
+    "e8d08fed6bfc7e61216233b00585a212eace629816b7f1ddb0c0d90eb6df32d4",
+    "e849dab1f372f16b782ba7528e7f70281e35425bd2d644c92a36fb95909f98db",
+    "e8f175054803997c98df5270fde711b7d6d9d5434194718df5a2b8527f654e7c",
+    "7444a78beeb01ee3baca58721b2b7c936521f05df1f6e322bbafe3c2d4114704",
+]
+env = {
+    "SKIA_SOURCE_DIR": f"/builddir/{pkgname}-{pkgver}/skia",
+    "SKIA_GN_COMMAND": f"/builddir/{pkgname}-{pkgver}/gn/out/gn",
+    "SKIA_NINJA_COMMAND": "ninja",
+    "SKIA_USE_SYSTEM_LIBRARIES": "1",
+}
+
+
+def post_extract(self):
+    from cbuild.core import chroot
+
+    # see cbuild/hooks/do_extract/000_sources.py
+    with self.pushd("_gn"):
+        self.mkdir("gn/out", True)
+        self.mv("last_commit_position.h", "gn/out")
+        ret = chroot.enter(
+            "tar",
+            "-x",
+            "--no-same-owner",
+            "--no-same-permissions",
+            "-f",
+            self.chroot_cwd / f"gn-{_gn[18:30]}.tar.gz",
+            "-C",
+            self.chroot_cwd / "gn",
+            ro_root=True,
+            unshare_all=True,
+        ).returncode
+        if ret != 0:
+            self.error(f"Could not extract gn-{_gn[18:30]}.tar.gz (exit {ret})")
+    self.mv("_gn/gn", "gn")
+    self.rm("_gn", True)
+
+
+def post_patch(self):
+    from cbuild.util import cargo
+
+    cargo.clear_vendor_checksums(self, "skia-bindings")
+
+
+def do_configure(self):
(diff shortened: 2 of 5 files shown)


contrib/neovide/template.py
```diff
@@ -0,0 +1,92 @@
+pkgname = "neovide"
+pkgver = "0.12.1"
+pkgrel = 0
+build_wrksrc = "neovide"
+build_style = "cargo"
+hostmakedepends = [
+    "cargo",
+    "ninja",
+    "libcxx-devel-static",
```
q66, line 9: ??

contrib/neovide/template.py
```diff
@@ -0,0 +1,92 @@
...
+makedepends = [
+    "fontconfig-devel",
+    "freetype-devel",
+    "libexpat-devel",
+    "libjpeg-turbo-devel",
+    "libpng-devel",
+    "zlib-devel",
+]
+checkdepends = ["neovim"]
+pkgdesc = "No Nonsense Neovim Client in Rust"
```
q66, line 21: this violates pkgdesc conventions in at least 3 ways (and nothing about this seems no-nonsense)

contrib/neovide/template.py
```diff
@@ -0,0 +1,92 @@
...
+maintainer = "Subhaditya Nath <sn03.general@gmail.com>"
+license = "MIT"
+url = "https://neovide.dev"
+_skia = (
```
q66, line 25: use individual variables

contrib/neovide/template.py
```diff
@@ -0,0 +1,92 @@
...
+source = [
+    f"https://github.com/neovide/neovide/archive/refs/tags/{pkgver}.tar.gz>{pkgname}-{pkgver}.tar.gz",
+    f"https://gn.googlesource.com/gn/+archive/{_gn}.tar.gz>gn-{_gn[:8]}.tar.gz",
+    f"https://github.com/rust-skia/skia/archive/{_skia[0]}.tar.gz",
+    f"https://github.com/google/wuffs-mirror-release-c/archive/{_skia[1]}.tar.gz",
+    f"https://github.com/KhronosGroup/SPIRV-Cross/archive/{_skia[2]}.tar.gz",
+    f"https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/archive/{_skia[3]}.tar.gz",
+]
+source_paths = [
+    "neovide",
```
q66, line 41: this can probably be . and then the build_wrksrc can be dropped

contrib/neovide/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://github.com/neovide/neovide/tags"
```
q66, line 1: this should probably work automatically

contrib/neovide/template.py
```diff
@@ -0,0 +1,92 @@
...
+    from cbuild.util import cargo
+
+    with self.pushd(build_wrksrc):
+        self.cargo.vendor()
+        cargo.setup_vendor(self)
+
+
+def do_configure(self):
+    with self.pushd("../gn"):
+        self.do("python", "build/gen.py")
```
q66, line 74: use wrksrc

contrib/neovide/template.py
nekopsykose: we could also make gn a separate template. they don't actually have breaking changes very often at all (i did that for a while for chromium and it built even with outdated stuff fine, up until rust was added), and all that would be needed is a source mirror (like github)

speaking of the gn.googlesource here is completely unusable because everything from googlesource is from gitiles and there are no stable tarballs (new sha every fetch)

contrib/neovide/template.py
```diff
@@ -0,0 +1,92 @@
...
+    "SKIA_GN_COMMAND": f"/builddir/{pkgname}-{pkgver}/gn/out/gn",
+    "SKIA_NINJA_COMMAND": "ninja",
+    "SKIA_USE_SYSTEM_LIBRARIES": "1",
+}
+
+
+def do_prepare(self):
+    from cbuild.util import cargo
+
+    with self.pushd(build_wrksrc):
```
q66, line 67: this shouldn't be needed with proper directory layout

contrib/neovide/template.py
```diff
@@ -0,0 +1,92 @@
+pkgname = "neovide"
+pkgver = "0.12.1"
+pkgrel = 0
+build_wrksrc = "neovide"
```
q66, line 4: this uses rustls so it should be in user/ repository and have restricted archs set

nekopsykose: rin builds on ppc64le/riscv64 since 0.17 however (this is 0.17.7). they merged the generic fallback

nekopsykose: (i.e. this is not restricted because of rustls at all. maybe other reasons..)

contrib/neovide/template.py
```diff
@@ -0,0 +1,92 @@
...
+def pre_build(self):
+    with self.pushd("../gn"):
+        self.do("ninja", "-C", "out")
+
+
+def do_install(self):
+    self.install_license("LICENSE")
+    self.install_bin("target/*-chimera-linux-*/release/neovide", glob=True)
+    self.install_file("assets/neovide.desktop", "usr/share/applications")
+    for f in self.find("assets", "neovide-*.png"):
```
nekopsykose, line 86: there is an svg too

contrib/neovide/template.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def pre_build(self):
+    with self.pushd("../gn"):
+        self.do("ninja", "-C", "out")
+
+
+def do_install(self):
+    self.install_license("LICENSE")
+    self.install_bin("target/*-chimera-linux-*/release/neovide", glob=True)
```
nekopsykose, line 84: self.profile().triplet instead of glob

contrib/neovide/template.py
subnut: Ah that's why the CI failed with mismatching SHA.

Didn't know about that.

contrib/neovide/update.py
```diff
@@ -0,0 +1,2 @@
...
+pattern = r'/archive/refs/tags/([\d.]+)(?=\.tar\.gz")'
```
nekopsykose, line 2:
```suggestion
pattern = r'/archive/refs/tags/([\d.]+).tar.gz'
```

you don't need fancy captures

contrib/neovide/template.py
q66: yeah but it's still statically linked tls so i don't want it in primary repos regardless

q66: archs should still be restricted because skia does not support big endian

contrib/neovide/template.py
subnut: oh wait this was written by past me

will fix

---

**TL;DR** yesterday i thought of writing a template for this package, and today I discovered that a template already exists! So I got to work trying to fix it's build without even looking at the rest of the stuff.

That's why I was initially building 0.11.2 instead of the latest 0.12.1 (if you saw that on IRC)

contrib/neovide/update.py
subnut: nah the rest of the URLs clobber it for some reason.

will try again tho.

contrib/neovide/template.py
nekopsykose: yeah, written as `if self.profile().endian == "big": broken=`

contrib/neovide/template.py
subnut: where should that go?

contrib/neovide/template.py
q66: no, written as archs= because skia quite literally has arch-specific code in it

contrib/neovide/template.py
subnut: oh icons/hicolor/scalable/apps/ okay

contrib/neovide/update.py
subnut: sorry... idk anything about the python-specific regex stuff  
(skill issue basically)

will fix.

contrib/neovide/patches/fix-skia-bindings.patch
```diff
@@ -0,0 +1,28 @@
...
+-                "obj/third_party/icu/icu.ninja".into(),
+                 "obj/modules/skunicode/skunicode.ninja".into(),
+             ]);
++            if !env::use_system_libraries() {
++                files.push("obj/third_party/icu/icu.ninja".into());
++            }
+         }
+         if features.svg {
+             files.push("obj/modules/svg/svg.ninja".into());
+--- a/neovide/vendor/skia-bindings/.cargo-checksum.json
```
q66, line 23: unneeded/unwanted, use cargo.clear_vendor_checksums

q66: also i'm about 95% sure you'll need to further patch skia (see ppc-musttail.patch in firefox/thunderbird, similar in libreoffice and chromium)

contrib/neovide/update.py
subnut: Nah it doesn't seem to be working

without `update.py` -
```
$ cbuild update-check contrib/neovide verbose
Checking for updates: neovide=0.12.1
Adding 'https://neovide.dev' for version check...
Adding 'https://github.com/neovide/neovide/archive/refs/tags/' for version check...
Adding 'https://gn.googlesource.com/gn/+archive/' for version check...
Adding 'https://github.com/rust-skia/skia/archive/' for version check...
Adding 'https://github.com/google/wuffs-mirror-release-c/archive/' for version check...
Adding 'https://github.com/KhronosGroup/SPIRV-Cross/archive/' for version check...
Adding 'https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/archive/' for version check...
Fetching 'https://neovide.dev' for version checks...
Fetching 'https://github.com/neovide/neovide/tags' for version checks...
Fetching 'https://gn.googlesource.com/gn/+archive/' for version checks...
Fetching 'https://github.com/rust-skia/skia/tags' for version checks...
Fetching 'https://github.com/google/wuffs-mirror-release-c/tags' for version checks...
Fetching 'https://github.com/KhronosGroup/SPIRV-Cross/tags' for version checks...
Fetching 'https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/tags' for version checks...
Checking found version: 0.3.0
Checking found version: 0.3.1
Checking found version: 0.3.2
Checking found version: 0.3.3
Checking found version: 0.10.0
Checking found version: 0.10.1
Checking found version: 0.10.2
Checking found version: 0.10.3
Checking found version: 0.10.4
Checking found version: 0.11.0
Checking found version: 0.11.1
Checking found version: 0.11.2
Checking found version: 0.12.0
Checking found version: 0.12.1
Checking found version: 1.0.0
Checking found version: 1.0.1
Checking found version: 2.0.0
Checking found version: 2.1.0
Checking found version: 2.2.0
Checking found version: 2.3.0
Checking found version: 3.0.0
Checking found version: 3.0.1
contrib/neovide: 0.12.1 -> 1.0.0
contrib/neovide: 0.12.1 -> 1.0.1
contrib/neovide: 0.12.1 -> 2.0.0
contrib/neovide: 0.12.1 -> 2.1.0
contrib/neovide: 0.12.1 -> 2.2.0
contrib/neovide: 0.12.1 -> 2.3.0
contrib/neovide: 0.12.1 -> 3.0.0
contrib/neovide: 0.12.1 -> 3.0.1
```

contrib/neovide/template.py
subnut: "Graphical client for neovim" sounds ok?

contrib/neovide/template.py
subnut: @q66 what should be the value of `archs =` ?

(I am a complete noob when it comes to architectures)

contrib/neovide/update.py
subnut: wait shouldn't it be `\.tar\.gz` ?

nekopsykose: there's no realistic case where the . matters for being dot-only or any

subnut:
Some changes made, some yet to be made.

subnut:
> yeah but it's still statically linked tls so i don't want it in primary repos regardless

@q66 if i replaced rustls with tls-native (i.e. openssl), would this package be accepted into `contrib` ?

contrib/neovide/template.py
subnut: switched to koji.fedoraproject.org for gn

contrib/neovide/patches/fix-skia-bindings.patch
subnut: added ppc-musttail.patch from firefox

PR #1338, opened 2024-01-28
contrib/rrdtool: new package (1.8.0)

yanchan09:


contrib/rrdtool/template.py
```diff
@@ -0,0 +1,30 @@
...
+    "bash",
+    "gettext-devel",
+    "groff",
+    "libtool",
+    "pkgconf",
+]
+makedepends = ["glib-devel", "libxml2-devel", "pango-devel", "python-devel"]
+pkgdesc = "Round Robin Database"
+maintainer = "yanchan09 <yan@omg.lol>"
+license = "custom:GPL-2.0-or-later-with-FLOSS-Exception"
```
yanchan09, line 17: bit unsure about this, maybe it'd be fine to just specify `GPL-2.0-or-later` or something along the lines of `GPL-2.0-or-later AND custom:FLOSS-Exception`

yanchan09: for reference, the license text is at https://github.com/oetiker/rrdtool-1.x/blob/master/COPYRIGHT

contrib/rrdtool/patches/0007-fix-libxml-cfi.patch
```diff
@@ -0,0 +1,369 @@
+diff --git a/src/rrd_restore.c b/src/rrd_restore.c
+index 14b9446b..9f94f2df 100644
+--- a/src/rrd_restore.c
++++ b/src/rrd_restore.c
+@@ -48,6 +48,11 @@ static int opt_force_overwrite = 0;
+  * Helpers
+  */
+ 
++__attribute__((no_sanitize("cfi-icall")))
```
yanchan09, line 9: I was told that @q66 probably wouldn't approve this :) But right now the alternatives are either patching libxml2 (to export memory management functions as real functions instead of `extern` function pointers) or disabling CFI altogether.

q66: this is technically fine, but i'd probably just disable CFI altogether here, it uses glib (as well as libxml2) and CFI with glib things is largely a futile effort (you can be about 80% sure any glib callback is function pointer UB) and overly fragile

contrib/rrdtool/template.py
```diff
@@ -0,0 +1,30 @@
...
+hostmakedepends = [
+    "automake",
+    "bash",
+    "gettext-devel",
+    "groff",
+    "libtool",
+    "pkgconf",
+]
+makedepends = ["glib-devel", "libxml2-devel", "pango-devel", "python-devel"]
+pkgdesc = "Round Robin Database"
```
subnut, line 15: Shouldn't this be "Round Robin Database Tool" ?

contrib/rrdtool/template.py
nekopsykose: yes, use that instead, like in console-setup

or WITH like in sane-backends which might be invalid but who cares

contrib/rrdtool/patches/0002-fix-visibility.patch
```diff
@@ -0,0 +1,364 @@
+diff --git a/src/rrd.h b/src/rrd.h
```
nekopsykose, line 1: there's no real point to fix vis with a 400 line patch when you can just not add vis. it doesn't really have any benefit whatsoever here

nekopsykose: if you want to send it upstream tho then go for it

contrib/rrdtool/template.py
```diff
@@ -0,0 +1,38 @@
...
+def post_install(self):
+    self.install_license("COPYRIGHT")
+
+
+@subpackage(f"{pkgname}-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage(f"{pkgname}-python")
```
nekopsykose, line 33: shouldn't this be python-rrdtool since it's technically named that on pypi

yanchan09: seemingly pretty much everything that doesn't get pulled from pypi uses -python as suffix

nekopsykose: maybe we should start changing that i dunno

PR #1339, opened 2024-01-28
contrib/lighttpd: new package (1.4.73)

yanchan09:


yanchan09:
Right now a _www user is added with sysusers, which would conflict with contrib/nginx (PR #1318). Probably a better solution is needed... See also the discussion at https://github.com/chimera-linux/cports/pull/1318/files#r1466716450

yanchan09:
Also I need to see if I can make lighttpd log to syslog, if not, maybe it would be good to add `log-type = buffer` to the service configuration.

gstrauss:
The current stable lighttpd release is lighttpd 1.4.76, released 11 Apr 2024.

PR #1340, opened 2024-01-28
contrib/redis: new package (7.2.4)

yanchan09:


contrib/redis/patches/0005-fix-nodes-conf-ub.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/src/cluster.c b/src/cluster.c
```
nekopsykose, line 1: you should also send this upstream if you want

mikoxyz: the cfi fixes might be work sending upstream, too

PR #1344, opened 2024-01-29
contrib/bemenu: new package (0.6.17)

flukeeey:


flukeeey:
> Shouldn't the license be `GPL-3.0-only AND LGPL-3.0-only`?

Fixed. Thanks!

contrib/bemenu/template.py
```diff
@@ -0,0 +1,55 @@
...
+    self.pkgdesc = f"{pkgdesc} (curses backend)"
+    self.depends = [f"{pkgname}={pkgver}-r{pkgrel}"]
+
+    return ["usr/lib/bemenu/bemenu-renderer-curses.so"]
+
+
+@subpackage(f"{pkgname}-wayland")
+def _wayland(self):
+    self.pkgdesc = f"{pkgdesc} (wayland backend)"
+    self.depends = [f"{pkgname}={pkgver}-r{pkgrel}"]
```
nekopsykose, line 45: these should probably default install by adding an install_if on the package itself (see python-pyclip) so they're automatically there by default (all three)

flukeeey: I tried this just now, but installing `bemenu-wayland` will now auto install the X11 and curses backends. What if I want just the wayland backend?

nekopsykose: `apk add bemenu "!bemenu-x11" "!bemenu-curses"`

nekopsykose: the reason for this is that bemenu-curses and x11 together make up like 60kb, and the extra dependencies they have (ncurses and libx11) are already present on every system anyway.. there is no point to not actually install them. so, this makes them all exist by default, but still leaves the !exclusion if someone really wants to remove them for some reason. the other alternative is just not splitting it

flukeeey: Yup, I checked other packages and it seems this is the best way to do it. I have seen other distros using a post-install message, e.g. 'You should also install a bemenu backend, ex: bemenu-{wayland,x11,curses}', but this seems like extra work for the user for the sake of saving 60kb. Probably not worth it, like you say.

I wasn't sure if _x11 sould check against `libx11` or `xserver-xorg-core`. The latter seems to be a more explicit signal that the user is actually using X11, and doesn't just have X11 libs installed as part of Xwayland, for example. But other templates seem to use libx11.

nekopsykose: > I wasn't sure if _x11 sould check against libx11 or xserver-xorg-core.  

neither, just `install_if = ["pkgname=.."]`

flukeeey: Are you sure? The `ibus` package from @q66 adds extra conditions as I mentioned.

q66: there is very little reason to add the extra conditions, as they are never really reliable (for instance wayland will get automatically pulled in through a ton of other things even if one does not have a compositor), should be dropped from ibus too

q66: i'd make this whole thing a loop, like so:

```
def _subpkg(sname, sdesc):
    @subpackage(f"bemenu-{sname}")
    def _spkg(self):
        self.pkgdesc = f"{pkgdesc} ({sdesc} backend)"
        self.depends = [f"{pkgname}={pkgver}-r{pkgrel}"]
        self.install_if = [f"{pkgname}={pkgver}-r{pkgrel}"]

        return ["usr/lib/bmenu/bemenu-renderer-{sname}.so"]

for _sname, _sdesc in [
    ("curses", "curses"),
    ("wayland", "Wayland"),
    ("x11", "X11"),
]:
    _subpkg(_sname, _sdesc)
```

contrib/bemenu/template.py
```diff
@@ -0,0 +1,58 @@
+pkgname = "bemenu"
+pkgver = "0.6.17"
```
q66, line 2: update version

contrib/bemenu/template.py
```diff
@@ -0,0 +1,58 @@
...
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+hostmakedepends = ["scdoc", "gmake", "pkgconf"]
```
q66, line 6: hostmakedepends should have wayland-progs

PR #1348, opened 2024-01-30
contrib/bind: new package (9.18.21)

z-erica:


mikoxyz:
> unsure what to do about the bind update check, since the "latest version" its complaining about isnt the stable release: https://www.isc.org/bind/

maybe add `ignore = ["9.19*"]` to update.py for the time being?

nekopsykose:
we had removed lmdb ages ago in 38139762f084e20b6a27fc6811c69c10c639037c . it's also optional afaik

nekopsykose:
all the odd (9.17, 9.19, ..) versions are dev too so you can preemptively ignore them with like `9.[0-9][13579].*`

contrib/bind/template.py
```diff
@@ -0,0 +1,40 @@
...
+    "heimdal-devel",
+    "libcap-devel",
+    "libuv-devel",
+    "libxml2-devel",
+    "lmdb-devel",
+    "json-c-devel",
+    "nghttp2-devel",
+    "openssl-devel",
+]
+checkdepends = ["python-pytest", "cmocka-devel", "python-dnspython"]
```
nekopsykose, line 26: cmocka-devel in makedepends

contrib/bind/template.py
```diff
@@ -0,0 +1,40 @@
...
+checkdepends = ["python-pytest", "python-dnspython"]
+pkgdesc = "ISC DNS server"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "MPL-2.0"
+url = "https://www.isc.org/bind"
+source = f"https://downloads.isc.org/isc/bind9/{pkgver}/bind-{pkgver}.tar.xz"
+sha256 = "a556be22505d9ea4f9c6717aee9c549739c68498aff3ca69035787ecc648fec5"
+
+
+def post_install(self):
```
nekopsykose, line 35: it would be nice to make a -progs that separates the debugging cli utilities (and depend on it in the parent)

see https://pkgs.alpinelinux.org/contents?file=&path=&name=bind-tools&branch=edge&arch=x86_64 for a list of these

(mostly for dig but also the others just because)

nekopsykose (approved):

nekopsykose:
why is the github ui so bad

contrib/bind/template.py
nekopsykose: and not the sbin ones from there

nekopsykose: that seems fine

PR #1350, opened 2024-01-30
contrib/python-dnspython: new package (2.5.0)

mikoxyz:


mikoxyz:
ah, that's funny, zerica opened a pr containing dnspython like 20 minutes before i opened this one (#1348). not sure if i should close this one or not

PR #1352, opened 2024-01-30
contrib/cronie: new package (1.7.1)

flukeeey:
I couldn't find a cron implementation in cports. I saw there was a lightweight alternative via `/etc/dinit.d/snooze-*` but I wanted anacron-like support for my laptop, which isn't always online.

I had a couple of questions in template.py if someone more experience with cports could check. Thanks. :)

### DIFF
--- a/contrib/cronie/template.py
+++ b/contrib/cronie/template.py
@@ -1,0 +1,41 @@
+pkgname = "cronie"
+pkgver = "1.7.1"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--enable-anacron",
+    "--with-inotify",
+    "--with-pam",
+    "--without-selinux",
+]
+make_cmd = "gmake"
+hostmakedepends = ["gmake", "autoconf", "automake"]
+makedepends = ["linux-pam-devel", "musl-obstack-devel"]
+depends = ["bash", "debianutils"]
+pkgdesc = "Cronie cron daemon project"
+maintainer = "flukey <flukey@vapourmail.eu>"
+license = "ISC AND GPL-2.0-only"
+url = "https://github.com/cronie-crond/cronie"
+source = f"{url}/releases/download/{pkgname}-{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "78033100c24413f0c40f93e6138774d6a4f55bc31050567b90db45a2f9f1b954"
+tool_flags = {"LDFLAGS": ["-lobstack"]}
+suid_files = ["usr/bin/crontab"]
+file_modes = {"usr/bin/crontab": ("root", "root", 0o6755)}
+hardening = ["vis", "cfi"]
+
+
+def post_install(self):
+    self.install_license("COPYING")
+
+    # NOTE: Sets '-s': log to syslog instead of mailer by default
+    self.install_service(self.files_path / "crond")
+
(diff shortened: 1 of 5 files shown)


contrib/cronie/template.py
```diff
@@ -0,0 +1,43 @@
...
+    "--without-selinux",
+]
+make_cmd = "gmake"
+hostmakedepends = ["gmake", "autoconf", "automake"]
+makedepends = ["linux-pam-devel"]
+depends = ["bash", "debianutils"]
+pkgdesc = "Cronie cron daemon project"
+maintainer = "flukey <flukey@vapourmail.eu>"
+# TODO: obstack in cports causes LTO build errors, bundled obstack can be used
+# instead, but its license is LGPL-2.1, which cbuild apparently doesnt like?
```
nekopsykose, line 18: because the actual licence is either lgpl-2.1-only or lgpl-2.1-or-later

what is the error though?

nekopsykose:
```
ld: error: undefined symbol: _obstack_begin
>>> referenced by readtab.c:400 (./build/../anacron/readtab.c:400)
```

this is because it's never linked. you need to add -lobstack somewhere

flukeeey: The error was: `ERROR: failed validating license: unknown token: LGPL-2.1`

I found 'LGPL-2.1' on the SPDX license list here: https://spdx.org/licenses/archive/archived_ll_v2.4/LGPL-2.1.html

But I've now added `-lobstack` to LDFLAGS, so this is no longer needed. Thanks!

nekopsykose: > The error was: ERROR: failed validating license: unknown token: LGPL-2.1

yes, as i said it's an incomplete token

> I found 'LGPL-2.1' on the SPDX license list here:

the url has `archived` in it :) see https://spdx.org/licenses/LGPL-2.1.html for what it "used to mean"

contrib/cronie/template.py
```diff
@@ -0,0 +1,41 @@
...
+configure_args = [
+    "--enable-anacron",
+    "--with-inotify",
+    "--with-pam",
+    "--without-selinux",
+]
+make_cmd = "gmake"
+hostmakedepends = ["gmake", "autoconf", "automake"]
+makedepends = ["linux-pam-devel", "musl-obstack-devel"]
+depends = ["bash", "debianutils"]
```
nekopsykose, line 14: afaict the bash here is just because the default 0hourly sets SHELL=bash, but could make it sh instead so it doesn't pull in bash just for the default config (and people can still set it if they want it)

contrib/cronie/template.py
```diff
@@ -0,0 +1,41 @@
...
+    # NOTE: Sets '-s': log to syslog instead of mailer by default
+    self.install_service(self.files_path / "crond")
+
+    self.install_file(self.files_path / "crontab", "etc")
+    self.install_file(self.files_path / "cron-deny", "etc", name="cron.deny")
+    self.install_file(
+        self.files_path / "default-anacron", "etc/default", name="anacron"
+    )
+
+    self.install_file("contrib/0anacron", "etc/cron.hourly")
```
ptrcnull, line 39: this should be installed as executable, otherwise `run-parts` doesn't pick it up

q66:
0201e2cbb5c9b597fa3136e9db6a7192f4eb7713

PR #1360, opened 2024-01-31
main/linux-stable: update to 6.7.5

mikoxyz:
[ci skip]


q66:
you didn't regenerate the configs

mikoxyz:
> you didn't regenerate the configs

shii, sorry!

mikoxyz:
also took the liberty to change some stuff over to modules in the x86_64 config. on another note, the different arch configs should probably get synced up at some point; some options differ here and there for no apparent reason

q66:
most of these module changes make no sense

mikoxyz:
> most of these module changes make no sense

mm modules,,

mikoxyz:
on a more serious note, why shouldn't all those be modules?

q66:
because all it does is shrink the kernel image by a minuscule amount in exchange for adding a larger amount in modules (and possibly regressing some functionality or at least further reducing the chance of being able to boot without a ramdisk which already isn't possible but eventually could be)

mikoxyz:
if being able to boot without an initrd is something you want I could look into getting that working instead

nekopsykose:
you could do that another time and not in an update mr which we could probably merge rn

q66:
it's a non-trivial thing with many caveats so it cannot really be done a a part of routine bump

q66:
fwiw some of these could be modules but some also not (e.g. nvme core should definitely stay inside kernel, but the touchscreen driver stuff is an oversight and should probably be a module)

and any changes need to be applied in all architectures

PR #1363, opened 2024-02-02
contrib/jfsutils: new package (1.1.15)

mikoxyz:


yanchan09:
why

mikoxyz:
> why

if we're gonna have it in the kernel we might as well have the userspace utils too

PR #1364, opened 2024-02-02
contrib/{evolution,ytnef}: new packages

mhmdanas:


contrib/evolution/template.py
```diff
@@ -0,0 +1,62 @@
...
+    "nss-devel",
+    "shared-mime-info",
+    "sqlite-devel",
+    "webkitgtk-devel",
+    "xorgproto",
+    "ytnef-devel",
+]
+pkgdesc = "Mail, calendar and address book suite"
+maintainer = "triallax <triallax@tutanota.com>"
+license = "LGPL-2.1-or-later"
```
mhmdanas, line 52: The licensing hurt my brain, need to check this later

mhmdanas: Okay I'm fairly sure it's `LGPL-2.1-or-later`

contrib/evolution/template.py
```diff
@@ -0,0 +1,60 @@
+pkgname = "evolution"
+pkgver = "3.50.4"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DWITH_OPENLDAP=OFF",
+    "-DWITH_NSPR_INCLUDES=/usr/include/nspr",
+    "-DWITH_NSS_INCLUDES=/usr/include/nss",
```
nekopsykose, line 8: these are not needed and it finds them anyway

PR #1369, opened 2024-02-02
main/lynx: enable CFI

mhmdanas:
`cfi-genptr` on its own without `cfi` is the same as no `cfi` at all, which seems to have been missed in bc086bc37c1fe196a2ec7869ee2d5264862805b7.

cc @nekopsykose


nekopsykose:
good catch, yes

PR #1370, opened 2024-02-02
main/mpv: enable CFI (without indirect function call checks)

mhmdanas:
Seems to work fine.


mikoxyz:
hmmm, can't you figure out the individual functions that are causing the problems and exempt them individually instead (like i did in #1304)?

mhmdanas:
Hmm, I could do that yeah, I'll take a look later.

PR #1371, opened 2024-02-03
contrib/libbsd: new package (0.12.1)

ttyyls:


contrib/libmd/template.py
```diff
@@ -0,0 +1,26 @@
...
+sha256 = "1bd6aa42275313af3141c7cf2e5b964e8b1fd488025caf2f971f43b00776b332"
+
+
+def post_install(self):
+    self.install_license("COPYING")
+
+
+@subpackage("libmd-devel-static")
+def _static(self):
+    return self.default_static()
```
nekopsykose, line 21:
```suggestion
```

this doesn't do anything

contrib/libmd/template.py
```diff
@@ -0,0 +1,26 @@
+pkgname = "libmd"
+pkgver = "1.1.0"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_gen = []
```
nekopsykose, line 5:
```suggestion
# why
configure_gen = []
```

contrib/libbsd/template.py
```diff
@@ -0,0 +1,35 @@
+pkgname = "libbsd"
```
nekopsykose, line 1: what is this actually for?

contrib/libbsd/patches/no_nlist.patch
```diff
@@ -0,0 +1,11 @@
+--- configure.ac	2024-01-08 02:16:08.000000000 +0100
++++ configure.ac	2024-02-03 00:40:40.012618732 +0100
+@@ -259,7 +259,7 @@
+ need_bsd_getopt=yes
+ need_progname=yes
+ need_md5=yes
+-need_nlist=yes
++need_nlist=no
```
nekopsykose, line 8: i don't see anyone else needing this patch so what is it for?

contrib/libbsd/template.py
ttyyls: @nekopsykose working on porting https://gameoftrees.org/ but it's still wip

contrib/libbsd/patches/no_nlist.patch
ttyyls: we don't have <nlist.h> and according to q66 there are no plans to add it to musl-bsd-headers

contrib/libbsd/patches/no_nlist.patch
```diff
@@ -0,0 +1,11 @@
+--- configure.ac	2024-01-08 02:16:08.000000000 +0100
++++ configure.ac	2024-02-03 00:40:40.012618732 +0100
```
nekopsykose, line 2: write patches as -p1

contrib/libbsd/patches/no_nlist.patch
ttyyls: there is probably a cleaner way to do it from the autoreconf cli interface

nekopsykose: try ac_cv_need_nlist=no

nekopsykose: from what i see though that doesn't matter? in alpine for instance nlist.h only comes from libbsd itself too, and nothing has to be passed

nekopsykose: yeah, this isn't needed at all for anything and builds fine

the test fails (

```
FAIL: nlist
===========

Assertion failed: rc == 0 (../../test/nlist.c: main: 72)
FAIL nlist (exit status: 134)
```
)

but that is because of lto, it works with `!lto`. this patch has nothing to do with it

nekopsykose: also something segfaults silently in install

```
--- install-exec-hook ---
if [ "/usr/lib" != "/usr/lib" ]; then  /builddir/libbsd-0.11.8/build-aux/install-sh -c -d /destdir/libbsd-0.11.8/usr/lib;  mv /destdir/libbsd-0.11.8/usr/lib/libbsd*.so.*  /destdir/libbsd-0.11.8/usr/lib/;  fi
soname=`readlink /destdir/libbsd-0.11.8/usr/lib/libbsd.so`;   /destdir/libbsd-0.11.8/usr/lib/libbsd.so;  (echo '/* GNU ld script';  echo ' * The MD5 functions are provided by the libmd library. */';  cat format.ld;  echo "GROUP(/usr/lib/$soname AS_NEEDED(-lmd))";  )>/destdir/libbsd-0.11.8/usr/lib/libbsd.so
Segmentation fault (core dumped)
```

nekopsykose: looking real quick the segfault is in libbsd itself somewhere, you'll have to figure out what that is

contrib/libmd/template.py
ttyyls: to get things like --prefix=/usr and so on, by default gnu_configure runs `autoreconf -if` which in this repo doesn't make sense.

nekopsykose: i'm aware, that's what the comment is for

nekopsykose: though now it seems like this is actually autotools, is this a new source, i'm confused haha

contrib/libmd/template.py
```diff
@@ -0,0 +1,23 @@
...
+def post_install(self):
+    self.install_license("COPYING")
+
+
+@subpackage("libmd-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+configure_gen = []
```
nekopsykose, line 23: this doesn't go at the bottom of the file, though it should also be possible to reconf unless it's broken (add a comment for that)

contrib/libbsd/template.py
```diff
@@ -0,0 +1,41 @@
...
+exec_wrappers = [("/usr/bin/gsed", "sed")]
+
+
+def post_install(self):
+    self.install_license("COPYING")
+
+
+@subpackage("libbsd-devel")
+def _devel(self):
+    self.depends = [f"{pkgname}={pkgver}-r{pkgrel}"]
```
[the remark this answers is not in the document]
ttyyls: in this case it was necessary since it wasn't a library but just an ld script

PR #1374, opened 2024-02-03
contrib/aubio: new package (0.4.9)

z-erica:
this was a dependency to qtractor i couldnt package before. i think ive figured it out now


contrib/aubio/template.py
```diff
@@ -0,0 +1,47 @@
...
+]
+sha256 = [
+    "d48282ae4dab83b3dc94c16cf011bcb63835c1c02b515490e1883049c3d1f3da",
+    "21199cd220ccf60434133e1fd2ab8c8e5217c3799199c82722543970dc8e38d5",
+]
+# tests run in do_build
+options = ["!check"]
+
+# don't use deprecated ffmpeg functions
+tool_flags = {"CFLAGS": ["-DFF_API_LAVF_AVCTX"]}
```
nekopsykose, line 30: why not? it shouldn't matter for the build itself

nekopsykose: ah i see

apply https://github.com/aubio/aubio/commit/8a05420e5dd8c7b8b2447f82dc919765876511b3 instead so it doesn't get lost in the future to be removed :)

PR #1377, opened 2024-02-04
contrib/jsoncpp: new package (1.9.5)

anixedn:


contrib/jsoncpp/template.py
```diff
@@ -0,0 +1,22 @@
...
+build_style = "meson"
+hostmakedepends = ["meson", "pkgconf"]
+pkgdesc = "C++ library for interacting with JSON"
+maintainer = "avgwst <avgwst@tutanota.de>"
+license = "MIT"
+url = "https://github.com/open-source-parsers/jsoncpp"
+source = f"https://github.com/open-source-parsers/{pkgname}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "f409856e5920c18d0c2fb85276e24ee607d2a09b5e7d5f0a371368903c275da2"
+# no checks
+options = ["!check"]
```
nekopsykose, line 13: i'm really confused how you check this considering it literally builds+runs the tests by default just by removing this line

anixedn: in which packages should i disable tests? or i shouldn't?

nekopsykose: you should always run the tests. !check is not the default because tests are there to be run. it doesn't make any sense to skip them by default.

!check is when the tests cannot be built/run for some specific reason. and if they build but only a few fail then you disable specific tests instead

contrib/jsoncpp/template.py
```diff
@@ -0,0 +1,21 @@
...
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = ["meson", "pkgconf"]
+pkgdesc = "C++ library for interacting with JSON"
+maintainer = "avgwst <avgwst@tutanota.de>"
+license = "MIT"
+url = "https://github.com/open-source-parsers/jsoncpp"
+source = f"https://github.com/open-source-parsers/{pkgname}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "f409856e5920c18d0c2fb85276e24ee607d2a09b5e7d5f0a371368903c275da2"
+hardening = ["!vis"]
```
nekopsykose, line 12:
```suggestion
# vis breaks symbols
hardening = ["!vis"]
```

PR #1376, opened 2024-02-04
contrib/spdlog: new package (1.13.0)

anixedn:
checkdepends on https://github.com/chimera-linux/cports/pull/1379

PR #1379, opened 2024-02-04
contrib/catch2: new package (3.5.2)

anixedn:


nekopsykose:
technically this leaves an empty parent package (has just some usr/share/Catch2 fluff) and then the resulting library is in -devel-static. i wonder which of

- BUILD_SHARED_LIBS=ON actually works (when used in the other mr), in which case could use that, or
- just not splitting it at all as it's a test-only library?

makes more sense

anixedn:
hm, i have a couple ideas then: 1) just to add a suffix -devel to package name (but i think it's more stupid idea)
2) split package not to -devel, but only -static subpackage (because it won't allow me to put static libs in the main package)

nekopsykose:
> just to add a suffix -devel to package name  

no, that isn't needed in general. just the original name is fine by itself for that kinda thing (like the existing header-only libraries)


>  but only -static subpackage

this doesn't really make sense because to use the thing at all you need those static libs (assuming shared doesn't work, but here it doesn't matter which since test only). so every usage becomes `catch2-static` and the parent package is pointless by itself. at that point pointless to split at all i guess. so personally i'd say just don't split it :)  

i'll wait for q66 to decide via random coin toss to the opposite of what i think though

nekopsykose:
turns out it works with build_shared_libs so it's added with https://github.com/chimera-linux/cports/pull/1470 instead. you can rebase https://github.com/chimera-linux/cports/pull/1376

PR #1381, opened 2024-02-04
contrib/furnace: update to 0.6.1

z-erica:


nekopsykose:
ah i missed it but this now needs system_freetype=on

PR #1384, opened 2024-02-06
user/wl-kbptr: new package (0.2.0)

ttyyls:


user/wl-kbptr/template.py
```diff
@@ -0,0 +1,22 @@
...
+hostmakedepends = ["meson", "pkgconf"]
+makedepends = [
+    "cairo-devel",
+    "libxkbcommon-devel",
+    "wayland-devel",
+    "wayland-protocols",
+]
+pkgdesc = "Control the mouse pointer with the keyboard"
+maintainer = "ttyyls <contact@behri.org>"
+license = "GPL-3.0-or-later"
```
nekopsykose, line 14: this is not actually specified whether it's -only or -or-later anywhere

ttyyls: afaik the default GPL license is in use with copyright assignment to the fsf and that should be synonymous with the -or-later option

ttyyls: even if it's not the case, the compatibility matrix says it's ok to relicense GPL-v3 to GPL-v3-or-later https://www.gnu.org/licenses/gpl-faq.html#compat-matrix-footnote-7

PR #1385, opened 2024-02-06
contrib/libdbusmenu: new package (16.04.0)

anixedn:


contrib/libdbusmenu-glib/patches/layout_fix_style.patch
```diff
@@ -0,0 +1,87 @@
+diff --git a/tests/test-glib-layout.h b/test-glib-layout.h
```
q66, line 1: are these patches necessary for build? (i.e. does the build fail without them in a way that cannot be fixed without introducing these patches?)

contrib/libdbusmenu-glib/template.py
```diff
@@ -0,0 +1,75 @@
...
+build_style = "gnu_configure"
+configure_args = [
+    "--disable-dumper",
+    "--disable-gtk-doc",
+    "--enable-tests",
+    "--enable-static",
+    "--enable-gtk",
+    "--enable-introspection",
+    "--enable-vala",
+    "HAVE_VALGRIND_FALSE=''",
```
q66, line 13: this doesn't look necessary?

contrib/libdbusmenu-glib/template.py
```diff
@@ -0,0 +1,75 @@
...
+def _gtk3(self):
+    self.pkgdesc = f"{pkgdesc} (GTK+3)"
+    self.depends = ["libdbusmenu-glib"]
+    return [
+        "usr/lib/libdbusmenu-gtk3.so*",
+        "usr/lib/girepository-1.0/DbusmenuGtk3-0.4.typelib",
+    ]
+
+
+@subpackage("libdbusmenu-gtk3-devel")
```
q66, line 61: a single devel package should be enough

contrib/libdbusmenu-glib/template.py
```diff
@@ -0,0 +1,75 @@
...
+    "json-glib-devel",
+    "vala-devel",
+]
+pkgdesc = "Library for passing menus over DBus"
+maintainer = "avgwst <avgwst@tutanota.de>"
+license = "GPL-3.0-or-later AND LGPL-3.0-or-later AND LGPL-2.1-or-later"
+url = "https://launchpad.net/libdbusmenu"
+source = f"{url}/16.04/{pkgver}/+download/libdbusmenu-{pkgver}.tar.gz"
+sha256 = "b9cc4a2acd74509435892823607d966d424bd9ad5d0b00938f27240a1bfa878a"
+# vis breaks symbols
```
q66, line 40: drop this, vis is only useful for cfi and we're never enabling that for gtk/glib things

contrib/libdbusmenu-glib/template.py
```diff
@@ -0,0 +1,75 @@
...
+hardening = ["!vis"]
+# 12/13 checks FAIL
```
q66, line 42: have you tried looking into why?

contrib/libdbusmenu-glib/template.py
```diff
@@ -0,0 +1,75 @@
...
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_license("COPYING-GPL3")
```
q66, line 47: GPL and related licenses should not be installed

contrib/libdbusmenu-glib/template.py
q66: (i assume you need at very least `make_check_wrapper = ["dbus-run-session"]` + `dbus` in `checkdepends` to have a chance)

contrib/libdbusmenu-glib/template.py
```diff
@@ -0,0 +1,75 @@
+pkgname = "libdbusmenu-glib"
```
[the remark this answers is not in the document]
anixedn: i also don't see the reason for naming with postfix, especially if we don't separate it into gtk subpackage

contrib/libdbusmenu-glib/patches/layout_fix_style.patch
anixedn: yes, it fails during the checks without them ("error: use of GNU old-style field designator extension [-Werror,-Wgnu-designator]")

contrib/libdbusmenu-glib/template.py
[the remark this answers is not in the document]
q66: the gtk subpackage should stay but for the postfix i see no good reason

contrib/libdbusmenu-glib/patches/layout_fix_style.patch
q66: that is because of Werror, which should be patched out

contrib/libdbusmenu-glib/template.py
```diff
@@ -0,0 +1,75 @@
...
+
+def post_install(self):
+    self.install_license("COPYING-GPL3")
+    self.install_license("COPYING.2.1")
+
+
+@subpackage("libdbusmenu-gtk3")
+def _gtk3(self):
+    self.pkgdesc = f"{pkgdesc} (GTK+3)"
+    self.depends = ["libdbusmenu-glib"]
```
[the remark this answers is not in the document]
anixedn: should i also remove self.pkgdesc?

contrib/libdbusmenu-glib/template.py
anixedn: removed last three lines, but instead of HAVE_VALGRIND_FALSE='' i made a patch, so with --disable-tests it wouldn't fail

contrib/libdbusmenu/template.py
```diff
@@ -0,0 +1,56 @@
...
+    "json-glib-devel",
+    "vala-devel",
+]
+pkgdesc = "Library for passing menus over DBus"
+maintainer = "avgwst <avgwst@tutanota.de>"
+license = "GPL-3.0-or-later AND LGPL-3.0-or-later AND LGPL-2.1-or-later"
+url = "https://launchpad.net/libdbusmenu"
+source = f"{url}/16.04/{pkgver}/+download/libdbusmenu-{pkgver}.tar.gz"
+sha256 = "b9cc4a2acd74509435892823607d966d424bd9ad5d0b00938f27240a1bfa878a"
+# Glib tests fails
```
q66, line 37: the failures are because you're missing `checkdepends = ["dbus", "bash"]`

after that it still fails, because of unpackaged dbus-test-runner (https://launchpad.net/dbus-test-runner)

contrib/libdbusmenu/template.py
```diff
@@ -0,0 +1,55 @@
...
+@subpackage("libdbusmenu-gtk3")
+def _gtk3(self):
+    self.pkgdesc = f"{pkgdesc} (GTK+3)"
+    return [
+        "usr/include/libdbusmenu-gtk3-0.4/libdbusmenu-gtk/*",
+        "usr/lib/girepository-1.0/DbusmenuGtk3-0.4.typelib",
+        "usr/lib/libdbusmenu-gtk3.so",
+        "usr/lib/pkgconfig/dbusmenu-gtk3-0.4.pc",
+        "usr/share/gir-1.0/DbusmenuGtk3-0.4.gir",
+        "usr/share/vala/vapi/DbusmenuGtk3-0.4.vapi",
```
nekopsykose, line 49:
```suggestion
        "usr/lib/girepository-1.0/DbusmenuGtk3-0.4.typelib",
        "usr/lib/libdbusmenu-gtk3.so.4*",
```

contrib/libdbusmenu/template.py
```diff
@@ -0,0 +1,55 @@
...
+    "pkgconf",
+]
+makedepends = [
+    "atkmm-devel",
+    "dbus-glib-devel",
+    "gettext-devel",
+    "gtk-doc-tools",
+    "gtk+3-devel",
+    "glib-devel",
+    "gobject-introspection",
```
nekopsykose, line 27: sort

contrib/libdbusmenu/template.py
```diff
@@ -0,0 +1,55 @@
...
+    "json-glib-devel",
+    "vala-devel",
+]
+checkdepends = ["dbus-test-runner", "bash", "xserver-xorg-xvfb"]
+pkgdesc = "Library for passing menus over DBus"
+maintainer = "avgwst <avgwst@tutanota.de>"
+license = "GPL-3.0-or-later AND LGPL-3.0-or-later AND LGPL-2.1-or-later"
+url = "https://launchpad.net/libdbusmenu"
+source = f"{url}/16.04/{pkgver}/+download/libdbusmenu-{pkgver}.tar.gz"
```
nekopsykose, line 36:
```suggestion
source = f"{url}/{pkgver[:pkgver.rfind('.')]}/{pkgver}/+download/libdbusmenu-{pkgver}.tar.gz"
```

contrib/libdbusmenu/template.py
anixedn: Yep, the problem was because of missing bash

PR #1389, opened 2024-02-07
main/musl: add ahosts/ahostsv4/ahostsv6/shadow to getent

nekopsykose:
imported from alpine aports getent.c  

---

i rolled the manpage update myself, so i'm not sure if i named the fields right (and the sp_ stuff is an eyesore and overflows the line length.. not sure how to improve that) :)


q66:
roll together with musl bump

PR #1391, opened 2024-02-07
contrib/wireshark: new package (4.2.2)

nekopsykose:


contrib/wireshark/wireshark.pre-install
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+echo "g _wireshark -" | \
+    /usr/bin/systemd-sysusers --replace /usr/lib/sysusers.d/wireshark.conf - || :
```
nekopsykose, line 4: this sux but just like in qemu i assume there's no other way to have a group on an owned file like this

contrib/wireshark/template.py
```diff
@@ -0,0 +1,82 @@
...
+pkgdesc = "Network protocol analyzer"
+maintainer = "psykose <alice@ayaya.dev>"
+license = "GPL-2.0-or-later"
+url = "https://www.wireshark.org"
+source = f"https://www.wireshark.org/download/src/all-versions/wireshark-{pkgver}.tar.xz"
+sha256 = "9e3672be8c6caf9279a5a13582d6711ab699ae2a79323e92a99409c1ead98521"
+# forbid non-wireshark-group users from reading all network packets
+file_modes = {
+    "usr/bin/dumpcap": ("root", "_wireshark", 0o750),
+}
```
nekopsykose, line 64: this actually doesn't work and the file is installed as 755. maybe it breaks because of file_xattrs at the same time?

nekopsykose: hm no, even with no xattrs it doesn't work

but it does work for dbus so i'm confused

q66: this is the executable perms sanitization code, it skips it for suid but does not account for xattrs (i'll fix that later)

q66: should be addressed properly by https://github.com/chimera-linux/cports/commit/37ef5a5f5aeac6592138a7eef05d7043b6611cae

contrib/wireshark/wireshark.pre-install
q66, line 4: sd-sysusers

nekopsykose: doh, yes

contrib/wireshark/files/wireshark.conf
```diff
@@ -0,0 +1,3 @@
+# Create wireshark system group
+
+g _wireshark -
```
q66, line 3: what's the - for? groups take two fields

contrib/wireshark/wireshark.pre-install
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+echo "g _wireshark -" | \
```
q66, line 3: ditto here

contrib/wireshark/files/wireshark.conf
q66: eh nevermind sorry, my brain did a moment

PR #1392, opened 2024-02-07
contrib/papirus-icon-theme: new package (20240201)

anixedn:


contrib/papirus-icon-theme/template.py
```diff
@@ -0,0 +1,25 @@
...
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "8ff3caded7862e5e6f531dbae54b213ff1cd3666d26f23357c6183173856f380"
+
+
+def do_build(self):
+    pass
+
+
+def do_check(self):
+    self.make.invoke("test-all")
```
nekopsykose, line 21: make_check_target = "test-all"

contrib/papirus-icon-theme/template.py
```diff
@@ -0,0 +1,25 @@
...
+pkgdesc = "Pixel perfect icon theme"
+maintainer = "avgwst <avgwst@tutanota.de>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/PapirusDevelopmentTeam/papirus-icon-theme"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "8ff3caded7862e5e6f531dbae54b213ff1cd3666d26f23357c6183173856f380"
+
+
+def do_build(self):
+    pass
```
nekopsykose, line 17: is this even needed? the default build target is just a no-op already

contrib/papirus-icon-theme/template.py
nekopsykose: although, what is the point? this is an icon theme, i don't see any point in testing it at all. could not patch anything too, there is nothing valuable in these tests since we're not building the icons or anything to begin with?

contrib/papirus-icon-theme/template.py
anixedn: i set it just because of the warning "gmake: Nothing to be done for 'all'."

Issue #283, opened 2023-08-06
cbuild/util/cargo: allow an easy way to wrap cargo native links


nekopsykose:
`.cargo/config.toml`:

```
[target.{pkg.profile().triplet}]
crate = { rustc-link-lib = ["libname"] }
```

could have a map of crate: libname

mostly useful to force libgit2 link on git2 using projects, since it then forcefully uses the system lib, and a few others

PR #1393, opened 2024-02-07
contrib/delta: new package (0.16.5)

ttyyls:


contrib/delta/template.py
```diff
@@ -0,0 +1,24 @@
...
+maintainer = "ttyyls <contact@behri.org>"
+license = "MIT"
+url = "https://github.com/dandavison/delta"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "00d4740e9da4f543f34a2a0503615f8190d307d1180dfb753b6911aa6940197f"
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+    self.install_completion("etc/completion/completion.bash", shell="bash")
```
q66, line 22: don't use named args for shell

contrib/delta/template.py
```diff
@@ -0,0 +1,24 @@
+pkgname = "delta"
+pkgver = "0.16.5"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo", "pkgconf"]
+makedepends = [
+    "libgit2-devel",
```
nekopsykose, line 7: technically this is not linked because the `git2`->`libgit2-sys` crate will refuse to use the system lib unless it's against the exact same ver (the bindings are `libgit2-sys v0.14.2+1.5.1` and we have 1.7.2). so you'd have to `cargo update -p git2` to get this to work

it's also possible to make it link anyway with .cargo/config overrides (https://github.com/chimera-linux/cports/issues/283) since the symbols are compatible and passing the link arg will make the linker add a needed and remove all the unused vendored libgit2 code that is compiled in, but we don't have that yet. also not sure if it's actually safe for 2 releases out of date in versions (it's definitely safe to force for e.g. +1.7.1 libgit2-sys while having 1.7.2, but not sure here, just fluff)

ttyyls: I updated git2 to 1.18.2 which pulls ligit-sys-0.16.2 corresponding to the systems's 1.7.2. Going forward I'll just update the patch or delete it tracking libgit2 updates or upstream releasing in sync with libgit2

nekopsykose: yea, sounds good

PR #1395, opened 2024-02-07
Prototype zig package manager integration

ifreund:
Upstream zig recently added support for a "system package mode" intended to integrate the package management with distro packaging. This PR uses these new features to implement a proof of concept for packaging/distributing zig packages using chimera's cbuild toolchain.

This MR successfully packages waylock's 1.2.0 release and its dependencies. Native compilation on x86_64 and cross compilation to aarch64 both work.

I've opened this MR to encourage collaboration between chimera's maintainers and upstream zig maintainers with the goal of finding how we can work together to ensure our respective end-users are best served. I personally have a foot in both camps being both a zig core team member and chimera user/contributor.

Here are the main issues I see with the current state of the zig package manager from a distro packaging point of view:

1. Since zig identifies packages by their hash, there is no way for a distro to bump a zig package from e.g version 1.0.0 to 1.0.1 without patching the build.zig.zon of all consumers of that package.
2. There is no easy way to obtain the hash of a zig package using the cli without copying into the global cache. It should be trivial to add an option to `zig fetch` for this purpose which would allow full automation of the `install` step of the `zig-wayland` and `zig-xkbcommon` templates for example. Possibly this would be made irrelevant to a solution to 1. however.
3. ~~There is no easy way to delete the files in a zig package that are not referenced by the build.zig.zon and included in the package hash. There is no reason for distros to include these files in the packages they redistribute. A more generally useful solution might be to add a CLI command to list files included in the package, which the distro packaging tooling could use as a whitelist for what to include without needing to semantically parse the build.zig.zon.~~ Edit: `zig fetch` does this, I just misunderstood how to use it.
4. While `--host-target`, `--host-cpu`, and `--host-dynamic-linker` were added, what toolchains like cbuild which support cross-compilation really need are `--target`, `--cpu`, and `--dynamic-linker` which describe the target system, not the host. This PR currently uses the `-Dtarget` and `-Dcpu` flags defined by waylock's build.zig instead but this is technically not portable.

### DIFF
--- a/src/cbuild/util/zig_build.py
+++ b/src/cbuild/util/zig_build.py
@@ -1,0 +1,54 @@
+def build(pkg, extra_args=[]):
+    profile = pkg.profile()
+    sysroot = profile.sysroot
+    with open(pkg.cwd / "cbuild_zig_libc.txt", mode="w") as f:
+        f.write(
+            f"""include_dir={sysroot}/usr/include
+sys_include_dir={sysroot}/usr/include
+crt_dir={sysroot}/lib
+msvc_lib_dir=
+kernel32_lib_dir=
+gcc_dir=
+
+"""
+        )
+
+    zig_arch = None
+    zig_cpu = None
+    match profile.arch:
+        # TODO other architectures
+        case "x86_64" | "aarch64":
+            zig_arch = profile.arch
+            zig_cpu = "baseline"
+
+    # The Zig build system only has a single install step, there is no
+    # way to build artifacts for a given prefix and then install those artifacts
+    # to that prefix at some later time. Therefore, we build and install to the zig-out
+    # directory and later copy the artifacts to the destdir in do_install().
+    # We use zig-out to avoid path conflicts as it is the default install
+    # prefix used by the zig build system.
+    pkg.do(
+        "zig",
+        "build",
+        "-j" + str(pkg.make_jobs),
+        "--sysroot",
+        sysroot,
+        "--libc",
+        "cbuild_zig_libc.txt",
+        "--search-prefix",
+        "/usr",
+        "--prefix",
+        "/usr",
+        "--system",
+        sysroot / "usr/src/zig/packages",
+        "--release=safe",
+        f"-Dtarget={zig_arch}-linux-musl",
+        f"-Dcpu={zig_cpu}",
+        *extra_args,
+        env={"DESTDIR": "zig-out"},
+    )
+
+
+def install(pkg):
+    for x in (pkg.cwd / "zig-out").iterdir():
+        pkg.install_files(x, ".")
--- a/user/zig-wayland/template.py
+++ b/user/zig-wayland/template.py
@@ -1,0 +1,21 @@
+pkgname = "zig-wayland"
+pkgver = "0.2.0"
+pkgrel = 0
+pkgdesc = "Zig bindings for libwayland"
+maintainer = "Isaac Freund <mail@isaacfreund.com>"
+license = "MIT"
+url = "https://codeberg.org/ifreund/zig-wayland"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "831ce41cb0aad8da97de5e27125cbdd80454e5da8fd52aa78e918c0e0c784d70"
+
+
+def do_install(self):
+    self.install_files(
+        ".",
+        "usr/src/zig/packages/",
+        name="122062beeb6fd2bb21c91e81acb3ea6cbba69d3c00a31b62732254e190b5fc7a934e",
+    )
+
+
+def post_install(self):
+    self.install_license("LICENSE")
--- a/user/zig-wayland/template.py
+++ b/user/zig-wayland/template.py
@@ -1,20 +1,14 @@
 pkgname = "zig-wayland"
 pkgver = "0.2.0"
 pkgrel = 0
+build_style = "zig_package"
+hostmakedepends = ["zig"]
 pkgdesc = "Zig bindings for libwayland"
 maintainer = "Isaac Freund <mail@isaacfreund.com>"
 license = "MIT"
 url = "https://codeberg.org/ifreund/zig-wayland"
 source = f"{url}/archive/v{pkgver}.tar.gz"
 sha256 = "831ce41cb0aad8da97de5e27125cbdd80454e5da8fd52aa78e918c0e0c784d70"
-
-
-def do_install(self):
-    self.install_files(
-        ".",
-        "usr/src/zig/packages/",
-        name="122062beeb6fd2bb21c91e81acb3ea6cbba69d3c00a31b62732254e190b5fc7a934e",
-    )
 
 
 def post_install(self):
--- a/src/cbuild/build_style/zig_build.py
+++ b/src/cbuild/build_style/zig_build.py
@@ -1,0 +1,14 @@
+from cbuild.util import zig_build
+
+
+def do_build(self):
+    zig_build.build(self, self.zig_build_args)
+
+
+def do_install(self):
+    zig_build.install(self)
+
+
+def use(tmpl):
+    tmpl.do_build = do_build
+    tmpl.do_install = do_install
--- a/src/cbuild/build_style/zig_package.py
+++ b/src/cbuild/build_style/zig_package.py
@@ -1,0 +1,5 @@
+from cbuild.util import zig_package
+
+
+def use(tmpl):
+    tmpl.do_install = zig_package.install
--- a/src/cbuild/core/template.py
+++ b/src/cbuild/core/template.py
@@ -487,6 +487,8 @@ core_fields = [
     ("go_mod_dl", None, str, False, False, False),
     ("go_build_tags", [], list, False, False, False),
     ("go_check_tags", [], list, False, False, False),
+    # zig
+    ("zig_build_args", [], list, False, False, False),
 ]
 
 # a field priority list, the second element indicates whether
@@ -521,6 +523,7 @@ core_fields_priority = [
     ("make_check_env", True),
     ("make_check_wrapper", True),
     ("make_use_env", True),
+    ("zig_build_args", True),
     ("cmake_dir", False),
     ("meson_dir", False),
     ("hostmakedepends", True),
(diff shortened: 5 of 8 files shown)


src/cbuild/util/zig_build.py
```diff
@@ -0,0 +1,54 @@
...
+        sysroot,
+        "--libc",
+        "cbuild_zig_libc.txt",
+        "--search-prefix",
+        "/usr",
+        "--prefix",
+        "/usr",
+        "--system",
+        sysroot / "usr/src/zig/packages",
+        "--release=safe",
```
nekopsykose, line 44: you probably know this already but i assume this is meant to be a variable like zig_release_target= or similar (or at least i imagine hypothetically something might want to pick another option than release-safe)

user/zig-wayland/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "zig-wayland"
+pkgver = "0.1.0"
```
nekopsykose, line 2: unrelated and you also probably know this already- for git hashes these are some `$lasttag _gitYYYYMMDD` form instead. but i see this is just a proof of concept so it doesn't matter for now :)

user/zig/template.py
```diff
@@ -24,14 +24,13 @@
...
-source = f"https://ziglang.org/download/{pkgver}/zig-{pkgver}.tar.xz"
-sha256 = "72014e700e50c0d3528cef3adf80b76b26ab27730133e8202716a187a799e951"
+source = f"https://ziglang.org/builds/zig-{pkgver}-dev.2631+3069669bc.tar.xz"
+sha256 = "395688a4a30e0afad2c6e2d0a4bded5d61c58d88d2e89362716732cb1c91b332"
 # lighten up the build, only applies to bootstrap
 hardening = ["!int", "!scp", "!var-init"]
 # lto only gets applied to the C bootstrap and slows down the build (doesn't
 # affect the zig output)
 options = ["!lto"]
-restricted = "work in progress"
```
nekopsykose, line 34: half of the WIP for zig itself (the other half being the build style you've now implemented) is that currently we don't really have plans to have multiple llvm versions at once. the current unwritten requirement for things linking libLLVM (zig also uses libclang and liblld but that's just fluff) is that they are generally ready by the time .3/.4 is out or so (when we upgrade to new majors). so far, zig isn't up to date with the latest llvm in a shipped release near these points (though i do see that on master it happens very fast). not sure how this is best resolved though. maybe eventually we'll just have to have multiple versions..

user/zig-wayland/template.py
nekopsykose: also for ci you'll want `ignore = True` in update.py for these

andrewrk:
Thanks for putting this together!

> Since zig identifies packages by their hash, there is no way for a distro to bump a zig package from e.g version 1.0.0 to 1.0.1 without patching the build.zig.zon of all consumers of that package.

Two points on this one:

1. The distro can choose to lie and provide whatever version they want for any particular hash, swapping one version for another. The zig build system by design only verifies hashes on fetch, and if the system provides a particular hash, it will be trusted and used.
2. An upcoming feature that will not land in 0.12.0 will address the general use case of multiple package versions existing in a dependency tree, and the ability to swap one package version for another. Probably at the same time this will come with CLI flags or other functionality that makes it easy for a distro to participate in this process.


> There is no easy way to obtain the hash of a zig package using the cli without copying into the global cache. It should be trivial to add an option to `zig fetch` for this purpose which would allow full automation of the `install` step of the `zig-wayland` and `zig-xkbcommon` templates for example. Possibly this would be made irrelevant to a solution to 1. however.

Would you mind writing up a zig issue for this one? I started working on it, but I found myself not able to explain the problem very well. In particular, what would you want to happen to the directory of files that was fetched in order to calculate the hash? I just need to understand the use case a bit more in order to solve it satisfactorily.

> There is no easy way to delete the files in a zig package that are not referenced by the build.zig.zon and included in the package hash. There is no reason for distros to include these files in the packages they redistribute. A more generally useful solution might be to add a CLI command to list files included in the package, which the distro packaging tooling could use as a whitelist for what to include without needing to semantically parse the build.zig.zon.

Only files that are referenced by the build.zig.zon are included in the package hash. When you use `zig fetch` to fetch something, it does delete everything not referenced by the paths field of build.zig.zon.

If it helps, we can certainly provide such a tool. I'm not sure I understand the workflow yet though. Is this accurate?
1. Distro starts by unpacking pristine upstream source tarball
2. Distro deletes files not referenced by build.zig.zon
3. Distro obtains zig package hash
4. Distro creates a directory named after the package hash with the remaining upstream source files
5. That directory plus some other metadata becomes the distro package

And then this point has to do with streamlining step 2?

> While `--host-target`, `--host-cpu`, and `--host-dynamic-linker` were added, what toolchains like cbuild which support cross-compilation really need are `--target`, `--cpu`, and `--dynamic-linker` which describe the target system, not the host. This PR currently uses the `-Dtarget` and `-Dcpu` flags defined by waylock's build.zig instead but this is technically not portable.

Hmm, I see your point. I'm not sure how I got fixated on this idea of overriding the host. In the face of this new use case, it seems like I went down a completely wrong path.

src/cbuild/util/zig_build.py
andrewrk: Note that this actually gives first choice to the upstream application author, and "safe" here is the fallback choice if the upstream author does not declare a default release mode. In any case, your point stands.

nekopsykose: ah, i didn't know that. that's good to know, and probably makes this a little less meaningful :)

user/zig/template.py
andrewrk: There is another possibility that we can explore, which is to take advantage of zig's C backend, and support for [building from source without LLVM](https://github.com/ziglang/zig/?tab=readme-ov-file#building-from-source-without-llvm).

Such a build of zig is not suitable for direct users of zig, however, it could be used by the distro to package zig applications by compiling them to C code and then compiling and linking that C code with the system toolchain.

nekopsykose: hmm.. if that actually works for all regular zig projects one would want in a distro-packaged way (river, ncdu2, ..) then it honestly doesn't seem like the worst idea :D. i had no idea it was possible

src/cbuild/util/zig_build.py
ifreund: In any case, templates may override this default by adding `--release=fast` or whatever to `zig_build_args`.

user/zig-wayland/template.py
ifreund: Ah, I did not know that actually. I tried to use the semver version strings I use for arbitrary commits of my projects (e.g. `0.3.0-dev.204+4685f69`, 204 commits since the last tag for ordering, plus the commit hash) but it appears apk doesn't support semver in it's entirety.

In any case, I don't really see a reason to fix this for now, I'll obvious tag actual releases for these projects before taking this beyond a proof of concept :)

user/zig/template.py
ifreund: I haven't tried that approach yet but it certainly sounds promising in theory. I suspect there may be some rough edges that need to be ironed out but that's why I opened this PR in the first place :)

user/zig-wayland/template.py
nekopsykose: > but it appears apk doesn't support semver in it's entirety.  

the unofficial (forgot if it was "really" written anywhere? might be wrong) spec is that it's "gentoo versions", [technically](https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/686a9416520c0e21d0d8edf539f46b44808374ff/src/version.c#L15)..

actual semver supports way more potentially-craazy stuff (in both metadata and prerelease section), and yes it's especially annoying for the latter (projects do e.g. 1.0.0-alpha.1 but then one has to remap to _a1 for apk)

of course, yes this doesn't really matter here, just fluff :)

ifreund:
> 1. The distro can choose to lie and provide whatever version they want for any particular hash, swapping one version for another. The zig build system by design only verifies hashes on fetch, and if the system provides a particular hash, it will be trusted and used.

Good point, however this approach would not be sufficient to solve the situation where packages A and B list two different versions of package C (say 1.0.0 and 1.0.1) in their respective build.zig.zon files. I don't see a good way around patching the build.zig.zon of either A or B in that case.

From the distro point of view, a possible fix for this would be for `--system foobar` to look for packages in the `foobar` directory by name rather than by hash. This would rely on the zig ecosystem using consistent naming for packages, with the distro patching inconsistencies/duplicate names. It's currently possible (and I'm pretty sure an intentional feature) for the consumer of a zig package to totally ignore the name the package gave itself and assign it any arbitrary name, I'd hope this feature wouldn't be abused however.

> 2. An upcoming feature that will not land in 0.12.0 will address the general use case of multiple package versions existing in a dependency tree, and the ability to swap one package version for another. Probably at the same time this will come with CLI flags or other functionality that makes it easy for a distro to participate in this process.

It sounds like this would in theory resolve this issue in the long term, I wonder about how this would be implemented however as it seems that it would require reliance on consistent package naming or some other new metadata.

> Would you mind writing up a zig issue for this one? I started working on it, but I found myself not able to explain the problem very well. In particular, what would you want to happen to the directory of files that was fetched in order to calculate the hash? I just need to understand the use case a bit more in order to solve it satisfactorily.

> Only files that are referenced by the build.zig.zon are included in the package hash. When you use zig fetch to fetch something, it does delete everything not referenced by the paths field of build.zig.zon.

Actually, I think I misunderstood how `zig fetch` works and think that `zig fetch --global-cache-dir /usr/src/zig/packages/ foobar` would actually do exactly what we want here. I'll give it a try :)

> Hmm, I see your point. I'm not sure how I got fixated on this idea of overriding the host. In the face of this new use case, it seems like I went down a completely wrong path.

To be fair, most distro packaging toolchains don't support cross compilation so it's easy to look at a few and assume they only care about targeting the host system.

ifreund:
Pushed a commit leveraging `zig fetch` to avoid hardcoding the package hashes in the templates. It feels a bit like an abuse of `--global-cache-dir` but perhaps it's fine.

Nyaa97:
I have been using zig integration locally for some time for my own packages. I prefer to parse zig.build.zon myself, over creating a package for each dependency.
https://github.com/Nyaa97/cports

q66:
i will close this for as long as there is no supported zig in chimera, we can revisit at another point

PR #1396, opened 2024-02-07
contrib/pkcs11-helper: new package (1.30.0)

vakhrushevdv:
Hi, I want to add openvpn package and this is one of the dependencies.

### DIFF
--- a/contrib/pkcs11-helper/template.py
+++ b/contrib/pkcs11-helper/template.py
@@ -1,0 +1,29 @@
+pkgname = "pkcs11-helper"
+pkgver = "1.30.0"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--enable-slotevent",
+    "--enable-threading",
+]
+make_cmd = "gmake"
+hostmakedepends = ["automake", "pkgconf", "libtool", "gmake"]
+makedepends = ["openssl-devel"]
+pkgdesc = "Helper library for multiple PKCS#11 providers"
+maintainer = "Dmitriy Vakhrushev <dvakhrushev@netgate.com>"
+license = "GPL-2.0-only AND BSD-3-Clause"
+url = "https://github.com/OpenSC/pkcs11-helper"
+source = f"{url}/archive/refs/tags/pkcs11-helper-{pkgver}.tar.gz"
+sha256 = "076c9f664812a45f2da25efc157338b0b8bb1949117f0144050fec176b6fdf78"
+# vis breaks symbols
+hardening = ["!vis"]
+
+
+def post_install(self):
+    self.install_license("COPYING")
+    self.install_license("COPYING.BSD")
+
+
+@subpackage("pkcs11-helper-devel")
+def _devel(self):
+    return self.default_devel()
--- a/contrib/pkcs11-helper-devel
+++ b/contrib/pkcs11-helper-devel
@@ -1,0 +1,1 @@
+pkcs11-helper


contrib/pkcs11-helper/template.py
```diff
@@ -0,0 +1,32 @@
+pkgname = "pkcs11-helper"
+pkgver = "1.30.0"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+        "--disable-static",
```
nekopsykose, line 6: don't think you have to disable this

contrib/pkcs11-helper/template.py
```diff
@@ -0,0 +1,32 @@
...
+sha256 = "076c9f664812a45f2da25efc157338b0b8bb1949117f0144050fec176b6fdf78"
+
+def post_install(self):
+    self.install_license("COPYING")
+    self.install_license("COPYING.BSD")
+    self.install_license("COPYING.GPL")
+
+@subpackage("pkcs11-helper-man")
+def _man(self):
+    return ["usr/share/man"]
```
nekopsykose, line 28:
```suggestion
```

contrib/pkcs11-helper/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://github.com/OpenSC/pkcs11-helper/tags"
+pattern = r"pkcs11-helper-([\d.]+)"
```
nekopsykose, line 2: you don't need this file at all, the defaults work

contrib/pkcs11-helper/template.py
```diff
@@ -0,0 +1,32 @@
...
+maintainer = "Dmitriy Vakhrushev <dvakhrushev@netgate.com>"
+license = "GPL-2.0-only AND BSD-3-Clause"
+url = "https://github.com/OpenSC/pkcs11-helper"
+source = f"{url}/archive/refs/tags/pkcs11-helper-{pkgver}.tar.gz"
+sha256 = "076c9f664812a45f2da25efc157338b0b8bb1949117f0144050fec176b6fdf78"
+
+def post_install(self):
+    self.install_license("COPYING")
+    self.install_license("COPYING.BSD")
+    self.install_license("COPYING.GPL")
```
nekopsykose, line 24: gpl doesn't need installing

contrib/pkcs11-helper/template.py
```diff
@@ -0,0 +1,32 @@
...
+pkgver = "1.30.0"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+        "--disable-static",
+        "--enable-threading",
+        "--enable-slotevent",
+]
+make_cmd = "gmake"
+make_dir = "."
```
nekopsykose, line 11: it works fine without overriding make_dir

contrib/pkcs11-helper/template.py
```diff
@@ -0,0 +1,32 @@
...
+hostmakedepends = ["automake", "pkgconf", "libtool", "gmake"]
+makedepends = ["openssl-devel"]
+pkgdesc = "Helper library for multiple PKCS#11 providers"
+maintainer = "Dmitriy Vakhrushev <dvakhrushev@netgate.com>"
+license = "GPL-2.0-only AND BSD-3-Clause"
+url = "https://github.com/OpenSC/pkcs11-helper"
+source = f"{url}/archive/refs/tags/pkcs11-helper-{pkgver}.tar.gz"
+sha256 = "076c9f664812a45f2da25efc157338b0b8bb1949117f0144050fec176b6fdf78"
+
```
nekopsykose, line 20: check if hardening= works

contrib/pkcs11-helper/template.py
vakhrushevdv: thanks, done.

contrib/pkcs11-helper/template.py
vakhrushevdv: it works at least locally, ks, but I've changed

nekopsykose: it's not related to the url (keep the `{url}`). see e.g. `contrib/ddrescue`. for `vis`, see if all the exported symbols are there (`nm -D` on the library before/after), for cfi, see if it works without crashing (usually a testsuite would catch that if one runs here)

contrib/pkcs11-helper-man
```diff
@@ -0,0 +1 @@
+pkcs11-helper
```
nekopsykose, line 1: things without a @subpackage in the template don't need a symlink on disk (the ci fails cause of that)

contrib/pkcs11-helper/template.py
vakhrushevdv: got it, thank you very match for the details.

vakhrushevdv: flags have been checked:

-  tests can be enabled only when some kind of pkcs11 provider is available on the system
   I've tested on that it works on ubuntu system with preinstalled OpenHSM module. I've tried to add this module in the package, but it isn't trivial and needed to spend more time, which I have no for now.

- so I've tested it just by seeing at least that the tests are started and dynamic linkage passed. These  hardening (vis,cif) makes tests failed on dynamic linking, because part of the necessary functions are hidden. I guess this is what you might have guessed.

- 'vis' is closing major part of functions

- adding 'cif' is closing another several functions

nekopsykose: makes sense, thanks :)

PR #1361, opened 2024-01-31
main/alsa*: update to 1.2.11

nekopsykose:
https://www.alsa-project.org/wiki/Changes_v1.2.10_v1.2.11

PR #1403, opened 2024-02-09
main/alsa-{ucm-conf,lib}: update to 1.2.11

mhmdanas:
https://www.alsa-project.org/wiki/Changes_v1.2.10_v1.2.11


mhmdanas:
https://github.com/chimera-linux/cports/pull/1361 sigh, i should check first

PR #1405, opened 2024-02-09
contrib/guitarix: new package (0.44.1)

z-erica:


contrib/guitarix/template.py
```diff
@@ -0,0 +1,45 @@
...
+    "zita-convolver-devel",
+    "zita-resampler-devel",
+]
+pkgdesc = "Virtual guitar amplifier"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "GPL-2.0-or-later"
+url = "https://guitarix.org"
+source = f"$(SOURCEFORGE_SITE)/guitarix/guitarix/guitarix2-{pkgver}.tar.xz"
+sha256 = "77e83d754f51ac38c5423f38eeb55de5b3e26128e60b511b02d2defcf36e6c18"
+tools = {"LD": "ld.bfd"}
```
nekopsykose, line 43: this seems like a huge hack

nekopsykose: (for fwupd-efi/gnu-efi this is because they do fancy bare elf things that are kinda hardcoded to binutils, but for a regular-ass userspace program i would really rather not see gnubinutils suddenly being used..)

nekopsykose: oh wow that's a lot of links, lemme see..

nekopsykose: uh based on https://github.com/brummer10/GxPlugins.lv2/issues/9#issuecomment-431634711 they use `-b -r` but those work in ld.lld?

nekopsykose: what error do you actually get without this?

nekopsykose: yeah, it's linking e.g. just a png into a .o. if you ass a real -m value (forgot the one for your computer) does it work?

nekopsykose: (you can test that by just running it in the build directory etc, no need to patch stuff)

nekopsykose: if not then yeah i guess lld just doesn't support whatever this is (never seen this before lol), very cursed

nekopsykose: hm ok so it does support it

maybe there's a better way to force it to output host elf, that doesn't involve an explicit -m..

or, maybe you can find out if/why this is even needed? i don't understand why it needs to start a -r object file with a png

nekopsykose:
yes, see the patch in alpine aports for it (there's more flags)

contrib/guitarix/template.py
nekopsykose: assuming the names are all magically correct (i thought there was weird mapping) this actually looks completely ok

nekopsykose:
that seems ok. did you look at merging the object creation, i.e. mixing the .png with actual .o?

q66:
mostly lgtm but zita-resampler fails to pass either CFLAGS or LDFLAGS on the command line when linking the shared lib, so that needs fixing

PR #1408, opened 2024-02-11
contrib/reaction: new package (1.3.1)

ttyyls:


contrib/reaction/template.py
```diff
@@ -0,0 +1,46 @@
...
+# no tests defined
+options = ["!check"]
+
+
+def pre_install(self):
+    from cbuild.util import compiler
+
+    cc = compiler.C(self)
+    cc.invoke(["helpers_c/ip46tables.c"], "ip46tables")
+    cc.invoke(["helpers_c/nft46.c"], "nft46")
```
nekopsykose, line 21: put this in post_build and the installs in post_install

contrib/reaction/files/reaction.yml
```diff
@@ -0,0 +1,100 @@
+---
```
nekopsykose, line 1: you could also install this file from ./app/example.yml afaik

ttyyls: I do, reaction.yml is just that file plus a header with more clarifications

ttyyls: the default config hard depends on an iptables setup, so just in case someone running nftables or firewalld installs the package and fails to start it, those will be a great indication of what went wrong

ttyyls: well i suppose since it doesn't change anything functionally might as well delete it.

contrib/reaction/files/reaction
```diff
@@ -0,0 +1,8 @@
+# reaction service
+
+type            = process
```
q66, line 3: unalign the service

PR #1409, opened 2024-02-11
contrib/megatools: new package (1.11.1)

ifreund:
Man pages aren't currently built as docbook2x is required and looks like a lot of work to package. I don't have plans to package it myself at the moment but thought I'd open this PR anyways in case it's acceptable to merge without man pages and to avoid duplicated effort if anyone else looks into packaging this software.

### DIFF
--- a/contrib/megatools/template.py
+++ b/contrib/megatools/template.py
@@ -1,0 +1,27 @@
+pkgname = "megatools"
+pkgver = "1.11.1"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "asciidoc",
+    "meson",
+    "pkgconf",
+    # TODO package docbook2x to allow man page compliation
+    # "docbook2x",
+]
+makedepends = [
+    "fuse-devel",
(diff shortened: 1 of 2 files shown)


nekopsykose:
>  in case it's acceptable to merge without man pages  

yeah, that's ok :)

q66:
fyi latest release of docbook2x was 2007

contrib/megatools/template.py
```diff
@@ -0,0 +1,27 @@
...
+    "gobject-introspection",
+    "libcurl-devel",
+    "libsodium-devel",
+    "openssl-devel",
+]
+pkgdesc = "Command line client for mega.nz"
+maintainer = "Isaac Freund <mail@isaacfreund.com>"
+license = "GPL-2.0-or-later"
+url = "https://megatools.megous.com"
+source = f"{url}/builds/megatools-{pkgver}.20230212.tar.gz"
```
nekopsykose, line 25: ah also missed this- this .20230212 can just be part of pkgver

PR #1411, opened 2024-02-11
contrib/soju: new package (0.8.0)

ttyyls:


contrib/soju/template.py
```diff
@@ -0,0 +1,79 @@
...
+pkgdesc = "IRC bouncer"
+maintainer = "ttyyls <contact@behri.org>"
+license = "AGPL-3.0-or-later"
+url = "https://soju.im"
+source = f"https://git.sr.ht/~emersion/soju/archive/v{pkgver}.tar.gz"
+sha256 = "089b900ccd7fa009b7ae50b9c2b9e77099b50b2499422649c3b27aff7d62d109"
+env = {"GOFLAGS": "-tags=libsqlite3"}
+file_xattrs = {
+    "usr/bin/soju": {
+        "security.capability": "cap_net_bind_service,cap_net_admin+ep",
```
nekopsykose, line 17: this absolutely does not need cap_net_admin

nekopsykose: it also makes more sense to use capsh in the service file than to set the capability here, since only the service might need it

ttyyls:
in golang.py, custom args have to precede tags and -o in order to be valid.
with that patched and args=["-buildmode=pie"] on the utils, it seems to build fine on my local testing.

nekopsykose:
yeah, i amended that a little in 83e5369bb2c474e7751da5ab8f61fe8d192e4b02 but it's annoying to get perfectly right

contrib/soju/template.py
```diff
@@ -0,0 +1,74 @@
...
+
+def post_build(self):
+    self.make_build_args = [
+        "./contrib/migrate-db",
+        "./contrib/migrate-logs",
+        "./contrib/znc-import",
+    ]
+    from cbuild.util import golang
+
+    golang.Golang(self).build(args=["-buildmode=pie"])
```
nekopsykose, line 43: if you have to pass this then something else is wrong as that should be the default

probably you have to init self.golang = in init_build or wherever so everything is kept between the steps

nekopsykose:
honestly reading this again this is easier to just use build_style=go for, pass all the paths as make_build_args for the executables, then do a post_build of self.do(make doc/soju.1 doc/sojuctl.1) and install everything manually

ttyyls:
i wrothe the initial template using build_style="go" but then i had to repass the ldflags from the makefile and hardcode them in the template. i shifted to this more convoluted way to make future updates less noisy if those ldflags were to be get changed.

nekopsykose:
nothing has to be passed tho  

here you go

```
pkgname = "soju"
pkgver = "0.7.0"
pkgrel = 0
build_style = "go"
make_build_args = ["./contrib/...", "./cmd/..."]
hostmakedepends = ["go", "scdoc"]
makedepends = ["sqlite-devel"]
depends = ["ca-certificates", "libcap-progs"]
go_build_tags = ["libsqlite3"]
go_check_tags = list(go_build_tags)
pkgdesc = "IRC bouncer"
maintainer = "ttyyls <contact@behri.org>"
license = "AGPL-3.0-or-later"
url = "https://soju.im"
source = f"https://git.sr.ht/~emersion/soju/archive/v{pkgver}.tar.gz"
sha256 = "089b900ccd7fa009b7ae50b9c2b9e77099b50b2499422649c3b27aff7d62d109"


def post_build(self):
    self.do("make", "doc/soju.1", "doc/sojuctl.1")


def do_install(self):
    self.install_service(self.files_path / "soju")
    self.install_file(
        self.files_path / "sysusers.conf",
        "usr/lib/sysusers.d",
        name="soju.conf",
    )
    self.install_file(
        self.files_path / "tmpfiles.conf",
        "usr/lib/tmpfiles.d",
        name="soju.conf",
    )
    for util in [
        "migrate-db",
        "migrate-logs",
        "znc-import",
    ]:
        self.install_bin(f"{self.make_dir}/{util}", name=f"soju-{util}")
    self.install_file("contrib/casemap-logs.sh", "usr/libexec", mode=0o755)
    self.install_man("doc/soju.1")
    self.install_man("doc/sojuctl.1")


@subpackage("soju-utils")
def _utils(self):
    self.depends = [f"{pkgname}={pkgver}-r{pkgrel}"]
    return [
        "usr/bin/soju-*",
        "usr/libexec/casemap-logs.sh",
    ]
```

nekopsykose:
(+ the config file)

contrib/soju/template.py
ttyyls: no, that seems to have been unecessary

contrib/soju/template.py
```diff
@@ -0,0 +1,54 @@
...
+    for util in [
+        "migrate-db",
+        "migrate-logs",
+        "znc-import",
+    ]:
+        self.install_bin(f"{self.make_dir}/{util}", name=f"soju-{util}")
+    self.install_file("contrib/casemap-logs.sh", "usr/libexec", mode=0o755)
+    self.install_man("doc/soju.1")
+    self.install_man("doc/sojuctl.1")
+    self.cp(self.destdir / "config.in", "etc/soju/config")
```
nekopsykose, line 45:
```suggestion
    self.install_file("config.in", "etc/soju", name="config")
```

contrib/soju/template.py
```diff
@@ -0,0 +1,54 @@
...
+        self.files_path / "tmpfiles.conf",
+        "usr/lib/tmpfiles.d",
+        name="soju.conf",
+    )
+    for util in [
+        "migrate-db",
+        "migrate-logs",
+        "znc-import",
+    ]:
+        self.install_bin(f"{self.make_dir}/{util}", name=f"soju-{util}")
```
nekopsykose, line 41: ah, also forgot to install_bind("soju*", glob=True)

nekopsykose:
`fix: custom args should go before -o` should say `cbuild: go custom args...`

contrib/soju/template.py
```diff
@@ -0,0 +1,65 @@
...
+sha256 = "089b900ccd7fa009b7ae50b9c2b9e77099b50b2499422649c3b27aff7d62d109"
+
+
+def do_build(self):
+    self.golang.build(
+        args=[
+            "-ldflags",
+            "-X 'git.sr.ht/~emersion/soju/config.DefaultPath=/etc/config/soju' \
+	        -X 'git.sr.ht/~emersion/soju/config.DefaultUnixAdminPath=/run/soju/admin'",
+        ],
```
nekopsykose, line 25: you can pass this in make_build_args

nekopsykose: see tailscale

ttyyls:
> nothing has to be passed tho
without specifying the paths like https://git.sr.ht/~emersion/soju/tree/master/item/Makefile#L14 the path to the config is unknown and calls of soju/sojuctl that omit them fail, like here https://git.sr.ht/~emersion/soju/tree/master/item/cmd/sojuctl/main.go#L39

contrib/soju/template.py
ttyyls: the order wouldn't be right as they'd be put after -o

ttyyls: with the previous patch to args, it works as expected

nekopsykose: you mean the thing you already patched here? or something else?

ttyyls: yes the first commit in this pr

nekopsykose: actually looking again i'm not sure what you're talking about. in tailscale the build line looks like

```
go build -p 16 -v -trimpath -o /builddir/tailscale-1.58.2/build/ -ldflags= -X tailscale.com/version.longStamp=1.58.2.222 -X tailscale.com/version.shortStamp=1.58.2.22222 ./cmd/tailscale ./cmd/tailscaled
```

(all the things are passed after `-o`)

and yet, it works as expected. the ldflags are read fine. what are you talking about?

nekopsykose: yeah, i just tested removing the patch from here and then  

```
make_build_args = [
    "-ldflags="
    " -X git.sr.ht/~emersion/soju/config.DefaultPath=/etc/config/soju2"
    " -X git.sr.ht/~emersion/soju/config.DefaultUnixAdminPath=/run/soju/admin",
    "./contrib/...",
    "./cmd/...",
]
```

and this works completely fine, so i'm confused what you were trying to fix

nekopsykose: also the go build command disagrees with you:  

```
$ go build
usage: go build [-o output] [build flags] [packages]
```

everything is supposed to come after `-o`. not before it

ttyyls: you're right. failure of those tests must've been due to sth unrelated to it.

ttyyls: > also the go build command disagrees with you:
just dropped it

contrib/soju/files/soju
```diff
@@ -0,0 +1,8 @@
+# soju system service
+
+type       = process
+command    = /usr/bin/capsh --inh=cap_net_bind_service --user=_soju -- -c "exec /usr/bin/soju -config /etc/soju/config"
+depends-on = network.target
+depends-on = pre-local.target
+logfile    = /var/log/soju.log
+
```
nekopsykose, line 8: trailing newline

contrib/soju/template.py
```diff
@@ -0,0 +1,60 @@
+pkgname = "soju"
+pkgver = "0.7.0"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    "-ldflags",
+    "-X 'git.sr.ht/~emersion/soju/config.DefaultPath=/etc/config/soju' -X 'git.sr.ht/~emersion/soju/config.DefaultUnixAdminPath=/run/soju/admin'",
```
nekopsykose, line 7:
```suggestion
    "-ldflags="
    " -X git.sr.ht/~emersion/soju/config.DefaultPath=/etc/config/soju"
    " -X git.sr.ht/~emersion/soju/config.DefaultUnixAdminPath=/run/soju/admin",
```

python joins adjacent strings with no separator so you can use that to split things over lines even when they normally can't be

contrib/soju/files/soju
```diff
@@ -0,0 +1,8 @@
+# soju system service
+
+type       = process
+command    = /usr/bin/capsh --inh=cap_net_bind_service --user=_soju -- -c "exec /usr/bin/soju -config /etc/soju/config"
```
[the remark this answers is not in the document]
ttyyls: `Unable to set group list for user: Operation not permitted`

contrib/soju/files/soju
ttyyls: ok

contrib/soju/template.py
ttyyls: adds a space after ldflags which seems to confuse the rest of the cli parsing
`flag provided but not defined: -X 'git.sr.ht/~emersion/soju/config.DefaultUnixAdminPath
`

nekopsykose: i already tested exactly what i wrote and it works

nekopsykose: probably, you are missing the `=` after ldflags

contrib/soju/files/soju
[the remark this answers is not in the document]
nekopsykose: i guess that would be a reason :)  

the existing thing you wrote (`--inh="cap_net_bind_service+eip"`) is a syntax error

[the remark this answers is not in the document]
nekopsykose: actually no i just typed it wrong

contrib/soju/files/tmpfiles.conf
```diff
@@ -0,0 +1,3 @@
+# create soju state directories
+
+d /var/lib/soju 0755 _soju _soju -
```
nekopsykose, line 3:
```suggestion
d /var/lib/soju 0755 _soju _soju -
d /run/soju 0750 _soju _soju -
```

contrib/soju/files/soju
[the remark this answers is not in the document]
nekopsykose: but anyway, this doesn't work. as the documentation says `--inh` doesn't actually grant a capability. if you put `listen http+prometheus://localhost:412` it will fail (this cap is only needed for low (<1024) port numbers)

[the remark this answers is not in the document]
nekopsykose: did a bit of reading and it seems you need something like  

```
capsh --caps="cap_net_bind_service,cap_setgid,cap_setuid,cap_setpcap+eip" --user=_soju --addamb='cap_net_bind_service'
```

this then:  

```
$ doas grep Cap /proc/432/status
CapInh:	00000000000005c0
CapPrm:	0000000000000400
CapEff:	0000000000000400
CapBnd:	000001ffffffffff
CapAmb:	0000000000000400
$ capsh --decode=0000000000000400
0x0000000000000400=cap_net_bind_service
```

only gives it that one. found this out via https://www.redhat.com/en/blog/linux-capabilities-in-openshift

very unintuitive :P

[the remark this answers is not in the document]
nekopsykose: ah, and the setpcap one isn't needed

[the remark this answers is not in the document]
ttyyls: makes sense

contrib/soju/files/tmpfiles.conf
ttyyls: ok.

contrib/soju/files/soju
```diff
@@ -0,0 +1,7 @@
+# soju system service
+
+type       = process
+command    = /usr/bin/capsh --caps="cap_net_bind_service,cap_setgid,cap_setuid+eip" --user=_soju --addamb='cap_net_bind_service' -- -c "exec /usr/bin/soju -config /etc/soju/config"
```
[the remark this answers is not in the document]
ttyyls: this iteration started producing the following error `capability ['cap_net_bind_service'] is unknown to libcap` so I shuffled them around and added --keep=1 while at it

PR #1418, opened 2024-02-13
user/motif: new package (2.3.8)

mikoxyz:
threw this in user/ due to how cursed motif is


q66:
do you have something actually reasonable to use this with or is this just a meme

mikoxyz:
> do you have something actually reasonable to use this with or is this just a meme

is cde reasonable?

q66:
if somebody's unironically going to use it

Alasofia:
Actively trying to build Motif software on Chimera and this would be *lovely*.
XFile my beloved!

Also looking into a commit for the [2.4.1 modernization update.](https://github.com/dimmus/motif)

PR #1423, opened 2024-02-13
main/p11-kit: update to 0.25.3

triallax:


nekopsykose:
does rebuilding things against libraries that link this work, and does it work at runtime? 0.25 was reverted because there was some weird stuff but i forgot now

PR #1424, opened 2024-02-13
contrib/mkvtoolnix: new package (82.0)

nekopsykose:


contrib/cmark/template.py
```diff
@@ -0,0 +1,33 @@
...
+    "pkgconf",
+]
+checkdepends = ["python"]
+pkgdesc = "C implementation of the CommonMark markdown specification"
+maintainer = "psykose <alice@ayaya.dev>"
+license = "BSD-2-Clause"
+url = "https://github.com/commonmark/cmark"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "bbcb8f8c03b5af33fcfcf11a74e9499f20a9043200b8552f78a6e8ba76e04d11"
+# defaults to Release which sets this, and the tests crash in an assert without it..
```
q66, line 18: and the assertion crash isn't a regular test failure that NDEBUG hides?

contrib/mkvtoolnix/template.py
```diff
@@ -0,0 +1,65 @@
...
+url = "https://mkvtoolnix.download/index.html"
+source = f"https://mkvtoolnix.download/sources/mkvtoolnix-{pkgver}.tar.xz"
+sha256 = "6a8615436406c7fa45bfb2b6270da1bf06ea54cfcd13c3699643833e1d73ecbc"
+
+
+@subpackage("mkvtoolnix-gui")
+def _gui(self):
+    self.depends += [
+        "qt6-qtsvg",
+        f"{pkgname}={pkgver}-r{pkgrel}",
```
q66, line 55: nit: put this first in the list

contrib/cmark/template.py
nekopsykose: technically, it is an actual assert (`src/blocks.c:262`), in api_test. but upstream never sets a build_type in their ci, so their own tests always run with ndebug from =release, so i bet it's actually broken for them too. a bit too lazy to ask however

q66: well ok, i guess that's fine

nekopsykose: actually i was wrong and `if(NOT CMAKE_BUILD_TYPE)` is false when `-DCMAKE_BUILD_TYPE=`(empty) which they pass, so they do have asserts enabled. will see what the issue is..

nekopsykose: durr, no. that's wrong, it indeed makes it Release for empty. i was just looking at something else

nekopsykose: so yeah, it's ok, and i'll open an issue anyway

PR #1429, opened 2024-02-14
contrib/muon: new package (0.2.0)

ttyyls:


contrib/muon/template.py
```diff
@@ -0,0 +1,30 @@
+pkgname = "muon"
+pkgver = "0.2.0"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+    "scdoc",
+]
```
mhmdanas, line 9:
```suggestion
hostmakedepends = ["meson", "pkgconf", "scdoc"]
```

contrib/muon/template.py
```diff
@@ -0,0 +1,30 @@
...
+makedepends = [
+    "libarchive-devel",
+    "libcurl-devel",
+    "pkgconf-devel",
+]
```
mhmdanas, line 14:
```suggestion
makedepends = ["libarchive-devel", "libcurl-devel", "pkgconf-devel"]
```

contrib/muon/patches/0001-hostfix.patch
```diff
@@ -0,0 +1,22 @@
+From 4679fa306905f89280dcdd9885482b8cdd498aa3 Mon Sep 17 00:00:00 2001
+From: ttyyls <contact@behri.org>
+Date: Wed, 14 Feb 2024 01:01:09 +0100
+Subject: [PATCH] hostfix
```
mhmdanas, line 4: This patch description isn't too clear, can you improve it and add more details in the patch description if needed?

contrib/muon/template.py
```diff
@@ -0,0 +1,30 @@
...
+sha256 = "d73db1be5388821179a25a15ba76fd59a8bf7c8709347a4ec2cb91755203f36c"
+# hidden visibility breaks almost all tests
+hardening = ["!vis"]
+
+
+def post_install(self):
+    self.install_license("LICENSES/Apache-2.0.txt")
+    self.install_license("LICENSES/GPL-3.0-only.txt")
+    self.install_license("LICENSES/MIT.txt")
+    self.install_license("LICENSES/Unlicense.txt")
```
mhmdanas, line 30: I don't think you need to install Apache-2.0, GPL-3.0-only, and (maybe?) Unlicense, but somebody correct me if I'm wrong.

ttyyls: I know that reproducing the unlincense and apchae at least isn't necessary. but the point of omitting licenses is to drop a -doc package, if we're gonna have one anyway might as well include them all

contrib/muon/patches/0001-hostfix.patch
ttyyls: it's unclear to me as well, I tested with bash, zsh and oksh set as the system shell and it keep getting reproduced, the only conclusion I have is the test is looking for the wrong value

contrib/muon/template.py
nekopsykose: you only need mit

contrib/muon/patches/0001-hostfix.patch
nekopsykose: it does  

```
[find_program('sh'), '-c', 'echo $#', '--', '', 'a', '', 'b', ''],
```

in most sh this is parsed as `--` for argv[0]. then, there are 5 args after (`'', 'a', '', 'b', ''`)

in freebsd sh this is parsed as `--` to stop reading args and then `''` for argv[0]. then, there are 4 args left, so the count is different.

the correct thing to do is replace `--` with `sh`. then it's 5 for both and behaves the same way anyway (this was not doing what was intended silently)

PR #1436, opened 2024-02-15
enable various tests that need cunit

mikoxyz:
depends on #1394

### DIFF
--- a/main/libtracefs/patches/0002-libtracefs-utest-Define-_LARGEFILE64_SOURCE-for-lsee.patch
+++ b/main/libtracefs/patches/0002-libtracefs-utest-Define-_LARGEFILE64_SOURCE-for-lsee.patch
@@ -1,0 +1,34 @@
+From 7b4a9c9c9b962929323f9d97e00fb5e51bb205b9 Mon Sep 17 00:00:00 2001
+From: Miko Larsson <mikoxyzzz@gmail.com>
+Date: Thu, 22 Feb 2024 11:12:45 +0100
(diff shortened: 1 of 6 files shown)


nekopsykose:
i already packaged cunit in https://github.com/chimera-linux/cports/pull/1394

main/libtracefs/patches/0002-libtracefs-utest-Define-_LARGEFILE64_SOURCE-for-lsee.patch
```diff
@@ -0,0 +1,34 @@
...
+diff --git a/utest/tracefs-utest.c b/utest/tracefs-utest.c
+index 07ecd32..737e54a 100644
+--- a/utest/tracefs-utest.c
++++ b/utest/tracefs-utest.c
+@@ -3,6 +3,8 @@
+  * Copyright (C) 2020, VMware, Tzvetomir Stoyanov <tz.stoyanov@gmail.com>
+  *
+  */
++#define _LARGEFILE64_SOURCE
++
```
nekopsykose, line 28: this is the wrong fix, it should remove lseek64 and use lseek + -D_FILE_OFFSET_BITS=64(for glibc; on musl it's always 64) instead

mikoxyz:
On Fri Feb 23, 2024 at 2:11 AM CET, alice wrote:
> this is the wrong fix, it should remove lseek64 and use lseek + -D_FILE_OFFSET_BITS=64(for glibc; on musl it's always 64) instead
meh, this patch was what got merged upstream, so ¯\_(ツ)_/¯

nekopsykose:
write the good one next time then :P

PR #1437, opened 2024-02-15
contrib/yggdrasil: new package (0.5.5)

jcgruenhage:
### Things that are working
 - building and running this natively on aarch64
 - building this natively on x86_64
 - logs are stored by dinit properly
 - detection whether config file exists and falling back to `-autoconf` works
### Things that aren't working
 - cross compiling from x86_64 to aarch64
 ```
 => yggdrasil-0.5.5-r0: running post_install hook: 007_strip_debug...
objcopy: error: SHT_STRTAB string table section [index 11] is non-null terminated
=> yggdrasil-0.5.5-r0: ERROR: failed to attach debug link to usr/bin/yggdrasilctl
```

### DIFF
--- a/contrib/yggdrasil/template.py
+++ b/contrib/yggdrasil/template.py
@@ -1,0 +1,32 @@
+pkgname = "yggdrasil"
+pkgver = "0.5.5"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    "-ldflags="
+    f" -X github.com/yggdrasil-network/yggdrasil-go/src/version.buildName={pkgname}"
+    f" -X github.com/yggdrasil-network/yggdrasil-go/src/version.buildVersion={pkgver}",
+    "./cmd/yggdrasil/",
+    "./cmd/yggdrasilctl/",
+]
+make_check_args = ["./src/..."]
+hostmakedepends = ["go"]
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: contrib/yggdrasil/template.py [hunk 1]
```diff
@@ -0,0 +1,28 @@
...
+source = f"https://github.com/yggdrasil-network/yggdrasil-go/archive/v{pkgver}.tar.gz"
+sha256 = "cdbb56b19b91b828afa282554862efb2a79dd4ada26dfb5a7bf3b0c5220f6c17"
+
+
+def pre_build(self):
+    self.do("rm", "build")
+
+
+def do_build(self):
+    self.golang.build(env={"PKGNAME": pkgname, "PKGVER": pkgver})
```
nekopsykose, line 21: `make_build_env =` instead of overriding do_build should work for this

nekopsykose: actually no, this doesn't work now either. these variables are read by `./build` to write go ldflags. nothing else reads the env, and we invoke go directly

jcgruenhage: Good catch, thanks. I'll set the ldflags instead then

PR #1438, opened 2024-02-15
main/python-wheel: enable tests

mikoxyz:


nekopsykose:
i don't think this is possible because of cycles

nekopsykose:
in this case i don't think it's possible, so unless you find arcane wizardry beyond mortal comprehension it's close

PR #1446, opened 2024-02-18
contrib/kakoune: new package (2023.08.05)

superwhiskers:


contrib/kakoune/patches/fix-issue-63281-llvm-llvm-project.patch
```diff
@@ -0,0 +1,137 @@
+diff --git a/src/command_manager.cc b/src/command_manager.cc
```
nekopsykose, line 1: where is this from? did you upstream it?

nekopsykose:
you should patch out this https://github.com/mawww/kakoune/blob/2d9c84e363d499eb699540c53d9f912f446a4d30/src/Makefile#L19

contrib/kakoune/patches/fix-issue-63281-llvm-llvm-project.patch
nekopsykose: ah, i see it's https://github.com/mawww/kakoune/commit/7577fa1b668ea81eb9b7b9af690a4161187129dd . please use the actual .patch from the url when using those at least, otherwise there is no way to know :p

nekopsykose:
and also pass `gzip_man=no` to make_install_args

contrib/kakoune/patches/fix-issue-63281-llvm-llvm-project.patch
superwhiskers: i made it myself. there are some differences in it in terms of line numbers

superwhiskers: if diff is capable of dealing with that though, i wasn't aware. do let me know and i'll use the patch directly from there

nekopsykose: > if diff is capable of dealing with that though,

it is

superwhiskers:
there. that should address everything

nekopsykose:
have to squash the commits into one too

superwhiskers:
there you go

nekopsykose:
thanks!

PR #1448, opened 2024-02-18
contrib/starship: new package (1.17.1)

ttyyls:


contrib/googletest/template.py
```diff
@@ -0,0 +1,32 @@
+pkgname = "googletest"
```
nekopsykose, line 1: this already exists as main/gtest

contrib/googletest/template.py
```diff
@@ -0,0 +1,32 @@
...
+options = ["empty"]
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+
+
+@subpackage("googletest-devel-static")
+def _static(self):
+    return self.default_static()
```
nekopsykose, line 27: for reference, the default-devel also splits -devel-static, you never need to declare this normally

contrib/starship/template.py
```diff
@@ -0,0 +1,45 @@
+pkgname = "starship"
+pkgver = "1.17.1"
+pkgrel = 0
+build_style = "cargo"
+make_build_args = [
+    "--no-default-features",
+    "--features=battery,notify,gix-faster",
```
nekopsykose, line 7: with gix-faster zlib-ng is not even used, so you can remove it from here

nekopsykose: similarly, cmake can be removed from hostmakedeps, and dbus-devel/libgit2-devel/openssl-devel can also be removed (they seem unused and not linked to?)

contrib/starship/template.py
```diff
@@ -0,0 +1,45 @@
...
+]
+make_install_args = list(make_build_args)
+make_check_args = make_build_args + [
+    # test involving writes to /tmp fail
```
nekopsykose, line 11:
```suggestion
    # /tmp is home of the cbuild user so tests that check for /tmp/dir get ~/dir as exact path
```

contrib/starship/template.py
ttyyls: the default is gix-max-perf which fails (afaict due to a version mismatch) perhaps we it could be kept and re-enbaled on the next point release assuming it's fixed by then?

nekopsykose: nah

contrib/starship/template.py
```diff
@@ -0,0 +1,45 @@
...
+    "--skip=modules::directory::tests::highlight_git_root_dir_zero_truncation_length",
+    "--skip=modules::directory::tests::truncated_directory_config_large",
+]
+hostmakedepends = ["cargo", "pkgconf"]
+makedepends = [
+    "dbus-devel",
+    "libgit2-devel",
+    "openssl-devel",
+    "rust-std",
+    "zlib-ng-devel",
```
nekopsykose, line 24:
```suggestion
    "rust-std",
    "zlib-devel",
```

contrib/starship/template.py
ttyyls: ok

contrib/starship/template.py
```diff
@@ -0,0 +1,39 @@
...
+    "--",
+    "--skip=modules::directory::tests::fish_directory_config_small",
+    "--skip=modules::directory::tests::fish_style_directory_config_large",
+    "--skip=modules::directory::tests::highlight_git_root_dir_zero_truncation_length",
+    "--skip=modules::directory::tests::truncated_directory_config_large",
+]
+hostmakedepends = ["cargo", "pkgconf"]
+makedepends = ["rust-std", "zlib-devel"]
+checkdepends = ["git"]
+pkgdesc = "Cross-Shell Prompt"
```
q66, line 21: title case bad

contrib/starship/template.py
```diff
@@ -0,0 +1,39 @@
...
+source = (
+    f"https://github.com/starship/starship/archive/refs/tags/v{pkgver}.tar.gz"
+)
+sha256 = "2b2fc84feb0197104982e8baf17952449375917da66b7a98b3e3fd0be63e5dba"
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+    for shell in ["bash", "fish", "zsh"]:
+        self.do(
```
q66, line 34: since it's doing that, it should also be !cross

PR #1449, opened 2024-02-18
fix: cross! packages that generate completions

ttyyls:
https://github.com/chimera-linux/cports/pull/1448#discussion_r1493792987


nekopsykose:
thanks, i added another missed one in there from way ago

PR #1450, opened 2024-02-18
contrib/stagit: new package (1.2)

ttyyls:


contrib/stagit/patches/0001-use-system-prefix.patch
```diff
@@ -0,0 +1,33 @@
...
+-PREFIX = /usr/local
+-MANPREFIX = ${PREFIX}/man
++PREFIX ?= /usr
++MANPREFIX = ${PREFIX}/share/man
+ DOCPREFIX = ${PREFIX}/share/doc/${NAME}
+ 
+-LIBGIT_INC = -I/usr/local/include
+-LIBGIT_LIB = -L/usr/local/lib -lgit2
++LIBGIT_INC = -I/usr/include
++LIBGIT_LIB = -L/usr/lib -lgit2
```
nekopsykose, line 27: you can pass all of these with make_build_args already. also for PREFIX the default non-make_use_env overrides it too

contrib/stagit/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "stagit"
+pkgver = "1.2"
+pkgrel = 0
+build_style = "makefile"
+make_build_args = [
+    "LIBGIT_INC=-I/usr/include",
+    "LIBGIT_LIB=-L/usr/lib -lgit2",
```
nekopsykose, line 7:
```suggestion
    "LIBGIT_LIB=-lgit2",
```

default lib path already

q66:
ditto here for strlcpy/cat

PR #1453, opened 2024-02-19
contrib/yt-dlp: enable tests (+ deps)

mikoxyz:
i'm a bit unsure about python-brotli, but everything else *should* be fine

### DIFF
--- a/contrib/python-brotli/template.py
+++ b/contrib/python-brotli/template.py
@@ -1,0 +1,23 @@
+pkgname = "python-brotli"
+pkgver = "1.1.0"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+    "python-wheel",
+]
+makedepends = ["python-devel"]
+checkdepends = ["python-pytest"]
+depends = ["python"]
+pkgdesc = "Python bindings for the Brotli compression library"
+maintainer = "miko <mikoxyzzz@gmail.com>"
+license = "MIT"
+url = "https://github.com/google/brotli"
+source = f"$(PYPI_SITE)/b/brotli/Brotli-{pkgver}.tar.gz"
+sha256 = "81de08ac11bcb85841e440c13611c00b67d3bf82698314928d0b676362546724"
+
+
+def post_install(self):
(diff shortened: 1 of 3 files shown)


nekopsykose:
all of these deps are optional and don't need to be depended on

contrib/python-certifi/template.py
```diff
@@ -0,0 +1,19 @@
+pkgname = "python-certifi"
```
nekopsykose, line 1: this needs a patch to pick up the system ca bundle instead of using the one that comes with it, see e.g. alpine etc

contrib/yt-dlp/template.py
```diff
@@ -9,13 +9,16 @@
...
 ]
-checkdepends = ["python-pytest", "python-nose", "flake8"]
-depends = ["python"]
+depends = [
+    "python-brotli",
+    "python-certifi",
+    "python-requests",
+    "python-websockets",
+]
+checkdepends = ["python-pytest"]
```
nekopsykose, line 18:
```suggestion
checkdepends = [
    "python-brotli",
    "python-certifi",
    "python-pytest",
    "python-requests",
    "python-websockets",
]
depends = ["python"]
```

contrib/python-brotli/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "python-brotli"
```
nekopsykose, line 1: this should probably be in main/, and have a comment added into main/brotli to update it at the same time, etc

contrib/python-certifi/template.py
```diff
@@ -0,0 +1,20 @@
...
+# only bump if there are any changes to the code
```
nekopsykose, line 2: technically it can always be bumped harmlessly, so not sure why this restriction

mikoxyz:
On Fri Feb 23, 2024 at 2:19 AM CET, alice wrote:
> technically it can always be bumped harmlessly, so not sure why this restriction
i mean, yeah, but it's also extremely pointless

mikoxyz:
On Fri Feb 23, 2024 at 2:18 AM CET, alice wrote:
> this should probably be in main/, and have a comment added into main/brotli to update it at the same time, etc
@q66, thoughts?

PR #1455, opened 2024-02-19
main/syslog-ng: update to 4.6.0

mhmdanas:


q66:
conf file version is not updated

mhmdanas:
Done.

PR #1456, opened 2024-02-19
contrib/sfeed: new package (2.0)

ttyyls:


q66:
can probably set `COMPATOBJ= COMPATSRC=` to skip building strlcat/strlcpy implementations (musl comes with them)

PR #1457, opened 2024-02-19
main/bash-completion: enable tests

mikoxyz:
and that's not even all the checkdepends of packages we have in main/ if we actually want as close ot full coverage as possible

im so sorry


q66:
no

PR #1460, opened 2024-02-20
user/darkhttpd: new package

ttyyls:


user/darkhttpd/template.py
```diff
@@ -0,0 +1,19 @@
...
+url = "https://unix4lyfe.org/darkhttpd"
+source = f"https://github.com/emikulic/darkhttpd/archive/v{pkgver}.tar.gz"
+sha256 = "ab97ea3404654af765f78282aa09cfe4226cb007d2fcc59fe1a475ba0fef1981"
+hardening = ["vis", "cfi"]
+# no tests defined
+options = ["!check"]
+
+
+def install(self):
+    self.install_license("COPYING")
```
triallax, line 18: `COPYING` doesn't contain a copyright notice, so you can either e.g. import my patch from https://github.com/emikulic/darkhttpd/pull/70 or probably do this instead:
```suggestion
# check: no tests defined
# distlicense: license file doesn't include copyright notice, see
# https://github.com/emikulic/darkhttpd/pull/70
options = ["!check", "!distlicense"]


def install(self):
```

ttyyls: the patch itself looks good and I suppose necessary for the license to be coherent. But a copyight notice is included in the executable itself and gets displayed first when run, as such Im not sure if it's necessary to block on it

PR #1467, opened 2024-02-21
main/firmware-linux: update to 20240220

nekopsykose:


mhmdanas:
Tested and it works fine (and saves 367 MiB of disk space!).

nekopsykose:
it still needs fixing in the things that replaces= this in two packages (brcm-rpi/steamdeck) so those are also following the same compressed extension

PR #1471, opened 2024-02-21
contrib/gamescope: update to 3.14.1

nekopsykose:


JamiKettunen:
`pkgname`s are e.g. `wlroots0.17` but commit messages have `wlroots-0.17` (same for 0.16)

nekopsykose:
true, thanks

contrib/wlroots0.16/template.py
```diff
@@ -27,18 +27,19 @@
     "xcb-util-renderutil-devel",
     "xcb-util-wm-devel",
 ]
+replaces = ["wlroots"]
```
q66, line 30: this is nowhere near enough (e.g. people who currently have wlroots-devel installed explicitly in world will have leftover files)

i'd rather have one of the versions (probably the newer one) provide unversioned names virtually for both the main package and the devel package, which will take care of a proper transition

contrib/wlroots0.16/template.py
```diff
@@ -27,18 +27,19 @@
...
 pkgdesc = "Modular Wayland compositor library"
```
q66, line 31: pkgdesc for both should say which branch it is (i.e. `(0.16.x)` at the end)

contrib/wlroots0.16/template.py
nekopsykose: sure, added. the replaces= also wasn't necessary since the files don't conflict with the old ones

PR #1472, opened 2024-02-21
contrib/chromium: update to 122.0.6261.57

nekopsykose:
[ci skip]  

still have to test it


nekopsykose:
works, also builds on ppc

PR #1481, opened 2024-02-22
contrib/fio: fix cfi

mikoxyz:


mikoxyz:
> Can't you have all these patches in one file or something?

yeah, i could, but i don't wanna

PR #1483, opened 2024-02-22
contrib/passt: new package (2024.02.20)

nekopsykose:


sbrivio-rh:
Hi, I just spotted this, thanks for making this package!

I saw the patches too, and I think it would be great to keep upstream portable by fixing those issues upstream. The issues you hit on Chimera are not clear to me though and I can't just use `gsed`, of course. Could you quickly describe them for me if you still vaguely remember them? Thanks.

q66:
hi @sbrivio-rh, i checked the behavior with both implementations of sed now and it seems to come out the same with the bsd version; there are still issues with `-i`, i have reduced the patch, see: 7cf0dddfadd251c932bf9c31538103d81c6a15bb

q66:
(i have verified both the standard output and the seccomp.h come out identical)

sbrivio-rh:
> hi @sbrivio-rh, i checked the behavior with both implementations of sed now and it seems to come out the same with the bsd version; there are still issues with `-i`, i have reduced the patch, see: [7cf0ddd](https://github.com/chimera-linux/cports/commit/7cf0dddfadd251c932bf9c31538103d81c6a15bb)

Ah, thanks, that makes sense, on BSDs it's `-i extension` while GNU sed has `-i[PREFIX]`.

Do you have a clue about the other patches/failures too?

q66:
the `cp-d.patch` could possibly be used upstream as is (`-P` is supported by all implementations with the same meaning)

the `basename.patch` can likewise be applied; GNU has two implementations with different semantics, the POSIX one is in `libgen.h` and the GNU one is in `string.h` where the GNU one is guaranteed not to modify its arguments, however, you are calling it on a mutable string so it's ok to just include `libgen.h` (musl recently dropped the GNU definition, as it was incorrect and was calling the POSIX version regardless, as having both would require some kind of strange dispatch mechanism and/or asm symbol redirects)

the fmt patch i don't know, this is just a difference in arguments between the BSD and GNU versions of it (`-t` in GNU seems to have the same meaning as `-p` in BSD)

sbrivio-rh:
Thanks a lot! I'll look into the `fmt` thing at some point and post patches then.

sbrivio-rh:
@nekopsykose, I just applied [this patch](https://archives.passt.top/passt-dev/20240511154555.22176-1-lemmi@nerd2nerd.org/) for `basename()`, so you can drop (well, it will conflict anyway) `basename.patch` at the next upstream release.

nekopsykose:
@sbrivio-rh thanks! i'll drop it next release then :)

PR #1485, opened 2024-02-23
contrib/hx: new package (1.0.14)

ttyyls:


contrib/hx/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "hx"
+pkgver = "1.0.14"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_build_args = [
+    f"hx_git_hash=chimera-{pkgver}-r{pkgrel}",
+    "hx_version=1.0.14",
```
mhmdanas, line 8:
```suggestion
    f"hx_version={pkgver}",
```

contrib/hx/template.py
```diff
@@ -0,0 +1,25 @@
...
+sha256 = "7b7c494df149535f768b3aa449159928aec92b468a17f9eedebcea3dd21343d1"
+hardening = ["vis", "cfi"]
+# no tests defined
+options = ["!check"]
+
+
+def do_install(self):
+    self.install_bin("hx", name="hxed")
+    self.install_man("hx.1")
+    self.install_license("LICENSE")
```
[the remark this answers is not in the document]
ttyyls: > Why not use the `install` make target?

name conflict and dubiouscompression of the manpage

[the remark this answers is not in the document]
mhmdanas: You can patch it though.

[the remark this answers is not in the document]
ttyyls: sure but why add 4 lines diff and clutter things up, after all it's just an executable and a manpage.

nekopsykose:
in alpine it was renamed to hx-editor for instance

ttyyls:
> this conflicts in name with helix for /usr/bin. not sure what to do about that tho

I handle it in do_install by installing the binary as hxed. hx-editor seemed a bit too clunky

nekopsykose:
you know, i can't read sometimes

nekopsykose:
i would also rename the manpage though then, so it matches (good) and just in case helix adds one later

nekopsykose:
you need a patch for that

ttyyls:
> you need a patch for that

fine :P

contrib/hx/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "hx"
+pkgver = "1.0.14"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_build_args = [
+    f"hx_git_hash=chimera-{pkgver}-r{pkgrel}",
```
q66, line 7: pkgrel does not belong in stuff like that

contrib/hx/template.py
```diff
@@ -0,0 +1,25 @@
...
+    f"hx_version={pkgver}",
+]
+hostmakedepends = ["gmake"]
+pkgdesc = "Terminal Hex Editor"
+maintainer = "ttyyls <contact@behri.org>"
+license = "MIT"
+url = "https://github.com/krpors/hx"
+source = f"https://github.com/krpors/hx/archive/refs/tags/v{pkgver}.tar.gz"
```
q66, line 15: {url}

contrib/hx/template.py
```diff
@@ -0,0 +1,25 @@
...
+sha256 = "7b7c494df149535f768b3aa449159928aec92b468a17f9eedebcea3dd21343d1"
+hardening = ["vis", "cfi"]
+# no tests defined
+options = ["!check"]
+
+
+def do_install(self):
```
q66, line 22: use install target

q66:
cflags should be passed for linking and are not

q66:
according to CI build log they are not

contrib/hx/template.py
nekopsykose: the install target would still need to manually rename the files after (or patch the makefile), and also uncompress the precompressed manpage. if anything it's more effort than installing 3 files here

q66: the makefile needs patching regardless to fix other things so

nekopsykose:
> according to CI build log they are not  

this seems to be a gmake default rule limitation, since we pass `LDFLAGS=ldflags CFLAGS=cflags` to gmake, and the project is using the default `hx: $(objects)` rule only. gmake itself seems to only pass ldflags for links..

should still be fixed, but gmake itself also has a bug here for that it seems

contrib/hx/template.py
q66: no incompressing is necessary, you just install one instead of the other

q66:
the implicit rules should never be used

ttyyls:
install target was a bit of a pain but I think the current patch addresses all of the other raised issues.

PR #1489, opened 2024-02-24
contrib/gopls: fix update-check

triallax:
The latest release stopped being on the first page of the tags page on
GitHub, causing the update-check to fail. The GitHub API returns way more
tags than the website's first page, so use that.


contrib/gopls/update.py
```diff
@@ -1 +1,3 @@
-pattern = r"/tags/gopls/v([\d.]+)\.tar\.gz"
+url = "https://api.github.com/repos/golang/tools/git/refs/tags"
+pattern = r"refs/tags/gopls/v([\d.\w-]+)"
```
mhmdanas, line 2: I could add a double quote to the end of the regex to get rid of matching the `\w-` and the `ignore` below, but idk

mhmdanas: Also the `refs` isn't strictly necessary, can remove it if you want

PR #1494, opened 2024-02-25
main/python-pyyaml: enable tests

mikoxyz:


nekopsykose:
funnily it logs a warning to run pytest instead

mikoxyz:
On Sun Feb 25, 2024 at 7:11 PM CET, alice wrote:
> funnily it logs a warning to run pytest instead
huh, that doesn't seem right. they moved over to pytest in a98fd6088e81,
which was pushed a couple months after 6.0.1 was tagged.

nekopsykose:
ah yes  

guess this works for now

PR #1495, opened 2024-02-25
contrib/nihtest: enable tests

mikoxyz:


nekopsykose:
hmm, that's a neat way

PR #1502, opened 2024-02-26
main/glslang: FIXME for !check

mikoxyz:


nekopsykose:
would rather actually fix them or something than just update a fixme

nekopsykose:
you can fix the suboverflow instead

PR #1504, opened 2024-02-26
main/grilo-plugins: new package (0.3.16) + main/totem: depend on grilo-plugins

mikoxyz:


main/grilo-plugins/patches/0002-tests-disable-tracker3-tests.patch
```diff
@@ -0,0 +1,25 @@
+From 33e65b2f1ef0ad434cef80f70b870c38c4284ec9 Mon Sep 17 00:00:00 2001
+From: miko <mikoxyzzz@gmail.com>
+Date: Mon, 26 Feb 2024 21:31:15 +0100
+Subject: [PATCH 2/2] tests: disable tracker3 tests
+
+requires a running dbus
```
q66, line 6: can't use `make_check_wrapper = ["dbus-run-session"]`?

main/grilo-plugins/template.py
```diff
@@ -0,0 +1,47 @@
...
+    # TODO: gom
+    "-Denable-bookmarks=no",
+    "-Denable-thetvdb=no",
+    # TODO: libdmapsharing
+    "-Denable-dmap=no",
+    # TODO: libgdata
+    "-Denable-youtube=no",
+    # TODO: libmediaart
+    "-Denable-local-metadata=no",
+    # TODO: libsoup24
```
q66, line 15: that's not TODO, there will never be libsoup2

PR #1506, opened 2024-02-27
contrib/agate: new package (3.3.7)

ttyyls:


contrib/agate/files/agate
```diff
@@ -0,0 +1,11 @@
+# agate system service
+
+type            = bgprocess
```
nekopsykose, line 3: why is this bgprocess? it doesn't even detach or anything

ttyyls: it should be a process

PR #1514, opened 2024-02-28
main/cryptsetup: update to 2.7.1, main/cryptsetup-scripts: update to 2.7.0.1

mhmdanas:


main/cryptsetup-scripts/update.py
```diff
@@ -1,5 +1,5 @@
-pattern = r"cryptsetup\ Debian\ release\ 2:([\d.\-~rc]+)"
-ignore = ["*rc*", "*~*"]
+pattern = r"cryptsetup\ Debian\ release\ 2:([\d.\-~rc\+exp]+)"
```
mhmdanas, line 1: How about we match all letters maybe, or even just match the version in URLs like https://salsa.debian.org/cryptsetup-team/cryptsetup/-/tags/debian%2F2%252.7.1-1+exp?

nekopsykose: hm not sure what the benefit would be but maybe i'm too tired. does the existing check fail in some case?

mhmdanas: If you mean the original, it matches experimental versions (like the version in the link above).

nekopsykose: hm seems ok then, sure

PR #1517, opened 2024-02-28
contrib/i3status-rust: enable more features and manpage

ttyyls:


contrib/i3status-rust/template.py
```diff
@@ -1,13 +1,18 @@
 pkgname = "i3status-rust"
 pkgver = "0.33.0"
-pkgrel = 0
+pkgrel = 1
 build_style = "cargo"
-make_build_args = ["--features=notmuch"]
+make_build_args = [
+    "--no-default-features",
+    "--features=notmuch,pipewire,pulseaudio,icu_calendar",
```
nekopsykose, line 7: why is pulseaudio here if you removed libpulse-devel

contrib/i3status-rust/template.py
```diff
@@ -16,9 +21,16 @@
 maintainer = "ttyyls <contact@behri.org>"
 license = "GPL-3.0-only"
 url = "https://github.com/greshake/i3status-rust"
-source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
-sha256 = "fd722f630080ef0b25558bec9b342a9fe8842c3af049d55a1370c6760ad84c67"
+source = [
+    f"{url}/archive/refs/tags/v{pkgver}.tar.gz",
+    f"!https://dev.alpinelinux.org/archive/i3status-rs/i3status-rs-{pkgver}.1",
```
nekopsykose, line 26: this is not guaranteed to be stable or exist

contrib/i3status-rust/template.py
nekopsykose: it's also missing maildir

contrib/i3status-rust/template.py
ttyyls: it gets updated when alpine updates their i3status package, assuming we do it in sync it should never cause any issues.

contrib/i3status-rust/template.py
ttyyls: enabling libpulse is the default, when disabled it breaks the audio source block so I chose to keep it

ttyyls: as for the maildir I just didn't see anything in the ports tree that could provide it so it seemed superfluous.

nekopsykose: you should keep libpulse-devel too then  

> as for the maildir I just didn't see anything in the ports tree that could provide it  

mbsync is one, some others can expose one too iirc

contrib/i3status-rust/template.py
ttyyls: other solutions could be to  

1. mirror it
2. build it ourselves using a similar setup
3. to simply drop it since a native build will be enabled anyway once a pandoc port is merged.

nekopsykose: i know, i'm the one that did that in the first place  

i would rather not rely on waiting for it tho. if upstreams don't want to ship a manpage without 'needing pandoc' then having no manpage at all is fine

contrib/i3status-rust/template.py
ttyyls: ok.

contrib/i3status-rust/template.py
ttyyls: just dropped it.

nekopsykose: > to simply drop it since a native build will be enabled anyway once a pandoc port is merged.  

most likely it wouldn't be, because pandoc would be limited to where ghc is which is just a few architectures. actually having it everywhere is no chance

contrib/i3status-rust/template.py
```diff
@@ -1,13 +1,19 @@
...
+make_check_args = list(make_build_args)
 hostmakedepends = ["cargo", "pkgconf"]
 makedepends = [
     "dbus-devel",
     "libcurl-devel",
     "libpulse-devel",
+    "pipewire-devel",
     "libsensors-devel",
     "notmuch-devel",
     "openssl-devel",
```
nekopsykose, line 19:
```suggestion
    "libsensors-devel",
    "notmuch-devel",
    "openssl-devel",
    "pipewire-devel",
```

ttyyls: mb

PR #1520, opened 2024-02-29
Fix adding printers in GNOME control center

mhmdanas:
- **main/python-pycups: new package (2.0.1)**
- **main/system-config-printer: new package (1.5.18)**
- **main/gnome-control-center: add depenency on system-config-printer**

### DIFF
--- a/main/python-pycups/template.py
+++ b/main/python-pycups/template.py
@@ -1,0 +1,20 @@
+pkgname = "python-pycups"
+pkgver = "2.0.1"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+    "python-wheel",
+]
+makedepends = ["cups-devel", "python-devel"]
+depends = ["python"]
+pkgdesc = "Python bindings for CUPS"
+maintainer = "triallax <triallax@tutanota.com>"
+license = "GPL-2.0-or-later"
+url = "https://github.com/OpenPrinting/pycups"
+source = f"$(PYPI_SITE)/p/pycups/pycups-{pkgver}.tar.gz"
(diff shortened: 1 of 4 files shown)


main/python-pycups/template.py
```diff
@@ -0,0 +1,20 @@
...
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+    "python-wheel",
+]
+makedepends = ["cups-devel", "python-devel"]
+depends = ["cups", "python"]
```
q66, line 12: should be cups-libs here

q66: or maybe not even that, it should get autodetected via shlibs

q66: and it does:

```
=> python-pycups-2.0.1-r0: running pre_pkg hook: 001_runtime_deps...
   SONAME: libcups.so.2 <-> cups-libs
```

mhmdanas: I'm already aware that `cups-libs` is autodetected as an shlib dependency, but I added a `cups` dependency because [the README states](https://github.com/OpenPrinting/pycups/blob/3a661e382d6ab8404c9d9327bf71e85abe154f42/README):

> CUPS needs to be installed and running for PyCUPS working correctly.

Though I _think_ it'd work regardless? I can still remove it if you think that's the right thing to do.

q66: looking at other distros, they don't have it, and it does not make sense to have it considering this is API bindings to libcups and doing this will make installing gnome pull in an entire cups server (which should be user choice)

q66: e.g. https://packages.debian.org/sid/python3-cups

PR #1522, opened 2024-03-01
contrib/chafa: new package (1.14.0)

ttyyls:


contrib/chafa/template.py
```diff
@@ -0,0 +1,48 @@
+pkgname = "chafa"
+pkgver = "1.14.0"
+pkgrel = 0
+build_style = "configure"
```
nekopsykose, line 4: this is gnu_configure

contrib/chafa/template.py
```diff
@@ -0,0 +1,48 @@
...
+sha256 = "faefc09807c955c882fa044947fc939952862f6e8d2e242262dbcc9f97e108ef"
+# FIXME: cfi
+hardening = ["vis", "!cfi"]
+
+
+def post_install(self):
+    self.install_license("COPYING")
+    self.install_license("COPYING.LESSER")
+
+
```
nekopsykose, line 45:
```suggestion
```

contrib/chafa/template.py
ttyyls: no, doing so trigger a secondary configure and fails.

contrib/chafa/template.py
ttyyls: ok.

contrib/chafa/template.py
ttyyls: autogen.sh calls configure at the very with all extra args https://github.com/hpjansson/chafa/blob/b77c28c85bcae5f7c5f42f1c820ac73393152865/autogen.sh#L110
doing a gnu_configure where the configure is skipped would boil down to the same thing

nekopsykose: > autogen.sh calls configure at the very end with all extra args

the thing you linked to has a `NOCONFIGURE` right there

> where the configure is skipped would boil down to the same thing  

except for the heaps of extra options that gnu_configure passes compared to configure

ttyyls: fair enough

contrib/chafa/patches/0001-disable-downloads-in-build-step.patch
```diff
@@ -0,0 +1,65 @@
...
+diff --git a/docs/Makefile.am b/docs/Makefile.am
+index fa8ee8b..e07ba79 100644
+--- a/docs/Makefile.am
++++ b/docs/Makefile.am
+@@ -70,10 +70,10 @@ XSLTPROC_FLAGS = \
+         --stringparam html.stylesheet manpage.css
+ 
+ .xml.1:
+-	$(AM_V_GEN) $(XSLTPROC) $(XSLTPROC_FLAGS) http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl $<
++	$(AM_V_GEN) $(XSLTPROC) $(XSLTPROC_FLAGS) /usr/share/xsl-nons/docbook/manpages/docbook.xsl $<
```
nekopsykose, line 55: this doesn't do anything, xsltproc looks up the path in filesystem from the url

nekopsykose: the reason the configure check above fails is because it silently fails from missing `xmlcatalog`, which is in `libxml2-progs`

contrib/chafa/template.py
```diff
@@ -0,0 +1,43 @@
...
+build_style = "gnu_configure"
+configure_args = [
+    "--enable-man",
+    "--without-imagemagick",
+]
+configure_env = {"NOCONFIGURE": "1"}
+configure_gen = ["./autogen.sh"]
+make_cmd = "gmake"
+hostmakedepends = [
+    "autoconf",
```
nekopsykose, line 13:
```suggestion
```

contrib/chafa/patches/0001-disable-downloads-in-build-step.patch
nekopsykose: maybe the -progs should be an xsl dependency already tbh

contrib/chafa/template.py
```diff
@@ -0,0 +1,43 @@
...
+    "libtiff-devel",
+    "libwebp-devel",
+]
+pkgdesc = "Character art facsimile generator"
+maintainer = "ttyyls <contact@behri.org>"
+license = "LGPL-3.0-or-later AND GPL-3.0-or-later"
+url = "https://hpjansson.org/chafa"
+source = f"https://github.com/hpjansson/chafa/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "faefc09807c955c882fa044947fc939952862f6e8d2e242262dbcc9f97e108ef"
+# FIXME: cfi
```
q66, line 37: glib application, don't bother

PR #1525, opened 2024-03-01
`main/musl` `main/musl-mallocng` `main/musl-cross`: use https instead of http for musl source

sidkshatriya:
main/musl main/musl-mallocng main/musl-cross: use https instead of http for musl source


nekopsykose:
`main/musl*: ...`

sidkshatriya:
@nekopsykose done ! :-)

PR #1536, opened 2024-03-05
main/gstreamer: update to 1.24.0

nekopsykose:


q66:
move from contrib should come with pkgrel bump

PR #1538, opened 2024-03-05
user/limine: new package (7.0.5)

nekopsykose:


q66:
why user?

nekopsykose:
dunno, just felt like it

q66:
i guess considering the lack of integration with our kernel hooks and whatnot it's the right call for now

nekopsykose:
it would never have any kernel hook integration since there's no config inclusion or config generator

q66:
how does it know about the kernels and their command lines and stuff then?

nekopsykose:
> about the kernels   

you write them into the config. currently this means you would manually need to adjust per version update (until globbing is added in the future), but that's ok  

> and their command lines  

you also write them into the config the same way. the config has the commandline and everything in it, see https://github.com/limine-bootloader/limine/blob/8ba483bb1611b0e74beb7bff74f99e3eabd8b3a1/test/limine.cfg for an example

q66:
how is that any different from generating grub.cfg then

nekopsykose:
grub-mkconfig and various grub config parts are shipped by upstream, and here nothing is shipped. i have zero interest in having a custom configuration generator downstream that does special stuff that is not 'there is one config file you manually write', which is what every limine user currently does

q66:
that you don't have interest in it does not mean there can "never be one"

nekopsykose:
i'm not saying it's technically impossible to have one. i'm saying that nobody else provides one and that's what everyone using this already expects (just writing one file themselves manually without distro-provided generators on top), because upstream doesn't provide one (or any conveniences like including config snippets), so i have no plans to add any such extra stuff. this is complete as-is.  

if for some reason you want to use this bootloader yourself and add a generator on top (with a way to disable it so it doesn't nuke the config file someone is manually using), go for it, but that is unrelated to this merge request

PR #1549, opened 2024-03-08
user/got: new package (0.96)

ttyyls:


user/got/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "libmd-devel",
+    "libbsd-devel",
+    "libevent-devel",
+    "linux-headers",
+    "musl-devel",
+    "ncurses-devel",
+    "libuuid-devel",
+    "zlib-devel",
+]
+depends = ["git", "outils"]
```
nekopsykose, line 24: it actually depends on git and outils?

ttyyls: both are soft dependencies. jot(1) from outils is listed in the readme as a dependency https://got.gameoftrees.org/?action=blob&commit=be9db6c2d00fb3a74f65c0ce4dc01359aa1986fc&file=README.portable&folder=&path=got-portable.git#line68  but from what I can tell isn't needed for basic functionality , and while got itself does not call for git, gotd (the cgi daemon) does hard depend on utilities in /usr/libexec/git-core

ttyyls: actually jot is a test dependency which is skipped anyway, and I chose not to build the gitwrapper which calls the git-receive-pack and git-upload-pack since we don't have a default git-daemon config/service. so it should be fine with neither of them

PR #1558, opened 2024-03-10
main/grub: `set -x` in grub-mkconfig

mikoxyz:


q66:
i don't see how set -x is supposed to be anyhow correct here (its purpose is to print out every command as it's run)

also i don't think it's enough to just do this regardless (other things in the grub hooks rely on the values that may error out, and will result in wrong behavior when unset)

so i don't think messing with the set -e is ever a good solution here

mikoxyz:
> i don't see how set -x is supposed to be anyhow correct here (its purpose is to print out every command as it's run)
> 
> also i don't think it's enough to just do this regardless (other things in the grub hooks rely on the values that may error out, and will result in wrong behavior when unset)
> 
> so i don't think messing with the set -e is ever a good solution here

hm, any suggestions on how to go about fixing this in a more proper way?

PR #1559, opened 2024-03-10
contrib/certbot: new package (2.10.0)

a16bitsysop:
and deps python-configargparse, python-distro

### DIFF
--- a/contrib/certbot/files/tmpfiles.conf
+++ b/contrib/certbot/files/tmpfiles.conf
@@ -1,0 +1,3 @@
+d /etc/letsencrypt
+d /var/lib/letsencrypt
+d /var/log/letsencrypt 0700
--- a/contrib/certbot/template.py
+++ b/contrib/certbot/template.py
@@ -1,0 +1,35 @@
+pkgname = "certbot"
+pkgver = "2.10.0"
+pkgrel = 0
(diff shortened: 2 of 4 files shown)


q66:
please split it up into batches of like max 5 templates, otherwise it's too much to take in for review

a16bitsysop:
Okay will try and split to related batches

a16bitsysop:
@q66 the other deps have been merged now

contrib/certbot/files/tmpfiles.conf
```diff
@@ -0,0 +1,3 @@
+d /etc/letsencrypt
+d /var/lib/letsencrypt
+d /var/log/letsencrypt 0700
```
nekopsykose, line 3: shouldn't certbot also have its own user and whatnot

nekopsykose: though maybe not, idk

a16bitsysop: the tmpfiles.conf is from arch linux which doesn't have a certbot user, neither does alpine.  I think in normal use it is creating certificates for things run as lots of different users.

nekopsykose: sure, makes sense

PR #1561, opened 2024-03-11
user/vgmplay: new package (0.51.1)

z-erica:


q66:
let's draft it for now then maybe?

q66:
this is probably because you are not compiling any audio output driver into libvgm, there is a few of them https://github.com/ValleyBell/libvgm/tree/master/audio

what we want is most likely pulse, the others don't seem relevant

PR #1568, opened 2024-03-11
main/cryptsetup-scripts: update to 2.7.1.1

mhmdanas:


q66:
you do realize there are 0 changes in this

mhmdanas:
https://salsa.debian.org/cryptsetup-team/cryptsetup/-/blob/debian/latest/debian/changelog:
>   * d/functions: get_mnt_devno(): Speed up execution time on large
>    /proc/mounts.

But fair, nothing too important.

PR #1573, opened 2024-03-11
contrib/dbus-test-runner: new package (19.04.0)

anixedn:
Required by https://github.com/chimera-linux/cports/pull/1385
I disabled checks for now, because i can't really figure out, why does it fail
Here is the log with results https://pastebin.com/ZWk3pGCQ

### DIFF
--- a/contrib/dbus-test-runner/template.py
+++ b/contrib/dbus-test-runner/template.py
@@ -1,0 +1,30 @@
+pkgname = "dbus-test-runner"
+pkgver = "19.04.0"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--disable-static"]
+make_cmd = "gmake"
+hostmakedepends = [
+    "automake",
+    "gettext-devel",
+    "gmake",
+    "intltool",
+    "libtool",
+    "pkgconf",
+]
+makedepends = [
+    "glib-devel",
+]
+depends = ["dbus"]
+checkdepends = ["bash", "dbus", "python-dbusmock"]
+pkgdesc = "Test executables under a new DBus session"
+maintainer = "avgwst <avgwst@tutanota.de>"
+license = "GPL-3.0-only"
+url = "https://launchpad.net/dbus-test-runner"
+source = f"{url}/{pkgver[:pkgver.rfind('.')]}/{pkgver}/+download/{pkgname}-{pkgver}.tar.gz"
+sha256 = "645a32fbd909baf2c01438f0cbda29dc9cd01a7aba5504c45610d88e8a57cb76"
+
+
+@subpackage("dbus-test-runner-devel")
+def _devel(self):
+    return self.default_devel()
--- a/contrib/dbus-test-runner-devel
+++ b/contrib/dbus-test-runner-devel
@@ -1,0 +1,1 @@
+dbus-test-runner
--- a/contrib/dbus-test-runner/patches/ignore-test-output.patch
+++ b/contrib/dbus-test-runner/patches/ignore-test-output.patch
@@ -1,0 +1,13 @@
+diff --git a/tests/Makefile.am b/tests/Makefile.am
+index 4f4aad9..275f809 100644
+--- a/tests/Makefile.am
++++ b/tests/Makefile.am
+@@ -154,7 +154,7 @@ test-no-tasks: Makefile.am
+ 	@chmod +x $@
+ XFAIL_TESTS += test-no-tasks
+ 
+-TESTS += test-output
++#TESTS += test-output
+ test-output: Makefile.am
+ 	@echo "#!/bin/sh -e" > $@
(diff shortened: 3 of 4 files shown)


q66:
i see only one FAIL and that's test-output, which fails because of gnu assumptions in head (normally it's not allowed to take a negative number): https://github.com/ubports/dbus-test-runner/blob/86d63d119566974bd841cacd3202599b3b1a845d/tests/Makefile.am#L162

the purpose is to cut off the last line, which you can do with `sed '$d'`

contrib/dbus-test-runner/template.py
```diff
@@ -0,0 +1,35 @@
...
+]
+makedepends = [
+    "dbus",
+    "dbus-glib-devel",
+    "gettext-devel",
+    "glib-devel",
+    "glibmm-devel",
+]
+checkdepends = ["bash", "python-dbusmock"]
+pkgdesc = (
```
q66, line 23: this is too long, something like `Run executables under a new DBus session` is enough

contrib/dbus-test-runner/template.py
```diff
@@ -0,0 +1,35 @@
...
+hostmakedepends = [
+    "autoconf",
+    "automake",
+    "gmake",
+    "intltool",
+    "libtool",
+    "pkgconf",
+]
+makedepends = [
+    "dbus",
```
q66, line 16: dbus should go in `depends` and `checkdepends`

contrib/dbus-test-runner/template.py
```diff
@@ -0,0 +1,35 @@
...
+    "dbus-glib-devel",
+    "gettext-devel",
+    "glib-devel",
+    "glibmm-devel",
```
q66, line 20: i don't think glibmm is necessary at all

contrib/dbus-test-runner/template.py
```diff
@@ -0,0 +1,35 @@
...
+    "automake",
+    "gmake",
+    "intltool",
+    "libtool",
+    "pkgconf",
+]
+makedepends = [
+    "dbus",
+    "dbus-glib-devel",
+    "gettext-devel",
```
q66, line 18: gettext is not necessary either

contrib/dbus-test-runner/template.py
```diff
@@ -0,0 +1,35 @@
...
+    "glib-devel",
+    "glibmm-devel",
+]
+checkdepends = ["bash", "python-dbusmock"]
+pkgdesc = (
+    "Little utility to test a couple of executables under a new DBus session"
+)
+maintainer = "avgwst <avgwst@tutanota.de>"
+license = "GPL-3.0-or-later"
```
q66, line 27: this is GPL-3.0-only

contrib/dbus-test-runner/template.py
```diff
@@ -0,0 +1,35 @@
...
+    "autoconf",
+    "automake",
+    "gmake",
+    "intltool",
+    "libtool",
+    "pkgconf",
+]
+makedepends = [
+    "dbus",
+    "dbus-glib-devel",
```
q66, line 17: dbus-glib is seemingly not even used, you just need to patch it out of configure.ac

contrib/dbus-test-runner/patches/fix-test-output.patch
```diff
@@ -0,0 +1,27 @@
+diff --git a/tests/Makefile.am b/tests/Makefile.am
+index 4f4aad9..5e2d0b7 100644
+--- a/tests/Makefile.am
++++ b/tests/Makefile.am
+@@ -159,17 +159,17 @@ test-output: Makefile.am
+ 	@echo "#!/bin/sh -e" > $@
+ 	@echo "$(DBUS_RUNNER) --task cat --parameter \"$(top_srcdir)/src/dbus-test-runner.c\" --task-name \"cat1\" --task cat --parameter \"$(top_srcdir)/src/dbus-test-runner.c\" --task-name \"cat2\" > testcat.output.txt" >> $@
```
q66, line 7: the test-output test is flaking here, often cat2 does not manage to print out anything (i guess it races)

maybe best to just disable test-output from the suite

contrib/dbus-test-runner/template.py
anixedn: should i also patch for not using gettext? configure requires gettext-devel

contrib/dbus-test-runner/template.py
```diff
@@ -0,0 +1,30 @@
...
+    "intltool",
+    "libtool",
+    "pkgconf",
+]
+makedepends = [
+    "gettext-devel",
+    "glib-devel",
+]
+depends = ["dbus"]
+checkdepends = ["dbus", "bash", "python-dbusmock"]
```
q66, line 19: sort

contrib/dbus-test-runner/template.py
q66: i guess leave it, but move it to hostmakedepends

PR #1574, opened 2024-03-11
contrib/qt6-qtmultimedia: new package (6.6.2)

JamiKettunen:
Fixes a bunch of the tests compared to the pkg of this included in #1391 and #1424.

Btw would a more generic update-check default for all these Qt6 submodules make sense (so the `update.py`'s can be dropped)? I see at least the multiple `url`s being a potential issue since everything else defines just one atm in [src/cbuild/core/update_check.py](https://github.com/chimera-linux/cports/blob/master/src/cbuild/core/update_check.py)

### DIFF
--- a/contrib/qt6-qtmultimedia-devel
+++ b/contrib/qt6-qtmultimedia-devel
@@ -1,0 +1,1 @@
+qt6-qtmultimedia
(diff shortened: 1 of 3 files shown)


nekopsykose:
>  since everything else defines just one atm  

implicitly (with no update.py). both the url= and source= urls are used for version checks (with multiple source=[] all of them, even). so technically this is completely fine

PR #1577, opened 2024-03-13
contrib/thermald: enable cfi

mikoxyz:
cfi-icall is disabled for parse_new_cdev due to some fuckery with xmlFree. everything else seems fine, though


q66:
no cfi on glib packages by policy

mikoxyz:
> no cfi on glib packages by policy

ah fuck, sorry, i forgot this used glib

PR #1579, opened 2024-03-13
contrib/nautilus-python: new package (4.0)

valpackett:
Required e.g. for [nautilus-taildrop](https://github.com/bahorn/nautilus-taildrop)

(explicit commit due to [tiny build patch right after the 4.0 tag](https://gitlab.gnome.org/GNOME/nautilus-python/-/commit/9a74cae1024aa279d2d70af2bf298cd1b57c1997))

### DIFF
--- a/contrib/nautilus-python/template.py
+++ b/contrib/nautilus-python/template.py
@@ -1,0 +1,23 @@
+pkgname = "nautilus-python"
+pkgver = "4.0"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+]
+makedepends = [
+    "glib-devel",
+    "nautilus-devel",
+    "python-devel",
+    "python-gobject-devel",
+]
+depends = [
+    "python-gobject",
+]
+pkgdesc = "Python plugin support for Nautilus"
+maintainer = "Val Packett <val@packett.cool>"
+license = "GPL-2.0-or-later"
+url = "https://gitlab.gnome.org/GNOME/nautilus-python"
+source = f"{url}/-/archive/{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "2f85820d5033383271032169b1fbf7ee71d495a4562f6f1912849b12d5c9f9cc"


nekopsykose:
> explicit commit due to

if you use an explicit commit that is not the real 4.0, then it should be `4.0_git...`. but in this case you should use the real 4.0 and put the patch in patches/

(yes, quite pedantic, i know..)

contrib/nautilus-python/template.py
```diff
@@ -0,0 +1,27 @@
...
+pkgdesc = "Python plugin support for Nautilus"
+maintainer = "Val Packett <val@packett.cool>"
+license = "GPL-2.0-or-later"
+url = "https://gitlab.gnome.org/GNOME/nautilus-python"
+source = f"{url}/-/archive/9a74cae1024aa279d2d70af2bf298cd1b57c1997/{pkgname}-9a74cae1024aa279d2d70af2bf298cd1b57c1997.tar.gz"
+sha256 = "0a78fede1d6de7d9a06bb28abc3090b1aef6ae9776c41b195a381270f1d99558"
+
+
+def post_install(self):
+    self.install_license("COPYING")
```
nekopsykose, line 27: don't have to install gpl

contrib/nautilus-python/template.py
```diff
@@ -0,0 +1,27 @@
...
+build_style = "meson"
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+]
+makedepends = [
+    "nautilus-devel",
+    "glib-devel",
+    "python-devel",
+    "python-gobject-devel",
```
nekopsykose, line 13:
```suggestion
    "glib-devel",
    "nautilus-devel",
    "python-devel",
    "python-gobject-devel",
```

sort

valpackett:
thankfully builds w/o patch so now using just 4.0 yay

nekopsykose:
lucky! :P

q66:
> > explicit commit due to
> 
> if you use an explicit commit that is not the real 4.0, then it should be `4.0_git...`. but in this case you should use the real 4.0 and put the patch in patches/
> 
> (yes, quite pedantic, i know..)

for these kind of cases i prefer a tag + patch

nekopsykose:
that is what i suggested yea

PR #1581, opened 2024-03-14
contrib/goattracker: new package (2.76)

z-erica:
includes a package for https://github.com/libsdl-org/sdl12-compat, since the program is sdl1.2 only

### DIFF
--- a/contrib/goattracker/files/goattracker.desktop
+++ b/contrib/goattracker/files/goattracker.desktop
@@ -1,0 +1,7 @@
+[Desktop Entry]
+Type=Application
+Name=GoatTracker
+Comment=C64 music editor
+Exec=goattracker %F
+Icon=goattracker
+Categories=AudioVideo;AudioVideoEditing
--- a/contrib/goattracker/patches/fix_makefile.patch
+++ b/contrib/goattracker/patches/fix_makefile.patch
@@ -1,0 +1,45 @@
+--- a/src/makefile.common	2021-05-15 14:59:32.000000000 +0200
++++ b/src/makefile.common	2024-03-14 11:31:54.572719809 +0100
+@@ -2,8 +2,7 @@
+ 
+ CC?=gcc
+ CXX?=g++
+-CFLAGS+=-Ibme -Iasm -O3
+-CXXFLAGS+=$(CFLAGS) -fpermissive
++CFLAGS+=-Ibme -Iasm
+ 
+ EXE=	$(PREFIX)goattrk2$(SUFFIX) \
+ 	$(PREFIX)gt2reloc$(SUFFIX) \
+@@ -23,7 +22,7 @@
+ asm/asmtab.o asm/chnkpool.o asm/expr.o asm/lexyy.o asm/log.o asm/membuf.o asm/membufio.o asm/namedbuf.o asm/parse.o \
+ asm/pc.o asm/vec.o \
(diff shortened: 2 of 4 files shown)


contrib/goattracker/template.py
```diff
@@ -0,0 +1,36 @@
...
+    self.install_file("goat_tracker_commands.pdf", "usr/share/doc/goattracker")
+    self.install_file("readme.txt", "usr/share/doc/goattracker")
+    self.install_file(
+        "examples/*.sng", "usr/share/doc/goattracker/examples", glob=True
+    )
+    self.install_file(
+        self.files_path / "goattracker.desktop", "usr/share/applications"
+    )
+    self.do("convert", "src/goattrk2.bmp", "goattracker.png")
+    self.install_file("goattracker.png", "usr/share/pixmaps")
```
nekopsykose, line 36: icons go to /usr/share/icons/..., not pixmaps

contrib/goattracker/files/goattracker.desktop
```diff
@@ -0,0 +1,9 @@
+[Desktop Entry]
+Type=Application
+Version=1.0
+Name=GoatTracker
+Comment=C64 music editor
+Exec=goattracker %F
+Icon=goattracker.png
```
nekopsykose, line 7: and following the below, you don't need the .png extension here, just the bare name (extension is auto picked)

contrib/goattracker/files/goattracker.desktop
```diff
@@ -0,0 +1,9 @@
...
+Terminal=false
```
nekopsykose, line 8: terminal=false is implicit pretty sure, so you don't need that

contrib/goattracker/template.py
nekopsykose: in this case, this icon is 2851x2851. you probably want to resize it (so it fits into one of the sized folders), and maybe even generate multiple sizes?  

also probably want -quality 95 or something (the first 9 is the zlib compression level)

contrib/goattracker/patches/fix_makefile.patch
```diff
@@ -0,0 +1,45 @@
...
++CFLAGS+=-Ibme -Iasm
+ 
+ EXE=	$(PREFIX)goattrk2$(SUFFIX) \
+ 	$(PREFIX)gt2reloc$(SUFFIX) \
+@@ -23,7 +22,7 @@
+ asm/asmtab.o asm/chnkpool.o asm/expr.o asm/lexyy.o asm/log.o asm/membuf.o asm/membufio.o asm/namedbuf.o asm/parse.o \
+ asm/pc.o asm/vec.o \
+ bme/bme_gfx.o bme/bme_snd.o bme/bme_win.o bme/bme_mou.o bme/bme_kbd.o bme/bme_io.o bme/bme_end.o bme/bme.o
+-	$(CXX) -o $@ $^ $(LIBS)
++	$(CXX) $(LDFLAGS) -o $@ $^ $(LIBS)
```
nekopsykose, line 18: ah, these ones where you added ldflags also need c/xxflags in there too depending on which is used

contrib/goattracker/template.py
nekopsykose: wow yes i was blind, haha nvm

nekopsykose:
up to you, this looks ready to me

PR #1583, opened 2024-03-14
contrib/gtultra: new package (1.5.5)

z-erica:
note that this package conflicts with contrib/goattracker from https://github.com/chimera-linux/cports/pull/1581 by installing some binaries with the same names. unsure if i should/how to prevent that


nekopsykose:
just pick which version you think is way better or somethin

nekopsykose:
to package

PR #1585, opened 2024-03-14
using_git: don't recommend git checkout

jcgruenhage:
`git checkout` is the worst offender when it comes to bad git UX, and we
shouldn't teach it to newcomers. They should learn `git switch` and `git
restore` instead, which are way less confusing.

### DIFF
--- a/USING_GIT.md
+++ b/USING_GIT.md
@@ -82,19 +82,19 @@ In order to be able to send your changes as a pull request, you will need
 a branch for that pull request, and the difference over `master` is what the
 pull request will contain.
 
-### Using checkout
+### Using switch
 
-To create a new branch in the current tree and switch to it, you can use
-the `checkout` command like so:
+To create a new branch in the current tree and switch to it, you can use the
+`switch` command like so:
 
 ```
-$ git checkout -b my-custom-branch-name
+$ git switch -c my-custom-branch-name
 ```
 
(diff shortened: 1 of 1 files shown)


q66:
considering git switch is fairly new and the documentation (even in the latest version) contains a big "THIS COMMAND IS EXPERIMENTAL. THE BEHAVIOR MAY CHANGE.", i don't think that's right; besides, the checkout command is ubiquitous pretty much everywhere and it's good to be at least somewhat familiar with it either way

jcgruenhage:
I didn't know about the bit that it's marked experimental, but tbf: That warning has been there for the >3 years that this existed, and the behaviour hasn't changed meaningfully.

I've been using `git switch` for these 3 years and won't ever look back.

Either way though: Documenting both seems like a fair compromise, thanks for the quick turnaround!

q66:
checkout can only have bad UX if you use it in sketchy ways, but for simple branch switching/creating it's gotcha-free and there really isn't any meaningful difference between the commands; fwiw, i've yet to use switch (there is very little that would justify changing the habit)

jcgruenhage:
If you only use `git checkout` to switch branches, sure, but the fact that the functionalities of both `git restore` and `git switch` are included in `git checkout` makes it bad from a UX perspective even without any big gotchas. They are different use-cases that have been stuffed into the same command, and I find `git switch` and `git restore` a lot more intuitive to use.

PR #1589, opened 2024-03-15
main/elogind: update to 252.23

triallax:


q66:
every elogind release breaks about a billion things so let's give it some time

triallax:
can't bother updating to latest release, somebody else can pick it up

PR #1590, opened 2024-03-15
contrib/gitoxide: new package (0.34.0)

jcgruenhage:


contrib/gitoxide/template.py
```diff
@@ -0,0 +1,23 @@
...
+build_style = "cargo"
+_gitoxide_features = [
+    "--no-default-features",
+    "--features=max-control,gix-features/zlib-stock,gitoxide-core-blocking-client,http-client-curl",
+]
+make_build_args = _gitoxide_features
+make_install_args = _gitoxide_features
+make_check_args = _gitoxide_features
+hostmakedepends = ["cargo", "cmake", "pkgconf"]
+makedepends = ["rust-std", "openssl-devel", "zlib-devel", "sqlite-devel"]
```
nekopsykose, line 13: missing libcurl-devel for it to use system curl

it also never ends up linking to openssl or sqlite, so either that silently vendors or the installed binaries specifically just don't use it and other build-time stuff does. you probably want --bins to only build the actually installed things and not the entire gix project?

contrib/gitoxide/template.py
```diff
@@ -0,0 +1,23 @@
...
+pkgdesc = "Idiomatic, lean, fast & safe pure Rust implementation of Git"
```
nekopsykose, line 14:
```suggestion
pkgdesc = "Rust implementation of Git"
```

contrib/gitoxide/template.py
```diff
@@ -0,0 +1,23 @@
...
+pkgver = "0.34.0"
+pkgrel = 0
+build_style = "cargo"
+_gitoxide_features = [
+    "--no-default-features",
+    "--features=max-control,gix-features/zlib-stock,gitoxide-core-blocking-client,http-client-curl",
+]
+make_build_args = _gitoxide_features
+make_install_args = _gitoxide_features
+make_check_args = _gitoxide_features
```
nekopsykose, line 11:
```suggestion
make_build_args = [
    "--no-default-features",
    "--features=max-control,gix-features/zlib-stock,gitoxide-core-blocking-client,http-client-curl",
]
make_install_args = list(make_build_args)
make_check_args = list(make_install_args)
```

contrib/gitoxide/template.py
nekopsykose: ah, for sqlite it's because of  

```
rusqlite = { version = "0.30.0", optional = true, features = ["bundled"] }
```

nekopsykose: (this both vendors it, and breaks on ppc64le because sqlite has broken endianness checks with clang and always assumes the target is big endian)

nekopsykose: for openssl, the linker probably optimises out the link because the system libcurl links it instead, so nothing special there

q66: i think https://github.com/chimera-linux/cports/pull/1591 will fix it

q66:
```
=> gitoxide-0.34.0-r0: running pre_pkg hook: 001_runtime_deps...
   SONAME: libssl.so.3 <-> libssl3
   SONAME: libcrypto.so.3 <-> libcrypto3
   SONAME: libz.so.1 <-> zlib
   SONAME: libunwind.so.1 <-> libunwind
   SONAME: libc.so <-> musl
```

jcgruenhage: yeah, I verified locally with #1591 it does end up linking sqlite properly

jcgruenhage: regarding libcurl: I'll look into that, thanks for the pointer

contrib/gitoxide/template.py
```diff
@@ -0,0 +1,27 @@
...
+pkgver = "0.34.0"
+pkgrel = 0
+build_style = "cargo"
+make_build_args = [
+    "--no-default-features",
+    "--features=max-control,gix-features/zlib-stock,gitoxide-core-blocking-client,http-client-curl",
+]
+make_install_args = list(make_build_args)
+make_check_args = list(make_install_args)
+hostmakedepends = ["cargo", "cmake", "pkgconf"]
```
nekopsykose, line 11:
```suggestion
hostmakedepends = ["cargo", "pkgconf"]
```

not used anymore with the other stuff fixed

PR #1586, opened 2024-03-14
OpenPGP card tooling

jcgruenhage:
- **contrib/pcsc-ccid: new package (1.5.5)**: Needed for using the two tools below
- **contrib/openpgp-card-tools: new package (0.10.0)**: Tool for managing openpgp smart cards
- **contrib/openpgp-card-ssh-agent: new package (0.2.3)**: SSH agent for OpenPGP authentication subkeys on OpenPGP smart cards.

PR #1592, opened 2024-03-15
contrib/sequoia-sq: new package (0.34.0)

jcgruenhage:
Split out of #1586

### DIFF
--- a/contrib/capnproto/template.py
+++ b/contrib/capnproto/template.py
@@ -1,0 +1,32 @@
+pkgname = "capnproto"
+pkgver = "1.0.2"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DBUILD_SHARED_LIBS=ON"]
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+makedepends = [
+    "libucontext-devel",
+    "linux-headers",
+    "openssl-devel",
+    "zlib-devel",
+]
+pkgdesc = "Cap'n Proto serialization/RPC system"
+maintainer = "Jan Christian Grünhage <jan.christian@gruenhage.xyz>"
+license = "MIT"
+url = "https://capnproto.org"
+source = f"{url}/capnproto-c++-{pkgver}.tar.gz"
+sha256 = "9057dbc0223366b74bbeca33a05de164a229b0377927f1b7ef3828cdd8cb1d7e"
+
+
+if self.profile().cross:
+    hostmakedepends += ["capnproto-devel"]
+    configure_args += ["-DEXTERNAL_CAPNP=ON", "-DBUILD_TESTING=OFF"]
+
+
+def post_install(self):
+    self.install_license("LICENSE.txt")
+
+
+@subpackage("capnproto-devel")
+def _devel(self):
+    return self.default_devel(extra=["usr/bin"])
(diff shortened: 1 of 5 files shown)


contrib/capnproto/template.py
```diff
@@ -0,0 +1,26 @@
+pkgname = "capnproto"
+pkgver = "1.0.2"
+pkgrel = 0
+build_style = "gnu_configure"
```
nekopsykose, line 4: you seem to have dropped the cmake from before (which works fine)

contrib/capnproto/template.py
```diff
@@ -0,0 +1,26 @@
...
+hostmakedepends = ["automake", "libtool", "pkgconf"]
```
nekopsykose, line 5: need to pass  

```
```

otherwise they're all only static

contrib/capnproto/template.py
```diff
@@ -0,0 +1,26 @@
...
+maintainer = "Jan Christian Grünhage <jan.christian@gruenhage.xyz>"
+license = "MIT"
+url = "https://capnproto.org"
+source = f"{url}/capnproto-c++-{pkgver}.tar.gz"
+sha256 = "9057dbc0223366b74bbeca33a05de164a229b0377927f1b7ef3828cdd8cb1d7e"
+
+
+if self.profile().cross:
+    hostmakedepends += ["capnproto", "capnproto-devel"]
+    configure_args = ["--with-external-capnp"]
```
nekopsykose, line 17:
```suggestion
    hostmakedepends += ["capnproto-devel"]
    configure_args = ["-DEXTERNAL_CAPNP=ON", "-DBUILD_TESTING=OFF"]
```

after fixing the above this works

contrib/capnproto/template.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def post_install(self):
+    self.install_license("LICENSE.txt")
+
+
+@subpackage("capnproto-devel")
+def _devel(self):
+    return self.default_devel()
```
nekopsykose, line 26:
```suggestion
    return self.default_devel(extra="usr/bin")
```

contrib/capnproto/template.py
```diff
@@ -0,0 +1,26 @@
+pkgname = "capnproto"
+pkgver = "1.0.2"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = ["automake", "libtool", "pkgconf"]
+makedepends = ["linux-headers"]
```
nekopsykose, line 6: you probably also want zlib-devel, libucontext-devel, openssl-devel?

contrib/sequoia-sq/template.py
```diff
@@ -0,0 +1,28 @@
...
+pkgrel = 0
+build_style = "cargo"
+make_build_env = {"ASSET_OUT_DIR": "assets"}
+hostmakedepends = ["cargo", "pkgconf", "capnproto"]
+makedepends = [
+    "nettle-devel",
+    "openssl-devel",
+    "sqlite-devel",
+    "capnproto-devel",
+    "rust-std",
```
nekopsykose, line 12: sort

contrib/capnproto/template.py
nekopsykose: (since i'm pretty sure these are all makedep binaries to use along with -devel, and don't have to be next to the library)

PR #1593, opened 2024-03-15
main/pipewire,wireplumber: add log-type buffer for user services

nekopsykose:
i've had log-file = /home/me/whatever locally for a while to look at these logs, but this is probably a bit better. a small scrollback is usually enough to see if something weird was happenening


q66:
fold with #1617

PR #1598, opened 2024-03-16
Auditable rust binaries using cargo-auditable

jcgruenhage:
#### Relevant commits
- **main/cargo-auditable-bootstrap: new package (0.6.2)**: Needed so that it's possible to build cargo-auditable with cargo-auditable
- **cbuild: add option to build auditable rust binaries**: Add an option to enable auditable builds. This is turned on per default, but can be turned off in cases where it breaks compilation or for cargo-auditable-bootstrap.
- **main/cargo-auditable: new package (0.6.2)**: cargo-auditable itself, built with cargo-auditable-bootstrap
- **user/frawk: build with cargo-auditable and remove legacy .cargo/config**: this failed to build due to the cbuild change in #1646 
- **contrib/git-branchless: build with cargo-auditable**: This commit additionally disables checks. They were also broken on master, but I haven't looked deeper into why.

### Example
```
jcgruenhage@fru-mate-citro ~/d/g/c/cports (auditable-rust-binaries)> syft scan --override-default-catalogers all /bin/cargo-auditable
 ✔ Indexed file system                                                                                                                                                      /usr/bin
 ✔ Cataloged contents                                                                                               307bdf695ab0e91c48699fd05abd87c2d1e27da101162b9d32eb45d075a237a4
   ├── ✔ Packages                        [28 packages]  
   └── ✔ Executables                     [1 executables]  
NAME              VERSION  TYPE         
adler             1.0.2    rust-crate    
ahash             0.8.3    rust-crate    
auditable-serde   0.6.1    rust-crate    
camino            1.1.1    rust-crate    
cargo-auditable   0.6.2    rust-crate    
cargo-platform    0.1.2    rust-crate    
cargo_metadata    0.15.0   rust-crate    
cfg-if            1.0.0    rust-crate    
crc32fast         1.3.2    rust-crate    
hashbrown         0.12.3   rust-crate    
hashbrown         0.13.2   rust-crate    
indexmap          1.9.1    rust-crate    
itoa              1.0.3    rust-crate    
memchr            2.5.0    rust-crate    
miniz_oxide       0.6.2    rust-crate    
object            0.30.3   rust-crate    
once_cell         1.14.0   rust-crate    
pico-args         0.5.0    rust-crate    
proc-macro2       1.0.43   rust-crate    
quote             1.0.21   rust-crate    
ryu               1.0.11   rust-crate    
semver            1.0.14   rust-crate    
serde             1.0.147  rust-crate    
serde_derive      1.0.147  rust-crate    
serde_json        1.0.85   rust-crate    
syn               1.0.99   rust-crate    
topological-sort  0.2.2    rust-crate    
unicode-ident     1.0.4    rust-crate  
```

[ci skip]


nekopsykose:
> Needed so that it's possible to build cargo-auditable with cargo-auditable  

i disagree that it's needed, can also just exclude the one single thing

nekopsykose:
>  Add an option to enable auditable builds. This is opt-in for now.  

i also don't think there's a point for this. cargo auditable is a fully functional wrapper so it's possible to replace the regular use of cargo entirely in the build_style  

> contrib/syft: new package (1.0.1)  

this can be in a separate mr

nekopsykose:
> it's possible to replace the regular use of cargo entirely in the build_style  

in alpine i did exactly this (replacing every `cargo` with `cargo auditable`) and everything worked fine with no issues for reference, so i don't see the point of it being an option

nekopsykose:
(of course, still have to manually get cargo-auditable to build if the build_style forced it, so perhaps leaving the option is fine just to set it to =false on it)

jcgruenhage:
Doing it entirely in the build style isn't possible, as we need to at least replace all the `cargo` with `cargo-auditable` in the templates after switching to `cargo auditable` in the build style.

Aside of that: Having an easy way to turn this on or off is nice IMO, as there have been cases in the past where some rust projects weren't able to build with it.

As for `cargo-auditable-bootstrap`: That's *only* necessary, if we also want to ship an auditable version of `cargo-auditable`, which I think would be nice.

jcgruenhage:
Ah, and `syft`, sure, that can be a separate PR. Want me to split it out?

nekopsykose:
> Doing it entirely in the build style isn't possible, as we need to at least replace all the cargo with cargo-auditable in the templates after switching to cargo auditable in the build style.

those can be swapped too after. the point is that adding a `cargo_auditable = True` in every single template is not great, so just don't have to do it with making it the default

or maybe you think i meant something else?


> Ah, and syft, sure, that can be a separate PR. Want me to split it out?  

yeah

nekopsykose:
> Aside of that: Having an easy way to turn this on or off is nice IMO, as there have been cases in the past where some rust projects weren't able to build with it.  

yeah, you can leave the option with default true and then make it =false in cargo-auditable

jcgruenhage:
I do want to make it the default eventually - if you want me to make it the default now I'm fine with that.

jcgruenhage:
syft split out into https://github.com/chimera-linux/cports/pull/1600

nekopsykose:
> I do want to make it the default eventually - if you want me to make it the default now I'm fine with that.  

yeah, i'd do that. for cargo-auditable, i'd probably build it with itself from ./target in the same build (but this might be too fiddly and maybe q66 doesn't like it), so up to you

nekopsykose:
just drop the pkgrel bumps and this looks fine (just needs for q66 to look at the cbuild change)

we'll have a world rebuild some other time in a few months or whatnot, it can be incremental until then

jcgruenhage:
> just drop the pkgrel bumps and this looks fine (just needs for q66 to look at the cbuild change)

So I should include the changes to the packages, but *not* include the pkgrel bumps? Or should I strip this down to just include the cbuild changes and the two cargo auditable packages and completely skip the rest for now?

nekopsykose:
> So I should include the changes to the packages, but not include the pkgrel bumps?

yeah

the other changes are needed since the hostmakedepends change

user/frawk/patches/cargo-config.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/.cargo/config b/.cargo/config
```
nekopsykose, line 1: you both patch this and self.rm it (just the self.rm is fine)

nekopsykose:
also squash the `build with cargo-auditable` commits to `*/*:`

user/frawk/patches/cargo-config.patch
nekopsykose: hm, if you prefer it that way that's ok, in that case you can remove the self.rm since the patch is removing it

jcgruenhage: it is actually not: The patch is clearing out the file, but it remains as an empty file afterwards, which still breaks.

After thinking about it some more: If they end up adding something to the cargo config that is required, it still compiling but still being broken subtly is probably very unlikely, so I'm fine with having the patch removed.

nekopsykose: i think this is a quirk of bsd patch, that it can't remove files then (since it was deleted in the patch itself) :)  

yeah, sounds good. thanks for checking!

Shnatsel:
> in alpine i did exactly this (replacing every cargo with cargo auditable)

@nekopsykose do I understand it right that Alpine also builds its packages with `cargo auditable` by default now? As the author of `cargo auditable` I would very much like to know :sweat_smile:

jcgruenhage:
Using this (potentially flawed) method of checking how many packages in Alpine Linux use `cargo build` vs `cargo auditable build`, it looks like the default is the packages being auditable.

```
aports on  master 
❯ grep -r "cargo auditable build" -l | sort -u | wc -l
     401

aports on  master 
❯ grep -r "cargo build" -l | sort -u | wc -l
      10
```

PR #1599, opened 2024-03-16
contrib/copyq: update to 8.0.0

ttyyls:
this update also drops X11 support alongside its build dependencies, and turn off notifications support explicitely (previously implicitelt disabled) since it relies on knotifications.


nekopsykose:
> this updated also drops X11 support  

why?

ttyyls:
> > this updated also drops X11 support
> 
> why?

no particular reason, the option just landed in this update https://github.com/hluk/CopyQ/releases/tag/v8.0.0 and I thought of taking advantage of it to lighten the build.

nekopsykose:
> no particular reason,  

then don't

nekopsykose:
(we don't disable it anywhere else globally either, the only thing i can think of similarly is not adding alsa to things where pulse/pipewire exists)

ttyyls:
reverted

PR #1600, opened 2024-03-16
contrib/syft: new package (1.0.1)

jcgruenhage:


contrib/syft/template.py
```diff
@@ -0,0 +1,31 @@
+pkgname = "syft"
+pkgver = "1.0.1"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    f"-ldflags= -X main.version={pkgver}",
+    "github.com/anchore/syft/cmd/syft",
```
nekopsykose, line 7:
```suggestion
    "./cmd/syft",
```

local path seems to work ok

contrib/syft/template.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def post_build(self):
+    self.do("sh", "-c", "build/syft completion bash > syft.bash")
+    self.do("sh", "-c", "build/syft completion fish > syft.fish")
+    self.do("sh", "-c", "build/syft completion zsh > _syft")
+
+
+def do_install(self):
+    self.install_bin("build/syft")
```
nekopsykose, line 28:
```suggestion
def post_install(self):
```

contrib/syft/template.py
```diff
@@ -0,0 +1,30 @@
...
+url = "https://github.com/anchore/syft"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "79852b27a0665da4e41139066c009aaeae488e82fd6b465129aed5734c5ac934"
+# Test suite depends on docker
+# generates manpages/completions with host bins
+options = ["!check", "!cross"]
+
+
+def post_build(self):
+    self.do("sh", "-c", "build/syft completion bash > syft.bash")
```
q66, line 22:
```
with open(self.cwd / "syft.bash", "w") as outf:
    self.do("build/syft", "completion", "bash", stdout=outf)
```

etc

q66: you already did it right in other places, why do this here

nekopsykose: currently it's 50/50 between the two

nekopsykose: (and they both do the same thing so i have no preference either way)

q66: ??? this is the only template that does this

jcgruenhage: I'll make a follow-up PR for this

nekopsykose: > ??? this is the only template that does this  

contrib/yq, contrib/starship

don't ??? me

q66: that's still nowhere near 50/50 regardless and both of these were introduced by one person

jcgruenhage:
```diff
diff --git a/contrib/syft/template.py b/contrib/syft/template.py
index 48dd4133a..240c9fb7d 100644
--- a/contrib/syft/template.py
+++ b/contrib/syft/template.py
@@ -16,15 +16,19 @@ sha256 = "79852b27a0665da4e41139066c009aaeae488e82fd6b465129aed5734c5ac934"
 # Test suite depends on docker
 # generates manpages/completions with host bins
 options = ["!check", "!cross"]
+_completion_shells = {
+    "bash": "syft.bash",
+    "fish": "syft.fish",
+    "zsh": "_syft",
+}
 
 
 def post_build(self):
-    self.do("sh", "-c", "build/syft completion bash > syft.bash")
-    self.do("sh", "-c", "build/syft completion fish > syft.fish")
-    self.do("sh", "-c", "build/syft completion zsh > _syft")
+    for shell, comp_file in _completion_shells.items():
+        with open(self.cwd / comp_file, "w") as outf:
+            self.do("build/syft", "completion", shell, stdout=outf)
 
 
 def post_install(self):
-    self.install_completion("syft.bash", "bash")
-    self.install_completion("syft.fish", "fish")
-    self.install_completion("_syft", "zsh")
+    for shell, comp_file in _completion_shells.items():
+        self.install_completion(comp_file, shell)
```

Is that a sane way to do it?

jcgruenhage: ah well, you were quicker ^^

jcgruenhage: also: github showed me none of the replies of this thread until I explicitly reloaded, sorry if my messages are a bit out of context..

PR #1603, opened 2024-03-17
contrib/wlroots0.17: add xcb-util-errors dep

ifreund:
I needed this package for my own wayland compositor development purposes so I thought I might as well enable it for the packaged wlroots versions as well to give developers better error messages in possible bug reports.

### DIFF
--- a/main/xcb-util-errors/template.py
+++ b/main/xcb-util-errors/template.py
@@ -1,0 +1,26 @@
+pkgname = "xcb-util-errors"
+pkgver = "1.0.1"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = [
+    "automake",
+    "libtool",
+    "pkgconf",
+    "xorg-util-macros",
+]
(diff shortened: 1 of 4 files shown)


main/xcb-util-errors/template.py
```diff
@@ -0,0 +1,24 @@
...
+def post_install(self):
+    self.install_license("COPYING")
+
+
+@subpackage("xcb-util-errors-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+configure_gen = []
```
nekopsykose, line 24: this being at the bottom of files is a relic from when gnu_configure started to reconf by default

but you also don't need it, add automake+libtool+xcb-util-macros to hostmakedepends instead

ifreund: Indeed, I should probably use my brain instead of just copying from other templates...

q66:
the package doesn't belong in main, literally nothing but wlroots uses it

PR #1609, opened 2024-03-17
contrib/oath-toolkit: new package (2.6.11)

jcgruenhage:
Currently not working, failing at https://gitlab.com/oath-toolkit/oath-toolkit/-/blob/master/libpskc/schemas/Makefile.am?ref_type=heads#L27-L36 and I have no clue how to fix that.


nekopsykose:
xmlcatalog is in libxml2-progs, not libxml2 for hostmakedepends

nekopsykose:
after that it gets further but fails in ./gdoc because that has no shebang seemingly, not sure about that

contrib/oath-toolkit/template.py
```diff
@@ -0,0 +1,24 @@
+pkgname = "oath-toolkit"
+pkgver = "2.6.11"
+pkgrel = 0
+build_style = "gnu_configure"
+make_cmd = "gmake"
+make_dir = "."
```
jcgruenhage, line 6: @nekopsykose ^ this bit fixes the `./gdoc` issue. Building out of tree isn't supported it seems

jcgruenhage:
I haven't investigated the check failures further, but at least `oathtool` works when used in `pass-otp`.

contrib/oath-toolkit/template.py
nekopsykose: nice :)

contrib/oath-toolkit/template.py
```diff
@@ -0,0 +1,33 @@
...
+url = "http://www.nongnu.org/oath-toolkit"
+source = f"http://download.savannah.gnu.org/releases/{pkgname}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "fc512a4a5b46f4c43ab0586c3189fece4d54f7e649397d6fa1e23428431e2cb4"
+# TODO: failing checks are yet to be investigated
+options = ["!check"]
+
+
+@subpackage("oathtool")
+def _oathtool(self):
+    return ["usr/bin/oathtool"]
```
nekopsykose, line 28: this also has to take the manpage

jcgruenhage: done

contrib/oath-toolkit/template.py
```diff
@@ -0,0 +1,44 @@
...
+build_style = "gnu_configure"
+make_cmd = "gmake"
+make_dir = "."
+hostmakedepends = [
+    "gmake",
+    "automake",
+    "libtool",
+    "gtk-doc-tools",
+    "pkgconf",
+    "libxml2-progs",
```
nekopsykose, line 13: sort

contrib/oath-toolkit/template.py
```diff
@@ -0,0 +1,44 @@
...
+source = f"http://download.savannah.gnu.org/releases/{pkgname}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "fc512a4a5b46f4c43ab0586c3189fece4d54f7e649397d6fa1e23428431e2cb4"
+# TODO: failing checks are yet to be investigated
+options = ["!check"]
+
+
+def post_install(self):
+    for lib in ["pskc", "oath"]:
+        self.install_man(f"lib{lib}/man/{lib}_*.3", glob=True)
+        self.install_man(f"{lib}tool/{lib}tool.1")
```
nekopsykose, line 29: is there a reason the build system doesn't install these?

contrib/oath-toolkit/template.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@subpackage("oathtool")
+def _oathtool(self):
+    return ["usr/bin/oathtool", "usr/share/man/man1/oathtool.1"]
+
+
+@subpackage("pskctool")
+def _pskctool(self):
+    return ["usr/bin/pskctool", "usr/share/man/man1/pskctool.1"]
```
nekopsykose, line 39: hmm.. thinking again, is there a need to split these? if the whole thing is like just a few mb then there's usually not much point

contrib/oath-toolkit/template.py
jcgruenhage: I'm not sure why it doesn't, but it doesn't. Alpine for example also doesn't ship the man pages.

contrib/oath-toolkit/template.py
jcgruenhage: I thought it was good practice to split them, but I have no problem merging them either.

nekopsykose: splitting is generally to split up large stuff, or things that pull in some big dep chain (e.g. maybe some gui component), or auto policy stuff (default splits, devel , etc). but for small programs like this it's usually more hassle (entire thing is very small overall and having to hunt for a few binaries instead of installing just 1 thing is usually an extra annoyance).

so compared to e.g. alpine there is less granular splitting here per-program or whatever as it's not as final-size focused

jcgruenhage: So I should get rid of this split and just keep the devel subpackage?

contrib/oath-toolkit/template.py
jcgruenhage: Is there anything I should do regarding this? Or is it also okay to keep it like this?

nekopsykose: the same is done for the .3's?

jcgruenhage: I called `apk manifest` on the packages and couldn't find the man pages, maybe I looked in the wrong place. Let me try again.

nekopsykose: yeah i removed this post_install and the same manpages are already installed

nekopsykose: alpine also has them https://pkgs.alpinelinux.org/contents?branch=edge&repo=community&arch=x86_64&page=2&name=oath%2dtoolkit%2ddoc

nekopsykose: > I called apk manifest on the packages and couldn't find the man pages,  

they are split to -man by default. for .3, it's -devel-man. in both cases, they autoinstall with some base- package when the respective one is installed

contrib/oath-toolkit/template.py
nekopsykose: yeah

contrib/oath-toolkit/template.py
jcgruenhage: yeah okay I apparently just checked the wrong packages, because the man pages are there.

contrib/oath-toolkit/template.py
jcgruenhage: So, `oath-toolkit-progs` or should they just go into the main package?

nekopsykose: if they're 45kb then there's no point to split that imo

PR #1610, opened 2024-03-18
Bump GNOME packages

mhmdanas:
- **main/evolution-data-server: update to 3.52.0**
- **contrib/evolution: update to 3.52.0**
- **main/gnome-calendar: bump for evolution-data-server 3.52.0**
- **main/gnome-shell: update to 45.5**
- **main/eog: update to 45.3**
- **main/mutter: update to 45.5**


main/evolution-data-server/template.py
```diff
@@ -1,10 +1,9 @@
 pkgname = "evolution-data-server"
-pkgver = "3.50.4"
+pkgver = "3.52.0"
 pkgrel = 0
 build_style = "cmake"
 # TODO: libgdata
```
nekopsykose, line 5: the todo itself also isn't needed

nekopsykose: (this was packaged at 3.42, in 3.46 gdata was removed in favor of internal stuff tho)

PR #1615, opened 2024-03-18
main/power-profiles-daemon: update to 0.20

mhmdanas:
cc @valpackett


valpackett:
Looks good. Any particular reason to use the fd notification over dbus readiness?

nekopsykose:
there is no dbus readiness if you mean waiting for the dbus name to come up. you would need a wrapper like https://github.com/chimera-linux/dbus-wait-for

valpackett:
oh, I thought we built that into dinit itself lol

nekopsykose:
well, once q66 adds that it becomes easy to just add it to command=, i forgot what his plans were :)

PR #264, opened 2023-08-05
contrib/btop: new package (1.2.13)

nekopsykose:
patches are just from master and will be in the next release anyway

PR #1622, opened 2024-03-19
user/btop: new package (1.3.2)

squili:
first contribution :3
cbuild is so nice to work with

### DIFF
--- a/user/btop/template.py
+++ b/user/btop/template.py
@@ -1,0 +1,18 @@
+pkgname = "btop"
+pkgver = "1.3.2"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DBTOP_LTO=OFF", "-DBTOP_FORTIFY=OFF"]
+hostmakedepends = ["cmake", "ninja"]
+pkgdesc = "TUI monitor of system resources"
+maintainer = "mia <mia@mia.jetzt>"
+license = "Apache-2.0"
+url = "https://github.com/aristocratos/btop"
+source = f"{url}/archive/v{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "331d18488b1dc7f06cfa12cff909230816a24c57790ba3e8224b117e3f0ae03e"
+hardening = ["cfi", "vis"]
+
+
+# no tests, just make sure binary runs
+def do_check(self):
+    self.do("./build/btop", "--version")
--- a/user/btop/patches/nomisc.patch
+++ b/user/btop/patches/nomisc.patch
@@ -1,0 +1,20 @@
+fstack-protector overwrites fstack-protector=strong and the others are already passed or useless
+--
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -100,15 +100,12 @@
+ endif()
+ 
+ if(NOT APPLE)
+-  target_compile_options(btop PRIVATE -fstack-clash-protection)
+ endif()
+ check_cxx_compiler_flag(-fstack-protector HAS_FSTACK_PROTECTOR)
+ if(HAS_FSTACK_PROTECTOR)
+-  target_compile_options(btop PRIVATE -fstack-protector)
+ endif()
+ check_cxx_compiler_flag(-fcf-protection HAS_FCF_PROTECTION)
+ if(HAS_FCF_PROTECTION)
+-  target_compile_options(btop PRIVATE -fcf-protection)
+ endif()
+ 
+ target_compile_definitions(btop PRIVATE


nekopsykose:
i actually did 'package' this twice before in #264 #1191 . the issue is that if you do something that spams a lot of jobs (perhaps also with long ass argv's, like building main/samba?), it crashes, somewhere in a libc++ assertion. so i never formally merged it since it always crashes after like an hour of uptime for me (if i build stuff; if i don't then it doesn't)

do you feel like debugging why that happens? didn't look very easy

nekopsykose:
unrelatedly you also want to steal the  

```
make_build_args = ["OPTFLAGS="]
make_use_env = True
```

so some unwanted stuff is left out and it doesn't override the project default cxxflags

squili:
ah, that's not good! i do not have the knowledge to debug something like that, unfortunately
maybe it would be a candidate for the user category instead? at least until someone takes up debugging
could also go and package bpytop instead, since it has most of the same features, but it doesn't seem to be actively being worked on

q66:
i think putting it in user first is reasonable enough

squili:
moved!

user/btop/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://api.github.com/repos/aristocratos/btop/git/refs/tags"
```
nekopsykose, line 1: this is only needed if the default detection doesn't work, and it works here:  

```
$ rm user/btop/update.py                
$ ./cbuild update-check user/btop s              
Checking for updates: btop=1.3.2
Adding 'https://github.com/aristocratos/btop' for version check...
Adding 'https://github.com/aristocratos/btop/archive/v1.3.2/' for version check...
Fetching 'https://github.com/aristocratos/btop/tags' for version checks...
Fetching 'https://github.com/aristocratos/btop/tags' for version checks...
Already fetched 'https://github.com/aristocratos/btop/tags', skipping...
Checking found version: 1.2.7
Checking found version: 1.2.8
Checking found version: 1.2.9
Checking found version: 1.2.10
Checking found version: 1.2.11
Checking found version: 1.2.12
Checking found version: 1.2.13
Checking found version: 1.3.0
Checking found version: 1.3.1
Checking found version: 1.3.2
```

user/btop/template.py
```diff
@@ -0,0 +1,20 @@
...
+hostmakedepends = ["gmake"]
+pkgdesc = "TUI monitor of system resources"
+maintainer = "mia <mia@mia.jetzt>"
+license = "Apache-2.0"
+url = "https://github.com/aristocratos/btop"
+source = f"{url}/archive/v{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "331d18488b1dc7f06cfa12cff909230816a24c57790ba3e8224b117e3f0ae03e"
+hardening = ["cfi", "vis"]
+make_build_args = ["OPTFLAGS="]
+make_use_env = True
```
nekopsykose, line 15: these go near the top (see ci fail)

nekopsykose: there's also some other thing to clean up in the makefile but i'll do that for you after

nekopsykose: > these go near the top (see ci fail)  

(after rebase, since it was broken on master)  

you can also run black+flake8 locally, same thing

nekopsykose: > there's also some other thing to clean up in the makefile but i'll do that for you after  

or well, you could use cmake instead since that looks better, for which it's just  

```
build_style = "cmake"
hostmakedepends = ["cmake", "ninja"]
```

nekopsykose: + `configure_args = ["-DBTOP_LTO=OFF", "-DBTOP_FORTIFY=OFF"]` and that's about it for cmake instead

nekopsykose: full template:  

```
pkgname = "btop"
pkgver = "1.3.2"
pkgrel = 0
build_style = "cmake"
configure_args = ["-DBTOP_LTO=OFF", "-DBTOP_FORTIFY=OFF"]
hostmakedepends = ["cmake", "ninja"]
pkgdesc = "TUI monitor of system resources"
maintainer = "mia <mia@mia.jetzt>"
license = "Apache-2.0"
url = "https://github.com/aristocratos/btop"
source = f"{url}/archive/v{pkgver}/{pkgname}-{pkgver}.tar.gz"
sha256 = "331d18488b1dc7f06cfa12cff909230816a24c57790ba3e8224b117e3f0ae03e"
hardening = ["cfi", "vis"]


# no tests, just make sure binary runs
def do_check(self):
    self.do("./build/btop", "--version")
```

note also the self.do, for cwd it's fine to use `./` instead of the longer form

squili: so many things!
it have put them into the git machine, thank you

nekopsykose: hah yeah, i talk a lot sorry

nekopsykose:
thanks!

PR #1623, opened 2024-03-19
contrib/pass-otp: new package (1.2.0)

jcgruenhage:


contrib/pass-otp/template.py
```diff
@@ -0,0 +1,18 @@
...
+license = "GPL-3.0-or-later"
+url = "https://github.com/tadfisher/pass-otp"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "5720a649267a240a4f7ba5a6445193481070049c1d08ba38b00d20fc551c3a67"
+# Has no test suite
+options = ["!check"]
+
+
+def do_build(self):
+    pass
```
jcgruenhage, line 18: Is this the right way to skip the build stage? This is just installing two files, one bash script (the pass extension itself) and one bash completion for pass with that extension

nekopsykose: yes

the other alternative is to not have a build style and manually call make install, but that is more clunky

nekopsykose: well, you can also not skip it? it doesn't look like it fails

nekopsykose: yeah it just prints some messages so you can let it do that :D

contrib/pass-otp/template.py
```diff
@@ -0,0 +1,18 @@
+pkgname = "pass-otp"
+pkgver = "1.2.0"
+pkgrel = 0
+build_style = "makefile"
+hostmakedepends = []
```
nekopsykose, line 5:
```suggestion
```

contrib/pass-otp/patches/0001-fix-bash-completions-dir.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/Makefile b/Makefile
+index a48cad1..185b446 100644
+--- a/Makefile
++++ b/Makefile
+@@ -4,7 +4,7 @@ DESTDIR ?=
+ LIBDIR ?= $(PREFIX)/lib
+ SYSTEM_EXTENSION_DIR ?= $(LIBDIR)/password-store/extensions
+ MANDIR ?= $(PREFIX)/share/man
+-BASHCOMPDIR ?= /etc/bash_completion.d
++BASHCOMPDIR ?= /usr/share/bash-completion/completions
```
nekopsykose, line 20: you don't need a patch for this, you can pass it in make_install_args

contrib/pass-otp/template.py
jcgruenhage: ¯\_(ツ)_/¯

PR #1628, opened 2024-03-20
contrib/dunst: new package (1.11.0)

ttyyls:


contrib/dunst/template.py
```diff
@@ -0,0 +1,47 @@
...
+url = "https://dunst-project.org"
+source = (
+    f"https://github.com/dunst-project/dunst/archive/refs/tags/v{pkgver}.tar.gz"
+)
+sha256 = "31c0eb749ca83dab7f5af33beb951c9f9a8451263fcee6cbcf8ba3dedbf2e1f1"
+env = {"SYSCONFDIR": "/etc"}
+hardening = ["vis", "cfi"]
+
+
+def pre_install(self):
```
nekopsykose, line 35: fix the makefile install instead

contrib/dunst/template.py
```diff
@@ -0,0 +1,47 @@
+pkgname = "dunst"
+pkgver = "1.11.0"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_check_target = "test"
+make_use_env = True
+hostmakedepends = ["gmake", "perl", "pkgconf"]
```
nekopsykose, line 8: + wayland-progs

ttyyls:
> doesn't this need a user service

nope. dbus activation takes care of that

nekopsykose:
it doesn't because you can have multiple things providing the notification endpoint and they won't necessarily pick what you want

contrib/dunst/files/dunst.user
```diff
@@ -0,0 +1,7 @@
+# dunst user service
+
+type               = process
+command            = /usr/bin/dunst
+depends-on         = dbus
+log-type           = buffer
+restart            = true
```
nekopsykose, line 7: restart =true is the default, and don't need to vertically align anymore

contrib/dunst/files/dunst.user
```diff
@@ -0,0 +1,6 @@
+# dunst user service
+
+type       = process
+command    = /usr/bin/dunst
+depends-on = dbus
+log-type   = buffer
```
nekopsykose, line 6:
```suggestion
type = process
command = /usr/bin/dunst
depends-on = dbus
log-type = buffer
```

PR #1634, opened 2024-03-20
contrib/ruff: new package (0.3.3)

nekopsykose:


q66:
this can be merged just fine

contrib/ruff/template.py
```diff
@@ -0,0 +1,49 @@
...
+        with open(self.cwd / f"ruff.{shell}", "w") as f:
+            self.do(
+                "./target/release/ruff",
+                "--generate-shell-completion",
+                shell,
+                stdout=f,
+            )
+
+
+def do_check(self):
```
q66, line 40: is there any actual need for this though

nekopsykose: you mean the tests at all, or the manual check? pytest doesn't run them

PR #1635, opened 2024-03-20
ci: use ruff format instead of black

nekopsykose:
black takes 50+ seconds. ruff takes less than one :)

also fix the single source of difference from black (which black doesn't fix back into)  

---

see https://docs.astral.sh/ruff/formatter/#black-compatibility for limitations. tl;dr not really anything  

we could also use it as a flake8 drop in, but  

- flake8 is not anywhere near as slow for now (though i would still like to change to this)
- have to fix a million things like:  
  ```
  main/zimg/template.py:22:2: F821 Undefined name `subpackage`
  main/zix/template.py:23:2: F821 Undefined name `subpackage`
  main/zlib/template.py:20:2: F821 Undefined name `subpackage`
  ```

i kept the black entry in pyproject.toml too so that still works

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,7 +28,7 @@ jobs:
         run: sudo apt-get update && sudo apt-get install python3-pip
 
       - name: Install dependencies
-        run: sudo pip3 install black flake8
+        run: sudo pip3 install flake8 ruff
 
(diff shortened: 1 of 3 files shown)


nekopsykose:
the entire ci lint step now takes 16s instead of 70+ :D

PR #1642, opened 2024-03-21
Protobuf 26

JamiKettunen:
Seems fine from a quick test

https://github.com/protocolbuffers/protobuf/releases/tag/v26.0


JamiKettunen:
I'll roll https://github.com/nmeum/android-tools/releases/tag/34.0.5 update also into this, gonna investigate what needs to be done with `ppc.patch` first though

PR #1644, opened 2024-03-22
blender dependencies: openimageio, opencolorio, robin-map, pystring

z-erica:
will update pr (maybe dropping some packages) as the actions run


contrib/opencolorio/patches/fix_opengl_target.patch
```diff
@@ -0,0 +1,20 @@
+--- a/src/libutils/oglapphelpers/CMakeLists.txt	2023-12-08 02:02:08.000000000 +0100
++++ b/src/libutils/oglapphelpers/CMakeLists.txt	2024-01-29 09:15:21.011903195 +0100
+@@ -72,7 +72,7 @@
+         target_link_libraries(oglapphelpers
+             PRIVATE
+                 OpenColorIO
+-                OpenGL::OpenGL
++                OpenGL::GL
```
nekopsykose, line 8: pass     "-DOPENGL_opengl_LIBRARY=/usr/lib/libGL.so", to cmake instead

contrib/pystring/template.py
```diff
@@ -0,0 +1,22 @@
...
+license = "BSD-3-Clause"
+url = "https://github.com/imageworks/pystring"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "49da0fe2a049340d3c45cce530df63a2278af936003642330287b68cefd788fb"
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+    # cmake does not install the header for some reason
+    self.install_file("pystring.h", "usr/include/pystring")
```
nekopsykose, line 17: apply https://github.com/imageworks/pystring/pull/44 instead

contrib/numpy/template.py
```diff
@@ -0,0 +1,31 @@
+pkgname = "numpy"
```
nekopsykose, line 1: python-numpy

this is also pretty useless without fortran and openblas. you should put this in a separate mr and we can work on those separately

also, i forget if python_pep517 even passes through cflags and stuff, does it?

nekopsykose: also the versions here are input wrong etc

contrib/embree/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "embree"
+pkgver = "4.3.1"
+pkgrel = 0
```
nekopsykose, line 3: this is iirc arch limited but we can check that later

contrib/embree/template.py
```diff
@@ -0,0 +1,25 @@
...
+    "libjpeg-turbo-devel",
+    "libpng-devel",
+    "glfw-devel",
+    "mesa-devel",
+]
+pkgdesc = "Ray tracing library"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "Apache-2.0"
+url = "https://www.embree.org"
+source = "https://github.com/embree/embree/archive/refs/tags/v4.3.1.tar.gz"
```
nekopsykose, line 19:
```suggestion
source = "https://github.com/embree/embree/archive/refs/tags/v{pkgver}.tar.gz"
```

etc

contrib/embree/template.py
```diff
@@ -0,0 +1,25 @@
...
+pkgver = "4.3.1"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DCMAKE_BUILD_TYPE=Release"]
+hostmakedepends = ["cmake", "ninja"]
+makedepends = [
+    "onetbb-devel",
+    "openimageio-devel",
+    "libjpeg-turbo-devel",
+    "libpng-devel",
```
nekopsykose, line 11: missing sort in half the lists

contrib/glfw/template.py
```diff
@@ -0,0 +1,27 @@
+pkgname = "glfw"
+pkgver = "3.3.9"
```
nekopsykose, line 2: you downgraded this again by accident, though of course you'll have to fix the 3.4 build

nekopsykose:
`git reflog` has older stuff if you lost things

contrib/embree/template.py
```diff
@@ -0,0 +1,25 @@
...
+pkgdesc = "Ray tracing library"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "Apache-2.0"
+url = "https://www.embree.org"
+source = f"https://github.com/embree/embree/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "824edcbb7a8cd393c5bdb7a16738487b21ecc4e1d004ac9f761e934f97bb02a4"
+
+
+@subpackage("embree-devel")
+def _devel(self):
```
nekopsykose, line 24: this also installs garbage to `/usr/embree-vars*` and also some /usr/src stuff (pass -DEMBREE_TUTORIALS=OFF for that)

there's also a ton of random bins that are probably useless. i would split progs here

contrib/openimageio/template.py
```diff
@@ -0,0 +1,43 @@
...
+    "libpng-devel",
+    "libraw-devel",
+    "libtiff-devel",
+    "libwebp-devel",
+    "mesa-devel",
+    "opencolorio-devel",
+    "openexr-devel",
+    "openjpeg-devel",
+    "onetbb-devel",
+    "python-pybind11-devel",
```
nekopsykose, line 22: also needs pugixml (and maybe USE_EXTERNAL_PUGIXML but not sure)

nekopsykose:
dropped embree from this since i want to test it more later (submit another mr for it)

still needs a bit more testing (i have a hunch some of this is broken with int and i want to check the tests), but looks good enough for now before this is used for anythin

PR #1646, opened 2024-03-23
contrib/gitui: update to 0.25.2

ttyyls:


ttyyls:
I wasn't able to reproduce the error locally, but I suspect that the  .cargo/config  shipped with the project is taking priority over the .cargo/config.toml which adds the vendored sources setting. Thus the build skips using the vendored sources on the install stage triggering a fetch from crates-io which ofc fails.
 https://github.com/chimera-linux/cports/actions/runs/8402217954/job/23011382337#step:8:945

ttyyls:
> Note: Cargo also reads config files without the .toml extension, such as .cargo/config. Support for the .toml extension was added in version 1.39 and is the preferred form. If both files exist, Cargo will use the file without the extension.

https://doc.rust-lang.org/cargo/reference/config.toml

q66:
i don't like that because it means silently removing potentially meaningful config

nekopsykose:
tomllib is python 3.11 so we could read the file for real and write out the section properly, not overwriting anything in place

nekopsykose:
well, didn't check how hard it is to write it back, since tomllib is read only

ttyyls:
Afaik it only ever gets used to define options specific to foreign targets for development purposes and it doesn't make sense to keep it anyway since it will shadow the config.toml.

In the offchance that some future project has some useful setting in there I suppose we could concatenate the two files then write before appending the vendored-sources snippet.

src/cbuild/util/cargo.py
```diff
@@ -71,6 +71,11 @@ def setup_vendor(pkg, vendor_path="vendor", wrksrc=None):
     # `parents` ensures the directory is allowed to exist already
     pkg.mkdir(dirn / ".cargo", parents=True)
     with open(dirn / ".cargo/config.toml", "a") as cf:
+        # add options from legacy .cargo/config it it exists;
+        legacy = dirn / ".cargo/config"
+        if legacy.exists():
+            with open(legacy, "r") as infile:
+                cf.write(infile.read())
```
ttyyls, line 78: @q66 does this address your concerns of losing meaningful configuration?

PR #1649, opened 2024-03-24
contrib/zola: new package (0.18.0)

jcgruenhage:


contrib/zola/template.py
```diff
@@ -0,0 +1,31 @@
...
+sha256 = "c0e1711a68bc005c2e0ecc76a468f3459739c9e54af34850cb725d04391e19b5"
+# generates completions with host bins
+options = ["!cross"]
+
+
+def post_build(self):
+    for shell in ["bash", "fish", "zsh"]:
+        with open(self.cwd / f"zola.{shell}", "w") as outf:
+            self.do(
+                f"target/{self.profile().triplet}/release/zola",
```
jcgruenhage, line 21: Is it preferable to have it like it is commited with a format string or joining paths like this alternative suggestion? Mostly a style question, shouldn't be a difference functionally
```suggestion
               self.cwd()
               / "target"
               / self.profile().triplet
               / "release"
               / "zola",
```

nekopsykose: generally we seem to do fstring

nekopsykose: and for something as short as this the multiline just makes it longer so i think that looks worse (but ofc a style preference; in other places i usually prefer multiline stuff..)

jcgruenhage: (Well except that `self.cwd` is not a valid path in the sandboxed container, but that's just a minor detail)

nekopsykose: for cwd there's no reason to not use ./ unless you need cwd/chroot_cwd for absolute path

nekopsykose: (e.g. if you put it an env var for some configuration, non-absolute is weirder)

jcgruenhage: I'll just keep it with the fstring then, shorter anyway :)

PR #1650, opened 2024-03-24
contrib/minijinja-cli: new package (1.0.15)

jcgruenhage:


contrib/minijinja-cli/template.py
```diff
@@ -0,0 +1,24 @@
+pkgname = "minijinja-cli"
+pkgver = "1.0.15"
+pkgrel = 0
+build_wrksrc = "minijinja-cli"
+build_style = "cargo"
+make_build_args = ["--features=completions"]
+make_build_env = {
+    "ASSET_OUT_DIR": "assets",
+}
```
jcgruenhage, line 9: This is intentionally only enabled for the build phase, which causes it to generate completions during the build phase. During `do_install`, cargo will build it again, but without the `completions` feature activated.

Is it okay to keep it like this? Or should I override `do_build` to be more explicit in that behaviour instead? I could explicitly add to cargo build invocations, one with the feature enabled and the env var set, and then a second one without both of these.

nekopsykose: i think this is ok (helps with installing the better more minimised package, and since all the steps are implicitly correct there is no need for rewriting the logic), but just add a comment like

```
# cargo install rebuilds without these args in install to not ship feature at runtime
```

etc

contrib/minijinja-cli/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://api.github.com/repos/mitsuhiko/minijinja/git/refs/tags"
```
nekopsykose, line 1: this file is not needed (i checked)

contrib/minijinja-cli/template.py
```diff
@@ -0,0 +1,25 @@
...
+build_wrksrc = "minijinja-cli"
+build_style = "cargo"
+# cargo install rebuilds without these args in install to not ship feature at runtime
+make_build_args = ["--features=completions"]
+make_build_env = {
+    "ASSET_OUT_DIR": "assets",
+}
+hostmakedepends = ["cargo", "pkgconf"]
+makedepends = ["rust-std", "zstd-devel"]
+pkgdesc = "CLI tool for minijinja, a minimal dependency jinja implementation"
```
nekopsykose, line 13:
```suggestion
pkgdesc = "Minimal jinja implementation"
```

best i can come up with but maybe you have even better ideas

jcgruenhage: fine with me, and no, don't have better ideas

PR #1651, opened 2024-03-24
contrib/typst: new package (0.11.0)

jcgruenhage:


nekopsykose:
manually install this one cause it relinks for a few minutes in cargo install

PR #1652, opened 2024-03-24
contrib/postgresql: new package (16.2)

squili:
still a lot to do here, but the core functionality seems to be working!


contrib/postgresql/template.py
```diff
@@ -0,0 +1,106 @@
...
+maintainer = "mia <mia@mia.jetzt>"
+license = "PostgreSQL"
+url = "https://www.postgresql.org"
+source = [
+    f"https://ftp.postgresql.org/pub/source/v{pkgver}/postgresql-{pkgver}.tar.bz2"
+]
+sha256 = ["446e88294dbc2c9085ab4b7061a646fa604b4bec03521d5ea671c2e5ad9b2952"]
+# TODO: checks rely on libpq already being installed
+# could maybe makedepend on libpq once it's in the repos?
+# note: alpine doesn't do checks either
```
nekopsykose, line 34: this just means to install for tests, i.e. running meson install as part of check and running a check on that tmpdir. it's not impossible, but let's leave this for later and decide then (i'm not sure if there is a minimised set- the full testsuite is unreasonable and takes forever iirc, since it's a whole database thing)

contrib/postgresql/template.py
```diff
@@ -0,0 +1,106 @@
+pkgname = "postgresql"
+pkgver = "16.2"
+pkgrel = 0
+build_style = "configure"
```
nekopsykose, line 4: meson

even if it wasn't, this is a gnu_configure project too. but meson is better here

contrib/postgresql/template.py
```diff
@@ -0,0 +1,106 @@
...
+configure_args = [
+    # TODO: review path choices - alpine puts pkgname in a few of these
```
nekopsykose, line 6: these are all the defaults (and map the same way to meson)- alpine changes them because alpine supports multiple versions

tbh, having a postgresql without packaging multiple versions (read: current, and N-1), to allow upgrading, is useless long term. on any major everyone would break their database because you need the old binaries around to upgrade it. so you actually do need to name this postgresql16 and do some setup to allow a 17 later, which does involve changing these. but we can think about it later

contrib/postgresql/template.py
```diff
@@ -0,0 +1,106 @@
...
+    "--with-system-tzdata=/usr/share/zoneinfo",
+]
+make_cmd = "gmake"
+makedepends = [
+    "gmake",
+    "pkgconf",
+    "icu-devel",
+    "readline-devel",
+    "zlib-devel",
+    "linux-headers",
```
nekopsykose, line 21: gmake, pkgconf are hostmakedepends

but here swap gmake for meson ofc

contrib/postgresql/template.py
```diff
@@ -0,0 +1,106 @@
...
+        "usr/lib/libpgfeutils.a",
+    ]
+
+
+@subpackage("libecpg")
+def _libecpg(self):
+    self.pkgdesc = "Embedded SQL in C"
+    return [
+        "usr/lib/libecpg.so*",
+        "usr/lib/libpgtypes.so*",  # TODO: should libpgtypes be it's own package?
```
nekopsykose, line 72: alternate question- what benefit do we get from splitting the two libs separately, instead of just one `postgresql-libs`?

contrib/postgresql/template.py
```diff
@@ -0,0 +1,106 @@
...
+options = ["!check"]
+
+
+# NOTE: most of these subpackages are based on whatever alpine is doing, so they might need tweaking
+
+
+@subpackage("libpq")
+def _libpq(self):
+    self.pkgdesc = "PostgreSQL client library"
+    return ["usr/lib/libpq.so*"]
```
nekopsykose, line 44: base .so should be in -devel, so this is `.so.*` instead usually, unless things dlopen the unversioned name (advice for in general)

contrib/postgresql/template.py
squili: some postgresql libraries seem to not be tied to specific postgresql versions. the alpine libpq package is pulled from the latest postgres. maybe this would be a good argument for it?
https://git.alpinelinux.org/aports/tree/main/postgresql16/APKBUILD#n55

squili: additionally, we could split out a `postgresql-libs` and a `postgresql16`/`postgresql17`

nekopsykose: yeah, that was the next part. if everything is fine using latest libpg/ecpg/pgtypes, that means only 'latest' should have. alpine has that and iirc that works fine, so indeed just one postgresql-libs should be ok. but i'm not sure what to call it, because postgresql-libs-devel is weird, and -devel against the libs (latest only) is separate from -devel for the db itself (used to build out of tree postgres plugins, not latest only). what should we call them?

nekopsykose: easy solution is to actually keep this lib splitting as is, with libpq-devel/libecpg-devel, i.e. ignoring what i said. that seems ok

nekopsykose: and the todo pgtypes should stay in ecpg so that is done

nekopsykose: for things latest only, you can do

```
@subpackage("libecpg-devel", _is_latest)
```

and put `_is_latest = True` somewhere. though to get this to work like in alpine you also need `external-libpq.patch` that will only start getting applied on the old version while pulling new libpq in makedep. but let's leave everything i wrote in this comment for later

squili:
i am not sure what to do about that run failure

nekopsykose:
you need to add an `update.py` (search the repo) that lets the latest version be detected. in this case you want `pkgname = "postgresql"` in the file, then maybe tweak url/pattern if needed. check with `./cbuild update-check ...` locally

squili:
im not sure how well pg_versions integrates with manpages on chimera, but i did fix a bug that would cause it to fail installing the package if you don't have man installed

squili:
also, the service file is very weak and should be improved with stuff like reloading

nekopsykose:
> also, the service file is very weak and should be improved with stuff like reloading

`dinitctl signal HUP postgresql` should already work from what i can tell, unless it behaves weirdly after

squili:
how would i expose that to users? i saw install_svscript but no packages appear to be currently using it so i don't have much to base that on

contrib/postgresql-common/files/postgresql.confd
```diff
@@ -0,0 +1,79 @@
+# Configuration for /etc/init.d/postgresql
```
nekopsykose, line 1: this entire file is unused (there is no init.d)

contrib/postgresql16/template.py
```diff
@@ -0,0 +1,198 @@
...
+    "zlib-devel",
+    "linux-headers",
+    "lz4-devel",
+    "zstd-devel",
+    "perl",
+    "python-devel",
+    "tcl-devel",
+    "openssl-devel",
+    "libxml2-devel",
+    "e2fsprogs-devel",
```
nekopsykose, line 48: sort

contrib/postgresql16/postgresql16.pre-deinstall
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
```
q66, line 1: remove this whole file

contrib/postgresql-common/files/pg_versions
```diff
@@ -0,0 +1,225 @@
+#!/bin/sh
```
q66, line 1: this is crude and we don't want that, any "alternatives management" should be done through package management and not some weird haphazard scripts

contrib/postgresql-common/postgresql-common.trigger
```diff
@@ -0,0 +1,72 @@
+#!/bin/sh
```
q66, line 1: this is also gross and not usable

if you're just taking packaging from alpine here, please don't do that

contrib/postgresql16/template.py
```diff
@@ -0,0 +1,198 @@
...
+
+@subpackage(f"{pkgname}-contrib")
+def _contrib(self):
+    def subpackage():
+        _subpackage_install(self, "contrib")
+        for subdir in _contrib_search(
+            self,
+            lambda name: name.endswith("_plperl") or name.endswith("_plpython"),
+        ):
+            self.parent.do(
```
q66, line 101: running commands in subpackage install hooks violates packaging guidelines, subpackage install hooks should *only* steal files

nekopsykose: an exception we have is for libreoffice that does the install file stuff. i think here it's "almost" similar in that all of these have a large amount of files (that change per version seemingly) that are laborious to manually separate instead of calling gmake install per subpackage. is there a way that doesn't involve listing a million files?

q66:
this is a draft, the alternatives management needs cleanroom design

contrib/postgresql16/template.py
q66: fwiw file lists can also be used anywhere else if necessary, the important part is that they are generated ahead of time

q66:
i'm finally getting to this one (apologies for taking so long, there were lots of things in the way :(...)

i won't merge it as it is but i'll rework things as needed myself (this is a complicated template and it's probably much faster and easier if i do that), i'll leave you as the author of the commit

thanks for working on it :)

q66:
needs more work but 07ab5153276d3d36cf17c67f54663c1317cca378

PR #1659, opened 2024-03-25
contrib/atuin: new package (18.1.0)

jcgruenhage:


contrib/atuin/template.py
```diff
@@ -0,0 +1,34 @@
+pkgname = "atuin"
+pkgver = "18.1.0"
+pkgrel = 0
+build_wrksrc = "atuin"
+build_style = "cargo"
+hostmakedepends = ["cargo"]
+makedepends = ["sqlite-devel"]
```
nekopsykose, line 7: --no-default-features to remove `check-update`

nekopsykose, line 7: should also add a patch like

```
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,12 +33,12 @@
 
 [workspace.dependencies.reqwest]
 version = "0.11"
-features = ["json", "rustls-tls-native-roots"]
+features = ["json", "default-tls"]
 default-features = false
 
 [workspace.dependencies.sqlx]
 version = "0.6"
-features = ["runtime-tokio-rustls", "chrono", "postgres"]
+features = ["runtime-tokio-native-tls", "chrono", "postgres"]
 
 [patch.crates-io]
 sqlx = { path = "./vendor/sqlx" }
```

to use openssl instead

nekopsykose: you can also replace build_wrksrc with `-p atuin` to the args while doing that, and then remove the `../` from the install below i think

nekopsykose: (since it also spends minutes in install, so replace that with do_install and a manual one as well)

jcgruenhage: I do understand patching tools to use system libraries instead of bundled versions of that same library, but why replace rustls with openssl?

PR #1669, opened 2024-03-26
contrib/php83: new package (8.3.6)

rbgarga:
There are 8 tests failing, test results with all details are available at https://gist.github.com/rbgarga/d6df49ce639a58555b09fd6e1a33488d any insight to solve them is much appreciated.


contrib/php/template.py
```diff
@@ -0,0 +1,276 @@
...
+    ("xmlreader", True),
+    ("xmlwriter", True),
+    ("xsl", False),
+    ("zip", False),
+    ("zlib", False),
+]:
+    _extension(_extn, _iif)
+
+
+@subpackage(f"{pkgname}-devel")
```
q66, line 269: php-devel is fine

contrib/php/template.py
```diff
@@ -0,0 +1,276 @@
...
+        extn = extso.removesuffix(".so")
+        with open(extcp / f"{extn}.ini", "w") as outf:
+            if extn == "opcache":
+                outf.write(f"zend_extension={extso}\n")
+            else:
+                outf.write(f"extension={extso}\n")
+
+
+def _extension(extn, iif):
+    @subpackage(f"{pkgname}-{extn}")
```
q66, line 208: `php-{extn}` is fine

contrib/php/template.py
```diff
@@ -0,0 +1,276 @@
...
+    def _ext(self):
+        self.pkgdesc = f"{pkgdesc} ({extn} extension)"
+
+        if iif:
+            self.install_if = [f"{pkgname}={pkgver}-r{pkgrel}"]
+
+        return [
+            f"etc/php/conf.d/{extn}.ini",
+            f"?usr/include/php/ext/{extn}",
```
q66, line 217: do we want to ship include files with the modules? this should probably remain in devel unless you have a good reason

contrib/php/template.py
```diff
@@ -0,0 +1,276 @@
...
+        "ext/posix/tests/posix_errno_variation1.phpt",
+        "ext/zip/tests/oo_encryption.phpt",
+        "sapi/cli/tests/009.phpt",
+        "sapi/cli/tests/012-2.phpt",
+        "sapi/fpm/tests/bug77780-header-sent-error.phpt",
+    ]:
+        self.rm(f)
+
+
+def pre_install(self):
```
q66, line 173: init_install

contrib/php/template.py
```diff
@@ -0,0 +1,276 @@
...
+    "sqlite-devel",
+    "unixodbc-devel",
+    "zlib-devel",
+]
+checkdepends = ["procps"]
+pkgdesc = "HTML-embedded scripting language"
+maintainer = "Renato Botelho do Couto <renato@netgate.com>"
+license = "PHP-3.01"
+url = "https://www.php.net"
+source = f"{url}/distributions/php-{pkgver}.tar.xz"
```
q66, line 119: tar.gz

contrib/php/files/php-fpm
```diff
@@ -0,0 +1,6 @@
+# php-fpm service
+
+type            = process
+command         = /usr/bin/php-fpm
```
q66, line 4: should force --nodaemonize here

contrib/php/files/php-fpm.conf
```diff
@@ -0,0 +1,143 @@
...
+; The value can vary from -19 (highest priority) to 20 (lowest priority)
+; Note: - It will only work if the FPM master process is launched as root
+;       - The pool process will inherit the master process priority
+;         unless specified otherwise
+; Default Value: no set
+; process.priority = -19
+
+; Send FPM to background. Set to 'no' to keep FPM in foreground for debugging.
+; Default Value: yes
+daemonize = no
```
q66, line 99: if forced in the service file, this can remain commented out

contrib/php/files/sysusers.conf
```diff
@@ -0,0 +1,4 @@
+# Create php-fpm system user
+
+u _php - "php-fpm www user" /nonexistent /usr/bin/nologin
```
q66, line 3: /var/empty for homedir

contrib/php/template.py
nekopsykose: that's just worse than the xz one, there's no point to use the gz tarball

contrib/php/template.py
```diff
@@ -0,0 +1,276 @@
+pkgname = "php"
+pkgver = "8.3.4"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    # common parameters
+    "--datadir=/usr/share/php",
```
q66, line 7: i'd also version these paths like `/usr/share/php8.3` etc

perhaps the package names too

q66: and we may want to have a `-default` metapackage system similar to openjdk and the likes

contrib/php/template.py
nekopsykose: i'm pretty sure nobody is deciding to unilaterally do that and it's a huge overreaction

contrib/php/template.py
rbgarga: Are you thinking to rename binaries as well to make it possible for multiple versions of PHP to co-exist?

rbgarga: I think it's all done

PR #1679, opened 2024-03-27
contrib/ruff: new package (0.3.4)

nekopsykose:


jcgruenhage:
Can we switch to `ruff` for formatting and the CI checks as well? It's a drop in replacement for `black` and a lot faster:

```text
❯ hyperfine 'find main contrib user src -name \'*.py\' -exec black --fast --check {} +' 'find main contrib user src -name \'*.py\' -exec ruff format --check {} +'
Benchmark 1: find main contrib user src -name '*.py' -exec black --fast --check {} +
  Time (mean ± σ):     527.9 ms ±   9.2 ms    [User: 415.1 ms, System: 112.8 ms]
  Range (min … max):   517.2 ms … 548.5 ms    10 runs
 
Benchmark 2: find main contrib user src -name '*.py' -exec ruff format --check {} +
  Time (mean ± σ):      81.4 ms ±   1.4 ms    [User: 59.1 ms, System: 74.0 ms]
  Range (min … max):    78.7 ms …  85.7 ms    34 runs
 
Summary
  find main contrib user src -name '*.py' -exec ruff format --check {} + ran
    6.48 ± 0.16 times faster than find main contrib user src -name '*.py' -exec black --fast --check {} +
```

q66:
i prefer to stick with black especially considering after the recent ci script changes it's no longer meaningfully slow

jcgruenhage:
ah, right, it's not slow anymore.. I just remembered the CI was slow and saw ruff being packaged, but the speed up that was achieved by other means is already good enough, that's true.

nekopsykose:
yeah, i had a pr for that already, and it made format go from `50s`(!) to what you see, but now black is also just a fraction of the whole 20s or so, so it doesn't matter as much. black has really bad symlink behaviour (it was reformatting the files like 12 times over or whatever)

though why this '12' made it go from '1s' to '50s' i have no idea

q66:
it was slow because it was handling every template like 6 times because it treats paths encountered through symlinks as separate (means that every template was done several times through its subpackage links and parent links for contrib etc)

nekopsykose:
that said it would be nice to use ruff instead of flake8 since it seems flake8 right now misses a lot of stuff (like the stuff i fixed in src/ that was never detected). or maybe our flake8 is misconfigured?

q66:
that's something flake8 should fix (maybe we should investigate and report)

jcgruenhage:
Actually: `ruff` without `find` is even quicker, by quite a bit as well, that pushes a full format check down to 60ms:

```text
❯ hyperfine 'black --check main contrib user src' 'ruff format --check main contrib user src'
Benchmark 1: black --check main contrib user src
  Time (mean ± σ):     16.543 s ±  0.211 s    [User: 14.707 s, System: 1.831 s]
  Range (min … max):   16.239 s … 16.836 s    10 runs
 
Benchmark 2: ruff format --check main contrib user src
  Time (mean ± σ):      60.4 ms ±   1.4 ms    [User: 78.3 ms, System: 69.8 ms]
  Range (min … max):    57.6 ms …  65.2 ms    45 runs
 
Summary
  ruff format --check main contrib user src ran
  273.88 ± 7.21 times faster than black --check main contrib user src
```

q66:
anyway I'll merge this as long as ppc64le builder is not busy

nekopsykose:
> Actually: ruff without find is even quicker, by quite a bit as well, that pushes a full format check down to 60ms:  

yeah, by your own measurement it's 60ms as opposed to black being 520ms from te find version above. though the 'exec overhead' is also far lower so i use ruff locally for editor autoformat (it's noticably 0-delay instead of a quarter of a second when triggering). for ci though that 500ms does not mean much to the 20 seconds

PR #1680, opened 2024-03-27
cbuild: enable keeping frame pointers by default

### LINKED ISSUE #1604: -fno-omit-frame-pointer, -Cforce-frame-pointers=yes
arch, fedora, and ubuntu now started doing this, so maybe we should too

long discussion (in arch): https://gitlab.archlinux.org/archlinux/rfcs/-/merge_requests/26  
some implementation notes: https://gitlab.archlinux.org/daandemeyer/rfcs/-/blob/fp/rfcs/0026-fno-omit-frame-pointer.rst  

tl;dr: profiling stuff using dwarf backtraces is often quite unreasonable, https://rwmj.wordpress.com/2023/02/14/frame-pointers-vs-dwarf-my-verdict/ is quite funny. having fp enabled makes it much easier to 'get an accurate flamegraph of why something is slow', quickly, and with low overhead. manually rebuilding stuff with fp locally is usually far too much hassle  

misc other info: https://www.brendangregg.com/blog/2024-03-17/the-return-of-the-frame-pointers.html

overall performance loss is dependent on package and the worst offenders seem to be '10%' for things like python- and that's only in synthetic benches. we could leave a per-package optout too. in general it seems the effect is not noticeable

nekopsykose:
[ci skip]

closes #1604  

---

this is the most naive implementation:  

- does this matter only for stage>0? (shouldn't make a meaningful difference)
- do we want the `-m` arg on every arch? (clang accepts it everywhere, but i don't know outside-of-x86_64 caveats)

after this, and rebuilding musl with it for a test on a simple program, `perf record -F100 -g ...` (`-g` implicit `--call-graph=fp`) works normally.

### DIFF
--- a/Packaging.md
+++ b/Packaging.md
@@ -1697,6 +1697,8 @@ the template including for subpackages:
   disable linker and LTO threads.
 * `linkundefver` *(false)* Pass `--undefined-version` to `ld.lld` to
   bypass version errors in affected packages.
+* `framepointer` *(true)* If enabled, frame pointers will be turned
+  on to make profiling of resultant binaries easier.
 
 The following options apply to a single package and need to be specified
 for subpackages separately if needed:
--- a/main/python/template.py
+++ b/main/python/template.py
@@ -63,6 +63,7 @@ sha256 = "be28112dac813d2053545c14bf13a16401a21877f1a69eb6ea5d84c4a0f3d870"
 # we cannot enable ubsan stuff because there is known UB where tests
 # are just skipped and so on, so be on the safe side for the time being
 hardening = ["vis", "!cfi", "!int"]
+options = ["!framepointer"]
 
 env = {
     # emulate python's configure stuff but with -O2
--- a/src/cbuild/core/profile.py
+++ b/src/cbuild/core/profile.py
@@ -81,6 +81,9 @@ def _get_archflags(prof, tmpl, hard):
     if not hard["ssp"]:
         sflags.append("-fno-stack-protector")
 
+    if opts["framepointer"]:
+        sflags.append("-fno-omit-frame-pointer")
+
     if hard["sst"]:
         sflags.append("-fsanitize=safe-stack")
 
@@ -234,6 +237,9 @@ def _get_rustflags(self, tmpl, name, extra_flags, debug, hardening, shell):
             "--sysroot",
             self.sysroot / "usr",
         ]
+
+    if opts["framepointer"]:
+        bflags += ["-Cforce-frame-pointers=true"]
 
     if tmpl.options["relr"] and self._has_relr(tmpl.stage):
         bflags += ["-Clink-arg=-Wl,-z,pack-relative-relocs"]
--- a/src/cbuild/core/template.py
+++ b/src/cbuild/core/template.py
@@ -392,6 +392,7 @@ default_options = {
     "ltostrip": (False, False),
     "linkparallel": (True, True),
     "linkundefver": (False, False),
+    "framepointer": (True, True),
 }
 
 core_fields = [


q66:
call it just framepointers, make the cflags go earlier in the sequence (right before first sanitize), make the rust flags go earlier in the sequence (before relr), add documentation

the leaf flag is unnecessary (it's the default when `-fno-omit-frame-pointer` is specified on its own), see the arguments passed to cc1 when compiling with `-v`

nekopsykose:
> the leaf flag is unnecessary (it's the default when -fno-omit-frame-pointer is specified on its own), see the arguments passed to cc1 when compiling with -v  

good to know! thanks for checking  

> call it just framepointers, make the cflags go earlier in the sequence (right before first sanitize), make the rust flags go earlier in the sequence (before relr), add documentation  

ack, on it

q66:
unless you intend to keep it for cases when something would explicitly specify `-momit-leaf-frame-pointer` in its project flags, but is that something that realistically happens?

nekopsykose:
hmm.. i can't say i've seen that one before. well, maybe in one or two projects ever? some projects like to pass -fomit-frame-pointer for fun, appended..

nekopsykose:
i guess it would be safer to leave it 'just in case' but it spams more flags, so i'm indifferent either way  

> see the arguments passed to cc1 when compiling with -v  

also weirdly this logs nothing about it (either -f or -fno)  

```
$ cc -O2 -fno-omit-frame-pointer -v main.c
clang version 17.0.6
Target: x86_64-chimera-linux-musl
Thread model: posix
InstalledDir: /usr/bin
System configuration file directory: /etc/clang
 "/usr/bin/clang-17" -cc1 -triple x86_64-chimera-linux-musl -emit-obj -dumpdir a- -disable-free -clear-ast-before-backend -disable-llvm-verifier -discard-value-names -main-file-name main.c -mrelocation-model pic -pic-level 2 -pic-is-pie -mframe-pointer=all -ffp-contract=on -fno-rounding-math -mconstructor-aliases -funwind-tables=2 -target-cpu x86-64 -tune-cpu generic -debugger-tuning=gdb -v -fcoverage-compilation-dir=/tmp/mytemp.PdCGNe -resource-dir /usr/lib/clang/17 -internal-isystem /usr/local/include -internal-externc-isystem /usr/include/fortify -internal-externc-isystem /usr/include -internal-isystem /usr/lib/clang/17/include -include stdc-predef.h -O2 -fdebug-compilation-dir=/tmp/mytemp.PdCGNe -ferror-limit 19 -stack-protector 2 -fgnuc-version=4.2.1 -fcolor-diagnostics -vectorize-loops -vectorize-slp -faddrsig -D__GCC_HAVE_DWARF2_CFI_ASM=1 -o /tmp/main-d22fcd.o -x c main.c
clang -cc1 version 17.0.6 based upon LLVM 17.0.6 default target x86_64-chimera-linux-musl
#include "..." search starts here:
#include <...> search starts here:
 /usr/local/include
 /usr/include/fortify
 /usr/include
 /usr/lib/clang/17/include
End of search list.
 "/usr/bin/ld" -pie -z now -z relro --hash-style=gnu --build-id --eh-frame-hdr -m elf_x86_64 -dynamic-linker /lib/ld-musl-x86_64.so.1 -o a.out /lib/Scrt1.o /lib/crti.o /usr/lib/clang/17/lib/x86_64-chimera-linux-musl/clang_rt.crtbegin.o -L/usr/lib/clang/17/lib/x86_64-chimera-linux-musl -L/lib/ -L/usr/lib/ -L/lib -L/usr/lib --as-needed /tmp/main-d22fcd.o /usr/lib/clang/17/lib/x86_64-chimera-linux-musl/libclang_rt.builtins.a --as-needed -lunwind --no-as-needed --push-state --as-needed -latomic --pop-state -lc /usr/lib/clang/17/lib/x86_64-chimera-linux-musl/libclang_rt.builtins.a --as-needed -lunwind --no-as-needed /usr/lib/clang/17/lib/x86_64-chimera-linux-musl/clang_rt.crtend.o /lib/crtn.o
```

q66:
actually just call it framepointer, makes more sense that way

PR #1685, opened 2024-03-28
certbot deps: python-time-machine, python-freezegun, python-distlib, python-pytest-env

a16bitsysop:


contrib/python-pytest-env/template.py
```diff
@@ -0,0 +1,23 @@
...
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-hatchling",
+    "python-hatch_vcs",
+    "python-wheel",
+]
+depends = ["python-pytest"]
+checkdepends = depends
```
nekopsykose, line 13:
```suggestion
checkdepends = list(depends)
```

contrib/python-py-cpuinfo/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "python-py-cpuinfo"
```
nekopsykose, line 1: added this for benchmark so you can rebase it out

contrib/python-distlib/template.py
```diff
@@ -0,0 +1,29 @@
...
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+    "python-wheel",
+]
+checkdepends = ["python-pytest"]
+pkgdesc = (
+    "Low-level functions for packaging and distribution of Python software"
+)
```
nekopsykose, line 18:
```suggestion
pkgdesc = "Functions for packaging and distribution of Python software"
```

contrib/python-pytest-env/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "python-pytest-env"
+pkgver = "1.1.3"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-hatchling",
+    "python-hatch_vcs",
+    "python-wheel",
```
nekopsykose, line 10: you can leave wheel out when hatchling is used

contrib/python-py-cpuinfo/template.py
a16bitsysop: Okay will drop

contrib/python-pytest-env/template.py
a16bitsysop: Okay will take out

PR #1688, opened 2024-03-29
contrib/git-grab: new package (2.1.0)

wezm:


contrib/git-grab/template.py
```diff
@@ -0,0 +1,19 @@
+pkgname = "git-grab"
+pkgver = "2.1.0"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo"]
+makedepends = [
+    "rust-std",
+]
```
nekopsykose, line 8:
```suggestion
makedepends = ["rust-std"]
```

when it's 1 you can leave it single line

wezm: Fixed now

nekopsykose: black doesn't have an opinion about it (if you have trailing , it always leaves it)

PR #1691, opened 2024-03-29
main/gnome-contacts: new package (46.0)

triallax:
In `main` because I plan to add it to `main/gnome-apps` eventually.


triallax:
#1741

PR #1698, opened 2024-03-29
fix drobilla.net update checks

jcgruenhage:
I've checked https://repo.chimera-linux.org/cports-updates/cports-updates.txt to see what packages I need to update and decided to fix a bunch of the `CAUTION: no version found for '<package>'` warnings in there.

The only thing we're still missing is libreoffice, but I really couldn't find a good url for fetching that.

### DIFF
--- a/contrib/suil/update.py
+++ b/contrib/suil/update.py
@@ -1,0 +1,2 @@
+url = "https://drobilla.net/category/suil/"
+pattern = r"suil-([\d.]+)\.tar\.xz"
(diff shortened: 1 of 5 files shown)


contrib/suil/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://drobilla.net/category/suil/"
+pattern = r"suil-([0-9.]+).tar.xz"
```
mhmdanas, line 2:
```suggestion
pattern = r"suil-([\d.]+)\.tar\.xz"
```

ditto for the others

jcgruenhage: I've applied that change, but we've got `[0-9.]` in 43 other `update.py` files. Should these all be changed as well to be consistent?

PR #1699, opened 2024-03-30
contrib/fastfetch: new package (2.8.10)

JamiKettunen:
The runtime soft-dep situation still needs to be cleared up but otherwise this should be ok. Check out `fastfetch -c all` for everything this can do
image


contrib/fastfetch/template.py
```diff
@@ -0,0 +1,62 @@
+pkgname = "fastfetch"
+pkgver = "2.8.10"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DCMAKE_BUILD_TYPE=RelWithDebInfo", "-DBUILD_TESTS=ON"]
```
nekopsykose, line 5:
```suggestion
configure_args = ["-DBUILD_TESTS=ON"]
```

this doesn't do anything

contrib/ddcutil/template.py
```diff
@@ -1,6 +1,6 @@
 pkgname = "ddcutil"
```
nekopsykose, line 1: this is unrelated to this mr

contrib/fastfetch/template.py
nekopsykose: (the -debug in the version is based on NDEBUG which you can put into cflags)

JamiKettunen:
I noticed `LM: turnstiled (unspecified)` looks odd for our the detected login manager: https://github.com/fastfetch-cli/fastfetch/blob/2.8.10/src/detection/lm/lm_linux.c#L137

I assume this will happen with other display managers also and it's not just a GDM thing. Checking `/var/run/systemd/sessions/` I have `2` and `c2` there (along with `.ref` counterparts) with my `XDG_SESSION_ID=c2`:
```
...
TYPE=unspecified
...
SERVICE=turnstiled
...
```
the `2` appears to be the one it gets everywhere else normally:
```
...
TYPE=wayland
...
SERVICE=gdm-password
...
```
just wondering if this is somehow fixable

contrib/fastfetch/template.py
```diff
@@ -0,0 +1,63 @@
...
+]
+pkgdesc = "Fast neofetch-like system information fetching tool"
+maintainer = "Jami Kettunen <jami.kettunen@protonmail.com>"
+license = "MIT"
+url = "https://github.com/fastfetch-cli/fastfetch"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "e3ea8388d79cc02ed1a6e45b8e5fec4068507ca03923386bdc9a7b3ff02d2a52"
+tool_flags = {"CFLAGS": ["-DNDEBUG"]}
+# CFI dies immediately (ffPlatformPathAddHome at FFlist.c:31:12)
+hardening = ["vis", "!cfi"]
```
mhmdanas, line 31: have a todo here?
```suggestion
# TODO: CFI dies immediately (ffPlatformPathAddHome at FFlist.c:31:12)
hardening = ["vis", "!cfi"]
```

contrib/fastfetch/template.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@subpackage("fastfetch-full")
```
q66, line 34: i don't much see the point of this package?

JamiKettunen: I guess most people can install the extra stuff like `ddcutil` themselves if they want display backlight level infos and whatnot (and almost everything else is automatically installed on a full desktop environment setup anyway). 

but should `lscpu` for example be in `depends` since it's a fallback to determine proper CPU model on some ARM/RISC-V systems? https://github.com/fastfetch-cli/fastfetch/blob/2.8.10/src/detection/cpu/cpu_linux.c#L276-L347

nekopsykose: lscpu seems ok

contrib/fastfetch/template.py
```diff
@@ -0,0 +1,32 @@
...
+    "vulkan-headers",
+    "vulkan-loader-devel",
+    "wayland-devel",
+    "xfconf-devel",
+    "zlib-devel",
+]
+depends = ["lscpu"]
+pkgdesc = "Fast neofetch-like system information fetching tool"
+maintainer = "Jami Kettunen <jami.kettunen@protonmail.com>"
+license = "MIT"
```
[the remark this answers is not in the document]
JamiKettunen: ci would be failing if it wasn't already being done: https://github.com/fastfetch-cli/fastfetch/blob/2.8.10/CMakeLists.txt#L1106-L1109

PR #1700, opened 2024-03-30
contrib/ddcutil: add trigger to load i2c-dev

JamiKettunen:
This is required to use ddcutil (via DDC/CI) and would already get loaded upon rebooting.
```
# ddcutil getvcp 1 10
   Unexpected error. Unable to open sysfs directory /sys/class/drm/card0-HDMI-A-1/ddc/i2c-dev: No such file or directory
No /dev/i2c devices exist.
ddcutil requires module i2c-dev.
```


q66:
fa451634f44c8e905f4ef313eea7d625cba7e30c

PR #1710, opened 2024-03-31
more precise update check regexes

jcgruenhage:


q66:
is there any particular reason for this? because if not, it just adds noise

nekopsykose:
the 0-9 -> \d looks fine, you can make that 1 commit i guess

the . -> \. is quite pointless because that will never falsely match something already in practice

jcgruenhage:
> is there any particular reason for this? because if not, it just adds noise

I've done this based on https://github.com/chimera-linux/cports/pull/1698#pullrequestreview-1969693765.

I had done `[0-9]` a bunch myself in other places, based on looking at other files, and if `\d` is preferred over `0-9` then we should have that consistent so that new contributors (like myself) don't do that based off of other files in this repo.

> the 0-9 -> \d looks fine, you can make that 1 commit i guess

I don't have a problem squashing this together, if that's preferred

> the . -> \. is quite pointless because that will never falsely match something already in practice

I can drop those if you want, again, doing this based on the review I got in https://github.com/chimera-linux/cports/pull/1698#pullrequestreview-1969693765 to make stuff consistent.

q66:
imo it doesn't matter at all, you can use `[0-9]` or `\d`, it's the same

jcgruenhage:
If y'all don't think it's worthwhile to make this more consistent then I'll close this. I don't particularly care either, the only reason I did it was the review telling me to change it for that other PR.

q66:
i dunno if @nekopsykose can think of a reason, so let's wait a bit; as i see it, there are places where striving for consistency is important, but here i cannot think of any particularly important/meaningful reason

jcgruenhage:
I personally also prefer `[0-9]`, same reasoning..

The reason on my side was that I had wrote `[0-9]` in a bunch of places myself, and after that PR review in the other one I felt like that was the less preferred variant in here. If there's no "less preferred variant", or if it's even the other way around, sure, these changes don't make sense at all.

nekopsykose:
i don't have any particular reason no, it's all the same to me for these. i type \d but if everyone wants 0-9 that's fine, or just leaving it inconsistent :)

jcgruenhage:
Then I'll just leave it incosistent. There's no reason to make this consistent if no one particularly cares.

jcgruenhage:
(We can still do this later, this was just a bunch of find + grep + sed and replicating it is a matter of minutes)

q66:
sgtm

PR #1716, opened 2024-04-01
contrib/rust-audit-info: new package (0.5.2)

jcgruenhage:


contrib/rust-audit-info/template.py
```diff
@@ -0,0 +1,16 @@
+pkgname = "rust-audit-info"
+pkgver = "0.5.2"
+pkgrel = 0
+build_wrksrc = "rust-audit-info"
+build_style = "cargo"
+hostmakedepends = ["cargo"]
```
nekopsykose, line 6: you also need makedepends = ["rust-std"] in all of these

jcgruenhage: which ones does "all of these" mean? Should I also open PRs for the ones you've already merged?

nekopsykose: yeah, i just missed those

Issue #1718, opened 2024-04-01
podman: cannot stop rootless container created with --pid host


JamiKettunen:
This is my last major blocker for adding `distrobox` into the repo, reproducible with:
```sh
podman create \
--name test --ipc host --network host \
--privileged --security-opt label=disable \
--user root:root --pid host \
--volume /:/run/host:rslave \
--volume /dev:/dev:rslave \
--volume /sys:/sys:rslave \
--volume /tmp:/tmp:rslave \
--volume "$HOME":"$HOME":rslave \
--volume /etc/hosts:/etc/hosts:ro \
--volume /etc/localtime:/etc/localtime:ro \
--volume /etc/resolv.conf:/etc/resolv.conf:ro \
--ulimit host --annotation run.oci.keep_original_groups=1 \
--mount type=devpts,destination=/dev/pts \
--userns keep-id \
alpine:latest \
sleep infinity

podman start test
podman ps
podman --log-level debug stop test
```
with the last command output containing:
```
DEBU[0000] Sending signal 15 to container ... 
DEBU[0010] Timed out stopping container ... with SIGTERM, resorting to SIGKILL: given PID did not die within timeout 
WARN[0010] StopSignal SIGTERM failed to stop container test in 10 seconds, resorting to SIGKILL 
DEBU[0010] Sending signal 9 to container ... 
Error: given PID did not die within timeout
```
workaround to get back to a stopped state: `pkill conmon`

fwiw attempting the same on Void musl `SIGTERM` is logged but it appears to go straight for `SIGKILL` immediately afterward which does work:
```
DEBU[0000] Stopping container ... (PID ...) 
DEBU[0000] Sending signal 15 to container ... 
DEBU[0000] Sending signal 9 to container ... 
DEBU[0000] Cleaning up container ... 
```

nekopsykose:
as a first step, could you juggle around the versions of stuff here to make podman etc match? maybe this is a 5.0 regression

q66:
i can reproduce this, interestingly using `crun kill` (or `runc kill`, whichever runtime you are using) works as expected

q66:
another thing: using `crun kill` works, but `crun kill --all` (which podman seemingly invokes) does not, both immediately exit but only one results in the container being stopped

q66:
when you look at the container status:

```
$ cat /run/user/1000/crun/4b92c2aeb85d3299596a764de4abf8886b5e3405c9503ea767779ff0c07957b9/status
{
[...]

PR #1726, opened 2024-04-01
contrib/distrobox: new package (1.7.2.1)

JamiKettunen:
Probably doesn't make sense to merge until #1718 is fixed so marking as a draft atm.

### DIFF
--- a/contrib/distrobox/template.py
+++ b/contrib/distrobox/template.py
@@ -1,0 +1,16 @@
+pkgname = "distrobox"
+pkgver = "1.7.2.1"
+pkgrel = 0
+depends = ["curl", "podman", "polkit"]
+pkgdesc = "Use any Linux distribution inside your terminal"
+maintainer = "Jami Kettunen <jami.kettunen@protonmail.com>"
+license = "GPL-3.0-only"
+url = "https://distrobox.it"
+source = f"https://github.com/89luca89/distrobox/archive/{pkgver}.tar.gz"
+sha256 = "ff2cca0c6334fff6ed577d23f68a6746ad4009f42d8a45eef5ca3850c895a4bb"
+
+
+def do_install(self):
+    self.do("./install", "--prefix", f"{self.chroot_destdir}/usr")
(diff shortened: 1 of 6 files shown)


contrib/distrobox/template.py
```diff
@@ -0,0 +1,28 @@
...
+
+def do_install(self):
+    self.do("./install", "--prefix", f"{self.chroot_destdir}/usr")
+    # conf
+    self.install_file(self.files_path / "distrobox.conf", "usr/share/distrobox")
+    # docs
+    self.install_file("docs/*.md", "usr/share/doc/distrobox", glob=True)
+
+
+@subpackage("distrobox-podman")
```
JamiKettunen, line 21: Technically `docker-cli` also should work here but I wasn't sure how to make the backend selection

contrib/distrobox/template.py
```diff
@@ -0,0 +1,28 @@
...
+source = f"https://github.com/89luca89/distrobox/archive/{pkgver}.tar.gz"
+sha256 = "71af97c7033cdd3f2679e922b52e5b6c801d697af63bbe3f9d771948c6575e90"
+
+
+def do_install(self):
+    self.do("./install", "--prefix", f"{self.chroot_destdir}/usr")
+    # conf
+    self.install_file(self.files_path / "distrobox.conf", "usr/share/distrobox")
+    # docs
+    self.install_file("docs/*.md", "usr/share/doc/distrobox", glob=True)
```
mhmdanas, line 18: those comments aren't too useful and should be removed i think

JamiKettunen: sure, added also missing `polkit` dep for rootful container `.desktop` launcher auths via `pkexec`

JamiKettunen:
Now this should be more or less usable, not entirely sure if the hacks for rootless container stuff are acceptable

contrib/distrobox/template.py
JamiKettunen: Ended up dropping this for now since apparently rootless containers don't work with it, let me know if I should make `podman` a soft-dep still

PR #1727, opened 2024-04-02
main/poppler: update and run tests; contrib/libreoffice: update

triallax:
`libtommath-devel` is only needed when embedded firebird is used, but we pass `--disable-firebird-sdbc` to configure and so firebird is disabled altogether.

[ci skip]

### DIFF
--- a/main/poppler/template.py
+++ b/main/poppler/template.py
@@ -1,5 +1,6 @@
 pkgname = "poppler"
-pkgver = "24.03.0"
+pkgver = "24.04.0"
+_test_commit = "400f3ff05b2b1c0ae17797a0bd50e75e35c1f1b1"
 pkgrel = 0
 build_style = "cmake"
 configure_args = [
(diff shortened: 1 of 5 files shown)


nekopsykose:
should roll poppler into this too

mhmdanas:
Note: tested inkscape build after poppler bump but not libreoffice

main/poppler/template.py
```diff
@@ -35,10 +36,21 @@
...
+source_paths = [".", "testdata"]
+sha256 = [
+    "1e804ec565acf7126eb2e9bb3b56422ab2039f7e05863a5dfabdd1ffd1bb77a7",
+    "0d850a1d06944671c991be6822b7146ade401b06aad560ff39b254a028074525",
+]
+
+
+def init_configure(self):
+    self.configure_args.append(
+        f"-DTESTDATADIR=/builddir/{self.wrksrc}/{source_paths[1]}"
```
nekopsykose, line 52:
```suggestion
        f"-DTESTDATADIR=/builddir/{self.wrksrc}/testdata"
```

PR #1729, opened 2024-04-02
contrib/furnace: update to 0.6.2

z-erica:


nekopsykose:
it's already the latest, isn't it

nekopsykose:
also incidentally you have the wrong checke summe's

PR #1738, opened 2024-04-02
contrib/hugo: new package (0.124.1)

ifreund:


contrib/hugo/template.py
```diff
@@ -0,0 +1,33 @@
...
+build_style = "go"
+hostmakedepends = ["go"]
+pkgdesc = "Fast and flexible static site generator"
+maintainer = "Isaac Freund <mail@isaacfreund.com>"
+license = "Apache-2.0"
+url = "https://gohugo.io"
+source = f"https://github.com/gohugoio/hugo/archive/v{pkgver}.tar.gz"
+sha256 = "0beb0436f6bd90abb425523229a37f1d31e2e9c7ba9fac4556a72aab3b11bfef"
+# tests require network access
+options = ["!check"]
```
mhmdanas, line 13:
```suggestion
# tests require network access
# manpages and completions are generated by running binary so no cross
options = ["!check", "!cross"]
```

mhmdanas (changes requested):

contrib/hugo/template.py
ifreund: Added, thanks

contrib/hugo/template.py
```diff
@@ -0,0 +1,34 @@
+pkgname = "hugo"
+pkgver = "0.124.1"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go"]
+pkgdesc = "Fast and flexible static site generator"
```
nekopsykose, line 6:
```suggestion
pkgdesc = "Static site generator"
```

ifreund: done :)

PR #1741, opened 2024-04-03
Add GNOME weather, contacts, and maps

triallax:


q66:
enable in gnome-apps and bump that

q66:
maybe even fold together with the gnome-maps thing if that works?

triallax:
i can do that, and gnome-contacts too?

q66:
yeah

q66:
gnome-maps needs a version update

PR #1747, opened 2024-04-03
main/cargo: fix build

rbgarga:
Without sqlite-devel it fails with this error:

note: ld: error: unable to find library -lsqlite3

### DIFF
--- a/main/cargo/template.py
+++ b/main/cargo/template.py
@@ -14,7 +14,7 @@ hostmakedepends = [
     "pkgconf",
     "zlib-devel",
 ]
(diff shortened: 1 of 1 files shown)


mhmdanas:
i think you want to bump the pkgrel for this because it changes the built package (links with system sqlite instead of using vendored)

q66:
it doesn't matter, it will update soon for new version anyway

q66:
this is missing sqlite-devel-static for bootstrap builds

q66:
hm, or maybe not actually, this will probably link dynamically in either case

PR #1750, opened 2024-04-03
main/libbs2b: fix -Werror=format-security

rbgarga:
Without this change it fails like:

```/builddir/libbs2b-3.1.0/src/bs2bconvert.c:156:11: error: format string is not a string literal (potentially insecure) [-Werror,-Wformat-security]
  156 |                 printf( sf_strerror( NULL ) );
      |                         ^~~~~~~~~~~~~~~~~~~
/builddir/libbs2b-3.1.0/src/bs2bconvert.c:156:11: note: treat the string as an argument to avoid this
  156 |                 printf( sf_strerror( NULL ) );
      |                         ^
      |                         "%s",```

### DIFF
--- a/main/libbs2b/patches/fix-build.patch
+++ b/main/libbs2b/patches/fix-build.patch
@@ -1,0 +1,13 @@
+diff --git a/src/bs2bconvert.c b/src/bs2bconvert.c
+index 24531b2..5e8c992 100644
+--- a/src/bs2bconvert.c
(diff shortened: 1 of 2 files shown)


nekopsykose:
these aren't 'clang 17' but rather -Werror=format-security instead

nekopsykose:
should also bump the rel, sorry didn't see that first time

q66:
bump rel

PR #1757, opened 2024-04-04
contrib/qt6: update to 6.7.0

nekopsykose:


nekopsykose:
> qt6-qt{multimedia,qt5compat,tools,wayland} seem to fail during configure due to not found libQt6QmlBuiltins.a (with qt6-qtdeclarative-devel-static ofc gets past but I'm unsure it's the right solution, qt6-qtdeclarative probably needs some fix)  

no, they changed the cmake config in declarative so it will always fail if the static libs aren't found. it works here because of !splitstatic

nekopsykose:
this should work, i don't see any other way to do it. it's not even a static/dynamic thing (qmlbuiltins is static-only afaict), so to leave qtdeclarative usable it can't be split to separate -static (or every consumer needs to always add -devel-static which is a bit pointless)

PR #1743, opened 2024-04-03
armhf support (base-cbuild + base-minimal)

z-erica:
blocked on:
- [ ] autoconf fails the native build check stage: https://pastebin.com/BvgzCC06
- [ ] zlib cross built is busted, makes apk.static segfault (its fine w a natively built zlib)
- [x] `llvm-17.0.6-r4: update available: 17.0.6 -> 18.1.2`
- [x] util-linux fails to cross build, bash-completion not found by configure script
- [x] util-linux fails to cross build: `dmesg-2.39.3-r0: ERROR: take(): path 'usr/share/bash-completion/completions/dmesg' did not match anything`, fixed by https://github.com/chimera-linux/cports/commit/626d44253fc27a0296c16dd6b729a8daad449d7b

additional base-core blockers:
- [ ] btrfs-progs cross build fails: `./btrfsutilpy.h:27:10: fatal error: 'Python.h' file not found`

PR #1759, opened 2024-04-04
armhf+armv7 support (base-cross)

z-erica:
split off from https://github.com/chimera-linux/cports/pull/1743
- [x] _possibly_ armv7 (armv6?) as well


q66:
this looks pretty much good to merge, however i don't see why 0007-compiler-rt-build-crt-in-runtimes-build.patch is needed for libcxx-cross

q66:
i have the necessary changes in my llvm18 batch, so i will close this PR

PR #1761, opened 2024-04-04
contrib/mdadm: new package (4.2)

LokiRazer:
need some help on this pr - sorry is my first PR on a OSS Project


q66:
there are a lot more things that one would need to do (dinit service, initramfs-tools hooks, etc) and it should probably be in main/ (and a dependency of base-core-fs)

q66:
i'll take care of it myself a bit later, this may be a bit too much for first time packaging

PR #1765, opened 2024-04-05
main/fprintd: update to 1.94.3

mhmdanas:


mhmdanas:
This version also supports `basu` for dbus instead of `elogind`, but `basu` is in contrib but we'd have to move it to `main`. What should I do?

PR #1776, opened 2024-04-07
Stockfish 16.1

shtayerc:


contrib/stockfish/template.py
```diff
@@ -0,0 +1,57 @@
...
+    case "ppc":
+        make_build_args.append("ARCH=ppc-32")
+    case "ppc64" | "ppc64le":
+        make_build_args.append("ARCH=ppc-64")
+    case "aarch64":
+        make_build_args.append("ARCH=armv8")
+    case "riscv64":
+        make_build_args.append("ARCH=riscv64")
+    case _:
+        broken = "arch not supported"
```
nekopsykose, line 43: this is a bit of a meme because nothing in the makefile actually does anything with these values. they just pass some not very interesting compiler flags and that's it. it doesn't seem to actually be arch restricted..

the same kind of thing also goes for an lto section in it. really the whole makefile is a mess. maybe you feel like just removing this check since it's not really needed?

if not, you should set archs= above too.

contrib/stockfish/template.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def post_extract(self):
+    # copy the files over
+    for s in source:
+        if s.startswith("!"):
+            s = s[1:]
+            self.cp(
+                self.sources_path / s[s.rfind("/") + 1 :], self.build_wrksrc
+            )
```
nekopsykose, line 53: there's only two things to copy, so just unroll the loop and do two simple self.cp for them

contrib/stockfish/template.py
```diff
@@ -0,0 +1,57 @@
...
+    f"!https://tests.stockfishchess.org/api/nn/{_net_file_big}",
+    f"!https://tests.stockfishchess.org/api/nn/{_net_file_small}",
+]
+sha256 = [
+    "a5f94793b5d4155310397ba89e9c4266570ef0f24cd47de41a9103556f811b82",
+    "b1a57edbea574ca8b88d6837473845791beb53d885f87f86d5ccdd5659fbf3b2",
+    "baff1ede1f90c1dd1b4f772f1eff29848821801e8186345da7f0eb4121bd6f63",
+]
+
+tool_flags = {"LDFLAGS": ["-lpthread", "-Wl,-z,stack-size=2097152"]}
```
nekopsykose, line 27: -lpthread here does nothing

contrib/stockfish/template.py
```diff
@@ -0,0 +1,57 @@
...
+    f"{url}/archive/sf_{pkgver}.tar.gz",
+    f"!https://tests.stockfishchess.org/api/nn/{_net_file_big}",
+    f"!https://tests.stockfishchess.org/api/nn/{_net_file_small}",
+]
+sha256 = [
+    "a5f94793b5d4155310397ba89e9c4266570ef0f24cd47de41a9103556f811b82",
+    "b1a57edbea574ca8b88d6837473845791beb53d885f87f86d5ccdd5659fbf3b2",
+    "baff1ede1f90c1dd1b4f772f1eff29848821801e8186345da7f0eb4121bd6f63",
+]
+
```
nekopsykose, line 26:
```suggestion
```

contrib/stockfish/template.py
```diff
@@ -0,0 +1,57 @@
...
+build_wrksrc = "src"
+build_style = "makefile"
+make_cmd = "gmake"
+make_build_target = "build"
+make_build_args = ["optimize=no"]
+hostmakedepends = ["gmake"]
+pkgdesc = "Free UCI chess engine derived from Glaurung"
+maintainer = "shtayerc <david.murko@mailbox.org>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/official-stockfish/Stockfish"
```
nekopsykose, line 13:
```suggestion
url = "https://stockfishchess.org"
```

contrib/python-chess/template.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def do_install(self):
+    self.do(
+        "python",
+        "setup.py",
+        "install",
+        "--prefix=/usr",
+        f"--root={self.chroot_destdir}",
+    )
```
nekopsykose, line 32: this should use pep_517 like the others, and it works fine. things don't need a pyproject.toml to use it, it uses setuptools normally

contrib/stockfish/template.py
shtayerc: I guess we can skip arch specific stuff without patching (SUPPORTED_ARCH)

contrib/python-chess/template.py
shtayerc: I could not get default pep517 check working. It does not detect any test file and then fails the build.

nekopsykose: setup.py test is fine here because that just invokes `python -m unittest -v`. i meant the rest of the stuff

nekopsykose:
you also need to add `$(CXXFLAGS)` to this line with a patch https://github.com/official-stockfish/Stockfish/blob/0716b845fdef8a20102b07eaec074b8da8162523/src/Makefile#L1051 so they're present at link time

PR #1781, opened 2024-04-08
contrib/{glycin-loaders,loupe}: new packages

triallax:
- **contrib/glycin-loaders: new package (1.0.1)**
- **contrib/loupe: new package (46.2)**

### DIFF
--- a/contrib/glycin-loaders/template.py
+++ b/contrib/glycin-loaders/template.py
@@ -1,0 +1,45 @@
+pkgname = "glycin-loaders"
+pkgver = "1.0.1"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Dtests=false"]
+hostmakedepends = [
+    "cargo",
+    "gettext",
+    "meson",
+    "pkgconf",
+]
+makedepends = [
+    "cairo-devel",
+    "libheif-devel",
+    "libjxl-devel",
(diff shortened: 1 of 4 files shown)


contrib/glycin-loaders/template.py
```diff
@@ -0,0 +1,50 @@
...
+    self.make_env.update(renv)
+
+
+def post_install(self):
+    for loader in [
+        "glycin-heif",
+        "glycin-image-rs",
+        "glycin-jxl",
+        "glycin-svg",
+    ]:
```
triallax, line 45: Do you think it'd be fine here to dynamically find the loaders that were built, so that we don't miss something added in the future inadvertently? If so, how can I do that?

nekopsykose: i think iterating on filepaths and checking for executable && !directory in target/release lets you do that

contrib/loupe/template.py
```diff
@@ -0,0 +1,50 @@
...
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "cargo",
+    "meson",
+    "pkgconf",
+    "glib",
+    "gettext",
+    "itstool",
+    "desktop-file-utils",
```
nekopsykose, line 12: sort

PR #1782, opened 2024-04-08
contrib/pgn-extract: new package (22.11)

shtayerc:


contrib/pgn-extract/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "pgn-extract"
+pkgver = "22.11"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_check_target = "all"
```
nekopsykose, line 6: pass make_build_args = ["OPTIMISE="] too  

also add `$(CFLAGS)` to

```makefile
pgn-extract : $(OBJS)
	$(CC) $(DEBUGINFO) $(ORIGCFLAGS) $(CPPFLAGS) $(LDFLAGS) $(OBJS) $(LIBS) -o pgn-extract
```

contrib/pgn-extract/template.py
```diff
@@ -0,0 +1,21 @@
...
+make_check_args = ["--directory", "test"]
+make_use_env = True
+hostmakedepends = ["gmake"]
+pkgdesc = "Portable Game Notation (PGN) Manipulator for Chess Games"
+maintainer = "shtayerc <david.murko@mailbox.org>"
+license = "GPL-3.0-or-later"
+url = "https://www.cs.kent.ac.uk/people/staff/djb/pgn-extract"
+source = f"{url}/pgn-extract-{pkgver.replace('.' , '-')}.tgz"
```
nekopsykose, line 14:
```suggestion
source = f"{url}/pgn-extract-{pkgver.replace('.', '-')}.tgz"
```

contrib/pgn-extract/template.py
shtayerc: If i just patch $(CFLAGS) here it breaks. I guess we only need cflags from cports

nekopsykose: ah, sure, cause `-c` is in them

PR #1783, opened 2024-04-09
contrib/nbfc-linux-*: new package

Sid127:
Installed with `apk --allow-untrusted` on local machine and tested thoroughly, works as intended.

### DIFF
--- a/contrib/nbfc-linux/files/nbfc
+++ b/contrib/nbfc-linux/files/nbfc
@@ -1,0 +1,7 @@
+# NoteBook FanControl service
+
+type = process
+command = /usr/bin/nbfc_service
+depends-on = local.target
+logfile = /var/log/nbfc.log
+logile-permissions = 644
--- a/contrib/nbfc-linux/template.py
+++ b/contrib/nbfc-linux/template.py
@@ -1,0 +1,21 @@
+pkgname = "nbfc-linux"
+pkgver = "0.1.15"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_build_args = ["confdir=/etc"]
+make_install_args = ["confdir=/etc"]
+hostmakedepends = ["gmake"]
+pkgdesc = "NoteBook FanControl for Linux"
+maintainer = "Sid Pranjale <mail@sidonthe.net>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/nbfc-linux/nbfc-linux"
+source = f"{url}/archive/{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "d1b37800886a66a6ef5a98ba45a88812fa451579028f5e4032375c62f46df535"
+# no tests
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_service(self.files_path / "nbfc")
+    self.rm(self.destdir / "usr/lib/systemd/system/nbfc_service.service")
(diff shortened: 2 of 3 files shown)


contrib/nbfc-linux/files/nbfc_service
```diff
@@ -0,0 +1,9 @@
+# NoteBook FanControl service
+
+type            = bgprocess
+command         = /usr/bin/nbfc start
+stop-command    = /usr/bin/nbfc stop
+pid-file        = /run/nbfc_service.pid
+restart         = true
```
nekopsykose, line 7:
```suggestion
```

default

contrib/nbfc-linux/template.py
```diff
@@ -0,0 +1,20 @@
+pkgname = "nbfc-linux"
+pkgver = "0.1.15"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_build_args = ["confdir=/etc"] 
+make_install_args = ["confdir=/etc"] 
+hostmakedepends = ["gmake"]
+pkgdesc = "Lightweight C port of NoteBook FanControl"
```
nekopsykose, line 9:
```suggestion
pkgdesc = "NoteBook FanControl for Linux"
```

nekopsykose:
it should be squashed to 1 commit, and the `-*` shouldn't be there

contrib/nbfc-linux/files/nbfc_service
```diff
@@ -0,0 +1,8 @@
+# NoteBook FanControl service
+
+type            = bgprocess
+command         = /usr/bin/nbfc start
```
nekopsykose, line 4: this should be named 'nbfc' rather than 'nbfc_service' pretty sure, since the _service doesn't mean anything

Sid127: I'm just following how upstream named the service for other init systems

nekopsykose:
set the lint ci too

contrib/nbfc-linux/files/nbfc_service
Sid127: I do think it is important to offer an interface that matches other distros, but sure, renamed

contrib/nbfc-linux/files/nbfc
```diff
@@ -0,0 +1,6 @@
+# NoteBook FanControl service
+
+type            = process
+command         = /usr/bin/nbfc_service
+before          = login.target
+depends-on      = early-fs-local.target
```
q66, line 6:
```suggestion
depends-on = pre-local.target
```

triallax:
commit message should be `contrib/nbfc-linux: new package (0.1.15)`, and also get rid of the unnecessary commit description (`contrib/nbfc-linux: update to correct pkg`)

contrib/nbfc-linux/files/nbfc
```diff
@@ -0,0 +1,8 @@
+# NoteBook FanControl service
+
+type               = process
```
q66, line 3: drop the alignment, we're not going to be doing that anymore

contrib/nbfc-linux/files/nbfc
```diff
@@ -0,0 +1,8 @@
...
+command            = /usr/bin/nbfc_service
+before             = login.target
```
q66, line 5: any particular reason to force it to start before login?

contrib/nbfc-linux/files/nbfc
```diff
@@ -0,0 +1,8 @@
...
+depends-on         = pre-local.target
```
q66, line 6: local.target is fine

contrib/nbfc-linux/files/nbfc
Sid127: nope, can be removed

PR #1785, opened 2024-04-09
contrib/intel-undervolt: new package

Sid127:
Confirmed tested and working as on arch linux. Did not test the -loop service but it should work the same. Tested by the following means:

- set a large compile with make -j9
- apply undervolt config
- notice thermal throttling to be reduced
- revert config
- throttling is back
- apply config again and reboot, then use `intel-undervolt read` to check if config is applied on boot

### DIFF
--- a/contrib/intel-undervolt/files/intel-undervolt
+++ b/contrib/intel-undervolt/files/intel-undervolt
@@ -1,0 +1,5 @@
+# intel-undervolt service
+
+type = scripted
+command = /usr/bin/intel-undervolt apply
+depends-on = local.target
--- a/contrib/intel-undervolt/files/intel-undervolt-loop
+++ b/contrib/intel-undervolt/files/intel-undervolt-loop
@@ -1,0 +1,5 @@
+# intel-undervolt daemon service
+
+type = process
+command = /usr/bin/intel-undervolt daemon
+depends-on = local.target
--- a/contrib/intel-undervolt/patches/configure-gsed.patch
+++ b/contrib/intel-undervolt/patches/configure-gsed.patch
@@ -1,0 +1,20 @@
+--- a/configure
++++ b/configure
+@@ -65,7 +65,7 @@ sedarg() {
+   printf '%s' "s,^\($1 =\).*$,\1 $2,"
+ }
+ 
+-sed Makefile.in \
++sed \
+ -e "`sedcond SYSTEMD "$enable_systemd"`" \
+ -e "`sedcond ELOGIND "$enable_elogind"`" \
+ -e "`sedcond OPENRC "$enable_openrc"`" \
+@@ -74,7 +74,7 @@ sed Makefile.in \
+ -e "`sedarg RUNSTATEDIR "$runstatedir"`" \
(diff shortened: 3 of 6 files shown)


triallax:
commit message should be `user/intel-undervolt: new package (1.7)`

user/intel-undervolt/files/intel-undervolt
```diff
@@ -0,0 +1,6 @@
+# Intel Undervolt service
+
+type       = scripted
```
q66, line 3: ditto here regarding paddings

user/intel-undervolt/files/intel-undervolt
```diff
@@ -0,0 +1,6 @@
...
+command    = /usr/bin/intel-undervolt apply
+before     = login.target
```
q66, line 5: ditto with the question

user/intel-undervolt/files/intel-undervolt
```diff
@@ -0,0 +1,6 @@
...
+depends-on = pre-local.target
```
q66, line 6: local.target is likewise fine here

user/intel-undervolt/files/intel-undervolt-loop
```diff
@@ -0,0 +1,6 @@
+# Intel Undervolt service
+
+type       = process
+command    = /usr/bin/intel-undervolt daemon
+before     = login.target
```
q66, line 5: ditto all the stuff here

user/intel-undervolt/files/intel-undervolt
```diff
@@ -0,0 +1,6 @@
+# Intel Undervolt service
```
q66, line 1: `# intel-undervolt service`

user/intel-undervolt/files/intel-undervolt-loop
```diff
@@ -0,0 +1,6 @@
+# Intel Undervolt service
```
q66, line 1: `# intel-undervolt daemon service`

user/intel-undervolt/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "intel-undervolt"
+pkgver = "1.7"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_build_args = ["CC=cc"]
```
q66, line 6: these are not necessary

user/intel-undervolt/template.py
```diff
@@ -0,0 +1,25 @@
...
+license = "GPL-3.0-or-later"
+url = "https://github.com/kitsunyan/intel-undervolt"
+source = f"{url}/archive/{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "29a7ebaee4830d65d0b5cefa6d497887d4f23f34659876dfe944f3a020cf33ff"
+# no tests
+options = ["!check"]
+
+
+def do_configure(self):
+    self.do("./configure")
```
q66, line 20: enable the elogind hook

PR #1789, opened 2024-04-10
main/bash-completion: update to 2.13.0

nekopsykose:


main/bash-completion/template.py
```diff
@@ -13,6 +13,10 @@
...
 url = "https://github.com/scop/bash-completion"
 source = f"{url}/releases/download/{pkgver}/{pkgname}-{pkgver}.tar.xz"
-sha256 = "73a8894bad94dee83ab468fa09f628daffd567e8bef1a24277f1e9a0daf911ac"
+sha256 = "c5f99a39e40f0d154c03ff15438e87ece1f5ac666336a4459899e2ff4bedf3d1"
 # what's the point (needs a truckload of checkdepends too)
 options = ["!check"]
+
+
+def post_install(self):
+    self.rm(self.destdir / "etc/bash_completion.d", recursive=True)
```
nekopsykose, line 22: this technically places a compat function thing in there but idk if it's that important

triallax: you can pass `compatdir=/usr/share/bash-completion/completions` in `make_install_args` like i do in https://github.com/chimera-linux/cports/pull/1826 i guess

triallax: hmm actually then you'd need to either patch `bash_completion` to use that or set the `BASH_COMPLETION_COMPAT_DIR` envvar somewhere to use it, so i guess we can just remove the file if it's not important

nekopsykose: yeah will just leave it out for now i guess

nekopsykose: ah no, _get_cword is compat and needed by some things

triallax: bleh

PR #1298, opened 2024-01-22
contrib/libvirt: new package (10.0.0)

NovaAndrom3da:

PR #1800, opened 2024-04-11
Libvirt, integration, and dnsmasq

cesorious:
IDK about the dinit unit files, they probably aren't right, as they are mostly an attempt at translating arch's systemd units. There was also some socket activation, that I couldn't figure out. requires adding yourself to the libvirt group and installing a qemu-system*.

Closes #1298

### DIFF
--- a/contrib/dnsmasq/files/sysusers.conf
+++ b/contrib/dnsmasq/files/sysusers.conf
@@ -1,0 +1,3 @@
+# Create dnsmasq system user
+
+u _dnsmasq - "dnsmasq user" - /usr/bin/nologin
--- a/contrib/dnsmasq/patches/copts.patch
+++ b/contrib/dnsmasq/patches/copts.patch
@@ -1,0 +1,12 @@
+diff --git a/src/config.h b/src/config.h
+index e722e98..5535d77 100644
+--- a/src/config.h
++++ b/src/config.h
+@@ -463,3 +463,7 @@ static char *compile_opts =
+ "dumpfile";
+ 
+ #endif /* defined(HAVE_DHCP) */
++#define HAVE_DNSSEC
++#define HAVE_DBUS
++#define HAVE_LIBIDN2
++#define HAVE_CONNTRACK
--- a/contrib/dnsmasq/template.py
+++ b/contrib/dnsmasq/template.py
@@ -1,0 +1,37 @@
+pkgname = "dnsmasq"
+pkgver = "2.90"
+pkgrel = 0
+build_style = "makefile"
+make_install_args = [
+    "BINDIR=/usr/bin",
+]
+hostmakedepends = [
+    "pkgconf",
+]
+makedepends = [
+    "dbus-devel",
+    "libidn2-devel",
+    "libnetfilter_conntrack-devel",
+    "linux-headers",
+    "nettle-devel",
+]
+depends = ["dbus"]
+pkgdesc = "DNS, DHCP, RA, TFTP and PXE server"
+maintainer = "cesorious <cesorious@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://www.thekelleys.org.uk/dnsmasq/doc.html"
+source = f"https://www.thekelleys.org.uk/dnsmasq/dnsmasq-{pkgver}.tar.gz"
+sha256 = "8f6666b542403b5ee7ccce66ea73a4a51cf19dd49392aaccd37231a2c51b303b"
+# no tests
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_file("dbus/dnsmasq.conf", "usr/share/dbus-1/system.d")
+
+    self.install_file(
+        self.files_path / "sysusers.conf",
+        "usr/lib/sysusers.d",
+        name="dnsmasq.conf",
+    )
+    self.install_service(self.files_path / "dnsmasq")
--- a/contrib/libvirt/files/sysusers.conf
+++ b/contrib/libvirt/files/sysusers.conf
@@ -1,0 +1,6 @@
+# Create libvirt-qemu system user and libvirt group
+
+u _libvirt-qemu - "libvirt-qemu user" /var/lib/libvirt /usr/bin/nologin
+m _libvirt-qemu kvm
+
+g libvirt - -
--- a/contrib/libvirt/files/virtnwfilterd
+++ b/contrib/libvirt/files/virtnwfilterd
@@ -1,0 +1,8 @@
+# virtnwfilterd daemon service
+
+type          = process
+command       = /usr/bin/virtnwfilterd
+depends-on    = network.target
+after         = dbus
+depends-on    = local.target
+before        = login.target
--- a/contrib/libvirt/files/virtstoraged
+++ b/contrib/libvirt/files/virtstoraged
@@ -1,0 +1,8 @@
+# virtstoraged daemon service
+
+type          = process
+command       = /usr/bin/virtstoraged
+depends-on    = network.target
+after         = dbus
+depends-on    = local.target
+before        = login.target
--- a/contrib/libvirt/template.py
+++ b/contrib/libvirt/template.py
@@ -1,0 +1,127 @@
+pkgname = "libvirt"
+pkgver = "10.3.0"
+pkgrel = 0
+build_style = "meson"
+configure_args = [
+    "-Dattr=enabled",
+    "-Dblkid=enabled",
+    "-Dcapng=enabled",
+    "-Dcurl=enabled",
+    "-Ddriver_qemu=enabled",
+    "-Ddriver_secrets=enabled",
+    "-Dfirewalld=enabled",
+    "-Dfirewalld_zone=enabled",
+    "-Dfuse=enabled",
+    "-Dlibiscsi=enabled",
+    "-Dlibnl=enabled",
+    "-Dlibssh2=enabled",
+    "-Dno_git=true",
+    "-Dnumactl=enabled",
+    "-Dpackager=Chimera Linux",
+    "-Dpciaccess=enabled",
+    "-Dstorage_dir=enabled",
+    "-Dstorage_disk=enabled",
+    "-Dstorage_fs=enabled",
+    "-Dstorage_lvm=enabled",
+    "-Dstorage_mpath=enabled",
+    "-Dstorage_scsi=enabled",
+    "-Dstorage_zfs=enabled",
+    "-Dudev=enabled",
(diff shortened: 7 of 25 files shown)


contrib/dnsmasq/template.py
```diff
@@ -0,0 +1,32 @@
+pkgname = "dnsmasq"
+pkgver = "2.90"
+pkgrel = 1
```
nekopsykose, line 3: pkgrel starts at 0 for everything

contrib/dnsmasq/template.py
```diff
@@ -0,0 +1,32 @@
...
+hostmakedepends = [
+    "pkgconf",
+]
+makedepends = [
+    "dbus-devel",
+    "libnetfilter_conntrack-devel",
+    "linux-headers",
+    "nettle-devel",
+]
+pkgdesc = "Lightweight DNS, DHCP, RA, TFTP and PXE server"
```
nekopsykose, line 17:
```suggestion
pkgdesc = "DNS, DHCP, RA, TFTP and PXE server"
```

contrib/libvirt/template.py
```diff
@@ -0,0 +1,136 @@
...
+    ]:
+        self.install_service(self.files_path / f"virt{service}d")
+
+
+@subpackage("libvirt-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage("libvirt-firewalld")
```
cesorious, line 122: for whatever reason the libvirt-firewalld subpackage only grabs `usr/lib/firewalld/policies/libvirt-routed-in.xml` and none of the other zones or policies.

cesorious:
libvirt requires gst-plugins-good for audio to work should I add this to depends?

contrib/libvirt/template.py
nekopsykose: the first return is all that happens, you can't return from a function multiple times here since it's not a generator (afaik)

just write out the whole list of five items

cesorious: I tried that too, but it still only returns the first file

nekopsykose: doing  

```
return [
	"a",
	"b",
	"c",
]
```

definitely returns all the files

nekopsykose:
```
return [
    "usr/lib/firewalld/policies/libvirt-routed-in.xml",
    "usr/lib/firewalld/policies/libvirt-routed-out.xml",
    "usr/lib/firewalld/policies/libvirt-to-host.xml",
    "usr/lib/firewalld/zones/libvirt-routed.xml",
    "usr/lib/firewalld/zones/libvirt.xml",
]
```

there you go

nekopsykose: or probably just `return ["usr/lib/firewalld"]` unless you're keeping other files in there in the main package

cesorious: woops, forgot  to reinstall the libvirt package after rebuilding...

contrib/dnsmasq/patches/copts.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/src/config.h b/src/config.h
+index e722e98..5535d77 100644
+--- a/src/config.h
++++ b/src/config.h
+@@ -463,3 +463,7 @@ static char *compile_opts =
+ "dumpfile";
+ 
+ #endif /* defined(HAVE_DHCP) */
++#define DHAVE_DNSSEC
```
q66, line 9: DHAVE_ ?

contrib/dnsmasq/files/sysusers.conf
```diff
@@ -0,0 +1,3 @@
+# Create dnsmasq system user
+
+u dnsmasq - "dnsmasq user" - /usr/bin/nologin
```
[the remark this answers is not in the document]
q66: the service should launch with something like `-k -u _dnsmasq -g _dnsmasq --enable-dbus`

it should `depends-on = dbus`, `before = network.target` and `depends-on = pre-network.target` probably (but not entirely sure)

contrib/dnsmasq/template.py
```diff
@@ -0,0 +1,32 @@
...
+hostmakedepends = [
+    "pkgconf",
+]
+makedepends = [
+    "dbus-devel",
+    "libnetfilter_conntrack-devel",
+    "linux-headers",
+    "nettle-devel",
+]
+pkgdesc = "DNS, DHCP, RA, TFTP and PXE server"
```
q66, line 17: since the service will depend on dbus, it should also runtime-depend on dbus here

contrib/libvirt/files/libvirtd
```diff
@@ -0,0 +1,10 @@
+# deprecated libvirtd daemon service
```
q66, line 1: if it's deprecated, why create it?

contrib/libvirt/files/sysusers.conf
```diff
@@ -0,0 +1,6 @@
+# Create libvirt-qemu system user and libvirt group
+
+u virtqemu - "virtqemu user" /var/lib/libvirt /usr/bin/nologin
```
q66, line 3: ditto for underscored accounts/groups

contrib/libvirt/files/sysusers.conf
```diff
@@ -0,0 +1,6 @@
...
+m virtqemu kvm
+
+g libvirt - -
```
q66, line 6: what is this group for?

contrib/libvirt/files/sysusers.conf
```diff
@@ -0,0 +1,6 @@
+# Create libvirt-qemu system user and libvirt group
+
+u virtqemu - "virtqemu user" /var/lib/libvirt /usr/bin/nologin
+m virtqemu kvm
```
q66, line 4: should probably also belong to disk

contrib/libvirt/files/sysusers.conf
q66: i'd also call it libvirt

contrib/libvirt/files/virtstoraged
```diff
@@ -0,0 +1,9 @@
+# virtstoraged daemon service
+# clonficts with libvirtd daemon
```
q66, line 2: clonflicts

contrib/libvirt/template.py
```diff
@@ -0,0 +1,126 @@
...
+depends = [
+    "dnsmasq",
+]
+pkgdesc = "API, daemon, and management tool for virtualization"
+maintainer = "cesorious <cesorious@gmail.com>"
+license = "LGPL-2.1-only"
+url = "https://libvirt.org"
+source = f"https://download.libvirt.org/libvirt-{pkgver}.tar.xz"
+sha256 = "215772bc5dc4a672e67ffa9de3774f05ed4b7ed282dbe296ec5c9fec01dd7ae3"
+options = ["!lto"]
```
q66, line 88: comment why

contrib/libvirt/files/virtstoraged
q66: (ditto for anywhere else, but i'd just drop the libvirtd service *and* these lines)

contrib/libvirt/files/virtnwfilterd
```diff
@@ -0,0 +1,9 @@
+# virtnwfilterd daemon service
+# clonficts with libvirtd daemon
+
+type          = process
+command       = /usr/bin/virtnwfilterd
+depends-on    = network.target
+after         = dbus
```
q66, line 7: idk if these request a name on the bus, but if they do, they should be `depends-on = dbus`

contrib/libvirt/template.py
```diff
@@ -0,0 +1,126 @@
...
+    "udev-devel",
+    "yajl-devel",
+]
+checkdepends = [
+    "pahole",
+    "python-black",
+    "python-flake8",
+    "python-pytest",
+]
+depends = [
```
q66, line 79: if services request dbus names this should depend on dbus too

contrib/libvirt/template.py
```diff
@@ -0,0 +1,126 @@
...
+        self.install_service(self.files_path / f"virt{service}d")
+
+
+@subpackage("libvirt-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage("libvirt-firewalld")
+def _firewalld(self):
```
q66, line 123: need pkgdesc

q66:
i left in a bunch of initial comments, there will probably be plenty more once you're done fixing these, but i don't want to cause too much of a mess for now

contrib/libvirt/files/sysusers.conf
cesorious: default seems to be libvirt-qemu, also it seems the libvirt group is hardcoded, so we'd still need a "libvirt" group along with an _libvirt group

contrib/libvirt/files/sysusers.conf
cesorious: to use libvirt without polkit iirc

contrib/libvirt/files/sysusers.conf
cesorious: this was mostly just copied from arch's sysusers file

contrib/libvirt/template.py
cesorious: leftover from Nova's initial template, seems to work fine with lto actually

contrib/libvirt/files/virtnwfilterd
cesorious: i do not know how to figure that out, I'll leave it as is for now

contrib/libvirt/template.py
cesorious: i do not know how to figure that out, I'll leave it as is for now

contrib/libvirt/files/sysusers.conf
cesorious: renamed to _libvirt-qemu as per the default for libvirt

cesorious:
Another question, libvirt comes with a `60-qemu-postcopy-migration.conf` file which sets `vm.unprivileged_userfaultfd = 1`, however, it seems that post-copy migration works fine on my system, at least with with linux-stable and libvirt states that this shouldn't be needed on modern kernels in their [10.1.0 release notes](https://libvirt.org/news.html). Should I just remove this file to avoid any potential security issues related to having unprivileged userfaultfd? (I am not actually sure how bad it may be, but remember reading about it in the past)

nekopsykose:
i think removing it is fine

PR #1801, opened 2024-04-11
Swtpm

cesorious:


contrib/swtpm/template.py
```diff
@@ -0,0 +1,61 @@
...
+@subpackage("libswtpm-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage("libswtpm-devel-static")
+def _devel(self):
+    self.depends = [f"libswtpm-devel~{pkgver}"]
+
+    return ["usr/lib/swtpm/libswtpm_libtpms.a"]
```
nekopsykose, line 61:
```suggestion
```

don't need this, default split does this from devel already

contrib/swtpm/template.py
```diff
@@ -0,0 +1,61 @@
...
+sha256 = "7bba52aa41090f75087034fac5fe8daed10c3e7e7234df7c9558849318927f41"
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+
+
+@subpackage("libswtpm")
+def _libs(self):
+    return ["usr/lib/swtpm/*.so.*"]
```
nekopsykose, line 49:
```suggestion
    return self.default_libs()
```

contrib/swtpm/template.py
```diff
@@ -0,0 +1,61 @@
...
+    "iproute2",
+    "trousers",
+]
+depends = ["gnutls-progs"]
+pkgdesc = "TPM emulator with socket, character device, and Linux CUSE interface"
+maintainer = "cesorious <cesorious@gmail.com>"
+license = "BSD-3-Clause"
+url = "https://github.com/stefanberger/swtpm"
+source = (
+    f"https://github.com/stefanberger/swtpm/archive/refs/tags/v{pkgver}.tar.gz"
```
nekopsykose, line 38:
```suggestion
    f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
```

contrib/swtpm/template.py
```diff
@@ -0,0 +1,61 @@
+pkgname = "swtpm"
+pkgver = "0.8.1"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--with-gnutls"]
+configure_env = {"NOCONFIGURE": "1"}
```
nekopsykose, line 6: don't think you need this since you're not running autogen.sh and autoreconf doesn't conf twice

contrib/libtpms/template.py
```diff
@@ -0,0 +1,32 @@
...
+    "openssl-devel",
+]
+checkdepends = [
+    "bash",
+]
+pkgdesc = "Software emulation of a Trusted Platform Module"
+maintainer = "cesorious <cesorious@gmail.com>"
+license = "BSD-3-Clause"
+url = "https://github.com/stefanberger/libtpms"
+source = f"https://github.com/stefanberger/libtpms/archive/refs/tags/v{pkgver}.tar.gz"
```
nekopsykose, line 22:
```suggestion
source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
```

contrib/libtpms/template.py
```diff
@@ -0,0 +1,32 @@
...
+    "automake",
+    "gmake",
+    "libtool",
+    "pkgconf",
+]
+makedepends = [
+    "openssl-devel",
+]
+checkdepends = [
+    "bash",
```
nekopsykose, line 16: when there's just one thing you can leave it as a single line instead of multiple

contrib/swtpm/template.py
cesorious: without this, cbuild fails with
 `=> swtpm-0.8.1-r0: static libraries should be in the -static package`

contrib/swtpm/template.py
cesorious: this doesn't seem to work because the libs are in a subdirectory, at least it fails with
`=> libswtpm-0.8.1-r0: ERROR: take(): path 'usr/lib/lib*.so.[0-9]*' did not match anything`

nekopsykose: oh right i missed the subdir

contrib/swtpm/template.py
nekopsykose: sure, that seems to happen because of the subdir

cesorious: so just leave it as is? or pull it into devel?

nekopsykose: i think you can put it into devel yeah

nekopsykose: or maybe not, if not then it's fine as is

cesorious: cbuild still complains if it's in devel, leaving it as is

contrib/swtpm/template.py
```diff
@@ -0,0 +1,58 @@
...
+url = "https://github.com/stefanberger/swtpm"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "7bba52aa41090f75087034fac5fe8daed10c3e7e7234df7c9558849318927f41"
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+
+
+@subpackage("libswtpm")
```
q66, line 44: swtpm-libs

contrib/swtpm/template.py
```diff
@@ -0,0 +1,58 @@
...
+def _libs(self):
+    return ["usr/lib/swtpm/*.so.*"]
+
+
+@subpackage("libswtpm-devel")
```
q66, line 49: swtpm-devel

contrib/swtpm/template.py
```diff
@@ -0,0 +1,58 @@
...
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage("libswtpm-devel-static")
```
q66, line 54: drop this one and instead the .a's in swtpm-libs (cbuild will split it automatically)

contrib/swtpm/template.py
```diff
@@ -0,0 +1,58 @@
+pkgname = "swtpm"
+pkgver = "0.8.1"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--with-gnutls"]
```
q66, line 5: the tss user should probably be configured to match trousers?

contrib/swtpm/template.py
nekopsykose: it doesn't cause it's in /usr/lib/swtpm (i had the same thought)

nekopsykose: or do you mean it works when it ends up in -libs proper?

q66: er i meant -devel, not -libs

q66: the static splitter does an rglob on usr/lib, it should work in subdirs

contrib/swtpm/template.py
```diff
@@ -0,0 +1,55 @@
...
+    "libtasn1-devel",
+    "libtpms-devel",
+    "linux-headers",
+    "openssl-devel",
+]
+checkdepends = [
+    "iproute2",
+    "trousers",
+]
+depends = ["gnutls-progs"]
```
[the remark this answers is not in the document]
cesorious: it seems void depends on it, but neither arch nor alpine does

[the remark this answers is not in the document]
q66: we probably want to in order to make sure the tss user is created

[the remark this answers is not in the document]
cesorious: swtpm now depends on trousers, also bumped to 0.8.2

q66:
this looks good to me now, thanks - i'll let the CI have a go at it and if that passes i think we're done

PR #1809, opened 2024-04-12
main/python-docutils: update to 1.21.1

triallax:


main/python-docutils/template.py
```diff
@@ -1,21 +1,24 @@
 pkgname = "python-docutils"
-pkgver = "0.20.1"
-pkgrel = 1
+pkgver = "0.21.1"
+pkgrel = 0
+# Remove on next pkgver bump
+_pkgver = f"{pkgver[:pkgver.rfind('.')]}.post{pkgver[-1]}"
```
nekopsykose, line 5: there's a real 0.21.1 on pypi though so it works without this afaict

main/python-docutils/template.py
```diff
@@ -1,21 +1,22 @@
...
+checkdepends = ["python-pillow", "python-pygments"]
 pkgdesc = "Python documentation utilities"
 maintainer = "q66 <q66@chimera-linux.org>"
 license = "custom:none AND BSD-2-Clause AND GPL-3.0-or-later AND Python-2.0"
 url = "http://docutils.sourceforge.net"
 source = f"$(PYPI_SITE)/d/docutils/docutils-{pkgver}.tar.gz"
-sha256 = "f08a4e276c3a1583a86dce3e34aba3fe04d02bba2dd51ed16106244e8a923e3b"
+sha256 = "65249d8a5345bc95e0f40f280ba63c98eb24de35c6c8f5b662e3e8948adea83f"
+# Some test files seem to be missing in the tarball
+options = ["!check"]
```
nekopsykose, line 19: should be present on github tag tarball instead

triallax: same errors with the upstream sourceforge tarball, and i can't see any tags in https://github.com/docutils/docutils/tags (gh mirror)

nekopsykose: right forgot this wasn't in github

PR #1810, opened 2024-04-12
cbuild: fix codeberg.org update-check

triallax:


nekopsykose:
hm but when does it fail? everything i tried still works

triallax:
try `./cbuild update-check foot v`

nekopsykose:
thanks

PR #1812, opened 2024-04-12
main/ijs: add !distlicense

triallax:


main/ijs/template.py
```diff
@@ -1,7 +1,7 @@
 pkgname = "ijs"
-_gsver = "9.55.0"
+_gsver = "10.03.0"
 pkgver = "0.35"
```
triallax, line 3: this doesn't feel right to me, there's no automated reminder (e.g. the update-check) to remind us to bump the ghostscript version (which btw wasn't bumped since the package was created in 290a7140f8c1f9ea8e398b69785c593388ff2c3f in january 2022)

suggestions?

q66: considering ijs hasn't really changed in years and is identical in these releases, it really does not matter

triallax: noted, keeping the pr as-is for the license thing

q66: just !distlicense it

PR #1816, opened 2024-04-13
Virt-manager

cesorious:
this template is worse than worse, help appreciated
depends on #1800

### DIFF
--- a/contrib/virt-manager/patches/no-gtk-update-icon-cache.patch
+++ b/contrib/virt-manager/patches/no-gtk-update-icon-cache.patch
@@ -1,0 +1,22 @@
+diff --git a/setup.py b/setup.py
+index b45d315..48d6e21 100755
+--- a/setup.py
++++ b/setup.py
+@@ -266,17 +266,6 @@ class my_install(setuptools.command.install.install):
+     def run(self):
+         setuptools.command.install.install.run(self)
+ 
+-        if not self.distribution.no_update_icon_cache:
+-            print("running gtk-update-icon-cache")
+-            icon_path = os.path.join(self.install_data, "share/icons/hicolor")
+-            self.spawn(["gtk-update-icon-cache", "-q", "-t", icon_path])
+-
+-        if not self.distribution.no_compile_schemas:
+-            print("compiling gsettings schemas")
+-            gschema_install = os.path.join(self.install_data,
+-                "share/glib-2.0/schemas")
+-            self.spawn(["glib-compile-schemas", gschema_install])
+-
+ 
+ ###################
+ # Custom commands #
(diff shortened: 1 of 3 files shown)


contrib/virt-manager/template.py
```diff
@@ -0,0 +1,104 @@
...
+    "libvirt-glib",
+    "spice-gtk",
+    "virt-manager-tools",
+    "vte-gtk3",
+]
+_checkdepends = [
+    "python-pytest",
+    "xorriso",
+]
+checkdepends = _checkdepends + _deps
```
JamiKettunen, line 34: Drop `_checkdepends`  and use `checkdepends = ["python-pytest", "xorriso"] + _deps`

JamiKettunen:
You can probably ditch the manual `do_{build,check,install}` (and start using just `make_check_args` similar to [`nicotine-plus`](https://github.com/chimera-linux/cports/blob/master/contrib/nicotine-plus/template.py)) after v4.2.0 releases soon :tm: if https://github.com/virt-manager/virt-manager/issues/652#issuecomment-2052551093 is anything to go by; alternatively you could backport the two commits mentioned on the issue too already, but the libvirt PR is still pending too atm.

About `make_check_args` you can eventually do something like this for all the tests (ideally each with some sort of explanation comment for why they're skipped):
```py
make_check_args = [
    "-k",
    "not (testDASDMdev or "  # fails because ???
    "testAPQNMdev or "
    "testPCIMdev or "
    # ...
    "testcli0424virt_clone)",
]
```

contrib/virt-manager/patches/no-gtk-update-icon-cache.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/setup.py b/setup.py
+index b45d315..48d6e21 100755
+--- a/setup.py
++++ b/setup.py
+@@ -266,17 +266,6 @@ class my_install(setuptools.command.install.install):
+     def run(self):
+         setuptools.command.install.install.run(self)
+ 
+-        if not self.distribution.no_update_icon_cache:
```
JamiKettunen, line 9: [Void avoids this patch](https://github.com/void-linux/void-packages/blob/master/srcpkgs/virt-manager/template) by passing `--no-update-icon-cache --no-compile-schemas` to `setup.py`

nekopsykose: we can't pass args to setup.py because the pep517 standard (invoked via python -m build) does not have a way to pass standard args yet..

nekopsykose: so the patch is fine, it's just a few lines

JamiKettunen: sad, but fair enough

PR #1828, opened 2024-04-15
contrib/qutebrowser: new package (3.1.0)

ttyyls:


ttyyls:
had to rebase and update the qt6 dependencies to be sync on 6.7.x 

python-pyqt6 stable pypi release is too old (december 2023 release) so I switched to the dev version.

From local testing it seems to be working fine and ready for merging at the exception of some !checks that may need to be fixed before that

image

q66:
webengine needs to be a completely separate PR because i will not review this massive batch, also i will not merge it unless it gets the chromium patch treatment (i.e. archs= parity, see https://github.com/chimera-linux/cports/blob/master/Packaging.md#targets-and-tiers)

ttyyls:
since python-pyqt6 released 6.7.0, I rolled back the switch to the dev version in this pr

ttyyls:
reverted python-pyqt6-webengine to 6.6.0, since the latest release (6.7.0) is broken

nekopsykose:
whats broken in it?

contrib/python-pyqt6-webengine/template.py
```diff
@@ -0,0 +1,48 @@
...
+    "python-pyqt-builder",
+    "python-pyqt6",
+    "python-sip",
+]
+makedepends = [
+    "python-devel",
+    "qt6-qtbase-devel",
+    "qt6-qtwebchannel-devel",
+    "qt6-qtwebengine-devel",
+    "python-pyqt6",
```
nekopsykose, line 21: sort this and host

contrib/qutebrowser/template.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def post_build(self):
+    self.do("a2x", "-f", "manpage", "doc/qutebrowser.1.asciidoc")
+
+
+def post_install(self):
+    self.install_man("doc/qutebrowser.1")
+    self.install_dir("usr/share/qutebrowser")
+    self.install_files("scripts", "usr/share/qutebrowser")
```
nekopsykose, line 52: dont need the install_dir here p sure

ttyyls: leftover from some previous glob install_file(s). I still don't think it's correct though since ideally I should also filter __pycache__ and test dirs and whatever like the makefile does here https://github.com/qutebrowser/qutebrowser/blob/main/misc/Makefile#L33 except to use the makefile i'd need to patch in an install -d "dir" for each since it relies on gnu behavior

nekopsykose: i would patch the makefile

nekopsykose: and remove the setup.py install from it too ofc

nekopsykose: you should also `make man` instead of manual a2x

ttyyls: ok

contrib/qutebrowser/template.py
```diff
@@ -0,0 +1,56 @@
...
+license = "GPL-3.0-only"
+url = "https://qutebrowser.org"
+source = f"https://github.com/qutebrowser/qutebrowser/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "9307d55f1d9157b7906ac08566e3951bffd6f5b753432deaa9a13681995ba3ca"
+# unpackaged dependencies
+options = ["!check"]
+
+
+def post_build(self):
+    self.do("gmake", "-f", "misc/Makefile", "man")
```
nekopsykose, line 46: actually since install deps on man you can just delete this

PR #1835, opened 2024-04-16
contrib/colordiff: new package (1.0.21)

subnut:


nekopsykose:
you should probably at least use the old version of the template from before it was removed in ca846329b1cedf90f3b42b3d5ce632a139c5a93a . but first see if diff --color doesn't do what you want

subnut:
Nah, `diff --color` is good, but not enough. (Thanks though, didn't know that existed!)

Also the older template built the manpages using `xmlto` AFAICT, which isn't needed because the manpages are already included in the tarball.

Still, let me know if I need to change it to the older template.

PR #1840, opened 2024-04-17
contrib/rofi: update to 1.7.5_p3

NovaAndrom3da:


NovaAndrom3da:
what the hell

it wouldn't build with the patch locally

now it wont build without the patch on there

mkay annoying

NovaAndrom3da:
what the hell do you want from me

PR #1843, opened 2024-04-17
main{libshumate,gnome-maps}: new packages

triallax:
- **main/protobuf: move from contrib**
- **main/protobuf-c: move from contrib**
- **main/libshumate: new package (1.2.0)**
- **main/gnome-maps: new package (46.0)**


triallax:
#1741

PR #1846, opened 2024-04-18
packaging: fix self.do usage example

NovaAndrom3da:


q66:
commit message

NovaAndrom3da:
What should I change it to?

PR #1848, opened 2024-04-18
contrib/bind: update to 9.18.26

z-erica:
one note: bind seems to depend on bind-devel, i think bc of weirdly versioned sos. is this something to be fixed?


q66:
yes, that needs to be fixed

q66:
i'd just make -devel not use default_devel and not add -libs at all

PR #1855, opened 2024-04-19
main/gnome-keyring: run checks in parallel

triallax:
Seems to work just fine here.


q66:
it's fine most of the time but there is some race going on

triallax:
hmm annoying

PR #1860, opened 2024-04-20
contrib/prismlauncher: add qt6-qtsvg as runtime depend

nopjmp:
qt6-qtsvg is not listed as a dependency as prismlauncher does not directly link to qt6-qtsvg. This causes the svg icons to be missing.

### DIFF
--- a/contrib/prismlauncher/template.py
+++ b/contrib/prismlauncher/template.py
@@ -18,6 +18,10 @@ makedepends = [
     "quazip-devel",
     "zlib-devel",
 ]
+depends = [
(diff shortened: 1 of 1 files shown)


q66:
i don't think this is quite right, the ability to render svg icons is a property of the toolkit and it should probably be installed by default with qtbase without having to add a dependency to every application

nopjmp:
> i don't think this is quite right, the ability to render svg icons is a property of the toolkit and it should probably be installed by default with qtbase without having to add a dependency to every application

I agree. I'll close this and I can discuss this on IRC and/or propose a patch for adding it as a default for qtbase.

PR #1862, opened 2024-04-21
main/grub: update to 2.12

q66:
Build:

* aarch64 now ok
* ppc64le ok
* riscv64 ok
* x86_64 ok

Functionality so far all untested.


nekopsykose:
this will most likely have the same issue all othet distros had where the new grub-mkconfig breaks booting until grub-install is reran. not sure we can do anything about that tho

q66:
i've included a patch that makes the generated configs compatible already, unless you mean something else?

nekopsykose:
hm sure, if that works then sounds good

PR #1866, opened 2024-04-21
user/taskwarrior: update to 3.0.1

jcgruenhage:
Submitted as a draft for now, so that @nekopsykose has a chance to confirm the intent behind the dependency bumps - the suspicion in IRC was that the intention was to get rid of ring 0.16.20 to fix the build on non-x86/arm architectures, which I've put into the message for the patch to help future me (or others) to understand why this is dependency bump is needed


nekopsykose:
that's correct, yeah

PR #1881, opened 2024-04-23
contrib/fzf: update to 0.50.0

wezm:
Does this still need `options = ["!strip"]`? @q66 added it in 1896da1 but I'm not sure what the reason was. As far as I can tell it builds and runs ok without it.

### DIFF
--- a/contrib/fzf/template.py
+++ b/contrib/fzf/template.py
@@ -1,5 +1,5 @@
 pkgname = "fzf"
-pkgver = "0.49.0"
+pkgver = "0.50.0"
 pkgrel = 0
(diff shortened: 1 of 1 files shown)


nekopsykose:
probably yes, though i would replace it with !debug instead

nekopsykose:
it just fails in ppc64le without it or whatever

wezm:
Ok I added `!debug`. Not sure what up with CI, seems like an issue with Ubuntu repos.

nekopsykose:
yeah, looks like gha microsoft ubuntu repo mirrors are down

PR #1883, opened 2024-04-23
contrib/wlsunset: update to 0.4.0

ttyyls:


contrib/wlsunset/patches/default.patch
```diff
@@ -0,0 +1,15 @@
+default to guessing the timezone to avoid sigill with int hardening  on a call without arguments
```
nekopsykose, line 1: hm but doesn't it then crash when !manual_time is set

ttyyls: from what i can tell you added an !int in #604 'cause it segfaulted when ran without any other arguments..
Afaict the segfault is due to the following:

first cfg.manual_time is unset and thus false, so this will always make a call to longitude_time_offset with a cfg.longitude that is also unset
```
	if (!cfg.manual_time) {
		ctx.longitude_time_offset = longitude_time_offset(cfg.longitude);
	} else {
		ctx.longitude_time_offset = -get_timezone();
	}

```
longitude_time_offset makes use of the longitude (which is also unset) in a return call and there goes the segfault.

manual_time is otherwise set to true if (and only if) both longitude and latitude are set or an explicit sunrise value is set. Both cases are not problematic either way.

The most conscise patch I could think of to work around this was to default to get_timezone for the offset.

ttyyls: now that i spelled it out, giving a default longitude based on the a get_timezone call might also work and be even more precise.

nekopsykose: sounds like you should think of an upstream fix?

ttyyls: would rather just revert to !int for now. might send a fix upstream by the next release.

nekopsykose: sure, fine by me

PR #1895, opened 2024-04-25
contrib/mpd: new package (0.23.15)

nullobsi:


contrib/mpd/template.py
```diff
@@ -0,0 +1,73 @@
...
+    "zlib-devel",
+]
+pkgdesc = "Music player daemon"
+maintainer = "nullobsi <nullobsi@unix.dog>"
+license = "GPL-2.0-or-later AND BSD-2-Clause"
+url = "https://www.musicpd.org"
+source = f"https://www.musicpd.org/download/mpd/{pkgver[:-3]}/mpd-{pkgver}.tar.xz"
+sha256 = "550132239ad1acf82ccf8905b56cc13dc2c81a4489b96fba7731b3049907661a"
+
+make_dir = "built"
```
nekopsykose, line 59: this goes near the top (i assume it fails since build/ already exists, but then add a comment for why)

contrib/mpd/files/mpd
```diff
@@ -0,0 +1,9 @@
+# MPD service
+
+type            = process
+command         = /usr/bin/mpd --no-daemon
+before          = network.target
+depends-on      = pre-network.target
+restart         = true
+smooth-recovery = true
+
```
nekopsykose, line 9: there's an extra trailing newline here, also don't need restart=true, and don't need to align service file = anymore

contrib/mpd/files/sysusers.conf
```diff
@@ -0,0 +1,5 @@
+# Create MPD system user
+
+u _mpd - "MPD user" /var/lib/mpd /usr/bin/nologin
+m _mpd audio
+
```
nekopsykose, line 5: extra trailing newlines here and in mpd.conf too

contrib/mpd/files/sysusers.conf
```diff
@@ -0,0 +1,5 @@
+# Create MPD system user
+
+u _mpd - "MPD user" /var/lib/mpd /usr/bin/nologin
```
nekopsykose, line 3: i think you need a tmpfiles entry for this dir too

contrib/mpd/template.py
```diff
@@ -0,0 +1,73 @@
...
+    self.install_service(self.files_path / "mpd")
+
+    self.install_file(self.files_path / "mpd.conf", "etc")
+    self.install_file(
+        self.files_path / "sysusers.conf",
+        "usr/lib/sysusers.d",
+        name="mpd.conf",
+    )
+    self.install_file("doc/mpdconf.example", "usr/share/doc/mpd")
+
```
nekopsykose, line 73:
```suggestion
    self.install_service(self.files_path / "mpd")
    self.install_file(self.files_path / "mpd.conf", "etc")
    self.install_file(
        self.files_path / "sysusers.conf",
        "usr/lib/sysusers.d",
        name="mpd.conf",
    )
    self.install_file("doc/mpdconf.example", "usr/share/doc/mpd")
```

contrib/mpd/files/mpd
```diff
@@ -0,0 +1,9 @@
+# MPD service
+
+type            = process
+command         = /usr/bin/mpd --no-daemon
+before          = network.target
```
nekopsykose, line 5: why is it before network? seems like it should be before login.target if anything

contrib/mpd/template.py
```diff
@@ -0,0 +1,73 @@
...
+    "fdk-aac-devel",
+    "ffmpeg-devel",
+    "flac-devel",
+    "fluidsynth-devel",
+    "fmt-devel",
+    "lame-devel",
+    "libcdio-paranoia-devel",
+    "libcurl-devel",
+    "libexpat-devel",
+    "libgme-devel",
```
nekopsykose, line 29: pass explicit meson -Dsomething=enabled for these if possible

nekopsykose: (like most of these libs, not just these lines, i commented in a random spot here)

contrib/mpd/files/sysusers.conf
nekopsykose: also, does it make /var/lib/mpd/music|playlists itself on start or do you need tmpfiles for that too?

nullobsi: i haven't specified a music directory in the config initially, but I might add that as a default and put them in tmpfiles

contrib/mpd/template.py
```diff
@@ -0,0 +1,169 @@
...
+    "-Dbzip2=enabled",
+    "-Diso9660=enabled",
+    "-Dzzip=disabled",
+    # Tag plugins
+    "-Did3tag=enabled",
+    "-Dchromaprint=enabled",
+    # Decoder plugins
+    "-Dadplug=disabled",
+    "-Daudiofile=enabled",
+    # FIXME: FFMPEG works but AAC is fairly common and this should be packaged
```
nekopsykose, line 49: i think ffmpeg working is fine? nothing wrong with the ffmpeg decoders after all

contrib/mpd/template.py
```diff
@@ -0,0 +1,168 @@
+pkgname = "mpd"
+pkgver = "0.23.15"
+pkgrel = 0
+build_style = "meson"
+# MPD source tree includes a clashing `build` dir
```
triallax, line 5: comment should be moved to above `make_dir`

contrib/mpd/template.py
```diff
@@ -0,0 +1,168 @@
...
+    "twolame-devel",
+    "wavpack-devel",
+    "yajl-devel",
+    "zlib-devel",
+]
+pkgdesc = "Music player daemon"
+maintainer = "nullobsi <nullobsi@unix.dog>"
+license = "GPL-2.0-or-later AND BSD-2-Clause"
+url = "https://www.musicpd.org"
+source = f"https://www.musicpd.org/download/mpd/{pkgver[:-3]}/mpd-{pkgver}.tar.xz"
```
triallax, line 150:
```suggestion
source = f"{url}/download/{pkgname}/{pkgver[:-3]}/{pkgname}-{pkgver}.tar.xz"
```

triallax (changes requested):

PR #1897, opened 2024-04-27
mingw libraries

z-erica:


contrib/clang-rt-crt-mingw-w64/template.py
```diff
@@ -0,0 +1,118 @@
...
+        f"mingw-w64-headers-{_an}",
+    ]
+    depends += [f"clang-rt-crt-mingw-w64-{_an}={pkgver}-r{pkgrel}"]
+
+
+def do_configure(self):
+    from cbuild.util import cmake
+
+    for an in _targets:
+        at = an + "-w64-mingw32"
```
q66, line 68: i think the targets for 32-bit aren't w64, are they

contrib/mingw-w64-headers/template.py
```diff
@@ -0,0 +1,70 @@
...
+license = "ZPL-2.1"
+url = "https://www.mingw-w64.org"
+source = f"$(SOURCEFORGE_SITE)/mingw-w64/mingw-w64-v{pkgver}.tar.bz2"
+sha256 = "3f66bce069ee8bed7439a1a13da7cb91a5e67ea6170f21317ac7f5794625ee10"
+options = ["empty"]
+
+_targets = ["x86_64", "i686", "aarch64", "armv7"]
+
+for _an in _targets:
+    self.depends += [f"mingw-w64-headers-{_an}={pkgver}-r{pkgrel}"]
```
q66, line 18: self does not belong here

q66: you still need to pre-declare an empty depends (can't append to an undeclared variable)

contrib/mingw-w64-winpthreads/template.py
```diff
@@ -0,0 +1,91 @@
...
+            # don't step on mingw-w64-winpthreads
+            for hdr in ["signal", "time", "unistd"]:
+                self.uninstall(f"usr/{at}/include/pthread_{hdr}.h")
+
+
+def _gen(an, at):
+    @subpackage(f"mingw-w64-winpthreads-{an}")
+    def _subp(self):
+        self.subdesc = an
+        # self.depends = [f"mingw-w64-crt-{an}"]
```
q66, line 82: why is this commented out?

contrib/clang-rt-builtins-mingw-w64/template.py
```diff
@@ -0,0 +1,117 @@
...
+                suffix = an
+            case "armv7":
+                suffix = "arm"
+            case "i686":
+                suffix = "i386"
+
+        self.subdesc = f"{an} support"
+        self.depends = ["clang"]
+        # strip not supported for COFF
+        self.nostrip_files = [
```
q66, line 108: i'd just make it `self.options = ["!strip"]` since there is only one file

q66:
also, can you rewrite all the instances of `whatever={pkgver}-r{pkgrel}` using the new functions to make that nicer? i.e. use `self.parent` where correct, otherwise `self.with_pkgver` and the likes

contrib/mingw-w64-winpthreads/template.py
q66: the dep seems reasonable to me, winpthreads should probably pull in crt always? but shrug, i don't deal much with windows

PR #1905, opened 2024-04-28
contrib/openjdk17: make -demos subpackage optional

JamiKettunen:
Just found it silly that this is a hard-dependency; now one can mask it out with `apk add '!openjdk17-demos'`, also switch only `{pkgname}-jdk` instances to `openjdk{_majver}-jdk` to match rest of the file.

Maybe it shouldn't even be an `install_if` by default? Idk. I could also drop the `pkgrel` bump since this isn't really anything *that* useful and @q66 was planning on adding the alternatives stuff sometime

### DIFF
--- a/contrib/openjdk17/template.py
+++ b/contrib/openjdk17/template.py
@@ -3,7 +3,7 @@ _majver = "17"
 _fver = f"{_majver}.0.11"
 _bver = "9"
 pkgver = f"{_fver}_p{_bver}"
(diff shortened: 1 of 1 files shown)


q66:
why wouldn't be? if you want just jdk or jre (which is most of the time) you don't install openjdk17 in the first place

q66:
anyway the template will undergo a rework based on the new alternatives system

PR #1906, opened 2024-04-28
contrib/mpdc: new package (0.35)

ttyyls:
confirmed to work with #1895

had to change the name mpc=>mpdc 'cause otherwise it would conflict with main/mpc


contrib/mpdc/template.py
```diff
@@ -0,0 +1,17 @@
+pkgname = "mpdc"
+pkgver = "0.35"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = ["meson", "pkgconf"]
+makedepends = ["libmpdclient-devel"]
+pkgdesc = "Minimalist command line mpd client"
```
nekopsykose, line 7:
```suggestion
pkgdesc = "Command line mpd client"
```

nekopsykose:
maybe we should rename main/mpc to gmpc or something instead but not sure

PR #1912, opened 2024-04-30
user/git-extras: update to 7.2.0

ttyyls:


user/git-extras/template.py
```diff
@@ -1,13 +1,47 @@
...
+            "bin/git-bulk",
+            "bin/git-extras",
+            "bin/git-force-clone",
+            "bin/git-fork",
+            "bin/git-get",
+            "bin/git-ignore",
+            "bin/git-setup",
+            "bin/git-standup",
+        ]:
+            pass
```
ttyyls, line 41: this whole thing acts a subtitute to https://github.com/tj/git-extras/blob/main/Makefile#L68

nekopsykose: why not just the makefile install for the entire thing again

ttyyls: needs a patch.. which i'll add instead of trying to recreate it in the template.

nekopsykose:
this installs to /etc/bash_completion which is wrong

ttyyls:
> this installs to /etc/bash_completion which is wrong
patched it out and added the bash_completion install to post_install instead

nekopsykose:
also the commit message should say `user/git-extras`

ttyyls:
> pretty sure you don't need a patch and manual install, setting `COMPL_DIR` in `make_install_args` should be enough

im already patching the makefile anyway, I just thought i%'d be more consistent this way

q66:
i think just adding git to hostdeps should avoid having to patch anything?

PR #1915, opened 2024-05-01
main/libadwaita-demo: new package (1.5.0)

atheriel:
This commit makes the `libadwaita` demo application available via a subpackage of `libadwaita`.

A packaged demo application is a useful companion when developing against `libadwaita`, especially when combined with the GTK inspector.

### DIFF
--- a/main/libadwaita/template.py
+++ b/main/libadwaita/template.py
@@ -1,9 +1,9 @@
 pkgname = "libadwaita"
 pkgver = "1.5.0"
-pkgrel = 0
+pkgrel = 1
 build_style = "meson"
 configure_args = [
-    "-Dexamples=false",
+    "-Dexamples=true",
     "-Dtests=true",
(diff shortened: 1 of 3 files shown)


nekopsykose:
rather than `new package` it's `add demo application` for the commit

main/libadwaita/template.py
```diff
@@ -38,3 +38,10 @@
 @subpackage("libadwaita-devel")
 def _devel(self):
     return self.default_devel()
+
+
+@subpackage("libadwaita-demo")
+def _demo(self):
+    self.pkgdesc = f"{pkgdesc} (demo applications)"
+    return ["usr/bin/adwaita-1-demo"]
```
nekopsykose, line 46:
```suggestion
    return [
        "usr/bin/adwaita-1-demo",
        "usr/share",
    ]
```

it has other files too

PR #1942, opened 2024-05-04
contrib/aria2: new package

nullobsi:
I believe aria2 has a library that could be built too, but I haven't figured out how to achieve this with a subpackage yet or if it should be -devel or -libs.

### DIFF
--- a/contrib/aria2/template.py
+++ b/contrib/aria2/template.py
@@ -1,0 +1,57 @@
+pkgname = "aria2"
+pkgver = "1.37.0"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--disable-libaria2",
+    "--disable-werror",
+    "--enable-bittorrent",
+    "--enable-metalink",
+    "--enable-ssl",
+    "--enable-websocket",
+    "--enable-year2038",
+    "--enable-epoll",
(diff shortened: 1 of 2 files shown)


nekopsykose:
just leave the lib out for now

ttyyls:
try this for an update.py
```
url = "https://github.com/aria2/aria2/tags"
pkgname = "release"
```

contrib/aria2/template.py
```diff
@@ -0,0 +1,56 @@
...
+]
+makedepends = [
+    "openssl-devel",
+    "zlib-devel",
+    "libxml2-devel",
+    "c-ares-devel",
+    "libuv-devel",
+    "libssh2-devel",
+    "cppunit-devel",
+    "sqlite-devel",
```
nekopsykose, line 47: sort these

contrib/aria2/template.py
```diff
@@ -0,0 +1,56 @@
...
+]
+pkgdesc = "Lightweight multi-protocol download utility"
```
nekopsykose, line 49:
```suggestion
pkgdesc = "Multi-protocol download utility"
```

nekopsykose:
seems to work with hardening = vis, cfi too

PR #1944, opened 2024-05-04
contrib/bibata-cursor: new package (2.0.6)

Ogromny:


contrib/bibata-cursor/template.py
```diff
@@ -0,0 +1,30 @@
+pkgname = "bibata-cursor"
+pkgver = "2.0.6"
+pkgrel = 0
+pkgdesc = "Open source, compact, and material designed cursor set"
```
nekopsykose, line 4:
```suggestion
pkgdesc = "Material design cursor set"
```

contrib/bibata-cursor/template.py
```diff
@@ -0,0 +1,30 @@
...
+        "Bibata-Original-Amber-Right",
+        "Bibata-Original-Classic",
+        "Bibata-Original-Classic-Right",
+        "Bibata-Original-Ice",
+        "Bibata-Original-Ice-Right",
+    ]
+    for theme in themes:
+        self.install_files(theme, "usr/share/icons")
+
+    self.install_license("LICENSE")
```
JamiKettunen, line 30: GPL license doesn't need to be installed for future reference, I'm also wondering if the theme install stuff could've been made more concise somehow with a `glob=True` and `Bibata-*` but I didn't seem to find anything fancy for that

PR #1945, opened 2024-05-04
contrib/minetest: new package (5.8.0)

ttyyls:


ttyyls:
I've made an attempt at providing a service file with a _minetest user and tmpfile entry [here](https://github.com/chimera-linux/cports/compare/master...ttyyls:cports:minetest) bu I rolled it back since it'd require manual installation and loading of individual games and then changing the service file to make them work.

contrib/minetest/template.py
```diff
@@ -0,0 +1,76 @@
...
+]
+sha256 = [
+    "610c85a24d77acdc3043a69d777bed9e6c00169406ca09df22ad490fe0d68c0c",
+    "2fde8e27144988210b9c0ff1e202905834d9d25aaa63ce452763fd7171096adc",
+]
+tool_flags = {"CXXFLAGS": ["-Wno-deprecated-declarations"]}
+hardening = ["!int"]
+
+
+def post_extract(self):
```
JamiKettunen, line 48: Drop whole `post_extract()` and `build_wrksrc` above and just use
```py
source_paths = [".", "lib/irrlichtmt"]
```

ttyyls: looks cleaner that way. thanks

contrib/minetest/template.py
```diff
@@ -0,0 +1,69 @@
...
+    self.pkgdesc = f"{pkgdesc} (common files)"
+    self.install_if = [f"{pkgname}={pkgver}-r{pkgrel}"]
+
+    return ["usr/share/minetest"]
+
+
+@subpackage("minetest-server")
+def _server(self):
+    self.pkgdesc = f"{pkgdesc} (server)"
+    self.depends = ["minetest-common"]
```
q66, line 67: needs to be versioned

q66: also probably main package should also depend on it, no?

ttyyls: >main package should also depend on it
yes and it does through the install_if on minetest-common

q66: does that make sense over just a hard dependency in the other direction though

ttyyls: i changed it into a hard dependency in the main package.

PR #1958, opened 2024-05-08
contrib/libsixel: new package (1.10.3)

Natthias:
tmux can be built with sixel support, but I wanted to check if this should go in contrib and if we even wanted tmux to have sixel support enabled

### DIFF
--- a/contrib/libsixel/template.py
+++ b/contrib/libsixel/template.py
@@ -1,0 +1,33 @@
+pkgname = "libsixel"
+pkgver = "1.10.3"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+]
+makedepends = [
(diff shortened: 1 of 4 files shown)


contrib/libsixel/template.py
```diff
@@ -0,0 +1,40 @@
...
+@subpackage("libsixel-utils")
+def _utils(self):
+    self.pkgdesc = f"{pkgdesc} (utilities)"
+    return [
+        "usr/bin/img2sixel",
+        "usr/bin/sixel2png",
+        "usr/share/man/man1",
+        "usr/share/zsh",
+        "usr/share/bash-completion",
+    ]
```
triallax, line 40:
```suggestion
@subpackage("libsixel-progs")
def _progs(self):
    return self.default_progs()
```

triallax (changes requested):

PR #1961, opened 2024-05-10
contrib/nerd-fonts: new package (3.2.1)

Ogromny:
Maybe the `license` is wrong (each font has its own), otherwise everything should be good


contrib/nerd-fonts/template.py
```diff
@@ -0,0 +1,299 @@
...
+    f"{pkgname}-hack={pkgver}-r{pkgrel}",
+    f"{pkgname}-hasklig={pkgver}-r{pkgrel}",
+    f"{pkgname}-heavy-data={pkgver}-r{pkgrel}",
+    f"{pkgname}-hermit={pkgver}-r{pkgrel}",
+    f"{pkgname}-ia-writer={pkgver}-r{pkgrel}",
+    f"{pkgname}-ibm-plex-mono={pkgver}-r{pkgrel}",
+    f"{pkgname}-inconsolata-go={pkgver}-r{pkgrel}",
+    f"{pkgname}-inconsolata-lgc={pkgver}-r{pkgrel}",
+    f"{pkgname}-inconsolata={pkgver}-r{pkgrel}",
+    f"{pkgname}-intel-one-mono={pkgver}-r{pkgrel}",
```
nekopsykose, line 40: there's no point to have both a depends= here and an install_if below

contrib/nerd-fonts/nerd-fonts-0xproto
```diff
@@ -0,0 +1 @@
+nerd-fonts-0xproto
```
nekopsykose, line 1: the two symlinks inside the folder are broken (you should run ./cbuild relink-subpkgs)

q66:
since #1988 is in a better shape, i will close this one

PR #1963, opened 2024-05-11
user/qbe: new package (1.2)

z-erica:


contrib/qbe/template.py
```diff
@@ -0,0 +1,14 @@
+pkgname = "qbe"
+pkgver = "1.2"
+pkgrel = 0
```
nekopsykose, line 3: archs = x8664, riscv64, aarch64

contrib/qbe/template.py
```diff
@@ -0,0 +1,14 @@
...
+build_style = "makefile"
```
nekopsykose, line 4: also need to add cflags into ldflags with a patch

contrib/qbe/patches/fix_asm_keyword.patch
```diff
@@ -0,0 +1,62 @@
+--- a/arm64/emit.c	2024-01-29 10:24:41.000000000 +0100
```
nekopsykose, line 1: add a link to where it's from

nekopsykose: enjoy sending it :3

user/qbe/template.py
```diff
@@ -0,0 +1,16 @@
+pkgname = "qbe"
+pkgver = "1.2"
+pkgrel = 0
+archs = ["x86_64", "riscv", "aarch64"]
```
nekopsykose, line 4:
```suggestion
archs = ["x86_64", "riscv64", "aarch64"]
```

contrib/qbe/patches/fix_asm_keyword.patch
nekopsykose: sure, though i would just send a new patch anyway because they are discussing random stuff in there

user/qbe/template.py
```diff
@@ -0,0 +1,16 @@
+pkgname = "qbe"
+pkgver = "1.2"
+pkgrel = 0
+archs = ["x86_64", "riscv64", "aarch64"]
```
q66, line 4: sort this list

user/qbe/patches/fix_asm_keyword.patch
```diff
@@ -0,0 +1,62 @@
+--- a/arm64/emit.c	2024-01-29 10:24:41.000000000 +0100
```
triallax, line 1: instead of this patch you could also set `tool_flags = {"CFLAGS": ["-std=c99"]}` in the template (which is what the upstream Makefile sets anyway) and remove this patch entirely, but this patch is also fine

PR #1968, opened 2024-05-12
contrib/yash: new package (2.56.1)

Ogromny:


contrib/yash/template.py
```diff
@@ -0,0 +1,30 @@
...
+]
+makedepends = ["ncurses-devel"]
+pkgdesc = "Yet another shell"
+maintainer = "ogromny <ogromnycoding@gmail.com>"
+license = "GPL-2.0-only"
+url = "https://github.com/magicant/yash"
+source = f"{url}/releases/download/{pkgver}/{pkgname}-{pkgver}.tar.xz"
+sha256 = "f7f5a1ffd246692568e4823a59b20357317d92663573bd1099254c0c89de71f5"
+# for some reasons with cbuild 6 tests fail, running the test manually works fine though
+options = ["!check"]
```
nekopsykose, line 26: should figure out why

contrib/yash/template.py
```diff
@@ -0,0 +1,30 @@
+pkgname = "yash"
+pkgver = "2.56.1"
+pkgrel = 0
+build_style = "configure"
```
q66, line 4: gnu_configure + `configure_gen = []`

contrib/yash/template.py
```diff
@@ -0,0 +1,30 @@
...
+    "--enable-dirstack",
+    "--enable-double-bracket",
+    "--enable-help",
+    "--enable-history",
+    "--enable-lineedit",
+    "--enable-printf",
+    "--enable-socket",
+    "--enable-test",
+    "--enable-ulimit",
+    "--prefix=/usr",
```
q66, line 16: after changing the buildstyle drop this

contrib/yash/template.py
Ogromny: I can't otherwise I get this error:
```
/builddir/yash-2.56.1/configure: --sysconfdir=/etc: unknown argument
```

yash's configure will return 1 if it doesn't recognize an argument:
https://github.com/magicant/yash/blob/trunk/configure#L156

q66: patch out the error

q66: hm, or actually maybe don't... doesn't seem that beneficial here, let's keep configure

PR #1969, opened 2024-05-12
main/musl: update to 1.2.5_git20240511

nekopsykose:
no particular reason other than rebuilding this since the framepointer changes


main/musl-mallocng/patches/getdents.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/dirent/posix_getdents.c b/src/dirent/posix_getdents.c
```
nekopsykose, line 1: any reason mallocng has a separate patch folder?

main/musl-mallocng/template.py
```diff
@@ -24,14 +24,20 @@
...
 # scp makes it segfault
 hardening = ["!scp"]
 # does not ship tests
 options = ["!check", "!lto"]
 
 
+def post_extract(self):
+    # reported in libc.so --version
+    with open(self.cwd / "VERSION", "w") as f:
+        f.write(pkgver)
```
nekopsykose, line 38: don't care too much about this (and left it out of cross), just a slightly nicer version display. could put the commit in it instead too..

main/musl-mallocng/patches/getdents.patch
q66: because the main patches won't apply unless compiler-rt sources are present

main/musl-mallocng/template.py
q66: this is ok

main/musl-mallocng/patches/getdents.patch
q66: (the alternative is to unify sources for mallocng and always apply scudo into the codebase, which is actually maybe better? idk, what do you think?)

nekopsykose: hm right

i think that is better since then we effectively have identical templates that differ only in a configure arg. or at least it seems better to me since there's less diff

nekopsykose: pushed what i think is correct, should be the same now. i didn't test it at runtime though..

nekopsykose: ok, tested. stuff uhh still works, and based on build log it's still mallocng and not scudo, looks fine

q66: looks good :)

q66:
why are the patches gone?

q66:
ah nevermind sorry, brainfart

main/musl/patches/getdents.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/dirent/posix_getdents.c b/src/dirent/posix_getdents.c
```
nekopsykose, line 1: i sent this in https://www.openwall.com/lists/musl/2024/05/12/2 for reference

q66: ty

main/musl-mallocng/patches/getdents.patch
nekopsykose: thanks for looking!

PR #1970, opened 2024-05-12
main/wasi-libc: update to 0.20240412

nekopsykose:
matches latest 'release' sdk-22 https://github.com/WebAssembly/wasi-libc/releases/tag/wasi-sdk-22


nekopsykose:
could probably add an update check based on the tags but not sure how to make it spit out back a date

nekopsykose:
idk if this needs compiler-rt/libcxx rebuilds either

PR #1974, opened 2024-05-13
contrib/gocryptfs: new package (2.4.0)

na-sa-do:


contrib/gocryptfs/template.py
```diff
@@ -0,0 +1,33 @@
...
+    f"{url}/releases/download/v{pkgver}/{pkgname}_v{pkgver}_src-deps.tar.gz"
+)
+sha256 = "45158daf20df7f94e0c9ec57ba07af21df2e25e15b8584bf3c7de96adbbc2efd"
+# requires fuse kernel module, also wants the no-OpenSSL build to be static but we don't care about that
+# (no standard build system means no default check phase anyway, but still...)
+options = ["!check"]
+
+
+def do_build(self):
+    self.do("./build.bash")
```
nekopsykose, line 27: this doesn't seem to do anything interesting that the go build_style wouldn't do already (just pass the paths to build in make_build_args)

you can set main.GitVersion in there too via ldflags for the version

nekopsykose: it also sets GOAMD64=v2 which is wrong

nekopsykose: in case you don't feel like figuring it out:  

```
pkgname = "gocryptfs"
pkgver = "2.4.0"
pkgrel = 0
build_style = "go"
make_build_args = [
    "-ldflags="
    + f" -X main.GitVersion=v{pkgver}"
    + " -X main.GitVersionFuse=[vendored]",
    "./",
    "./gocryptfs-xray",
]
hostmakedepends = [
    "go",
    "pkgconf",
]
makedepends = ["openssl-devel"]
pkgdesc = "Encrypted overlay filesystem"
maintainer = "Nasado <hi@nasado.name>"
license = "MIT"
url = "https://github.com/rfjakob/gocryptfs"
source = (
    f"{url}/releases/download/v{pkgver}/{pkgname}_v{pkgver}_src-deps.tar.gz"
)
sha256 = "45158daf20df7f94e0c9ec57ba07af21df2e25e15b8584bf3c7de96adbbc2efd"
# requires fuse kernel module, also wants the no-OpenSSL build to be static but we don't care about that
# (no standard build system means no default check phase anyway, but still...)
options = ["!check"]


def post_install(self):
```

note fuse-devel is unused. also you probably want `depends = ["fuse"]`?

contrib/gocryptfs/template.py
```diff
@@ -0,0 +1,33 @@
...
+pkgdesc = "Encrypted overlay filesystem"
+maintainer = "Nasado <hi@nasado.name>"
+license = "MIT"
+url = "https://github.com/rfjakob/gocryptfs"
+source = (
+    f"{url}/releases/download/v{pkgver}/{pkgname}_v{pkgver}_src-deps.tar.gz"
+)
+sha256 = "45158daf20df7f94e0c9ec57ba07af21df2e25e15b8584bf3c7de96adbbc2efd"
+# requires fuse kernel module, also wants the no-OpenSSL build to be static but we don't care about that
+# (no standard build system means no default check phase anyway, but still...)
```
nekopsykose, line 22:
```suggestion
# requires fuse kernel module
```

contrib/gocryptfs/template.py
nekopsykose: you also need

```
    self.install_man("Documentation/gocryptfs.1")
    self.install_man("Documentation/gocryptfs-xray.1")
```

in post_install

contrib/gocryptfs/template.py
```diff
@@ -0,0 +1,37 @@
...
+makedepends = [
+    "openssl-devel",
+]
+depends = [
+    "fuse",
+]
+pkgdesc = "Encrypted overlay filesystem"
+maintainer = "Nasado <hi@nasado.name>"
+license = "MIT"
+url = "https://github.com/rfjakob/gocryptfs"
```
nekopsykose, line 25:
```suggestion
url = "https://nuetzlich.net/gocryptfs"
```

na-sa-do: I saw that website but by the looks of it it hasn't been updated in literal years, so I figured linking to the github would make more sense, but idk.

nekopsykose: ah, you are right

nekopsykose:
thanks!

PR #1979, opened 2024-05-13
contrib/certbot: enable plugins

ttyyls:


nekopsykose:
maybe we should refactor python into `util.` so it's possible to call build/check/install from it instead of it being hardcoded in the build style. check being implemented as separate util steps would be useful in some other places too..

ttyyls:
@nekopsykose i agree. What I did here is anything but ideal hence the draft status.

nekopsykose:
yeah, looks about as good as it could get

contrib/certbot/template.py
```diff
@@ -25,11 +42,72 @@
...
+            "build",
+            "--wheel",
+            "--no-isolation",
+            f"../{plugin}",
+        )
+
+
 def post_install(self):
     self.install_license("LICENSE.txt")
+    from cbuild.util import python
```
nekopsykose, line 63: imports at the top (above install_license)

contrib/certbot/template.py
```diff
@@ -25,11 +42,72 @@
...
+            f"../{plugin}/dist/{plugin.replace('-','_')}-{pkgver}-py3-none-any.whl",
+        )
+    python.precompile(self, "usr/lib")
+
+
+@subpackage("certbot-plugins")
+def _pluggie(self):
+    self.pkgdesc = "certbot plugins"
+    self.install_if = [f"{pkgname}={pkgver}-r{pkgrel}"]
+    self.depends = depends + _plugins
```
nekopsykose, line 81: this isn't strictly correct as _plugins should be the name=pkgver-rpkgrel form for each

nekopsykose: it's probably easier to put an install_if in the _genmod on [pkgname=pkgver-rpkgrel, certbot-plugins=pkgver-rpkgrel]

ttyyls: i could probably omit the self.depends line then since it's inherited from the templates' depends

nekopsykose: each plugin should individually depend on pkgname=pkgver-rpkgrel

nekopsykose: (so in genmod)

ttyyls: sure

contrib/certbot/template.py
```diff
@@ -25,11 +42,72 @@
...
+            "installer",
+            "--destdir",
+            self.chroot_destdir,
+            f"../{plugin}/dist/{plugin.replace('-','_')}-{pkgver}-py3-none-any.whl",
+        )
+    python.precompile(self, "usr/lib")
+
+
+@subpackage("certbot-plugins")
+def _pluggie(self):
```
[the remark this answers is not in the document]
ttyyls: ikr :3

q66:
is there a reason to split all the plugins?

contrib/certbot/template.py
```diff
@@ -25,11 +42,77 @@
...
+
 def post_install(self):
+    from cbuild.util import python
+
+    for plugin in _plugins:
+        self.do(
+            "python",
+            "-m",
+            "installer",
+            "--destdir",
```
q66, line 69: add --compile-bytecode, drop the manual precompile

ttyyls: ok

contrib/certbot/template.py
```diff
@@ -25,11 +42,75 @@
...
+            f"certbot-plugins={pkgver}-r{pkgrel}",
+        ]
+        self.depends += [f"{pkgname}={pkgver}-r{pkgrel}"]
+
+        return [
+            f"usr/lib/python*/site-packages/certbot_{pname.replace('-', '_')}*"
+        ]
+
+
+for _plugin, _desc in [
```
q66, line 99: if these remain split, these descs are weird (DNS should be capitalized, Ovh should be OVH, etc.)

the plugins metapackage pkgdesc is also bad

ttyyls:
> is there a reason to split all the plugins?

not really, but it's very reasonable for someone to use just one of the webserver integrations with the builtin http-01 authenticators or use a platform specific dns-01 one.

q66:
but the plugins seemingly have no dependencies, so is there any harm in just having them be there

ttyyls:
> but the plugins seemingly have no dependencies, so is there any harm in just having them be there

no. i suppose i should the metapackage install_if then?

q66:
if they don't harm anything (don't add deps, don't get in the way, don't grow the package too much, etc) i just wouldn't bother splitting them at all

ttyyls:
well not splitting anything grows the installed size from 1876 KiB to 4520 KiB. But that's it

ttyyls:
looking back at this after some testing the full package will pull an aws sdk and its dependencies, cloudflare sdk, some dns53, alternative config parsing libraries since some have their own format and the overall package would be quite heavy.

Thus I resplit it and I'm currently adding them conditional by matching on the plugin name. and I elect to **not** install_if making people either opt for installing the base package with ther plugin of choice or all the plugins. the latter is bound to be a buggier experience that most people fine with just using the default nginx/apache config would rather avoid.

nekopsykose:
ah, right cause the plugins are missing the actual deps they use at runtime

ttyyls:
I believe I've caught all runtime dependencies (barring the google plugin which i yanked out).

This may be further cherry-picked to yank out outdated plugins or ones with undesirable dependencies, but for the most part I believe this to be complete

contrib/python-jmespath/template.py
```diff
@@ -0,0 +1,25 @@
...
+pkgdesc = "JSON matching expressions"
+maintainer = "ttyyls <contact@behri.org>"
+license = "MIT"
+url = "https://jmespath.org"
+source = (
+    f"https://github.com/jmespath/jmespath.py/archive/refs/tags/{pkgver}.tar.gz"
+)
+sha256 = "6a02470b1716ec7a32abe89a873a4795c41c938468225f8a53d860980ec9e3c6"
+# tests hang forever
+options = ["!check"]
```
nekopsykose, line 21: figure out which one and skip it or whatever

contrib/python-digitalocean/template.py
```diff
@@ -0,0 +1,20 @@
...
+depends = ["python-requests", "python-jsonpickle", "python-pyyaml"]
+checkdepends = depends + ["python-pytest", "python-responses"]
+pkgdesc = "Python module to manage Digital Ocean droplets"
+maintainer = "ttyyls <contact@behri.org>"
+license = "LGPL-3.0-or-later"
+url = "https://github.com/koalalorenzo/python-digitalocean"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "9c9c788ae03a088d0c03a9a59ff7ac6c492caadd4942d4fc58795ee859fc228f"
+# failing tests
+options = ["!check"]
```
nekopsykose, line 20: well figure it out

nekopsykose:
sort the depends lists

contrib/python-digitalocean/template.py
ttyyls: some tests were trying to do network calls so i skipped them

PR #1985, opened 2024-05-13
contrib/cmatrix: new package (2.0)

ttyyls:


nekopsykose:
not related to the packaging (looks fine), but is there a way to get `-c` to work? i never figured it out and it seems broken upstream

nekopsykose:
hm maybe its https://github.com/abishekvashok/cmatrix/commit/7647b4774c72a0803b36d89ece4edd95f5dbb97f

nekopsykose:
yep, here's a rebased version

```
From 7647b4774c72a0803b36d89ece4edd95f5dbb97f Mon Sep 17 00:00:00 2001
From: patatahooligan <filwn.oikonomou@gmail.com>
Date: Thu, 20 Oct 2022 07:23:15 +0300
Subject: [PATCH] Fix unicode character printing (#112)

* Fix unicode character printing

Fix several issues that prevented it from working

  * link to the wide-character version of ncurses
  * define NCURSES_WIDECHAR
  * use a function that can print wide characters
  * fix the character range. I don't know what the original was supposed
    to be, but half-width kana (which is what the movie reportedly used)
    was not at that range.

* Fix half-width kana character range

I'm not very familiar with katakana, but I think the extra characters I
was using are punctuation or other marks. So let's remove them. Note
that this is the same range space-pagan is using.
---
 CMakeLists.txt |  1 +
 cmatrix.c      | 21 +++++++++++++++++----
 2 files changed, 18 insertions(+), 4 deletions(-)

diff --git a/CMakeLists.txt b/CMakeLists.txt
index 6c50d7b..68ee176 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ if	(HAVE_GETOPT_H)
 endif	()
 
 Set(CURSES_NEED_NCURSES TRUE)
+Set(CURSES_NEED_WIDE TRUE)
 find_package(Curses)
 include_directories(${CURSES_INCLUDE_DIR})
 add_definitions(-DHAVE_NCURSES_H)
diff --git a/cmatrix.c b/cmatrix.c
index 7e3fbb9..7d8ca39 100644
--- a/cmatrix.c
+++ b/cmatrix.c
@@ -37,6 +37,8 @@
 #include "config.h"
 #endif
 
+#define NCURSES_WIDECHAR 1
+
 #ifdef HAVE_NCURSES_H
 #include <ncurses.h>
 #else
@@ -437,8 +439,8 @@
     /* Set up values for random number generation */
     if(classic) {
         /* Japanese character unicode range [they are seen in the original cmatrix] */
-        randmin = 12288;
-        highnum = 12351;
+        randmin = 0xff66;
+        highnum = 0xff9d;
     } else if (console || xwindow) {
         randmin = 166;
         highnum = 217;
@@ -739,7 +741,10 @@
                         } else if (lambda && matrix[i][j].val != ' ') {
                             addstr("λ");
                         } else {
-                            addch(matrix[i][j].val);
+                            wchar_t char_array[2];
+                            char_array[0] = matrix[i][j].val;
+                            char_array[1] = 0;
+                            addwstr(char_array);
                         }
                         if (bold == 2 ||
                             (bold == 1 && matrix[i][j].val % 2 == 0)) {
```

nekopsykose:
hm or is it, it's still like half-discolored

contrib/cmatrix/template.py
```diff
@@ -0,0 +1,13 @@
+pkgname = "cmatrix"
+pkgver = "2.0"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = ["automake"]
```
nekopsykose, line 5: cmake

ttyyls:
i could probably just switch the template to using that latest commit?

nekopsykose:
just that patch is the same effect, but yeah just _git this given how old it is and put it in user/

also, like

- on 2.0, -c is broken. ncurses is black bg, cmake is gray
- with those commits, -c is fixed, but the same bg stuff happens. why is cmake a different bg color?

ttyyls:
i guess let's not instead

nekopsykose:
haha, sure

PR #1986, opened 2024-05-13
main/libsoup: update to 3.5.1

triallax:


nekopsykose:
i wasnt sure if this is 'beta' or not, did they stop doing oddvers

triallax:
good point, i'm not quite sure so i guess i'll ask in the gnome matrix room real quick

q66:
https://gitlab.gnome.org/GNOME/libsoup/-/tags

>3.5.1 (dev)

triallax:
guess we'll want to ignore oddvers in update.py then?

PR #1670, opened 2024-03-26
contrib/wezterm: new package (20240203)

ttyyls:

PR #1990, opened 2024-05-14
contrib/fcitx5: new package (5.1.9)

ifreund:
Not very generally useful yet as `fcitx5-configtool` is missing and looks painful to package. This is enough for me to test the thing I needed fcitx5 for though and is a valuable starting point for anyone wanting to take this further.


contrib/fcitx5/template.py
```diff
@@ -0,0 +1,55 @@
+pkgname = "fcitx5"
+pkgver = "5.1.9"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DENABLE_EMOJI=OFF", "-DENABLE_X11=OFF"]
```
nekopsykose, line 5: is there a reason x11 (and emoji) is disabled?

ifreund: None other than me being too lazy to package the required dependencies and the fact that they aren't required for what I wanted fcitx for.

nekopsykose: i guess you'd need xcb-imdkit, i'll add that

nekopsykose: added

emoji also seems to build but you'd know more than me

nekopsykose:
i'll add configtool after to flesh it out a bit myself

contrib/fcitx5/template.py
ttyyls: @nekopsykose xcb-imdkit is part of my wezterm pr on #1670

ifreund: > emoji also seems to build but you'd know more than me

Indeed, I think I read some outdated information somewhere. It used to required additional dependencies but no longer does it seems: https://github.com/fcitx/fcitx5/commit/fb56879d0fb05984656fe4a103f3a6fabe37c7fb

Thanks for taking care of xcb-imdkit, I've pushed a new version with X11 and emoji support enabled.

nekopsykose:
thanks!

PR #1992, opened 2024-05-15
contrib/python-numpy: new package (1.26.4)

z-erica:


contrib/openblas/template.py
```diff
@@ -0,0 +1,39 @@
...
+# FIXME: flang support
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+pkgdesc = "Optimized BLAS library"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "BSD-3-Clause"
+url = "https://www.openblas.net"
+source = f"https://github.com/OpenMathLib/OpenBLAS/releases/download/v{pkgver}/OpenBLAS-{pkgver}.tar.gz"
+sha256 = "aa2d68b1564fe2b13bc292672608e9cdeeeb6dc34995512e65c3b10f4599e897"
+
+match self.profile().arch:
```
q66, line 16: afaict

1) `DYNAMIC_ARCH` does not apply everywhere; it can only be applied on aarch64, x86_64, riscv64, and ppc64le for us so i'd move it to the arch conditionals
2) for x86_64 the TARGET should probably be GENERIC, not CORE2
3) for ppc64 the target should be PPC970MP

q66: i'd also do USE_OPENMP=1 in general

q66: SSE2 is a part of x86_64 base spec, i don't think you can have it without that

nekopsykose: > (and I think with dynamic arch support all the simd stuff gets compiled in regardless)

yeah it does

> SSE2 is a part of x86_64 base spec, i don't think you can have it without that  

indeed it is, but clearly they have generic separately and core2 is a different path, so maybe it's just not used anyway, and so core2 skips a bit of baseline stuff that would be unused

would probably have to dig in the whole thing to see what the real difference is

q66: looks ok yea

q66:
the tests for numpy don't fully pass rn (the suite does not error/abort the build, but a bunch of tests fail)

but i guess that's ok for now (we're probably best off fixing stuff up when we have better flang support)

PR #1994, opened 2024-05-15
main/mandoc: move apropos stuff to main pkg, enable int+lto+vis+cfi

triallax:
The motivation for moving `mandoc-apropos`'s contents (save the trigger) to the main package is to allow people to do `apk add !mandoc-apropos` but still be able to run `makewhatis`, e.g. in a cron job, instead of using the apk trigger (because the command can take quite a bit).


triallax:
i was thinking about renaming `mandoc-apropos` to `mandoc-apropos-trigger` as well but there doesn't seem to be any precedent for that, and also `mandoc-apropos-trigger.trigger` is pretty ugly so idk, let me know what you think

PR #1996, opened 2024-05-15
contrib/minify: update to 2.20.22

ifreund:


contrib/minify/template.py
```diff
@@ -1,5 +1,5 @@
 pkgname = "minify"
-pkgver = "2.20.19"
+pkgver = "2.20.22"
 pkgrel = 1
```
nekopsykose, line 3:
```suggestion
pkgrel = 0
```

ifreund: done, thanks.

PR #1997, opened 2024-05-15
contrib/sysprof: new package (46.0)

nekopsykose:


contrib/sysprof/template.py
```diff
@@ -0,0 +1,44 @@
...
+pkgdesc = "System-wide profiler for Linux"
+maintainer = "psykose <alice@ayaya.dev>"
+license = "GPL-2.0-or-later"
+url = "https://www.sysprof.com"
+source = f"$(GNOME_SITE)/sysprof/{'.'.join(pkgver.rsplit('.')[:-1])}/sysprof-{pkgver}.tar.xz"
+sha256 = "73aa7e75ebab3e4e0946a05a723df7e6ee4249e3b9e884dba35500aba2a1d176"
+
+
+def post_install(self):
+    self.rm(self.destdir / "usr/systemd", recursive=True)
```
nekopsykose, line 39: this is a service for sysprofd which is a way for a container sysprof to request perms on the host- you don't need this with a system package so i didn't make a dinit version (though it would be trivial)

nekopsykose: added anyway

nekopsykose: to be exact- if someone ran sysprof from a container they would want this to exist on the host, so it's still useful

PR #2013, opened 2024-05-19
cbuild: make maintainer global to cports clone

q66:
The maintainer field in packages for now still exits but is ignored. All packages get a fallback name that primarily exists to identify the origin cports instance. That is, people are free to choose a name for their local repos.

Our builders will get a `Chimera Linux <cports@chimera-linux.org>` set on each builder.

This is a part of an effort to drop the concept of template ownership. Any authorship can be tracked through git already, and this deals with the issue of orphaning, especially considering in practice nobody really owns templates in cports and this has existed only because the concept exists in apk itself.

### DIFF
--- a/etc/config.ini.example
+++ b/etc/config.ini.example
@@ -52,6 +52,8 @@ remote = yes
 categories = main contrib user
 # whether restricted packages can be considered for building
 allow_restricted = no
+# maintainer field to use for packages, you can set it for personal repos
+maintainer = unknown <cports@local>
 
 # flags passed to tools
 [flags]
--- a/src/cbuild/core/build.py
+++ b/src/cbuild/core/build.py
@@ -17,6 +17,7 @@ def build(
     no_update=False,
     update_check=False,
     accept_checksums=False,
+    maintainer=None,
 ):
     if chost:
         depn = "host-" + pkg.pkgname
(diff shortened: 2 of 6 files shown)


q66:
Opening this for feedback. I will leave it open a few days, then merge and drop the field globally from all templates, unless there are any reasonable objections.

wezm:
No objections from me. Will you retire https://repo.chimera-linux.org/cports-updates/cports-updates.txt or rebuild it based on git?

q66:
it'll just become a flat list

a16bitsysop:
Would there be another way to check what packages needed updating that you maintain?

q66:
probably not for now, though i'm thinking maybe we could implement update-check sorting based on which people touched it in git, would probably be sufficient enough...

a16bitsysop:
> depends on what you mean by "maintain" though? because if this pr is merged we wouldn't have an official concept of maintainership anymore

Yes I understand that, but still there might be packages you might want to keep up to date that you are using etc.  Or some packages aren't straightforward to update.

subnut:
### Please pin this issue for higher visibility.

nekopsykose:
you can't pin pull requests

q66:
done in 147c731364d29c42322b233273b2c84ce216554d, field/update-check for now stays until i come up with something better, but generated packages ignore the maintainer field

PR #2017, opened 2024-05-19
main/ncurses: update to 6.5

nekopsykose:


main/ncurses/template.py
```diff
@@ -1,8 +1,10 @@
 pkgname = "ncurses"
-pkgver = "6.4"
-pkgrel = 2
+pkgver = "6.5"
+pkgrel = 0
 build_style = "gnu_configure"
 configure_args = [
+    "--disable-root-access",
```
nekopsykose, line 6: there's also root-environ but idk if that should also be disabled

q66: all i can find is that arch enables it and then https://github.com/kovidgoyal/kitty/issues/6842

they still enable it https://gitlab.archlinux.org/archlinux/packaging/packages/ncurses/-/blob/main/PKGBUILD?ref_type=heads#L55

q66:
i also wonder if `--enable-fvisibility` could be used... or if it causes some stuff to break

main/ncurses/template.py
nekopsykose: better to keep it on then :)

nekopsykose:
> i also wonder if --enable-fvisibility could be used... or if it causes some stuff to break  

afaict it does the same thing hardening=vis would do so i would set that instead. since it's now upstream it probably doesn't break anything...

q66:
if using vis we'd still need the configure arg for robustness since the build system can take different choices based on the setting and ncurses build system is not exactly known for not doing sketchy things

nekopsykose:
sure

i checked symbol diff and is the same

q66:
nice

main/ncurses/template.py
q66: debian disables root-access but not root-environ

nekopsykose: guess we'll match that then

PR #2019, opened 2024-05-19
contrib/yadm: new package (3.2.2)

hgersen:


contrib/yadm/template.py
```diff
@@ -0,0 +1,19 @@
+pkgname = "yadm"
+pkgver = "3.2.2"
+pkgrel = 0
+depends = ["git"]
```
nekopsykose, line 4: also depends on bash since it's actually a bash script

q66:
you should set up the test suite

nekopsykose:
also need to squishy the commits into 1

contrib/yadm/template.py
hgersen: 1) I've added bash as a dependency

2) The script has both "gawk" and "awk" on [this line](https://github.com/TheLocehiliosan/yadm/blob/master/yadm#L50). However in multiple places in the script it is explicitly calling only the [first element of that array](https://github.com/TheLocehiliosan/yadm/blob/0a5e7aa353621bd28a289a50c0f0d61462b18c76/yadm#L455) including when checking availability of awk.  This is weird, but might indeed mean it needs gawk for templates.

nekopsykose: it means it checks gawk first. it might still work with either- that's for you to check :)  

the testsuite would probably catch it too

hgersen: 1) I've tested that the default template functionality works with just awk. 

2) Getting this testsuite to work looks like a headache and likely first needs fixes to the yadm Makefile that I avoided because it isn't installing completions

nekopsykose: hm yeah this looks a bit hard to run

PR #2022, opened 2024-05-20
libkkc

na-sa-do:
Building fails with "ERROR: can't resolve libraries to shared libraries: m, atomic". Not sure what to do about that; manually depending on libatomic-chimera(-devel) doesn't help and IIUC libm is just part of libc under musl.

### DIFF
--- a/contrib/libkkc/template.py
+++ b/contrib/libkkc/template.py
@@ -1,0 +1,45 @@
+pkgname = "libkkc"
+pkgver = "0.3.5"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_env = {"MAKE": "gmake"}
+make_cmd = "gmake"
+hostmakedepends = [
+    "autoconf",
+    "automake",
+    "gettext",
+    "gmake",
+    "gnome-common",
+    "gobject-introspection",
+    "intltool",
+    "libtool",
+    "pkgconf",
+    "vala",
+]
+makedepends = [
+    "gettext-devel",
+    "json-glib-devel",
+    "libgee-devel",
+    "marisa-trie-devel",
+]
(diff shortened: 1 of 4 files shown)


contrib/libkkc/template.py
```diff
@@ -0,0 +1,45 @@
...
+
+    with open("README.md") as f:
+        readme_lines = f.read().split("\n")
+        relevant_lines = readme_lines[60:-1]
+        assert relevant_lines[0] == "GPLv3+"
+        assert not any(lambda x: x == "```", relevant_lines)
+    with open("extra-license") as f:
+        for line in relevant_lines:
+            f.write(line + "\n")
+    self.install_file("extra-license", f"{license_dir}/custom")
```
nekopsykose, line 45: you definitely dont need any of this

contrib/marisa-trie/template.py
```diff
@@ -0,0 +1,33 @@
+pkgname = "marisa-trie"
+pkgver = "0.2.6"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--enable-static"]
+make_dir = "."
+hostmakedepends = [
+    "autoconf",
+    "automake",
+    "clang",
```
nekopsykose, line 10:
```suggestion
    "automake",
```

nekopsykose: well, just copy the version from https://github.com/chimera-linux/cports/pull/1957 for this, though i think i will add this myself for something else first

nekopsykose:
ah it was this; fixing that isnt too hard but this then fails in like 10 spots later because kkc is an ancient codebase and idk how to fix any of the other parts :D

PR #2026, opened 2024-05-20
contrib/bats: new package (1.11.0)

hgersen:


nekopsykose:
not sure if it should be bats or bats-core but 90% of distros picked bats so i guess this is better

PR #2027, opened 2024-05-20
main/musl-locales: new package (0.1.0)

triallax:
- **main/musl-locales: new package (0.1.0)**
- **main/base-files: install musl-locales with base-locale**
- **main/musl: add default MUSL_LOCPATH for musl-locales**
- **main/musl-mallocng: add default MUSL_LOCPATH for musl-locales**


triallax:
testing would be appreciated, i tried this and it seems to work fine

PR #2037, opened 2024-05-22
contrib/shotman: update to 0.4.6

triallax:


triallax:
haven't tested this so that would be appreciated

i did a cross build for ppc64 (commenting out the stuff that required !cross) and it worked fine, but i'm not sure if the issue solved by the patch only appeared on a native ppc64 build

nekopsykose:
works fine

PR #2049, opened 2024-05-23
contrib/kanata: new package (1.6.1)

hgersen:
Note: currently works for root only, which does the job for my use. Adding the ability to run as user required a fix in uaccess input rules and suitable udev rule.

### DIFF
--- a/contrib/kanata/template.py
+++ b/contrib/kanata/template.py
@@ -1,0 +1,12 @@
+pkgname = "kanata"
+pkgver = "1.6.1"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo-auditable"]
+makedepends = ["rust-std"]
+pkgdesc = "Software keyboard remapper"
+maintainer = "hge <h.gersen@gmail.com>"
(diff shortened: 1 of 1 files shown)


contrib/kanata/template.py
```diff
@@ -0,0 +1,12 @@
+pkgname = "kanata"
+pkgver = "1.6.1"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo-auditable"]
+makedepends = ["rust-std"]
+pkgdesc = "Command-line benchmarking tool"
```
nekopsykose, line 7:
```suggestion
pkgdesc = "Keyboard remapping tool"
```

nekopsykose:
i'll wait to add the uaccess uinput rule for this first

> Adding the ability to run as user required a fix in uaccess input rules and suitable udev rule.

it also needs to be in input group for raw access to /dev/input** devices, since there's no other way to get those i guess. but with those two things it works as user yeah

q66:
sorry this took me so long, it disappeared from my mind and i finally forced myself to address this now :P

PR #2055, opened 2024-05-24
main/e2fsprogs: update to 1.47.1

nekopsykose:
also add fuse2fs

this now also ships 2 udev rules:

```
# Try to hide our fsck snapshots from udev's /dev/disk linking...
ACTION=="add|change", ENV{DM_LV_NAME}=="*.e2scrub", ENV{UDISKS_IGNORE}="1", OPTIONS="link_priority=-100"
```

```
# Don't let udisks automount ext4 filesystems without even asking a user.
# This doesn't eliminate filesystems as an attack surface; it only prevents
# evil maid attacks when all sessions are locked.
#
# According to http://storaged.org/doc/udisks2-api/latest/udisks.8.html,
# supplying UDISKS_AUTO=0 here changes the HintAuto property of the block
# device abstraction to mean "do not automatically start" (e.g. mount).
SUBSYSTEM=="block", ENV{ID_FS_TYPE}=="ext2|ext3|ext4|ext4dev|jbd", ENV{UDISKS_AUTO}="0"
```

not sure if we want the latter (the former is old and just wasn't installed because of missing udev-devel)


q66:
the latter is kinda sketch, still don't see why this should be guarded on file system type... may wanna drop it

nekopsykose:
sure. no strong feelings about it

PR #2058, opened 2024-05-25
contrib/flite: new package (2.2)

nekopsykose:
no idea how to fix the fortify meme tbh


q66:
funnily flite was already packaged once (removed in 0cbe25c2f3901c353815278bbb173b66d0441bc8), if we're reintroducing it it should probably go in main (as it can be used from gst plugin that is in main)

i originally nuked it because it seemed kinda dead, and generally a pain (and it did not build with then-gmake at all)

i have no idea about the const/fortify thing, i'll figure something out...

nekopsykose:
> as it can be used from gst plugin that is in main  

and webkit uses it too i think

nekopsykose:
but mostly this is also used by speechd, and speechd is a dep of firefox that it warns you about on first start that it needs it to do TTS. unless there was more speechd backends..

nekopsykose:
ah, seems it can use espeak-ng and flite is optional and prolly worse even

nekopsykose:
yea don't need this since espeak-ng fills the only relevant things

PR #2059, opened 2024-05-25
contrib/etckeeper: new package (1.18.21)

hgersen:


contrib/etckeeper/patches/use_apk.patch
```diff
@@ -0,0 +1,16 @@
+--- etckeeper.conf.orig	2024-05-21 18:34:47.193839464 +0100
++++ etckeeper.conf	2024-05-21 18:35:43.190313880 +0100
```
nekopsykose, line 2:
```suggestion
--- a/etckeeper.conf.orig	2024-05-21 18:34:47.193839464 +0100
+++ b/etckeeper.conf	2024-05-21 18:35:43.190313880 +0100
```

these should all be p1 form (bsd patch seems to accept it even when -p1 is passed.. but iirc gnu patch will not)

contrib/etckeeper/template.py
```diff
@@ -0,0 +1,19 @@
...
+pkgver = "1.18.21"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_install_args = [
+    "PYTHON=/bin/false",
+    "zshcompletiondir=${prefix}/share/zsh/site-functions",
+]
+make_check_target = "test"
+hostmakedepends = ["gmake", "gsed", "perl"]
```
q66, line 11: where is the gsed being used?

hgersen: I thought it was needed [here in the makefile](https://git.joeyh.name/index.cgi/etckeeper.git/tree/Makefile#n91), but it indeed works without both gsed and perl.

I've pushed the fix removing these two dependencies

PR #2071, opened 2024-05-27
contrib/libvips: Update source to prevent error 404.

claybie:


nekopsykose:
ah nice they fixed it

PR #2074, opened 2024-05-27
main/tzutils: install leap-seconds.list

triallax:


main/tzutils/template.py
```diff
@@ -1,18 +1,15 @@
 pkgname = "tzutils"
 pkgver = "2024a"
-pkgrel = 3
+pkgrel = 4
 build_style = "makefile"
 make_build_args = ["KSHELL=/bin/sh"]
 make_install_args = ["ZICDIR=/usr/bin", "ZFLAGS=-b fat"]
-hostmakedepends = []
```
q66, line 7: is wrong

triallax: ah right, missed that it's modified below

PR #2079, opened 2024-05-29
contrib/kea: new package (2.6.0)

rbgarga:


contrib/kea/template.py
```diff
@@ -0,0 +1,36 @@
...
+    "log4cplus",
+    "openssl-devel",
+    "python-devel",
+]
+pkgdesc = "Alternative DHCP implementation by ISC"
+maintainer = "Renato Botelho do Couto <renato@netgate.com>"
+license = "MPL-2.0"
+url = "https://kea.isc.org"
+source = (
+    f"https://downloads.isc.org/isc/{pkgname}/cur/2.4/{pkgname}-{pkgver}.tar.gz"
```
nekopsykose, line 21:
```suggestion
    f"https://downloads.isc.org/isc/kea/cur/{pkgver[:pkgver.rfind('.')]}/kea-{pkgver}.tar.gz"
```

contrib/kea/files/kea-dhcp-ddns
```diff
@@ -0,0 +1,7 @@
+# kea-dhcp-ddns service
+
+type       = process
+command    = /usr/sbin/kea-dhcp-ddns -c /etc/kea/kea-dhcp-ddns.conf
+depends-on = network.target
+waits-for  = kea-dhcp4
```
nekopsykose, line 6: waits-for means that:  

- if kea-dhcp4 fails to start, it won't block kea-dhcp-ddns (which will still start)
- stopping kea-dhcp4 will not stop dhcp-ddns (it would be allowed)

is that correct?

nekopsykose: if you want the second, but not the first, then depends-ms is the same but does care about failure

otherwise if you want neither you want just a regular depends-on

q66: yes, waits-for just means "wait for the named service to succeed or fail" but does not imply any dependency

q66: there is also depends-ms, which like waits-for does not imply a dependency, but requires the named service to succeed starting

nekopsykose:
put `ignore = ["*.*[13579].*]` in update.py for unstable versions

contrib/kea/files/kea-dhcp-ddns
rbgarga: I just want to decide the order services will start without creating any dependency so I'll replace it by `after` and I think it's enough

rbgarga:
> this currently skips all the tests which you might want to fix, but iirc that takes like over an hour to run

let me get it fixed, thanks!

contrib/kea/files/kea-dhcp4
```diff
@@ -0,0 +1,6 @@
+# kea-dhcp4 service
+
+type       = process
+command    = /usr/sbin/kea-dhcp4 -c /etc/kea/kea-dhcp4.conf
+depends-on = network.target
+after      = time-sync.target
```
nekopsykose, line 6:
```suggestion
type = process
command = /usr/sbin/kea-dhcp4 -c /etc/kea/kea-dhcp4.conf
depends-on = network.target
after = time-sync.target
```

don't need alignment anymore for all of these

rbgarga: good to know.  Thanks!

contrib/kea/update.py
```diff
@@ -0,0 +1 @@
+ignore = ["*.*[13579].*"]
```
nekopsykose, line 1: you also have to set url= so it picks up the new 2.6 version (since it doesn't currently)

so,

```
url = "https://downloads.isc.org/isc/kea"
pattern = r">([\d\.]+)/<"
ignore = ["*.*[13579].*", "201*"]
```

rbgarga: oh I missed 2.6.0 was released 2 days ago.  Good catch, thanks!  I will work on adding 2.6.0 instead

PR #2088, opened 2024-05-30
contrib/fastfetch: update to 2.14.0

JamiKettunen:
https://github.com/fastfetch-cli/fastfetch/blob/2.14.0/CHANGELOG.md

### DIFF
--- a/contrib/fastfetch/template.py
+++ b/contrib/fastfetch/template.py
@@ -1,5 +1,5 @@
 pkgname = "fastfetch"
-pkgver = "2.13.0"
+pkgver = "2.14.0"
 pkgrel = 0
 build_style = "cmake"
 configure_args = ["-DBUILD_TESTS=ON"]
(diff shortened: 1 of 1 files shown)


JamiKettunen:
Interesting there's now a `-DBINARY_LINK_TYPE=dynamic` configure options is available, I guess we should switch to it? Currently this uses whatever libs it finds at runtime to display information, this would make it depend on quite a bit but always unlike the soft-dep I was thinking around when this was initially merged
```
fastfetch-2.14.0-r0 depends on:
lscpu
so:libEGL.so.1
so:libMagickCore-7.Q16HDRI.so.10
so:libOSMesa.so.8
so:libOpenCL.so.1
so:libX11.so.6
so:libXrandr.so.2
so:libc.so
so:libchafa.so.0
so:libdbus-1.so.3
so:libdconf.so.1
so:libddcutil.so.5
so:libdrm.so.2
so:libgio-2.0.so.0
so:libglib-2.0.so.0
so:libgobject-2.0.so.0
so:libnm.so.0
so:libpulse.so.0
so:libvulkan.so.1
so:libwayland-client.so.0
so:libxcb-randr.so.0
so:libxcb.so.1
so:libxfconf-0.so.3
so:libz.so.1
```

nekopsykose:
by dynamic i mean the current dlopen not =dynamic

nekopsykose:
(in practice for what people care about they're gonna have installed already given how all the libs are pulled in..)

PR #2102, opened 2024-05-30
Protobuf 27.0

JamiKettunen:
https://github.com/protocolbuffers/protobuf/releases/tag/v27.0

`android-tools` seems to need `protobuf-devel-static` now due to:
```
CMake Error at /usr/lib/cmake/protobuf/protobuf-targets.cmake:135 (message):
  The imported target "protobuf::libupb" references the file

     "/usr/lib/libupb.a"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "/usr/lib/cmake/protobuf/protobuf-targets.cmake"

  but not all the files it references.

Call Stack (most recent call first):
  /usr/lib/cmake/protobuf/protobuf-config.cmake:13 (include)
  vendor/CMakeLists.txt:92 (find_package)
```
though I guess this may need other fixing since https://github.com/nmeum/android-tools/blob/35.0.1/vendor/CMakeLists.txt#L92 just calls `find_package(Protobuf CONFIG)`

see also: https://github.com/protocolbuffers/protobuf/blob/v27.0/cmake/libupb.cmake#L14-L15
> Note: upb does not support shared library builds, and is intended to be statically linked as a private dependency.


nekopsykose:
> see also: [protocolbuffers/protobuf@v27.0/cmake/libupb.cmake#L14-L15](https://github.com/protocolbuffers/protobuf/blob/v27.0/cmake/libupb.cmake?rgh-link-date=2024-05-30T23%3A26%3A57Z#L14-L15)  

it looks like utf8_range? see the patch

for utf8_range, it's "static" and installed, but in reality all this does is add it to the cmake files.. which nothing consumes since only protobuf itself links it. i.e. you could patch it the same way and it just wont install it but also nothing should break

the only case in which something might break from doing that is for statically linking protobuf if the libprotobuf*.a files don't fat link the libs into them (since then they're missing for actual static links later) but doesn't matter too much (could also test it)

also this needs an opencv rebuild

nekopsykose:
elected to just always pull static instead

PR #2103, opened 2024-05-30
main/polkit: update to 124

triallax:


q66:
please rebase

triallax:
rebased

triallax:
just realised that it seems like everything else installs pam rules to /etc/pam.d, should i do that here too?

q66:
i'm thinking maybe i should have all distro pam go in /usr instead, but still thinking about it

q66:
doesn't work (pkexec doesn't work, device automount doesn't work, etc)

PR #2113, opened 2024-06-01
user/blink: new package (1.1.0)

z-erica:
this cant actually run chimera-native executables, so mostly here as a curiosity


nekopsykose:
sure, throw it in user

PR #2117, opened 2024-06-01
contrib/upterm: new package (0.14.3)

ttyyls:


contrib/upterm/template.py
```diff
@@ -0,0 +1,47 @@
...
+    self.install_sysusers(self.files_path / "sysusers.conf")
+    self.install_tmpfiles(self.files_path / "tmpfiles.conf")
+    self.install_man("etc/man/man1/*.1", glob=True)
+    for shell in ["bash", "fish", "zsh"]:
+        self.install_completion(f"upterm.{shell}", shell)
+
+
+@subpackage("uptermd")
+def _daemon(self):
+    self.pkgdesc = f"{pkgdesc} daemon"
```
triallax, line 41:
```suggestion
    self.pkgdesc = f"{pkgdesc} (daemon)"
```

contrib/upterm/template.py
```diff
@@ -0,0 +1,33 @@
...
+make_install_args = make_build_args
+hostmakedepends = ["go"]
+pkgdesc = "Instant terminal sharing"
+maintainer = "ttyyls <contact@behri.org>"
+license = "Apache-2.0"
+url = "https://github.com/owenthereal/upterm"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "3ca2bb64943ba8ecd31ae83036794749bb32618a1929f4b22d535cfb28dda367"
+# cross: generates completions with host binary
+# check: uses -race that conflicts with -buildmode=pie
```
nekopsykose, line 15: then just don't pass -race

ttyyls: most of them fail without it

nekopsykose: i just ran them and they work fine, you just forgot bash in checkdepends

nekopsykose: and when they fail (by hanging) the failure has nothing to do with -race

PR #2119, opened 2024-06-02
contrib/pipx: fix source url and change hash (error 404)

claybie:


nekopsykose:
missing a `:` after `contrib/pipx`

also bump the pkgrel too

nekopsykose:
and squishy the commits into 1

PR #2126, opened 2024-06-03
git: add myself to .mailmap

triallax:


.mailmap
```diff
@@ -8,3 +8,4 @@ q66 <q66@chimera-linux.org> Daniel Kolesa <daniel@octaforge.org>
 q66 <q66@chimera-linux.org> Daniel Kolesa <q66@chimera-linux.org>
 q66 <q66@chimera-linux.org> Daniel Kolesa <dkolesa@igalia.com>
 q66 <q66@chimera-linux.org> q66 <daniel@octaforge.org>
+triallax <triallax@tutanota.com> Mohammed Anas <triallax@tutanota.com>
```
nekopsykose, line 11: i am 95% sure `triallax <triallax@tutanota.com> <triallax@tutanota.com>` works the same way as the name is not required

nekopsykose: you can find this out by searching for the name, then putting that mailmap rule and searching the log again

triallax: yeah that works, thanks

PR #2127, opened 2024-06-03
user/got: update to 0.99

ttyyls:


q66:
why was the symlink patch reverted, does not say anywhere and should

nekopsykose:
(could patch it again but also just keeping ldscript is zero effort since it's how upstream does it anyway)

PR #2128, opened 2024-06-03
contrib/kscreenlocker: install PAM config files

triallax:


contrib/kscreenlocker/files/kde-fingerprint.pam
```diff
@@ -0,0 +1,14 @@
+#%PAM-1.0
+
+auth       required                    pam_shells.so
+auth       requisite                   pam_nologin.so
+auth       requisite                   pam_faillock.so      preauth
+-auth      required                    pam_fprintd.so
+auth       optional                    pam_permit.so
+auth       required                    pam_env.so
+
```
nekopsykose, line 9: should probably stop aligning these because it looks atrocious

PR #2140, opened 2024-06-04
main/iproute2: remove header file

triallax:
we remove man3 for it anyway


nekopsykose:
the only reference i can find for anything using this is dpdk and they vendor their own copy, how funny https://github.com/DPDK/dpdk/blob/76cef1af8bdaeaf67a5c4ca5df3f221df994dc46/drivers/net/tap/bpf/bpf_elf.h#L7

triallax:
@nekopsykose lol

PR #2141, opened 2024-06-04
contrib/xfdashboard: use better url

triallax:
oopsie


nekopsykose:
this really xf'd my dashboard

triallax:
rip

PR #2145, opened 2024-06-04
contrib/snapshot: new package (46.3)

triallax:


contrib/snapshot/template.py
```diff
@@ -0,0 +1,43 @@
...
+    "gst-plugins-bad-devel",
+    "gst-plugins-base-devel",
+    "libadwaita-devel",
+]
+depends = [
+    "gstreamer-libcamera",
+    "gstreamer-pipewire",
+    "libcamera",
+    "pipewire",
+    "wireplumber",
```
triallax, line 25: could remove `libcamera` and `pipewire` here because they get pulled in by the other stuff

nekopsykose: yeah it's fine if they crom from gst

i think you don't need wireplumber either

triallax: i added wireplumber because it was mentioned as a dep but sure

nekopsykose: it's a dep of anything that plays audio too, yet it's not added there either

same goes for like.. a font, an icon theme, ... base deps are not useful to add

PR #2147, opened 2024-06-05
contrib/nginx: use system user in service

ttyyls:
fixes a typo where the system user would otherwise be installed as `_nignx`
i also used that opportunity to unalign the service and run it as said user.

### DIFF
--- a/contrib/nginx/files/nginx
+++ b/contrib/nginx/files/nginx
@@ -1,6 +1,6 @@
 # nginx service
 
-type            = process
-command         = /usr/bin/nginx -g "daemon off;"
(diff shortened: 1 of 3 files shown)


nekopsykose:
this doesn't work and breaks nginx, it has to drop privs itself (it does it for spawned workers)

nekopsykose:
the user _nginx; is what does it

rest looks good

q66:
oopsie

nekopsykose:
it's fixed now so it's fine

nekopsykose:
though the commit message now doesn't make sense

PR #2153, opened 2024-06-06
contrib/firefox: backport fix for segfaults on loading XML

### LINKED ISSUE #2133: contrib/firefox: repeatable tab crash
I'm reporting this here as I can visit this page without crashing on an Arch system. 

Visting https://thisweekinbevy.com/issue/2024-06-03-tiny-glade-vj-performances-and-2d-lighting on an up-to-date (as of this morning) x86_64 Chimera system reliably crashes the Firefox tab. I tried to collect a stack trace but did not succeed.

The kernel logs say:

> 2024-06-04 09:03:55 info kernel: traps: Isolated Web Co[7750] general protection fault ip:7cb01c159bd4 sp:7ffef807df60 error:0 in libxul.so[7cb016ae0000+5825000]

Firefox says (this was a different instance hence the different PIDs):

> [Parent 9019, IPC I/O Parent] WARNING: process 9538 exited on signal 11: file ipc/chromium/src/base/process_util_posix.cc:335

This is what I tried:

- Install `firefox-dbg`
- Start Firefox
- `lldb --attach-pid <pid>` for each process listed by `pgrep Iso`

It seems this isn't enough as it spaws a new process for the tab and that's the one that crashes.

I also tried to turn on core dumps but I'm not sure if this is the right way, since I never saw a core file:

`ulimit -c unlimited`

Any guidance on how to further debug this would be most welcome.

ptrcnull:
https://phabricator.services.mozilla.com/D210814

closes #2133

### DIFF
--- a/contrib/firefox/patches/backport-D210814.patch
+++ b/contrib/firefox/patches/backport-D210814.patch
@@ -1,0 +1,1864 @@
+From 6dd67745b8036c2f0ebec0daf8eaf8e9e7b2a7ee Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Emilio=20Cobos=20=C3=81lvarez?= <emilio@crisal.io>
+Date: Sun, 19 May 2024 21:24:16 +0000
+Subject: [PATCH] Bug 1897361 - Share counter-style / list-style-type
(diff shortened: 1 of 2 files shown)


wezm:
Thank you!

PR #2161, opened 2024-06-07
linux-rpi for armhf

z-erica:


q66:
i reviewed the config a bit, it seems there is still a big amount of difference from aarch64 in generic things (please try a bit to sync it, i'm not asking for it to be 100% but at least in various generic kernel behaviors and features the configs should be equivalent)

PR #2166, opened 2024-06-07
contrib/nfs-utils: new package (2.6.4)

mikoxyz:


yanchan09:
https://gist.github.com/yanchan09/21a538cd313e7a444aaaa64211feb886 with heimdal-devel and heimdal-devel-static allows building rpc.gssd/rpc.svcgssd

yanchan09:
tested myself, both NFSv4 client & sever work, both without and with sec=krb5p

PR #2173, opened 2024-06-08
contrib/dnsmasq: install config file

Natthias:
noticed this file was missing when setting up libvirt

### DIFF
--- a/contrib/dnsmasq/patches/conf.patch
+++ b/contrib/dnsmasq/patches/conf.patch
@@ -1,0 +1,37 @@
+diff --git a/dnsmasq.conf.example b/dnsmasq.conf.example
+index f1fb2d1..ebca0e9 100644
+--- a/dnsmasq.conf.example
++++ b/dnsmasq.conf.example
+@@ -22,7 +22,7 @@
+ 
+ # Uncomment these to enable DNSSEC validation and caching:
+ # (Requires dnsmasq to be built with DNSSEC option.)
+-#conf-file=%%PREFIX%%/share/dnsmasq/trust-anchors.conf
++#conf-file=/usr/share/dnsmasq/trust-anchors.conf
+ #dnssec
+ 
+ # Replies which are not DNSSEC signed may be legitimate, because the domain
+@@ -106,8 +106,12 @@
+ 
+ # If you want dnsmasq to change uid and gid to something other
+ # than the default, edit the following lines.
+-#user=
+-#group=
++#user=_dnsmasq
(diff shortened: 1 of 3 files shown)


nekopsykose:
you should patch the file to change the defaults too (fill out default `#user=_dnsmasq`, uncomment conf-dir /etc/dnsmasq.d, add a keep-empty dir for dnsmasq.d, make it local-service by default, etc

Natthias:
yea, i forgot to add the patch, i'll also be adding files for dnssec in a bit

contrib/dnsmasq/patches/conf.patch
```diff
@@ -0,0 +1,33 @@
...
+ 
+ # Replies which are not DNSSEC signed may be legitimate, because the domain
+@@ -106,8 +106,8 @@
+ 
+ # If you want dnsmasq to change uid and gid to something other
+ # than the default, edit the following lines.
+-#user=
+-#group=
++user=_dnsmasq
++group=_dnsmasq
```
nekopsykose, line 21: does it work by default by using these (since i meant to leave it commented?)

if it binds e.g. :25 before dropping privs it should be fine, in which case keep it- do check though

contrib/dnsmasq/patches/conf.patch
```diff
@@ -0,0 +1,33 @@
...
+ 
+ # If you want dnsmasq to listen for DHCP and DNS requests only on
+ # specified interfaces (and the loopback) give the name of the
+@@ -675,7 +675,7 @@
+ 
+ # Include another lot of configuration options.
+ #conf-file=/etc/dnsmasq.more.conf
+-#conf-dir=/etc/dnsmasq.d
++conf-dir=/etc/dnsmasq.d
+
```
nekopsykose, line 31: also add `local-service` so by default it's localhost only (adding any interface= line or whatever later in .d by a user undoes it)

contrib/dnsmasq/patches/conf.patch
Natthias: seems to work fine, I could also just patch the config.h with these groups like alpine?

nekopsykose: good idea to do that too yeah

nekopsykose: if it works fine you should leave it uncommented

nekopsykose: actually yea it uses the patch default so nvm

PR #2174, opened 2024-06-09
[DRAFT] contrib/nbfc-linux: update to 0.2.7

Sid127:


Sid127:
I do not have a chimera build env at the moment, and the makefile patch needs an update. Will look into it in the next few days

nekopsykose:
superseded by the other mr

PR #2178, opened 2024-06-09
contrib/python-coverage: remove package

triallax:
not used by anything, if it is then that use should most likely be
patched out anyway


nekopsykose:
i don't see a reason to remove it even if it isn't used

triallax:
> i don't see a reason to remove it even if it isn't used

meh

PR #2182, opened 2024-06-09
main/heimdal-kdc: add service files

yanchan09:


main/heimdal/files/heimdal-kadmind
```diff
@@ -0,0 +1,6 @@
+# heimdal-kadmind service
+
+type = process
+command = /usr/libexec/kadmind
+depends-on = network.target
```
nekopsykose, line 5: +local, network is supplementary

yanchan09: what

nekopsykose:
```suggestion
depends-on = network.target
depends-on = local.target
```

PR #2186, opened 2024-06-09
main/libcamera: fix segfault in option handling

triallax:


nekopsykose:
you should upstream this with the correct fix, see how smartmontools fixed it upstream

triallax:
yeah, maybe whenever i feel like sending a patch to a mailing list

PR #2196, opened 2024-06-10
contrib/mercurial: new package (6.8)

ttyyls:


contrib/mercurial/template.py
```diff
@@ -0,0 +1,26 @@
...
+    "python-wheel",
+]
+makedepends = [
+    "gettext-devel",
+    "musl-devel",
+    "python-devel",
+    "zlib-devel",
+    "zstd-devel",
+]
+checkdepends = ["python-pytest"]
```
q66, line 18: runtime dependencies are missing

contrib/mercurial/template.py
```diff
@@ -0,0 +1,26 @@
...
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+    "python-wheel",
+]
+makedepends = [
+    "gettext-devel",
```
q66, line 12: pretty sure you don't need gettext-devel, and you never need musl-devel, and i'm not sure about the others considering according to the ci output they don't even get linked

contrib/mercurial/template.py
```diff
@@ -0,0 +1,26 @@
...
+    "zstd-devel",
+]
+checkdepends = ["python-pytest"]
+pkgdesc = "Distributed source control management"
+maintainer = "ttyyls <contact@behri.org>"
+license = "GPL-2.0-or-later"
+url = "https://www.mercurial-scm.org"
+source = f"https://www.mercurial-scm.org/release/mercurial-{pkgver}.tar.gz"
+sha256 = "00196944ea92738809317dc7a8ed7cb21287ca0a00a85246e66170955dcd9031"
+# vendored dependencies tests are broken on python 3.11 and up
```
q66, line 25: could probably still be made to work with enough checkdepends

ttyyls: no they use python's imp which was part of the stdlib until 3.11 where it was deprecated. the vendored version of zstd here and upstream both rely on it for all their tests. skipping zstd tests from mercurial causes everything else to be skipped so i thought to just !check it

contrib/mercurial/template.py
ttyyls: i was building the rust/C client components but i removed them before submitting the port, forgot to cleanup the deps too

contrib/mercurial/template.py
ttyyls: which?

q66: like... all of them

look at the package, it has none

```
=> mercurial-6.7.4-r0: running pre_pkg hook: 001_runtime_deps...
   SONAME: libc.so (provider: musl)
```

first of all, python

second of all, you add these makedepends like zlib/zstd and whatever but then the package contains no reference to them whatsoever

ttyyls: i removed zlib/zstd and added python to depends

q66:
install manpages

contrib/mercurial/template.py
```diff
@@ -0,0 +1,40 @@
...
+def post_install(self):
+    self.do(
+        "gmake",
+        "install",
+        f"DESTDIR={self.chroot_destdir}",
+        "PREFIX=/usr",
+        wrksrc="doc",
+    )
+    self.install_completion("contrib/bash_completion", "bash")
+    self.install_completion("contrib/zsh_completion", "zsh")
```
triallax, line 40:
```suggestion
    self.install_completion("contrib/bash_completion", "bash", "hg")
    self.install_completion("contrib/zsh_completion", "zsh", "hg")
```

triallax (changes requested):

contrib/mercurial/template.py
nekopsykose: that means you didn't run them correctly because that 'imp' error is inside the vendored zstd tests.. which are not related to the actual mercurial tests which can only be ran with ./tests/run-tests.py and fail for completely unrelated reasons

PR #2200, opened 2024-06-11
contrib/dopewars: new package (1.6.2)

mikoxyz:


nekopsykose:
what

contrib/dopewars-console/files/dopewars
```diff
@@ -0,0 +1,8 @@
+# dopewars service
+
+type        = bgprocess
+command     = /usr/bin/dopewars -S -r /run/dopewars/dopewars.pid
+run-as      = _dopewars
+pid-file    = /run/dopewars/dopewars.pid
+depends-on  = local.target
+after       = network.target
```
q66, line 8: targets are always depends-on

contrib/dopewars-console/files/dopewars
```diff
@@ -0,0 +1,8 @@
+# dopewars service
+
+type        = bgprocess
```
q66, line 3: figure out how to foreground it, we don't take bgprocess services in general

contrib/dopewars-console/template.py
```diff
@@ -0,0 +1,47 @@
+pkgname = "dopewars-console"
```
q66, line 1: what is the purpose of a separate console package

PR #2209, opened 2024-06-12
contrib/miniupnpd: new package (2.3.6)

rbgarga:


contrib/miniupnpd/template.py
```diff
@@ -0,0 +1,38 @@
+pkgname = "miniupnpd"
+pkgver = "2.3.6"
+pkgrel = 0
+build_wrksrc = pkgname
```
nekopsykose, line 4:
```suggestion
build_wrksrc = "miniupnpd"
```

contrib/miniupnpd/template.py
```diff
@@ -0,0 +1,38 @@
...
+build_style = "configure"
+configure_args = [
+    "--firewall=nftables",
+    "--ipv6",
+    "--portinuse",
+]
+make_cmd = "gmake"
+hostmakedepends = [
+    "git",
```
nekopsykose, line 13: as far as i can tell git is unused here

contrib/miniupnpd/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "libuuid-devel",
+    "linux-headers",
+]
+depends = ["nftables"]
+checkdepends = ["iproute2"]
+pkgdesc = "UPnP and NAT-PMP daemon for gateway routers"
+maintainer = "Renato Botelho do Couto <renato@netgate.com>"
+license = "BSD-3-Clause"
+url = "https://github.com/miniupnp/miniupnp"
+source = f"https://github.com/miniupnp/miniupnp/archive/refs/tags/{pkgname}_{pkgver.replace('.', '_')}.tar.gz"
```
nekopsykose, line 32:
```suggestion
source = f"{url}/archive/refs/tags/{pkgname}_{pkgver.replace('.', '_')}.tar.gz"
```

contrib/miniupnpd/update.py
```diff
@@ -0,0 +1 @@
+pattern = r"/refs/tags/miniupnpd_(\d+_\d+_\d+)\.tar\.gz"
```
nekopsykose, line 1: also override fetch_versions to map the _ back to . (see e.g. opensubdiv update.py)

contrib/miniupnpd/template.py
rbgarga: Yeah, I've added it while doing development and forgot to remove

nekopsykose: same for procps i think

rbgarga: It uses `sysctl` binary from procps

nekopsykose: in a bad way- it uses the build server configuration to configure itself for --v6sockets-v6only

contrib/miniupnpd/files/miniupnpd
```diff
@@ -0,0 +1,6 @@
+# miniupnpd service
+
+type = process
+command = /usr/bin/miniupnpd -f /etc/miniupnpd/miniupnpd.conf -d
```
nekopsykose, line 4: you can also pass --disable-fork to configure

aside from that, i think some other configure options might be useful?

contrib/miniupnpd/template.py
rbgarga: Do you think it would be better to just patch to remove that part of the code?

contrib/miniupnpd/files/miniupnpd
rbgarga: good catch, I missed `--disable-fork`.  Any particular configure option you believe we should enable?

contrib/miniupnpd/template.py
nekopsykose: i'm not sure, i think the daemon itself is hardcoded to only work with one configuration so the other would be broken at runtime

but most people are probably not running v6only so it's better to patch it out to guarantee that it doesn't enable that i guess

contrib/miniupnpd/files/miniupnpd
nekopsykose: --vendorcfg allows editing mfg info in the config file manually later

maybe regex/pcp-peer/igd2? but i'm not sure how useful they are

rbgarga: I never tried them as well, let me take a look

rbgarga: I've checked alpine, arch and ubuntu and after that decided to enable vendorcfg, igd2, pvp-peer and leasefile.  I didn't see regex being used by anyone

contrib/miniupnpd/files/miniupnpd
```diff
@@ -0,0 +1,6 @@
+# miniupnpd service
+
+type = process
+command = /usr/bin/miniupnpd -f /etc/miniupnpd/miniupnpd.conf
+depends-on = pre-network.target
```
q66, line 5: auxilitary target; you need thís and probably local.target dependencies

contrib/miniupnpd/files/miniupnpd
```diff
@@ -0,0 +1,6 @@
...
+log-type = buffer
```
q66, line 6: how much does it log? because buffer logtype only captures enough to fill the buffer and does not scroll

contrib/miniupnpd/files/miniupnpd
rbgarga: got it

contrib/miniupnpd/files/miniupnpd
nekopsykose: well you'd probably want logfile= then

also i think you forgot to commit the changes for these

rbgarga: buffer will be enough for it to throw possible error messages

rbgarga:
> @rbgarga: For the sake of completeness, since PCP is supported, the `pkgdesc` in:
> 
> https://github.com/chimera-linux/cports/blob/fae74602602e60fd103f7eb91f7ad04d20bc16b3/contrib/miniupnpd/template.py#L31
> 
> could perhaps be changed to:
> 
> ```
> pkgdesc = "UPnP IGD & PCP/NAT-PMP daemon for gateway router" 
> ```
> 
> If you do not want this cosmetic change, simply ignore it.

PR submitted https://github.com/chimera-linux/cports/pull/2279

Thanks!

PR #2210, opened 2024-06-12
contrib/binutils: fix riscv64 asm edge case

triallax:


contrib/binutils/patches/int.patch
```diff
@@ -0,0 +1,23 @@
...
++  upper.X_add_number = (bfd_vma)upper.X_add_number - lower.X_add_number;
+ 
+   if (ep->X_op != O_constant)
+     {
+@@ -2009,7 +2009,7 @@ load_const (int reg, expressionS *ep)
+   if (xlen > 32 && !IS_SEXT_32BIT_NUM (ep->X_add_number))
+     {
+       /* Reduce to a signed 32-bit constant using SLLI and ADDI.  */
+-      while (((upper.X_add_number >> shift) & 1) == 0)
++      while ((((bfd_vma)upper.X_add_number >> shift) & 1) == 0)
```
triallax, line 20: without this, the loop runs indefinitely until shift overflows

is this actually due to `int`? i'm not sure about that

nekopsykose: more likely clang optimises the ub so it does that

you should send it upstream

triallax: ok, i'll send the changes upstream soon

PR #2225, opened 2024-06-13
main/curl: enable tests

mikoxyz:


q66:
there is like no way we are moving all this stuff

nekopsykose:
vast majority of curl tests don't need any of this whatsoever and already run, why

nekopsykose:
anyway no

PR #2230, opened 2024-06-13
*/*: !installroot the kernels

nekopsykose:
should save some time on the really long install step without fakeroot overhead that is unneeded  

[ci skip]

didnt test this yet


nekopsykose:
seems to not affect the do_install time at all so i guess it's not that relevant :p

PR #2231, opened 2024-06-14
contrib/openldap: new package (2.6.8)

rbgarga:
Ressurrect openldap, previously available at main/openldap, but this
time add only client binaries and libraries.  There are better options
in the field for people that needs an LDAP server.

### DIFF
--- a/contrib/openldap/template.py
+++ b/contrib/openldap/template.py
@@ -1,0 +1,59 @@
+pkgname = "openldap"
+pkgver = "2.6.8"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--disable-debug",
+    "--disable-slapd",
+    "--disable-sql",
+    "--disable-static",
+    "--disable-versioning",
+    "--disable-wt",
+    "--enable-backends=mod",
+    "--enable-crypt",
+    "--enable-dynamic",
+    "--enable-modules",
+    "--enable-overlays=mod",
+    "--enable-rlookups",
+    "--enable-spasswd",
+    "--enable-syslog",
+    "--with-cyrus-sasl",
+    "--with-tls=openssl",
+]
+make_cmd = "gmake"
+hostmakedepends = [
+    "automake",
+    "gmake",
+    "libtool",
+    "mandoc",
+    "pkgconf",
+]
+makedepends = [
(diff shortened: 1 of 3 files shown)


contrib/openldap/template.py
```diff
@@ -0,0 +1,70 @@
...
+pkgdesc = "FOSS implementation of the Lightweight Directory Access Protocol"
+maintainer = "Renato Botelho do Couto <renato@netgate.com>"
+license = "OLDAP-2.8"
+url = "https://www.openldap.org"
+source = (
+    f"{url}/software/download/OpenLDAP/{pkgname}-release/{pkgname}-{pkgver}.tgz"
+)
+sha256 = "48969323e94e3be3b03c6a132942dcba7ef8d545f2ad35401709019f696c3c4e"
+# test suite takes ages
+options = ["!check"]
```
nekopsykose, line 48: afaik this is mostly the server part, but maybe there's library-only tests that are more reasonable?

contrib/openldap/template.py
```diff
@@ -0,0 +1,70 @@
...
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage("openldap-progs")
+def _progs(self):
+    return [
+        "usr/bin/ldap*",
+        "usr/share/man/man1/ldap*.1",
+    ]
```
nekopsykose, line 70: what other bins/manpages are there that default_progs() would be bad for?

contrib/openldap/template.py
```diff
@@ -0,0 +1,70 @@
...
+    "libsodium-devel",
+    "openssl-devel",
+]
+depends = [f"openldap-progs={pkgver}-r{pkgrel}"]
+pkgdesc = "FOSS implementation of the Lightweight Directory Access Protocol"
+maintainer = "Renato Botelho do Couto <renato@netgate.com>"
+license = "OLDAP-2.8"
+url = "https://www.openldap.org"
+source = (
+    f"{url}/software/download/OpenLDAP/{pkgname}-release/{pkgname}-{pkgver}.tgz"
```
nekopsykose, line 44:
```suggestion
    f"{url}/software/download/OpenLDAP/openldap-release/openldap-{pkgver}.tgz"
```

PR #2232, opened 2024-06-14
user/vim: upgrade to 9.1.0485, fixed some tests

subnut:


nekopsykose:
!check is still set so this isn't running anything

you should also upstream these because otherwise i don't see the point

nekopsykose:
or well, what i mean to say is maintaining a bunch of patches to fix a lot of tests isn't that great

PR #2234, opened 2024-06-14
main/gnome-console: change default icon

z-erica:
`>_<`


q66:
ty, should be better this way though :)

PR #2247, opened 2024-06-15
contrib/plasma-workspace: use our wallpapers as default, improve license

triallax:


contrib/plasma-workspace/files/metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+    "KPlugin": {
+        "Authors": [
+            {
+                "Email": "triallax@tutanota.com",
+                "Name": "triallax"
```
triallax, line 6: i should probably put the author of the wallpapers here instead?

contrib/plasma-workspace/template.py
```diff
@@ -132,6 +136,41 @@ def post_install(self):
...
+    self.install_link(
+        "usr/share/wallpapers/Chimera/contents/images/2560x1600.svg",
+        "/usr/share/backgrounds/chimera/bg-l.svg",
+        absolute=True,
+    )
+    self.install_link(
+        "usr/share/wallpapers/Chimera/contents/images_dark/2560x1600.svg",
+        "/usr/share/backgrounds/chimera/bg-d.svg",
+        absolute=True,
+    )
```
triallax, line 158: using symlinks here feels like a hack (also because of the `broken_symlinks` thing above, but idk

also should this be moved into a subpackage? or maybe even into chimera-artwork itself? it's just metadata.json and a couple symlinks so i think putting it into chimera-artwork is reasonable

triallax:
i went with this method of modifying the breeze themes, the other way that upstream recommends is creating new global themes, copied from the breeze ones, and changing their default
 wallpapers, as mentioned in https://bugs.kde.org/show_bug.cgi?id=487816, if you want me to do that instead then that's fine with me

triallax:
also not directly related to this pr's changes, but we could probably move the breeze themes into a dedicated subpackage and use install_if to make that optional

contrib/plasma-workspace/template.py
nekopsykose: doesn't have to be absolute i think

contrib/plasma-workspace/files/metadata.json
q66: put generic info there, i.e. `info@chimera-linux.org` and `Chimera Linux`

contrib/plasma-workspace/template.py
```diff
@@ -132,6 +136,41 @@ def post_install(self):
 
     self.rm(self.destdir / "usr/lib/systemd/user", recursive=True)
 
+    self.mkdir(
```
q66, line 139: use install_dir and the likes

contrib/plasma-workspace/template.py
```diff
@@ -132,6 +136,41 @@ def post_install(self):
...
+        self.destdir / "usr/share/wallpapers/Chimera/contents/images",
+        parents=True,
+    )
+    self.mkdir(
+        self.destdir / "usr/share/wallpapers/Chimera/contents/images_dark"
+    )
+    self.install_file(
+        self.files_path / "metadata.json", "usr/share/wallpapers/Chimera"
+    )
+    self.install_link(
```
q66, line 149: what's the deal with the 2560x1600 number

triallax: when i checked breeze's wallpapers, i noticed that for many of them (e.g. `/usr/share/wallpapers/Path/contents/images`), the various wallpapers sizes were all symlinks to `2560x1600.jpg` (and in only one case, `3200x2000.jpg`), so that's why I chose `2560x1600` as the size

PR #2248, opened 2024-06-16
user/hare{c,}: new package (0.24.0)

triallax:


user/hare/files/config.mk
```diff
@@ -0,0 +1,43 @@
+# install locations
+PREFIX = /usr/local
+BINDIR = $(PREFIX)/bin
+MANDIR = $(PREFIX)/share/man
+SRCDIR = $(PREFIX)/src
+STDLIB = $(SRCDIR)/hare/stdlib
```
triallax, line 6: i could get rid of the variables here that get overriden by cbuild anyway, if you want (or maybe just comment them out)

user/hare/template.py
```diff
@@ -0,0 +1,42 @@
...
+        _qbe_arch = "rv64"
+    case _:
+        _qbe_arch = ""
+make_build_args = [
+    f"ARCH={self.profile().arch}",
+    f"QBEFLAGS=-t{_qbe_arch}",
+    f"{self.profile().arch.upper()}_CC=cc",
+    f"{self.profile().arch.upper()}_LD=ld",
+]
+# TODO: make it use clang assembler (at build time and also at runtime)
```
triallax, line 22: @nekopsykose i vaguely remember you said you had done this already? not sure

user/harec/patches/link-cflags.patch
```diff
@@ -0,0 +1,26 @@
+diff --git a/Makefile b/Makefile
+index a7b4ed3..fe27227 100644
+--- a/Makefile
++++ b/Makefile
+@@ -36,7 +36,7 @@ harec_objects = \
+ $(BINOUT)/harec: $(harec_objects)
+ 	@mkdir -p -- $(BINOUT)
+ 	@printf 'CCLD\t%s\n' '$@'
+-	@$(CC) $(LDFLAGS) -o $@ $(harec_objects) $(LIBS)
++	@$(CC) $(CFLAGS) $(LDFLAGS) -o $@ $(harec_objects) $(LIBS)
```
triallax, line 10: we have a qbe patch doing the same thing so i did it here as well, but not sure if it has any effect on the build

user/hare/template.py
```diff
@@ -0,0 +1,42 @@
...
+hostmakedepends = [f"binutils-{self.profile().arch}", "harec", "qbe", "scdoc"]
+depends = ["binutils", "clang", "harec", "qbe", "tzdata"]
+checkdepends = ["tzdata"]
+pkgdesc = "Hare programming language"
+maintainer = "triallax <triallax@tutanota.com>"
+license = "MPL-2.0 AND GPL-3.0-only"
+url = "https://harelang.org"
+source = f"https://git.sr.ht/~sircmpwn/hare/archive/{pkgver}.tar.gz"
+sha256 = "7061dad3c79cca51a1662a71b1c6f8ec001f52ef3053dd3c2dbb95ae9beff7bc"
+tools = {"AS": f"{self.profile().triplet}-as"}
```
triallax, line 32: maybe we can also have `hare-{arch}` subpackages for cross-building like we have for binutils?

user/harec/template.py
```diff
@@ -0,0 +1,24 @@
...
+sha256 = "add6a7c4cbfd130c5e9fcecd2d43bec39640ed9f9cfbe9166e4b7e945a46b7de"
+tool_flags = {
+    # Taken from configs/linux.mk
+    "CFLAGS": ["-std=c11", "-D_XOPEN_SOURCE=700", "-Iinclude"],
+}
+hardening = ["vis", "cfi"]
+
+
+def pre_build(self):
+    self.cp("configs/linux.mk", "config.mk")
```
triallax, line 24: could also copy the config file over into the tree like i do for `hare`, but this also works just fine

user/hare/template.py
nekopsykose: it doesn't work very well (forgot why), upstream requires gas basically so no need for the todo

user/harec/patches/link-cflags.patch
nekopsykose: i think this was actually a bug workaround back in like llvm 14-16 and is not actually required anymore and yields an identical (i.e. bit for bit) result as the individual .o's have their flags stored to use for lto and whatever

user/hare/template.py
nekopsykose: sure, see the alpine install of hare that adds config files for other arches

nekopsykose: ah you added that already

hm, maybe then..

user/hare/patches/linker-script-execstack.patch
```diff
@@ -0,0 +1,13 @@
+otherwise cbuild flags hare and haredoc binaries as using execstack
```
nekopsykose, line 1: should also upstream this i think, think nobody has just noticed this yet?

user/harec/patches/install-target.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/Makefile b/Makefile
+index a7b4ed3..00c8728 100644
+--- a/Makefile
++++ b/Makefile
+@@ -67,6 +67,7 @@ check: $(BINOUT)/harec $(tests)
+ 	@./tests/run
+ 
+ install: $(BINOUT)/harec
++	install -d $(DESTDIR)$(BINDIR)
+ 	install -Dm755 $(BINOUT)/harec $(DESTDIR)$(BINDIR)/harec
```
nekopsykose, line 10: think upstream in this case would take a fix for this (regular install -d + install -m after etc)

triallax: already sent

user/hare/template.py
triallax: i mean sure, but still sucks to depend on it :/

i'll get rid of the todo for now but upstream told me they're open to making it work with clang so i'll look into it it at some point

user/hare/template.py
triallax: maybe i'm overthinking this, but installing e.g. `hare-riscv64` will also pull in `clang` and `binutils` for the host arch, we might want to make it possible to install the deps for just a given arch without also installing the ones for the host arch

triallax: ok i ended up having to add `hare-{arch}` anyway, to make the list of subpackages consistent across architectures

triallax:
why is this still a draft

user/hare/files/config.mk
```diff
@@ -0,0 +1,43 @@
...
+
+# variables used during build
+PLATFORM = linux
+ARCH = $(CBUILD_TARGET_MACHINE)
+HAREFLAGS =
+HARECFLAGS = -a$(ARCH)
+QBEFLAGS =
+ASFLAGS =
+LDLINKFLAGS = --gc-sections -z noexecstack
```
nekopsykose, line 15: since it uses ld.lld by default you should put -z pack-relative-relocs here too

user/hare/template.py
```diff
@@ -0,0 +1,66 @@
...
+        _qbe_arch = "amd64_sysv"
+    case "aarch64":
+        _qbe_arch = "arm64"
+    case "riscv64":
+        _qbe_arch = "rv64"
+    case _:
+        _qbe_arch = ""
+make_build_args = [
+    f"ARCH={self.profile().arch}",
+    f"QBEFLAGS=-t{_qbe_arch}",
```
nekopsykose, line 18: put this match below the main block (below tools=)

user/hare/patches/linker-script-execstack.patch
triallax: will do yeah

triallax:
the `hare` and `haredoc` binaries are still in the `hare` metapackage which as of now requires `hare-<native arch>`, do you think it might be a good idea to move them (and the standard library etc.) to a separate package?

user/hare/files/config.mk
nekopsykose: seems fine as is

user/harec/template.py
nekopsykose: fine as is

nekopsykose:
>the hare and haredoc binaries are still in the hare metapackage which as of now requires hare-<native arch>, do you think it might be a good idea to move them (and the standard library etc.) to a separate subpackage?

hm, not sure really. can always change it later

PR #2260, opened 2024-06-17
neovim: add missing dependency

subnut:
v0.10.0 added requirement of markdown treesitter parser.

### DIFF
--- a/contrib/neovim/template.py
+++ b/contrib/neovim/template.py
@@ -1,7 +1,7 @@
 # nb: neovim requires either lua5.1 or luaJIT (a mess)
 pkgname = "neovim"
 pkgver = "0.10.0"
-pkgrel = 1
+pkgrel = 2
 build_style = "cmake"
 configure_args = [
     "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
@@ -27,7 +27,11 @@ makedepends = [
 ]
 depends = [
     "lua5.1-lpeg",
+    "tree-sitter-bash",
+    "tree-sitter-c",
     "tree-sitter-lua",
+    "tree-sitter-markdown",
+    "tree-sitter-python",
     "tree-sitter-query",
     "tree-sitter-vimdoc",
 ]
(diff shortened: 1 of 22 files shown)


subnut:
The tests for inline-markdown's extensions seems to be failing even though I have enabled all extensions via the patch...

@nekopsykose any idea what the problem might be?

nekopsykose:
you cant update query because neovim depends on a specific version

contrib/tree-sitter-markdown/patches/extensions.patch
nekopsykose, line 1: you can just these env vars instead of patching it at all

subnut: I tried, it didn't work

(although this doesn't work either, lol)

contrib/neovim/template.py
```diff
@@ -28,7 +28,8 @@
 depends = [
     "lua5.1-lpeg",
     "tree-sitter-lua",
-    "tree-sitter-query",
+    "tree-sitter-markdown",
+    "tree-sitter-query<0.4.0",
```
nekopsykose, line 32:
```suggestion
    "tree-sitter-query",
```

don't need explicit versions

contrib/tree-sitter-markdown/template.py
```diff
@@ -0,0 +1,57 @@
...
+pkgver = "0.2.3"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_check_target = "test"
+hostmakedepends = [
+    "gmake",
+    "nodejs",
+    "tree-sitter-cli",
+    "pkgconf",
```
nekopsykose, line 11:
```suggestion
    "pkgconf",
    "tree-sitter-cli",
```

contrib/neovim/template.py
subnut: right. forgot to remove this after adding `ignore = True`.

removed :+1:

PR #2261, opened 2024-06-17
contrib/php8.3: add ldap extension

rbgarga:


q66:
you don't need the openldap commit

nekopsykose:
you should add libsasl-devel to the depends of openldap-devel and remove it from php, it's just missing from that devel package (in .pc it has -lsasl2)

PR #2267, opened 2024-06-18
 contrib/python-numpy: update to 2.0.0 

z-erica:


q66:
does this actually work? numpy 2 brings like a ton of api breakage

nekopsykose:
figured out how to run the numpy tests in parallel. it also makes them actually fail on failure :)

blender still works, opencv supports it, inkscape optimised svg export works too. i guess we have nothing really depending on this yet so it's alright

PR #2272, opened 2024-06-18
Linux 6.9

q66:
For now WIP until tested at least on my laptop and my altra.


q66:
x86_64 is ok, aarch64 pending

q66:
does not boot on the aarch64, investigation pending (need to get a cable :/)

q66:
15cff244e5be59574a0b24f7585db0f20626acdd

PR #2273, opened 2024-06-19
contrib/fish-shell: don't install config to /usr/etc

z-erica:


q66:
this was addressed differently in cdccd7addb95178fe74ed639a49cb2d7d73bed0c

PR #2285, opened 2024-06-20
main/libcanberra: move a bunch of files to the correct subpackages

triallax:


q66:
hm i don't think this is the right fix

triallax:
hm, why not?

also this isn't really a fix for any issue

triallax:
i only found out about it because of this line in my `/var/log/messages` btw:
```
2024-06-20 22:02:04 warning gnome-session-binary[11624]: WARNING: Desktop file /usr/share/gnome/autostart/libcanberra-login-sound.desktop for application libcanberra-login-sound.desktop could not be parsed or references a missing TryExec binary
```

PR #2286, opened 2024-06-20
main/python-psutil: update to 6.0.0

mikoxyz:


nekopsykose:
does everything that use this still work with it

nekopsykose:
gotta find out then

nekopsykose:
orca uses it to list cpu/memory/battery status somewhere and that's about it

nekopsykose:
checked and none of those use the changed api stuff so is k

PR #2306, opened 2024-06-23
contrib/git-lfs: new package (3.5.1)

subnut:


triallax:
commit message should be 'contrib/git-lfs: new package (3.5.1)`

contrib/git-lfs/template.py
```diff
@@ -0,0 +1,36 @@
+pkgname = "git-lfs"
+pkgver = "3.5.1"
+pkgrel = 0
+build_style = "go"
+make_cmd = "gmake"
+make_build_args = [
+    "-ldflags",
+    "-X github.com/git-lfs/git-lfs/v3/config.Vendor=ChimeraLinux",
```
triallax, line 8:
```suggestion
    "-ldflags=-X github.com/git-lfs/git-lfs/v3/config.Vendor=ChimeraLinux",
```

contrib/git-lfs/template.py
```diff
@@ -0,0 +1,36 @@
...
+pkgdesc = "Git extension for versioning large files"
+maintainer = "Subhaditya Nath <sn03.general@gmail.com>"
+license = "MIT"
+url = "https://git-lfs.com"
+source = f"https://github.com/git-lfs/git-lfs/releases/download/v{pkgver}/git-lfs-v{pkgver}.tar.gz"
+sha256 = "fc19c7316e80a6ef674aa4e1863561c1263cd4ce0588b9989e4be9461664d752"
+
+
+def post_build(self):
+    self.log("Building man pages")
```
triallax, line 22: don't need this
```suggestion
    self.log("Building man pages")
```

contrib/git-lfs/template.py
```diff
@@ -0,0 +1,36 @@
...
+pkgver = "3.5.1"
+pkgrel = 0
+build_style = "go"
+make_cmd = "gmake"
+make_build_args = [
+    "-ldflags",
+    "-X github.com/git-lfs/git-lfs/v3/config.Vendor=ChimeraLinux",
+]
+hostmakedepends = ["asciidoctor", "go"]
+checkdepends = ["bash", "git", "gmake"]
```
triallax, line 11: tests pass for me without gmake
```suggestion
checkdepends = ["bash", "git"]
```

triallax (changes requested):

triallax:
also a good idea to generate the completions, see e.g. https://git.alpinelinux.org/aports/tree/community/git-lfs/APKBUILD#n62 (don't forget !cross)

and also alpine bumps a few things for cve's (https://git.alpinelinux.org/aports/tree/community/git-lfs/APKBUILD#n49), maybe we want to do that too?

contrib/git-lfs/template.py
triallax: actually, by using the go build style for this we're missing out on running some tests (at least integration tests), maybe we should make the whole template use the make build style instead

contrib/git-lfs/template.py
subnut: Ah yes, I added it before the `*[for ...]` optimization was added, because that step was taking a LOT longer.

Removed :+1:

subnut:
The CVEs mentioned in the APKBUILD seem to have been patched already, but there's a different one present in x/net/http it seems -

https://deps.dev/go/github.com%2Fgit-lfs%2Fgit-lfs%2Fv3/v3.5.1  
https://pkg.go.dev/vuln/GO-2024-2687

Although it's listed as moderate, and it's seemingly aimed for DoS-ing servers, so it should not be affecting git-lfs AFAICT ? :thinking: 

---

Good tip about checking the APKBUILD though. I'd forgotten about that.  
(I seem to have gotten a bit rusty ngl.)

subnut:
Correct me if I'm wrong, but AFAICT, if we don't add the completions, then the template doesn't need `!nocross`, right?

contrib/git-lfs/template.py
subnut: Oh! I'll look into it.

subnut: Actually I'd initially made the template using make build style.

But the included makefile is a mess IMHO, so I just reverted to building it using the go build style and doing the other bits myself.

subnut: _Latest commit needs gmake to run tests._

subnut:
- [x] Shell completions included
- [ ]

contrib/git-lfs/template.py
```diff
@@ -0,0 +1,46 @@
...
+hostmakedepends = ["asciidoctor", "go"]
+checkdepends = ["bash", "curl", "git", "gmake", "perl"]
+depends = ["git"]
+pkgdesc = "Git extension for versioning large files"
+maintainer = "Subhaditya Nath <sn03.general@gmail.com>"
+license = "MIT"
+url = "https://git-lfs.com"
+source = f"https://github.com/git-lfs/git-lfs/releases/download/v{pkgver}/git-lfs-v{pkgver}.tar.gz"
+sha256 = "fc19c7316e80a6ef674aa4e1863561c1263cd4ce0588b9989e4be9461664d752"
+env = {"GOMODCACHE": "/cbuild_cache/golang/pkg/mod"}
```
nekopsykose, line 18:
```suggestion
```

put 

```
    from cbuild.util import golang

    self.env.update(golang.get_go_env(self))
```

in do_check

contrib/git-lfs/template.py
```diff
@@ -0,0 +1,47 @@
...
+
+def post_build(self):
+    self.mkdir("man")
+    self.do(
+        "asciidoctor",
+        "-bmanpage",
+        "-Dman",
+        "-a",
+        f"mansource={pkgname} {pkgver}",
+        *[f for f in self.find("docs/man", "*.adoc")],
```
nekopsykose, line 28: use a normal loop

subnut: IIRC a normal loop took MUCH longer, and that's why I did this optimization.

The normal loop took about 10-15 seconds on my system, whereas this one takes less than a second, IIRC.

subnut: Just checked - 
- Normal loop takes ~8 seconds
- List unpacking takes ~0.5 seconds

:shrug:

subnut: Should I keep this or switch to normal loop?

subnut: ah that's true



i'd forgotten about them lol.

PR #2307, opened 2024-06-23
 contrib/sccache: enable distributed server 

z-erica:


nekopsykose:
this is like x86_64-only isn't it? you need to guard it

q66:
why is it x86_64-only?

nekopsykose:
ask upstream https://github.com/mozilla/sccache/blob/35c502af4bb6fe12beeba28982e6915894433bf2/src/bin/sccache-dist/main.rs#L38

PR #2308, opened 2024-06-23
 contrib/icecream: new package (1.4) 

z-erica:


q66:
docbook2x is dead af (the release is from 2007), and icecc in git has long switched away from it since years ago, i'm not sure how good of an idea is to add that

also all the restart = true can go from services

q66:
do the logfiles get written? since it's dropping privs it may not have permissions to create /var/log/icecc (depending on whether it does so before or after dropping privs)

q66:
from what i can see it only does that when detach is true, and detach is only true when daemonizing, so it won't do that

q66:
via tmpfiles.d

contrib/icecream/files/tmpfiles.conf
```diff
@@ -0,0 +1,3 @@
+# icecream tmpfiles setup
+
+d /var/log/icecc 1770 _icecc _icecc -
```
q66, line 3: upstream uses 0755 here as far as i can tell, which is i think fine? there isn't anything sensitive in the logs

q66:
if you're patching in ugetopt you should also add the runtime dependency (see the other stuff that adds it)

PR #2319, opened 2024-06-24
contrib/fnc: new package (0.16)

z-erica:


nekopsykose:
this has an sqlite you could devendor

nekopsykose:
do check it works since some of those are random options i haven't seen so they might be required on sqlite proper

nekopsykose:
sounds good

Issue #304, opened 2023-08-10
compress manpages


nekopsykose:
default manpage (and info) split could pass a `gzip -9n` for free, since mandoc transparently decompresses it anyway

nekopsykose:
the only actual handling that has to be done is manpages can be symlinks i.e.  

```
xyz.1 -> abc.1
```

in which case, `abc.1` becomes `abc.1.gz`, and xyz.1 becomes a broken link (it has to be recreated to point to `abc.1.gz`).

nekopsykose:
if the links are handled correctly in a check then using `LINK_PAGES=symlink` for `man-pages` becomes viable, since most of them are just manpage links and make it like half the size from that alone

PR #2324, opened 2024-06-24
cbuild: compress manpages

q66:
This is a pass right before the autosplit logic but after everything else, which compresses manpages (which are guaranteed not to be compressed as they are extracted in an earlier hook) and retargets any potential symlinks.

Fixes https://github.com/chimera-linux/cports/issues/304

### DIFF
--- a/src/cbuild/hooks/post_install/199_recompress_manpages.py
+++ b/src/cbuild/hooks/post_install/199_recompress_manpages.py
@@ -1,0 +1,25 @@
+from cbuild.core import chroot
+
+
+def invoke(pkg):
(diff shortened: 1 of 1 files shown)


q66:
There is a chance that I am missing something here, so this will need a bunch of testing, and maybe somebody is aware of cases where this could break, so I'll leave this here for a while.

q66:
Also the logic is done using `gzip` in-chroot as the native python interface does not have a way to omit metadata. Also, this may be broken for bootstrap packages during stage 0 so might disable the hook for that, or even for anything <stage2. But I'll do that right after we are sure that this change is otherwise good.

PR #2328, opened 2024-06-25
main/gtk4: enable int and tests

triallax:
tested only on x86_64


triallax:
ci skipped latest push because it was just a dependency reordering

PR #2336, opened 2024-06-26
main/exiv2: run tests, remove unneeded update.py

triallax:


main/exiv2/template.py
```diff
@@ -5,19 +5,17 @@
...
     "-DEXIV2_BUILD_SAMPLES=OFF",
     "-DEXIV2_ENABLE_BMFF=ON",
-    "-DEXIV2_BUILD_UNIT_TESTS=OFF",
+    "-DEXIV2_BUILD_UNIT_TESTS=ON",
 ]
-make_check_target = "unit_test"
 hostmakedepends = ["cmake", "ninja", "pkgconf", "python"]
-makedepends = ["libexpat-devel", "zlib-devel", "brotli-devel", "inih-devel"]
+makedepends = ["brotli-devel", "inih-devel", "libexpat-devel", "zlib-devel"]
+checkdepends = ["gtest-devel", "python"]
```
nekopsykose, line 12: gtest to makedepends

python is already in host

the reason for that is that since unit_tests=on, it will fail if gtest is not found, but checkdepends are skipped when !check, so !check would break the build if it's in checkdepends

PR #2342, opened 2024-06-26
contrib/hostapd: new package (2.10)

rbgarga:


contrib/hostapd/template.py
```diff
@@ -0,0 +1,27 @@
...
+license = "BSD-3-Clause"
+url = "https://w1.fi/hostapd"
+source = f"https://w1.fi/releases/{pkgname}-{pkgver}.tar.gz"
+sha256 = "206e7c799b678572c2e3d12030238784bc4a9f82323b0156b4c9466f1498915d"
+# No tests available
+options = ["!check"]
+
+
+def init_configure(self):
+    self.cp("defconfig", ".config")
```
nekopsykose, line 22: think you want to enable some stuff that's not enabled by default like

- CONFIG_SAE
- CONFIG_ACS
- CONFIG_MBO
- CONFIG_DRIVER_WIRED/NONE
- CONFIG_RADIUS_SERVER
- CONFIG_IEEE80211* (even AC is not enabled by default, so only b/g/n wifi)
- CONFIG_FULL_DYNAMIC_VLAN
- CONFIG_VLAN_NETLINK
- CONFIG_P2P/MANAGER
- CONFIG_SUITEB192
- CONFIG_WPA_CLI_EDIT
- CONFIG_OWE

etc. note that some of these are not even written in defconfig but are respected anyway

contrib/hostapd/template.py
```diff
@@ -0,0 +1,27 @@
...
+make_cmd = "gmake"
+make_install_args = ["BINDIR=/usr/bin"]
+make_use_env = True
+hostmakedepends = ["pkgconf"]
+makedepends = ["gmake", "libnl-devel", "linux-headers", "openssl-devel"]
+pkgdesc = "IEEE 802.11 AP, IEEE 802.1X/WPA/WPA2/EAP/RADIUS Authenticator"
+maintainer = "Renato Botelho do Couto <renato@netgate.com>"
+license = "BSD-3-Clause"
+url = "https://w1.fi/hostapd"
+source = f"https://w1.fi/releases/{pkgname}-{pkgver}.tar.gz"
```
nekopsykose, line 15:
```suggestion
source = f"https://w1.fi/releases/hostapd-{pkgver}.tar.gz"
```

rbgarga: Out of curiosity, what is the reason to not use pkgname on source URLs?

nekopsykose: we just decided to not do that anymore at some point. it's not really a variable since if the template is copied and the pkgname changed the source becomes wrong, and in cases where you have like github/project/{pkgname} you lose the ability to copy the url quickly

PR #2354, opened 2024-06-27
main/ldb: move comment to correct place

triallax:


triallax:
lint will fail for builds targeting ppc64 because comment should be above main options declaration

PR #2355, opened 2024-06-27
main/libavif: run tests, enable libsharpyuv support

triallax:


main/libavif/template.py
```diff
@@ -1,35 +1,34 @@
...
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
 makedepends = [
-    "gdk-pixbuf-devel",
     "dav1d-devel",
+    "gdk-pixbuf-devel",
+    "gtest-devel",
     "libaom-devel",
     "libjpeg-turbo-devel",
     "libpng-devel",
+    "libwebp-devel",  # for libsharpyuv
```
triallax, line 21: i also thought of splitting a libsharpyuv subpackage from libwebp, what do you think?

nekopsykose: i don't see the point (this is pulled in every deptree ever)

don't need the comment either since it's just a webp lib and comes from it so it's not special

PR #2362, opened 2024-06-28
main/python-cffi: run tests

triallax:


main/python-cffi/patches/clang-errors.patch
```diff
@@ -0,0 +1,24 @@
...
+diff --git a/src/cffi/recompiler.py b/src/cffi/recompiler.py
+index 4167bc05..ac6c163e 100644
+--- a/src/cffi/recompiler.py
++++ b/src/cffi/recompiler.py
+@@ -953,7 +953,7 @@ def _struct_ctx(self, tp, cname, approxname, named_ptr=None):
+                 if cname is None or fbitsize >= 0:
+                     offset = '(size_t)-1'
+                 elif named_ptr is not None:
+-                    offset = '((char *)&((%s)0)->%s) - (char *)0' % (
++                    offset = '(size_t)(((char *)&((%s)4096)->%s) - (char *)4096)' % (
```
triallax, line 21: i _think_ this is correct c++ code that behaves the same as the original, but do let me know if i'm wrong

nekopsykose: it should probably be just `offsetof(%s, %s)`

triallax: for posterity, `offsetof` wouldn't work here: https://github.com/python-cffi/cffi/pull/78#issuecomment-2198107460

PR #2373, opened 2024-06-29
contrib/python-automat: enable tests

mikoxyz:


contrib/python-graphviz/patches/0001-nuke-the-cov-meme.patch
```diff
@@ -0,0 +1,35 @@
...
+
+diff --git a/setup.cfg b/setup.cfg
+index a575f6c..dd99809 100644
+--- a/setup.cfg
++++ b/setup.cfg
+@@ -17,18 +17,12 @@ addopts =
+ 	--doctest-continue-on-failure
+ 	-r fsxX
+ 	--durations=10
+-	--cov --cov-report=term --cov-report=html
```
nekopsykose, line 18: you only need to patch this and not the stuff below pretty sure so you can make the diff smaller

PR #2381, opened 2024-07-01
contrib/bat: specify directory for generated assets

tranzystorekk:


triallax:
you should format the template with black/ruff

PR #2392, opened 2024-07-02
Assorted LXQt components

ifreund:
Someone has reported issues with lxqt-panel and lxqt-runner under river and I think I need to run them under a debugger to figure out what's going on. Unfortunately these packages don't work yet and seem to require further LXQt components in order to function. I guess I'll keep packaging LXQt stuff until I get them running properly or get bored.

### DIFF
--- a/contrib/lxqt-panel/template.py
+++ b/contrib/lxqt-panel/template.py
@@ -1,0 +1,38 @@
+pkgname = "lxqt-panel"
+pkgver = "2.0.1"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DALSA_PLUGIN=NO",
+    "-DCPULOAD_PLUGIN=NO",
+    "-DNETWORKMONITOR_PLUGIN=NO",
+    "-DSYSSTAT_PLUGIN=NO",
+    "-DVOLUME_USE_ALSA=NO",
+]
+hostmakedepends = [
+    "cmake",
+    "lxqt-build-tools",
+    "ninja",
+    "perl",
+    "pkgconf",
+]
+makedepends = [
+    "kwindowsystem-devel",
+    "layer-shell-qt-devel",
+    "libdbusmenu-lxqt-devel",
+    "liblxqt-devel",
+    "libpulse-devel",
+    "libsensors-devel",
+    "lxqt-globalkeys-devel",
+    "lxqt-menu-data",
+    "qt6-qttools-devel",
+    "solid-devel",
+    "xcb-util-devel",
+]
+depends = ["lxqt-menu-data"]
+pkgdesc = "LXQt desktop panel"
+maintainer = "Isaac Freund <mail@isaacfreund.com>"
+license = "LGPL-2.1-or-later"
+url = "https://github.com/lxqt/lxqt-panel"
+source = f"{url}/releases/download/{pkgver}/lxqt-panel-{pkgver}.tar.xz"
+sha256 = "73483c36e411496f8e958b7e56ba8bb06ae0b4300a62cf4c4a78964da6a59407"
--- a/contrib/lxqt-themes/template.py
+++ b/contrib/lxqt-themes/template.py
@@ -1,0 +1,17 @@
+pkgname = "lxqt-themes"
+pkgver = "2.0.0"
+pkgrel = 0
+build_style = "cmake"
+hostmakedepends = [
+    "cmake",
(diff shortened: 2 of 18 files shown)


ifreund:
This appears to be enough to test the things I wanted to test. I don't plan on packinging further LXQt components at this time. What's here should already be quite useful for anyone wanting to take this further.

contrib/muparser/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "muparser"
+pkgver = "2.3.4"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DENABLE_OPENMP=OFF"]
```
nekopsykose, line 5: this just needs libomp-devel in makedepends and then it works (most likely)

contrib/qtxdg-tools/template.py
```diff
@@ -0,0 +1,19 @@
...
+    "lxqt-build-tools",
+]
+makedepends = [
+    "libqtxdg-devel",
+]
+pkgdesc = "User tools for libqtxdg"
+maintainer = "Isaac Freund <mail@isaacfreund.com>"
+license = "LGPL-2.1-or-later"
+url = "https://github.com/lxqt/qtxdg-tools"
+source = f"{url}/releases/download/{pkgver}/{pkgname}-{pkgver}.tar.xz"
```
nekopsykose, line 18:
```suggestion
source = f"{url}/releases/download/{pkgver}/qtxdg-tools-{pkgver}.tar.xz"
```

(for everything)

contrib/lxqt-panel/template.py
```diff
@@ -0,0 +1,38 @@
+pkgname = "lxqt-panel"
+pkgver = "2.0.1"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DCPULOAD_PLUGIN=NO",
+    "-DNETWORKMONITOR_PLUGIN=NO",
+    "-DSYSSTAT_PLUGIN=NO",
```
nekopsykose, line 8: any reason to disable these

contrib/lxqt-panel/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "layer-shell-qt-devel",
+    "libdbusmenu-lxqt-devel",
+    "liblxqt-devel",
+    "libsensors-devel",
+    "lxqt-globalkeys-devel",
+    "lxqt-menu-data",
+    "qt6-qttools-devel",
+    "solid-devel",
+    "xcb-util-devel",
+    "libpulse-devel",
```
nekopsykose, line 30:
```suggestion
    "kwindowsystem-devel",
    "layer-shell-qt-devel",
    "libdbusmenu-lxqt-devel",
    "liblxqt-devel",
    "libpulse-devel",
    "libsensors-devel",
    "lxqt-globalkeys-devel",
    "lxqt-menu-data",
    "qt6-qttools-devel",
    "solid-devel",
    "xcb-util-devel",
```

(sort all the lists)

contrib/muparser/template.py
ifreund: Honestly, I really don't think a parsing library that is only used to implement a calculator in the `lxqt-runner` program needs multithreading...

If enabling it doesn't make things more fragile I guess it wouldn't hurt though.

contrib/qtxdg-tools/template.py
ifreund: sure, the status quo did make copy pasting to create these templates easier though.

contrib/lxqt-panel/template.py
ifreund: The reason is that I was too lazy to package the required dependencies

contrib/muparser/template.py
nekopsykose: if they have it on by default then may as well leave it on

contrib/lxqt-themes/template.py
```diff
@@ -0,0 +1,19 @@
...
+build_style = "cmake"
+hostmakedepends = [
+    "cmake",
+    "lxqt-build-tools",
+    "ninja",
+    "perl",
+    "pkgconf",
+]
+makedepends = []
+depends = []
```
nekopsykose, line 13:
```suggestion
```

contrib/muparser/template.py
ifreund: done

PR #2393, opened 2024-07-02
contrib/ipmitool: new package (1.8.19)

rbgarga:


contrib/ipmitool/template.py
```diff
@@ -0,0 +1,29 @@
...
+hostmakedepends = ["automake", "gmake", "libtool"]
+makedepends = [
+    "linux-headers",
+    "openssl-devel",
+    "readline-devel",
+]
+pkgdesc = "CLI to manage IPMI systems"
+maintainer = "Renato Botelho do Couto <renato@netgate.com>"
+license = "BSD-3-Clause"
+url = "https://sourceforge.net/projects/ipmitool"
```
nekopsykose, line 23:
```suggestion
url = "https://codeberg.org/IPMITool/ipmitool"
```

it says sourceforge isn't used anymore

this also has 1.8.19

nekopsykose:
you'll have to add https://www.iana.org/assignments/enterprise-numbers.txt to source too and put it in ./enterprise-numbers i guess

rbgarga:
> you'll have to add https://www.iana.org/assignments/enterprise-numbers.txt to source too and put it in ./enterprise-numbers i guess

I needed to patch `configure.ac` to fix it when neither curl or wget are present on system and then build passed.  My previous push was missing the deletion of obsoleted patches too

nekopsykose:
sure :) does it work correctly without the enterprise numbers later though? normally it downloads and then installs them itself into /usr/share/misc/enterprise-numbers

rbgarga:
> sure :) does it work correctly without the enterprise numbers later though? normally it downloads and then installs them itself into /usr/share/misc/enterprise-numbers

good question, let me add it anyway

nekopsykose:
seems they're used around here https://codeberg.org/IPMITool/ipmitool/src/commit/eb1df8d6a608074c05a1db768830747c01927aaa/lib/ipmi_strings.c#L1509

contrib/ipmitool/patches/fix-enterprise-numbers.patch
```diff
@@ -0,0 +1,49 @@
...
+-	}
+-	@mv tmpfile.$$PPID $@
+-
+-install-pen-database: enterprise-numbers
++install-pen-database:
+ 	mkdir -m 755 -p $(DESTDIR)$(IANADIR)
+-	$(INSTALL_DATA) $< $(DESTDIR)$(IANADIR)/
++	$(INSTALL_DATA) enterprise-numbers $(DESTDIR)$(IANADIR)/
+ 
+ uninstall-pen-database:
```
nekopsykose, line 32: you can make this patch shorter by just replacing `if DOWNLOAD` with `if 1`, since the enterprise-numbers: rule will always be 'made' with the file being there, and then the rest just works

rbgarga:
For the record, code first tries to read `enterprise-numbers` from `$HOME/.local/usr/share/misc` so I went ahead and removed it from package to keep it simple

PR #2397, opened 2024-07-02
contrib/qemu: put edk2-fw in its own package

nekopsykose:
mostly to fix the case of creating an x86_64 uefi vm in virt-manager and it needing i386-vars specifically, which would require installing qemu-system-i386

this makes it work ootb

[ci skip]

also some other fixes


nekopsykose:
could also just pull it in every -system instead of only the same as the previous ones

PR #2399, opened 2024-07-03
main/libwacom: run more tests, fix installed python scripts

triallax:


main/libwacom/template.py
```diff
@@ -10,12 +10,13 @@
     "libgudev-devel",
     "libxml2-devel",
 ]
-checkdepends = ["bash"]
+depends = ["python-libevdev", "python-pyudev", "udev"]
```
triallax, line 13: is adding a udev rdep okay here?

triallax:
note: libwacom-update-db will still typically initially fail because it needs /etc/udev/hwdb.d to already exist

main/libwacom/template.py
nekopsykose: you don't need it

triallax:
above issue doesn't apply anymore after https://github.com/chimera-linux/cports/commit/649cff96962d40cb6e590f0b8083c7e287287661

main/libwacom/template.py
```diff
@@ -10,12 +10,13 @@
     "libgudev-devel",
     "libxml2-devel",
 ]
-checkdepends = ["bash"]
+depends = ["python-libevdev", "python-pyudev"]
+checkdepends = ["bash", "python-libevdev", "python-pytest", "python-pyudev"]
```
nekopsykose, line 14:
```suggestion
depends = ["python-libevdev", "python-pyudev"]
checkdepends = ["bash", "python-pytest"] + depends
```

PR #2400, opened 2024-07-03
contrib/fwupd-efi: update to 1.6

nekopsykose:
i have no idea how to actually test this at runtime since fwupd only seems to use it if you actually have an efi update ready...


nekopsykose:
got an efi update and this version hangs forever while 1.4 works, nice

nekopsykose:
doesn't work so good luck to whoever want to try

PR #2408, opened 2024-07-03
contrib/clucene: move cmake config to -devel

triallax:


triallax:
just checked and alpine removes this file entirely, we could get away with that too maybe

nekopsykose:
you could move it to /usr/lib/cmake

nekopsykose:
nvm it's like completely useless

nekopsykose:
er, i thought you had removed it already, i'll do that, derp

PR #2419, opened 2024-07-05
main/glib: don't skip gio contenttype test

triallax:


triallax:
worked locally but fails in ci apparently

PR #2425, opened 2024-07-06
cbuild: actually strip bins in strip

nekopsykose:
strip --strip-debug keeps .strtab/.symtab/.comment  

---

- maybe just `--strip-unneeded --keep-section=.gnu_debuginfo` is better?
- i'm not sure if the objcopy to itself is valid


nekopsykose:
kept strip ig

PR #2428, opened 2024-07-06
user/nmap: new package (7.95)

Gnarwhal:


user/nmap/template.py
```diff
@@ -0,0 +1,33 @@
+pkgname = "nmap"
+pkgver = "7.95"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--with-openssl",
+    "--without-zenmap",
+    "--without-ndiff",
+    "--without-ncat",
```
nekopsykose, line 9:
```suggestion
```

no reason to disable these

user/nmap/template.py
```diff
@@ -0,0 +1,33 @@
...
+]
+configure_gen = []
```
nekopsykose, line 11:
```suggestion
```
fix the regen instead

user/nmap/template.py
```diff
@@ -0,0 +1,33 @@
...
+make_cmd = "gmake"
+make_dir = "."
+hostmakedepends = [
+    "automake",
+    "gmake",
+]
+makedepends = [
+    "linux-headers",
+    "openssl-devel",
```
nekopsykose, line 20: you are missing a bunch of libraries and configure args to make it not use its own vendored libs

user/nmap/template.py
```diff
@@ -0,0 +1,33 @@
...
+]
+pkgdesc = "Utility for network discovery and security auditing"
+maintainer = "Gnarwhal <git.aspect893@passmail.net>"
+license = "custom:Nmap"
+url = "https://nmap.org"
+source = f"https://nmap.org/dist/nmap-{pkgver}.tar.bz2"
+sha256 = "e14ab530e47b5afd88f1c8a2bac7f89cd8fe6b478e22d255c5b9bddb7a1c5778"
+options = ["!parallel"]
```
nekopsykose, line 28:
```suggestion
```

builds fine in parallel

user/nmap/template.py
nekopsykose: perhaps

```
		--with-openssl=yes
		--with-libpcap=yes
		--with-libpcre=yes
		--with-libz=yes
		--with-libssh2=yes
		--with-liblua=/usr/lua5.4
```

with the respective libs

user/nmap/template.py
nekopsykose: ( i think you only need -j1 for make_install_args, but maybe not even that)

user/nmap/template.py
nekopsykose: you'll need to rebase

https://github.com/nmap/nmap/commit/828ab48764b82d0226e860c73c5dac5b11f77385
https://github.com/nmap/nmap/commit/4c0ed54f6848ddf1222c7d4828dc2d652c2f5f72
https://github.com/nmap/nmap/commit/c6ffdbf7ae35e397b663f787fc4934f01458f945

for pcre2 to work

user/nmap/template.py
```diff
@@ -0,0 +1,48 @@
...
+    "--without-zenmap",
+]
+configure_gen = [
+    "autoreconf",
+    "-if",
+    "--no-recursive",
+]
+make_cmd = "gmake"
+make_dir = "."
+make_build_args = ["-j1"]
```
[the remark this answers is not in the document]
nekopsykose: (the part that fails is because they call `make`, so you want `exec_wrappers = [("/usr/bin/gmake", "make")]`)

user/nmap/template.py
nekopsykose: hm, interestingly it links it anyway

user/nmap/template.py
```diff
@@ -0,0 +1,48 @@
...
+    "gmake",
+    "gettext-devel",
+]
+makedepends = [
+    "linux-headers",
+    "openssl-devel",
+    "libpcap-devel",
+    "pcre2-devel",
+    "libssh2-devel",
+    "libtool",
```
nekopsykose, line 33:
```suggestion
    "libtool",
```

libtool is hostmakedepends

user/nmap/template.py
```diff
@@ -0,0 +1,48 @@
...
+    "lua5.4-devel",
+    "zlib-ng",
```
nekopsykose, line 35:
```suggestion
    "zlib-ng-compat-devel",
```

user/nmap/template.py
nekopsykose: i still think you need parts of those commits though

Gnarwhal: I'm sorry I'm not sure what this means. Would this mean applying the changes in those commits via patch files to the source after downloading?

user/nmap/template.py
Gnarwhal: Sorry for the newbie question/wall of text. `autoreconf` is failing currently because the included libssh2's `Makefile.am` is requiring the [`tests/` and `example/`](https://github.com/nmap/nmap/blob/667527c4b46abb9cceef7a8b30a1f8a7e5d04d49/libssh2/Makefile.am#L3) directories which don't exist in the nmap included libssh2. My first inclination is to stop `autoreconf` from configuring that subdirectory since nmap is using the host libssh2 anyway. The problem is I don't know how to go about doing this :/ Insofar as I can tell `autoreconf` only has the `--no-recursive` flag which isn't useful because some subdirectories should be configured. I have maybe a few ideas, but I they all seem like a bodge. 
1. Modify the root `configure.ac` file to remove the `AC_CONFIG_SUBDIRS` (via patch file?)
2. Modify the `libssh2/Makefile.am` file to remove the requirement for `tests/` and `example/`
3. Run `autoreconf` with `--no-recursive` and then have cbuild manually recurse into only the subdirectories that actually need to be configured.

Getting your thoughts on this would be greatly appreciated. 💜

P.S. Is it better to ask this kind of thing in the IRC/matrix?

user/nmap/template.py
Gnarwhal: Part of the ncat test requires doing a dns resolution. I'm guessing this means that the ncat tests cannot be run since the cbuild sandbox has no network access during the check phase. Assuming that's true, is there a natural/obvious way to disable tests more granularly than `options = ["!check"]`?

user/nmap/template.py
nekopsykose: you would put patches in a folder called patches/ and they get applied for you, yeah

user/nmap/template.py
nekopsykose: > Assuming that's true, is there a natural/obvious way to disable tests more granularly than options = ["!check"]?  

if the build system exposes one, sure. if not, then no. dunno if there is a way in this case

user/nmap/template.py
nekopsykose: > P.S. Is it better to ask this kind of thing in the IRC/matrix?  

either place is alright :)  

> Modify the root configure.ac file to remove the AC_CONFIG_SUBDIRS (via patch file?)

yeah, a patch to remove the _subdirs call to libssh2 sounds ideal in this case

Gnarwhal:
No idea why but `'CPP=clang-cpp'` breaks configuration for `libdnet-stripped`

nekopsykose:
i couldn't figure it out either, looks like something is broke in the autoconf scripts i guess but those are impossible to debug. that workaround seems fine

user/nmap/template.py
Gnarwhal: From what I can tell

https://github.com/nmap/nmap/commit/828ab48764b82d0226e860c73c5dac5b11f77385 - This is already included in release 7.95
https://github.com/nmap/nmap/commit/4c0ed54f6848ddf1222c7d4828dc2d652c2f5f72 - The bad patterns don't exist in 7.95
https://github.com/nmap/nmap/commit/c6ffdbf7ae35e397b663f787fc4934f01458f945 - I've now included a patch for this with the exception of the changes to `configure.ac` (autoheader breaks if they were included)

nekopsykose: thanks!

PR #2438, opened 2024-07-08
 main/libgit2: update to 1.8.1 

z-erica:


nekopsykose:
yes, every single rust thing needs a patch to the lockfile, and things that consume the cargo crate as a dep (cargo-edit/cargo-flamegraph) need to update the cargo crate to a version that has libgit2-1.8.1.. and that version is released next rust update :D  

bumping it to a git version would work except then cargo vendor doesn't work as the cargo repository doesn't like that since they put invalid files in it or something

jcgruenhage:
Both cargo 0.80.0 and 0.81.0 which have been tagged since this conversation still depend on `libgit2-sys 0.16.2+1.7.2`. I think we might need another solution here.

What do people think about doing a libgit2 split like we also do for wlroots? Void Linux is also doing that: https://github.com/void-linux/void-packages/pull/50901#issuecomment-2267505849

nekopsykose:
> Both cargo 0.80.0 and 0.81.0 which have been tagged since this conversation still depend on libgit2-sys 0.16.2+1.7.2. 

yeah, i thought they'd include it in 0.81 but it'll only be 0.82

> What do people think about doing a libgit2 split like we also do for wlroots?  

i guess i'll just do that later

q66:
i don't see a problem with just sticking with the older version until next cargo?

nekopsykose:
> i don't see a problem with just sticking with the older version until next cargo?

multiple things using libgit2-sys want the newer one instead

jcgruenhage:
It's also possible to patch those to use an older git2 release, as I've just done with git-cliff. Annoying, but I don't know whether it's more or less annoying than a libgit2 split

PR #2443, opened 2024-07-08
 user/drawterm: new package (0_git20240703) 

z-erica:


user/wlr-protocols/template.py
```diff
@@ -0,0 +1,19 @@
+pkgname = "wlr-protocols"
```
q66, line 1: don't really like this, i've never seen anything directly use this and it's yet another thing with no releases that nothing really uses

user/drawterm/template.py
```diff
@@ -0,0 +1,33 @@
...
+make_use_env = True
+hostmakedepends = [
+    "pkgconf",
+    "wayland-progs",
+]
+makedepends = [
+    "libxkbcommon-devel",
+    "linux-headers",
+    "pipewire-devel",
+    "wayland-devel",
```
q66, line 16: building the wayland gui means it's likely tied to just wayland compositors and quite possibly only wlroots which is sketch

q66:
also https://github.com/void-linux/void-packages/blob/master/srcpkgs/drawterm/patches/posix_generic.patch

user/wlr-protocols/template.py
q66: you can add the one xml file drawterm needs into its sources and then pass `WLR_VIRTUAL_POINTER` on make commandline instead

nekopsykose: wlr-protocols is not meant to be packaged

user/drawterm/template.py
nekopsykose: nothing, seems fine

nekopsykose: for linux, they explicitly only build GUI=wl. i guess you could do a second build with GUI=x11

user/drawterm/template.py
```diff
@@ -0,0 +1,44 @@
...
+    "wayland-protocols",
+]
+pkgdesc = "Interface to Plan 9 systems"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "MIT"
+url = "https://drawterm.9front.org"
+source = [
+    f"https://git.9front.org/git/plan9front/drawterm/{_commit}/snap.tar.gz",
+    "!https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/raw/2b8d43325b7012cc3f9b55c08d26e50e42beac7d/unstable/wlr-virtual-pointer-unstable-v1.xml",
```
nekopsykose, line 26: you have to rename the file (`>newname` on the end, etc) so it changes name on update

i guess if we're pulling it like this anyway then may as well actually package it since it's less effort (regardless what i said before)

nekopsykose: the name of the downloaded file is `wlr-virtual-pointer-unstable-v1.xml`, so if you change the entire hash in front of it to download some hypothetical different file it would still reuse the cached one

(for this specific case this is more an imaginary problem, but in general)

user/drawterm/template.py
```diff
@@ -0,0 +1,53 @@
...
+    self.cp(
+        self.sources_path / f"{_wlr_virtual_pointer_commit}.xml",
+        "wlr-virtual-pointer-unstable-v1.xml",
+    )
+
+
+def init_build(self):
+    self.make_build_args += [
+        f"WLR_VIRTUAL_POINTER={self.chroot_cwd}/wlr-virtual-pointer-unstable-v1.xml"
+    ]
```
nekopsykose, line 47: you could probably not copy it at all and just self.chroot_sources_path / in here

PR #2445, opened 2024-07-08
contrib/php8.3: add pear subpackage

rbgarga:
- **contrib/php8.3: sort configure_args entries**
- **contrib/php8.3: disable one more failing test**
- **contrib/php8.3: add pear subpackage**


contrib/php8.3/template.py
```diff
@@ -244,6 +249,19 @@ def post_install(self):
...
+        ".depdb",
+        ".depdblock",
+        ".filemap",
+        ".lock",
+        f"usr/share/php{_majver}/pear/.channels",
+        f"usr/share/php{_majver}/pear/.filemap",
+        f"usr/share/php{_majver}/pear/.lock",
+        f"usr/share/php{_majver}/pear/.registry",
+    ]:
+        self.rm(f"{self.destdir}/{f}", recursive=True)
```
nekopsykose, line 264:
```suggestion
        self.uninstall(f, recursive=True)
```

but is there a way to make them somehow not installed in the first place? if it's a small build system patch it's a bit easier to not miss it later

contrib/php8.3/template.py
```diff
@@ -328,6 +347,20 @@ def _ext(self):
...
+def _pear(self):
+    self.pkgdesc = f"PHP{_majver} Extension and Application Repository"
+    self.depends = [f"{pkgname}={pkgver}-r{pkgrel}", f"{pkgname}-xml"]
+    self.install_if = [f"{pkgname}={pkgver}-r{pkgrel}"]
+
+    return [
+        f"etc/php{_majver}/pear.conf",
+        f"usr/bin/pear{_majver}",
+        f"usr/bin/peardev{_majver}",
+        f"usr/bin/pecl{_majver}",
```
nekopsykose, line 360: this leaves behind a usr/share/php8.3/pear/ in the main package. think you have to move that too

nekopsykose: also not sure if /usr/share/php8.3/pear/test/ is required or if you want to add an uninstall for that above

nekopsykose: also, when the same version is not the default, it seems to not work:  

```
$ pear8.3
exec: /usr/bin/php: not found
```

so i think this will break when there is more than one php version. looking at the script it's because it hardcodes a /usr/bin/php

nekopsykose: same goes for peardev/pecl of course

contrib/php8.3/template.py
rbgarga: I didn't find an easy way.  I also noticed inside pear .phar installer that the RPM spec file also removes these files after installation so I think it's the only way of doing that unfortunately

nekopsykose: that's ok :) uninstall is fine then

nekopsykose:
looks good, can drop the # tmp !check

rbgarga:
> looks good, can drop the # tmp !check

ooops.  done

PR #2462, opened 2024-07-10
main/modemmanager: add service file

alkaline-system:
still learning to write services, complain if something isn't right

### DIFF
--- a/main/modemmanager/files/modemmanager
+++ b/main/modemmanager/files/modemmanager
@@ -1,0 +1,6 @@
+type = process
+command = /usr/bin/ModemManager
+before = network.target
+depends-on = dbus
+depends-on = pre-network.target
+smooth-recovery = true
--- a/main/modemmanager/template.py
+++ b/main/modemmanager/template.py
@@ -1,6 +1,6 @@
 pkgname = "modemmanager"
 pkgver = "1.22.0"
-pkgrel = 0
+pkgrel = 1
 build_style = "meson"
 configure_args = [
     "-Ddbus_policy_dir=/usr/share/dbus-1/system.d",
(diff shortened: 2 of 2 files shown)


NovaAndrom3da:
Can you squash the commits please

nekopsykose:
i think by default git pull will create merge commits like that, so you need `git pull --rebase` or similar

main/modemmanager/files/modemmanager
```diff
@@ -0,0 +1,9 @@
+# modemmanager service
+
+type            = process
+command         = /usr/bin/modemmanager
+before          = network.target
+depends-on      = dbus
+depends-on      = pre-network.target
+restart         = true
+smooth-recovery = true
```
nekopsykose, line 9:
```suggestion
type = process
command = /usr/bin/modemmanager
before = network.target
depends-on = dbus
depends-on = pre-network.target
smooth-recovery = true
```

don't need indents or restart=true anymore (yeah should update the other files eventually..)

main/modemmanager/template.py
```diff
@@ -47,6 +47,10 @@
 sha256 = "6c8f8720737a3788e394c700f36236278c9de09d76069a079e6f1daaf08b2768"
 
 
+def post_install(self):
+    self.install_service(self.files_path / "modemmanager", enable=True)
```
nekopsykose, line 51:
```suggestion
    self.install_service(self.files_path / "modemmanager")
```

shouldn't be enabled by default on install

alkaline-system:
redid the commit (with suggested changes)
we good now

alkaline-system:
no, we're good NOW
sorry lol

main/modemmanager/template.py
nekopsykose: still have to remove this

main/modemmanager/files/modemmanager
```diff
@@ -0,0 +1,6 @@
+type = process
+command = /usr/bin/modemmanager
```
nekopsykose, line 2:
```suggestion
command = /usr/bin/ModemManager
```

the lowercase version doesn't exist

main/modemmanager/template.py
alkaline-system: oops

main/modemmanager/files/modemmanager
```diff
@@ -0,0 +1,6 @@
+type = process
```
triallax, line 1: service files usually have a comment at the beginning

```suggestion
# modemmanager service

type = process
```

nekopsykose: tbh they're all just `# {filename} service`, not really super insightful

could probably just start getting rid of those too

nekopsykose: will just think about that later

triallax: sure

PR #2463, opened 2024-07-10
user/vim: fixed issue with system vimrc location

Gnarwhal:


nekopsykose:
have to bump the pkgrel

PR #2464, opened 2024-07-10
ci: make ruff output github inline comments

nekopsykose:


nekopsykose:
format doesn't support this but maybe once they unify the frontend...

PR #2466, opened 2024-07-10
user/ddclient: new package (3.11.2)

Gnarwhal:


user/ddclient/files/ddclient
```diff
@@ -0,0 +1,7 @@
+# ddclient service
+
+type       = process
+command    = /usr/bin/ddclient -foreground
+depends-on = network.target
+restart    = true
+logfile    = /var/log/ddclient.log
```
nekopsykose, line 7:
```suggestion
type = process
command = /usr/bin/ddclient -foreground
depends-on = network.target
logfile = /var/log/ddclient.log
```

user/ddclient/template.py
```diff
@@ -0,0 +1,29 @@
...
+]
+depends = [
+    "curl",
+    "perl",
+]
+pkgdesc = "Client used to update dynamic DNS entries"
+maintainer = "Gnarwhal <git.aspect893@passmail.net>"
+license = "GPL-2.0-or-later"
+url = "https://ddclient.net"
+source = "https://github.com/ddclient/ddclient/archive/refs/tags/v3.11.2.tar.gz"
```
nekopsykose, line 22:
```suggestion
source = f"https://github.com/ddclient/ddclient/archive/refs/tags/v{pkgver}.tar.gz"
```

user/ddclient/template.py
```diff
@@ -0,0 +1,29 @@
...
+build_style = "gnu_configure"
+make_cmd = "gmake"
+make_dir = "."
+hostmakedepends = [
+    "automake",
+    "gmake",
+]
+makedepends = [
+    "curl",
+]
```
triallax, line 13: looks like a mistake
```suggestion
```

nekopsykose: it should be in hostmakedepends

user/ddclient/template.py
```diff
@@ -0,0 +1,29 @@
+pkgname = "ddclient"
+pkgver = "3.11.2"
+pkgrel = 0
+build_style = "gnu_configure"
+make_cmd = "gmake"
+make_dir = "."
```
nekopsykose, line 6:
```suggestion
```

these are not needed

nekopsykose:
should be in contrib/

user/ddclient/files/ddclient
Gnarwhal: Why shouldn't restart be true?

user/ddclient/template.py
```diff
@@ -0,0 +1,29 @@
...
+pkgdesc = "Client used to update dynamic DNS entries"
+maintainer = "Gnarwhal <git.aspect893@passmail.net>"
+license = "GPL-2.0-or-later"
+url = "https://ddclient.net"
+source = "https://github.com/ddclient/ddclient/archive/refs/tags/v3.11.2.tar.gz"
+sha256 = "243cd832abd3cdd2b49903e1b5ed7f450e2d9c4c0eaf8ce4fe692c244d3afd77"
+
+
+def post_install(self):
+    self.install_license("COPYING")
```
triallax, line 27: don't need to install license for gpl
```suggestion
```

user/ddclient/files/ddclient
nekopsykose: it is the default

user/ddclient/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "ddclient"
+pkgver = "3.11.2"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = [
+    "automake",
+    "curl",
+    "gmake",
```
nekopsykose, line 8:
```suggestion
```

since it's not used anymore

PR #2468, opened 2024-07-10
main/fortify-headers: update to 2.3.1

nekopsykose:
overlap checks got some fixes, mem* null checks are now off by default without a pedantic define set


nekopsykose:
also might want to use the USE_NATIVE_CHK? they might be a bit worse but if they improve diagnostics (should test) it would be nice, because the current diagnostics are quite bad

nekopsykose:
some quick stuff (and openssl) still seems to pass tests so i guess it's fine. the only changes are vasprintf/asprintf being added and the reenablement of the overlap checks

nekopsykose:
nvm, the native chk requires libc support

q66:
we should probably disable the printf stuff for now too https://github.com/jvoisin/fortify-headers/issues/62

nekopsykose:
seems even more issues showed up :D  

will just wait for next release

nekopsykose:
seems those are disabled on clang anyway so, no real changes

PR #2473, opened 2024-07-10
contrib/ytnef: split progs, add ytnefprocess rdeps

triallax:


contrib/ytnef/template.py
```diff
@@ -14,3 +14,10 @@
 @subpackage("ytnef-devel")
 def _devel(self):
     return self.default_devel()
+
+
+@subpackage("ytnef-progs")
+def _devel(self):
```
nekopsykose, line 20:
```suggestion
def _progs(self):
```

PR #2474, opened 2024-07-10
contrib/gdm-settings: new package (4.4)

Gnarwhal:


user/gdm-settings/template.py
```diff
@@ -0,0 +1,32 @@
...
+build_style = "meson"
+hostmakedepends = [
+    "blueprint-compiler",
+    "meson",
+    "gobject-introspection",
+]
+makedepends = [
+    "appstream-glib",
+    "desktop-file-utils",
+    "gettext",
```
nekopsykose, line 13: hostmake

user/gdm-settings/template.py
```diff
@@ -0,0 +1,32 @@
...
+    "glib-devel",
+    "gtk4-devel",
+    "libadwaita-devel",
+    "python-gobject-devel",
+    "python-pkgconfig",
```
nekopsykose, line 18: host

nekopsykose:
put it in contrib/

user/gdm-settings/template.py
```diff
@@ -0,0 +1,32 @@
...
+]
+depends = [
+    "gdm",
+]
+pkgdesc = "Settings app for GDM"
+maintainer = "Gnarwhal <git.aspect893@passmail.net>"
+license = "AGPL-3.0-or-later"
+url = "https://github.com/gdm-settings/gdm-settings"
+source = f"https://github.com/gdm-settings/gdm-settings/archive/refs/tags/v{pkgver}.tar.gz"
```
nekopsykose, line 27:
```suggestion
source = f"{url}/gdm-settings/archive/refs/tags/v{pkgver}.tar.gz"
```

user/gdm-settings/template.py
nekopsykose: though i think this is not used? you only want regular `pkgconf` afaict, could be wrong just had a quick look

nekopsykose:
i guess the workaround works but this code doesn't make any sense to me. what is the point of calling `cp` `mkdir` `rm` from python? the stdlib has versions of all of this..

e.g. `cp -r a -t b` can be `shutil.copy_tree(a, b)` (they both copy a tree recursively and fail when `b` exists). i would rather just see that fixed upstream instead of shell workarounds, but it would take quite a bit of refactoring around CommandList to also invoke a passed function instead of only `Command('sh', ...`

Gnarwhal:
My only guess for why it is done this way is so that gdm-settings can be run without root and then use pkexec to apply changes as root? But I don't know python well enough to know if there is a better work around ¯\_(ツ)_/¯

nekopsykose:
i guess that's a somewhat interesting reason that makes it (even) more work to do something else with

Gnarwhal:
New proposal: what if there was a supplementary program (call it `gdm-settings-applicator`) pre-written. And instead of adding shell commands directly to a list that get written to a temporary file and `sh`ed, settings.py setup command line arguments for the applicator and then the applicator as a whole was `pkexec`ed.

nekopsykose:
sure, ask upstream :D

nekopsykose:
that said this is fine for now :) just nice if made better later

PR #2490, opened 2024-07-13
bump gjs up to mozjs128

q66:


q66:
This is an experimental change. Works here (gnome-shell is stable) and the primary purpose is to nuke python 3.11. Testing would be appreciated

PR #2494, opened 2024-07-13
contrib/pipx: add dependency on pip

na-sa-do:


triallax:
what's the reasoning behind this change? i don't see pip in the list of dependencies upstream (https://github.com/pypa/pipx/blob/273c558fb5f183e16f396dbeeba2a951cfce678a/pyproject.toml#L38), and pipx works fine on my system without it

na-sa-do:
hm. I tried to install a tool with pipx and it errored because it couldn't find pip, and it worked after I installed pip, so I did this. but after uninstalling both pip and pipx, and reinstalling only pipx, it works fine... maybe I just misread the error message

PR #2496, opened 2024-07-13
contrib/hikari: new package (2.3.3)

ttyyls:


nekopsykose:
this is basically abandoned and i don't want even older half-broken wlroots versions

PR #2497, opened 2024-07-13
user/rspamd: update to 3.9.0

a16bitsysop:
* https://github.com/rspamd/rspamd/releases/tag/3.9.0


a16bitsysop:
the tests past on alpine:
https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1452333#L2546

nekopsykose:
the check that fails is `CHECK(fstr->len == 26);` because it expectes the zlib buffer to be precisely 26 but zlib-ng specifically allocates more intentionally so the buffer is not 26

a16bitsysop:
ah okay, maybe should be reported upstream then

nekopsykose:
it doesn't affect anything at runtime so just skipping the test is fine

PR #2498, opened 2024-07-13
contrib/*: disable cfi and remove cfi todos for all kde stuff

triallax:
cfi causes a lot of crashes for kde projects, and fixing them just doesn't seem worth it

[ci skip]


nekopsykose:
vis does do something if the project doesn't already set it (since it will remove some symbols from shared libraries so they end up smaller via pruning symbols that are otherwise just default exported). so since it works might as well leave it for the libraries specifically

nekopsykose:
> if the project doesn't already set it

(but i'm pretty sure almost every kde project does pass it themselves already)

triallax:
@nekopsykose sure, i'll keep this in mind for other packages as well

nekopsykose:
forgor 2 merge this before the plasma bump, just rebase it then merge it

PR #2499, opened 2024-07-13
contrib/localectl: new package (1.0.0)

Gnarwhal:
Hopefully it is in the realm of what you were looking for

Here is the source url: https://github.com/Gnarwhal/localectl-chimera

### DIFF
--- a/contrib/localectl/template.py
+++ b/contrib/localectl/template.py
@@ -1,0 +1,23 @@
+pkgname = "localectl"
+pkgver = "1.0.0"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "meson",
+    "gobject-introspection",
+    "pkgconf",
+]
+depends = [
+    "glib",
+]
+pkgdesc = "Control the system locale and keyboard layout settings"
+maintainer = "Gnarwhal <git.aspect893@passmail.net>"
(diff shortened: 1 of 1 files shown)


q66:
as i previously mentioned this should be a part of openrc-settingsd and not a separate project

Gnarwhal:
Oh you sure did. I'm so sorry I misinterpreted that. Sorry another probably base question then, but are you saying you would fork https://gitlab.com/postmarketOS/openrc-settingsd/-/tree/main/src?ref_type=heads and add localectl to it?

q66:
yes, and ideally upstream it, but that can come later (let's fork it downstream first and i'll review the code)

Gnarwhal:
Ok I believe I've got it working.
https://gitlab.com/Gnaaarwhal/openrc-settingsd/

q66:
thanks, i'll review it this weekend

q66:
sorry for taking ages, i went through the source this evening and cleaned a lot of it up; i kept your commit authorship and added myself as co-authored

i'd like to get this upstreamed so we should coordinate this somehow

Gnarwhal:
No worries at all :)
Thank you for taking the time to review <3

Getting it upstreamed would be nice, but I'll be the first to admit I don't know what that process looks like.

PR #2500, opened 2024-07-14
IIO sensor proxy

JamiKettunen:
Running tests for `iio-sensor-proxy` with `-Dtests=true` requires at least `umockdev` and it'd probably be nicer to have this under `main/` for the auto-soft-deps on GNOME at least.

As a part of adding `iio-sensor-proxy-meta` to `plasma-desktop-meta` deps also make it easier to mask out potentially unwanted plasma apps `-meta` packages basically turning .e.g.
```
apk add plasma-desktop '!'plasma-desktop{,-apps,-multimedia,-games,-accessibility,-kdepim,-devtools}-meta
```
into just:
```
apk add plasma-desktop '!plasma-desktop-meta'
```
While on my Minisforum V3 tablet I still don't have automatic rotation (yet) working due to missing accelerometer sensor from driver side I think it's fair enough to add it as default for KDE Plasma too even though it doesn't seem to e.g. have automatic brightness as an option which I could successfully test in GDM/GNOME already; someone else could try this too instead :)

Marking as a draft for now because for some reason after installing `plasma-desktop` I've not seen ALS work in GDM/GNOME despite `monitor-sensor` actively claiming otherwise...

### DIFF
--- a/main/iio-sensor-proxy/files/iio-sensor-proxy
+++ b/main/iio-sensor-proxy/files/iio-sensor-proxy
@@ -1,0 +1,5 @@
+# iio-sensor-proxy service
+
+type = process
+command = /usr/libexec/iio-sensor-proxy
+depends-on = dbus
--- a/main/iio-sensor-proxy/template.py
+++ b/main/iio-sensor-proxy/template.py
@@ -1,0 +1,45 @@
+pkgname = "iio-sensor-proxy"
+pkgver = "3.5"
+pkgrel = 0
+build_style = "meson"
+configure_args = [
+    "-Dgeoclue-user=_geoclue",
+    "-Dsystemdsystemunitdir=",
+    "-Dtests=true",
+]
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+]
+makedepends = [
+    "glib-devel",
+    "libgudev-devel",
+    "linux-headers",
+    "polkit-devel",
+    "udev-devel",
+]
+checkdepends = [
+    "python-dbusmock",
+    "python-gobject",
+    "python-psutil",
+    "umockdev",
+]
+install_if = [f"iio-sensor-proxy-meta={pkgver}-r{pkgrel}"]
+pkgdesc = "IIO accelerometer sensor to input device proxy"
+maintainer = "Jami Kettunen <jami.kettunen@protonmail.com>"
+license = "GPL-3.0-or-later"
+url = "https://gitlab.freedesktop.org/hadess/iio-sensor-proxy"
+source = f"{url}/-/archive/{pkgver}/iio-sensor-proxy-{pkgver}.tar.gz"
+sha256 = "8689425f2287626a95d95b1e1e5b62e497d09dd08cf411084ed22166d4a49da5"
+hardening = ["vis", "cfi"]
+
+
+def post_install(self):
+    self.install_service(self.files_path / "iio-sensor-proxy")
+
+
+@subpackage("iio-sensor-proxy-meta")
+def _(self):
+    self.subdesc = "recommends package"
+    self.options = ["empty"]
+    return []
--- a/contrib/plasma-desktop/template.py
+++ b/contrib/plasma-desktop/template.py
@@ -1,6 +1,6 @@
 pkgname = "plasma-desktop"
 pkgver = "6.1.4"
-pkgrel = 0
+pkgrel = 1
 build_style = "cmake"
 # FIXME: missing layout memory xml file? QTemporaryFile broken?
 make_check_args = ["-E", "kcm-keyboard-keyboard_memory_persister_test"]
@@ -151,7 +151,7 @@ def _(self):
         # non-kde, misc integrations
         "desktop-file-utils",
         "fprintd-meta",
-        # "iio-sensor-proxy",  # FIXME: package and test on device with accelerometer
+        "iio-sensor-proxy-meta",
         "power-profiles-daemon-meta",  # battery power saving
     ]
     self.options = ["empty"]
@@ -177,7 +177,7 @@ def _(self):
 @subpackage("plasma-desktop-apps-meta")
 def _(self):
     self.subdesc = "apps recommends package"
-    self.install_if = [self.parent]
+    self.install_if = [self.with_pkgver("plasma-desktop-meta")]
     self.depends = [
         # - core
         "discover",  # extra app management
@@ -253,7 +253,7 @@ def _(self):
 @subpackage("plasma-desktop-multimedia-meta")
 def _(self):
     self.subdesc = "multimedia recommends package"
-    self.install_if = [self.parent]
+    self.install_if = [self.with_pkgver("plasma-desktop-meta")]
     self.depends = [
         "audiocd-kio",  # kio plugin for audio cds
         "audiotube",  # youtube music client
@@ -272,7 +272,7 @@ def _(self):
 @subpackage("plasma-desktop-devtools-meta")
 def _(self):
     self.subdesc = "devtools recommends package"
-    self.install_if = [self.parent]
+    self.install_if = [self.with_pkgver("plasma-desktop-meta")]
     self.depends = [
         "heaptrack",
         "kcachegrind",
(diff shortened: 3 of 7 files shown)


nekopsykose:
> Running tests for iio-sensor-proxy with -Dtests=true requires at least umockdev

well, move it then

> Marking as a draft for now because for some reason after installing plasma-desktop I've not seen ALS work

do any of those location services work anymore?

main/iio-sensor-proxy/files/iio-sensor-proxy
```diff
@@ -0,0 +1,6 @@
+# iio-sensor-proxy service
+
+type            = process
+command         = /usr/libexec/iio-sensor-proxy
+depends-on      = local.target
+smooth-recovery = true
```
nekopsykose, line 6:
```suggestion
type = process
command = /usr/libexec/iio-sensor-proxy
depends-on = local.target
smooth-recovery = true
```

main/iio-sensor-proxy/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "linux-headers",
+    "polkit-devel",
+    "udev-devel",
+]
+install_if = [f"iio-sensor-proxy-meta={pkgver}-r{pkgrel}"]
+pkgdesc = "IIO accelerometer sensor to input device proxy"
+maintainer = "Jami Kettunen <jami.kettunen@protonmail.com>"
+license = "GPL-3.0-or-later"
+url = "https://gitlab.freedesktop.org/hadess/iio-sensor-proxy"
+source = f"https://gitlab.freedesktop.org/hadess/iio-sensor-proxy/-/archive/{pkgver}/iio-sensor-proxy-{pkgver}.tar.gz"
```
triallax, line 25:
```suggestion
source = f"{url}/-/archive/{pkgver}/iio-sensor-proxy-{pkgver}.tar.gz"
```

main/iio-sensor-proxy/template.py
```diff
@@ -0,0 +1,45 @@
...
+maintainer = "Jami Kettunen <jami.kettunen@protonmail.com>"
+license = "GPL-3.0-or-later"
+url = "https://gitlab.freedesktop.org/hadess/iio-sensor-proxy"
+source = f"{url}/-/archive/{pkgver}/iio-sensor-proxy-{pkgver}.tar.gz"
+sha256 = "8689425f2287626a95d95b1e1e5b62e497d09dd08cf411084ed22166d4a49da5"
+hardening = ["vis", "cfi"]
+
+
+def post_install(self):
+    self.install_service(self.files_path / "iio-sensor-proxy", enable=True)
```
nekopsykose, line 38: should it be enable=true?

main/gdm/template.py
```diff
@@ -55,6 +55,7 @@
     "gnome-settings-daemon",
     "gnome-shell",
     "gsettings-desktop-schemas",
+    "iio-sensor-proxy-meta",
```
nekopsykose, line 58: does gdm use it? i couldn't find anything for it

nekopsykose: (in contrast to g-c-c which looks stuff up in net.hadess.SensorProxy, etc)

JamiKettunen: ambient light sensor certainly reacts for me there and adjusts the brightness, that was kinda my basis for adding it there since I still can't test auto-rotation until my accelerometer sensor on the tablet is functional on linux

nekopsykose: sure, but it might not do it directly? there are zero references to 'sensor' in gdm

main/iio-sensor-proxy/template.py
JamiKettunen: idk, ideally it'd be dbus activated similar to https://gitlab.freedesktop.org/hadess/iio-sensor-proxy/-/blob/master/data/iio-sensor-proxy.service.in, ig until we have that with `dbus-broker` and the like sensor functionality should need `doas dinitctl enable iio-sensor-proxy`

main/gdm/template.py
nekopsykose: from what i can tell it's g-s-d that does things like that and probably automatically changes the brightness in settings

main/iio-sensor-proxy/template.py
nekopsykose: you can do that by adding a dbus service file (since they don't ship one) and putting dinitctl start in it (see some other stuff that does that)

nekopsykose: i would guess one isn't shipped because the udev rules file adds a systemd wants for the service, the dbus= systemd type is just for readiness notification

(and that has the effect of only starting it if sensors were detected on systemd)

best we can do is just add a normal dbus service that calls dinitctl start when something looks up the name and hence autostarts it

main/gdm/template.py
JamiKettunen: sure I'll move it to be a dep of `gnome-settings-daemon` instead

main/iio-sensor-proxy/template.py
JamiKettunen: made it a triggered service that's enabled by default (on desktops it'll show up as `STARTING` and makes `doas dinitctl list` look meh) where the udev rule does `dinitctl trigger iio-sensor-proxy` to actually launch it when sensors are detected similar to systemd.

alternatively it could just be a default-disabled service and have the udev rule `dinitctl start iio-sensor-proxy` on detected sensors but not sure which is more preferred

nekopsykose:
turns out the start from udev wont work because it runs very early and it would be cyclic with the service to be started since it depends on dbus, but that early dbus is not there yet, etc

so i think for now we can't actually do that. could leave the service disabled by default (and no patch to .rules), since i can't think of a way to make it work

JamiKettunen:
the service starts just fine on my tablet though even without `smooth-recovery = true` so :shrug: works perfectly fine (explicitly started) on my end and doesn't start in a VM (service not loaded) for example as expected due to no found sensors

nekopsykose:
hmm, wonder how that happens

JamiKettunen:
accelerometer sensor now works on the tablet, and I can finally reproduce boot dying at `early-udev-settle` timeout upon installing `iio-sensor-proxy`, I'll maybe play around with it a bit and if I can't figure out anything "smart" it shall be a manually enabled service

JamiKettunen:
I'm not entirely happy with how this ended up but oh well. I think it's about time to get it in even if it requires manually `doas udevadm trigger && doas dinitctl enable iio-sensor-proxy` post-install, probably worth documenting in wiki since it's not necessarily a good default behavior for those who actually have e.g. an accelerometer for auto-rotate and are wondering why "it's not working" in GNOME/KDE at least where `iio-sensor-proxy` is pulled in by default by the `-meta` pkg and I've confirmed auto-rotate to work in both on my tablet (and in GNOME/GDM also ambient light sensor though with weird handling of the once-a-second reported values).

q66:
you should re-draft it, i'll be adding a mechanism for triggering services from udev into dinit-chimera soon

main/iio-sensor-proxy/files/iio-sensor-proxy
```diff
@@ -0,0 +1,5 @@
+# iio-sensor-proxy service
+
+type = process
+command = /usr/libexec/iio-sensor-proxy
+depends-on = dbus
```
nekopsykose, line 5: is this enough? i think you also need local.target but no idea

q66: it's technically enough because you get the local.target dependency through dbus, but is always good to be explicit anyway

JamiKettunen: the PR currently contains my locally working setup yes

nekopsykose: ah, sure, just saw elogind had local.target on itself too

JamiKettunen:
integrated as a part of https://github.com/chimera-linux/cports/compare/6f65fa9c0f98...1ccd9a33c664

PR #2504, opened 2024-07-15
contrib/nushell: new package (0.95.0)

jcgruenhage:
- **contrib/nushell: new package (0.95.0)**: There's a patch in here for autoloading. The patch is upstream, but not yet released. I've tested it locally, it seems to work fine as far as I can tell.
- **cbuild: split nucomp**: I've done this based on 0744645a09d3b4aff93fda6665db9e248f8ef993 and 8536d97eb4bc1eaad7c6af9a33e5b329856896d5, and while I think I understand what's happening there for the most part, I'd appreciate another look from someone else here.
- **contrib/minijinja-cli: update to 2.0.3**: Only included as an example of a tool that I know has nushell completions, allowing to try this out easily.

### DIFF
--- a/contrib/nushell/patches/autoload.patch
+++ b/contrib/nushell/patches/autoload.patch
@@ -1,0 +1,228 @@
+From f241110005006f10e16c6080ebab7abec6f3e5ce Mon Sep 17 00:00:00 2001
+From: Darren Schroeder <343840+fdncred@users.noreply.github.com>
+Date: Tue, 25 Jun 2024 20:31:54 -0500
+Subject: [PATCH] implement autoloading (#13217)
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+# Description
+
+This PR implements script or module autoloading. It does this by finding
+the `$nu.vendor-autoload-dir`, lists the contents and sorts them by file
+name. These files are evaluated in that order.
+
+To see what's going on, you can use `--log-level warn`
+```
+❯ cargo r -- --log-level warn
+    Finished dev [unoptimized + debuginfo] target(s) in 0.58s
+     Running `target\debug\nu.exe --log-level warn`
+2024-06-24 09:23:20.494 PM [WARN ] nu::config_files: set_config_path() cwd: "C:\\Users\\fdncred\\source\\repos\\nushell", default_config: config.nu, key: config-path, config_file_specified: None
+2024-06-24 09:23:20.495 PM [WARN ] nu::config_files: set_config_path() cwd: "C:\\Users\\fdncred\\source\\repos\\nushell", default_config: env.nu, key: env-path, config_file_specified: None
+2024-06-24 09:23:20.629 PM [WARN ] nu::config_files: setup_config() config_file_specified: None, env_file_specified: None, login: false
+2024-06-24 09:23:20.660 PM [WARN ] nu::config_files: read_config_file() config_file_specified: None, is_env_config: true
+Hello, from env.nu
+2024-06-24 09:23:20.679 PM [WARN ] nu::config_files: read_config_file() config_file_specified: None, is_env_config: false
+Hello, from config.nu
+Hello, from defs.nu
+Activating Microsoft Visual Studio environment.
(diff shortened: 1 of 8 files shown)


src/cbuild/core/template.py
```diff
@@ -2125,6 +2137,7 @@ def take_progs(self, man="18"):
         self.take("usr/share/zsh", missing_ok=True)
         self.take("usr/share/fish/completions", missing_ok=True)
         self.take("usr/share/fish/vendor_completions.d", missing_ok=True)
+        self.take("usr/share/nushell/vendor/autoload", missing_ok=True)
```
jcgruenhage, line 2140: This bit specifically I don't quite understand: Why are we taking completions in `take_progs`, but also in the various completions packages?

contrib/nushell/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "nushell"
+pkgver = "0.95.0"
+pkgrel = 0
+build_style = "cargo"
+make_build_env = {
+    "PREFIX": "/usr",
+}
+make_install_env = make_build_env
```
jcgruenhage, line 8: If we don't explicitly set this env var during compile time, nushell will use `/usr/local` as a prefix, resulting in autoloading of completions happening from `/usr/local/share/nushell/vendor/autoload` instead of `/usr/share/nushell/vendor/autoload`

q66: you can just make it `make_env` single variable

src/cbuild/core/template.py
q66: the completions packages are automatic, i.e. if `foo-progs` takes them, you will then get a `foo-progs-nucomp` package

q66: that said, i don't think anything at all right now even provides these?

jcgruenhage: ahh, okay, that makes sense. Thanks :)

jcgruenhage: > that said, i don't think anything at all right now even provides these?

No, nothing ships nushell completions at the moment, but that's probably due to there not being a place to put them yet. Completions themselves do exist already. The previous way to use nushell completions was manually loading every single one of them from your config file, and that didn't even support globbing or anything like that, you manually had to list every single one of them.

I've included a patch here from upstream that's not released yet but already merged, which supports loading these from a system directory.

contrib/nushell/template.py
```diff
@@ -0,0 +1,22 @@
+pkgname = "nushell"
+pkgver = "0.95.0"
+pkgrel = 0
+build_style = "cargo"
+make_env = {
+    "PREFIX": "/usr",
+}
+hostmakedepends = ["cargo-auditable", "pkgconf"]
+makedepends = ["openssl-devel", "libgit2-devel", "zstd-devel", "sqlite-devel"]
```
triallax, line 9: please sort

contrib/nushell/patches/autoload.patch
```diff
@@ -0,0 +1,228 @@
+From f241110005006f10e16c6080ebab7abec6f3e5ce Mon Sep 17 00:00:00 2001
+From: Darren Schroeder <343840+fdncred@users.noreply.github.com>
+Date: Tue, 25 Jun 2024 20:31:54 -0500
+Subject: [PATCH] implement autoloading (#13217)
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+# Description
```
triallax, line 9: the pr template here is really annoying me but idk

jcgruenhage: If you want me to drop it I'll do that, I just fetched the patch from github as is

triallax: it's ok

contrib/nushell/template.py
jcgruenhage: done

jcgruenhage:
I've added a separate commit that includes patches likely to land upstream that also touch the automatic loading of completions. I've kept it in a separate commit because I'm not sure whether we want to include these in chimera yet or just wait for the next nushell release which will likely include them.

PR #2507, opened 2024-07-15
user/clevis: new package (20)

natthias:
Cross compiled on all t1 and t2 architectures.
Tang stuff remains completely untested, the dinit service also does not work.
With clevis, tpm2 luks2 decryption has been tested, most tests are also just skipped.

It might also make sense to split up some of the clevis stuff


Natthias:
Moved to user, adjusted to recent cports changes

natthias:
One minor issue that is still present is that in the local-bottom script it will loop through all network interfaces and attempt to bring them down. This causes an `error: ipv4 FIV table does not exist` to be printed, but the error is harmless.
Not sure exactly how to deal with this. On way is to do what the debain dropbear package does and make the interfaces it brings up/down configurable

natterangell:
Nice work! Let me know if you need help testing anything. I'm only really after the tpm unlock functionality, although I can see the value in the tang pin as well.

PR #2508, opened 2024-07-15
main/openrc-settingsd: fixed localed reading kbd-model map from wrong location

Gnarwhal:
I think `openrc-settingsd` is mistakenly being appended twice. `meson.build` sets `PKGDATADIR` to `/usr/share/openrc-settingsd` and installs `kbd-model-map` there. Then localed looks for `kbd-model-map` in `PKGDATADIR "/openrc-settingsd"` which obviously it doesn't find. This looks like it might be an upstream issue too, but I couldn't say that for certain.


triallax:
needs relbump

Gnarwhal:
Oh yes sorry. One day I'll remember this ;-;

nekopsykose:
pkgdatadir is datadir/package (i.e. usr/share + some name), so yeah looks like an upstream issue

PR #2512, opened 2024-07-17
contrib/forgejo: new package (7.0.5)

triallax:


contrib/forgejo/files/forgejo
```diff
@@ -0,0 +1,8 @@
+type = process
+command = /usr/bin/forgejo web
```
triallax, line 2: i've seen other distros pass `-c /etc/forgejo/app.ini` here but that doesn't seem necessary at all

also might want to depend on postgres here

triallax (commented):
forgot this but i should also update the default values in the example config since we patch them

contrib/forgejo/files/forgejo
nekopsykose: it can use any db including sqlite so not sure why it would depend on postgres

nekopsykose: most you can put without also installing postgres is after =

contrib/forgejo/template.py
```diff
@@ -0,0 +1,54 @@
...
+makedepends = ["linux-pam-devel"]
+depends = ["git", "git-lfs"]
+pkgdesc = "Git forge"
+maintainer = "triallax <triallax@tutanota.com>"
+license = "MIT"
+url = "https://forgejo.org"
+source = f"https://codeberg.org/forgejo/forgejo/archive/v{pkgver}.tar.gz"
+sha256 = "9b949f1f501911e278a709d78c885c411ad76ed1031888d106fae539086ce021"
+env = {
+    "EXTRA_GOFLAGS": "-trimpath -buildmode=pie",
```
nekopsykose, line 19: you probably want to append to this with self.get_goflags (and then you don't need the buildmode)

why trimpath isn't there and is instead in the util/ itself i have no idea..

contrib/forgejo/files/forgejo
triallax: i was using the term "depend" loosely here, i was thinking more about `after` like you mention

PR #2515, opened 2024-07-17
DRAFT: linux 6.10

q66:
Preliminary packaging for 6.10.


q66:
To be merged once testing proves it good and we're at ~6.10.2.

sidkshatriya:
(suggestion; will break out to issue if suggestion is found useful)

Currently `linux-stable` means 6.9.x and after this change it will mean 6.10.x

How about creating a separate package definition for `linux-6.9`, `linux-6.10` etc. Then symlink `linux-stable` to `linux-6.9`. When linux stable becomes linux 6.10 then you can simply symlink `linux-stable` to the `linux-6.10` package definition.

Why ? 

Sometimes you may want to keep around linux 6.9 series on your machine and maintain it yourself locally by updating `template.py` yourself. This workflow will simplify that.

q66:
that's not really how cbuild works, and we have no plans to package the individual series separately

PR #2518, opened 2024-07-17
contrib/strongswan: new package (5.9.14)

rbgarga:


contrib/strongswan/template.py
```diff
@@ -0,0 +1,112 @@
...
+pkgdesc = "Open Source IKEv2 IPsec-based VPN solution"
+maintainer = "Renato Botelho do Couto <renato@netgate.com>"
+license = "GPL-2.0-or-later"
+url = "https://www.strongswan.org"
+source = f"https://download.strongswan.org/{pkgname}-{pkgver}.tar.bz2"
+sha256 = "728027ddda4cb34c67c4cec97d3ddb8c274edfbabdaeecf7e74693b54fc33678"
+
+
+def post_install(self):
+    for d in [
```
q66, line 85: these should all be tmpfiles.d

contrib/strongswan/template.py
```diff
@@ -0,0 +1,112 @@
+pkgname = "strongswan"
+pkgver = "5.9.14"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--libexecdir=/usr/lib",
+    "--with-ipsecdir=/usr/lib/strongswan",
+    "--with-capabilities=libcap",
+    "--with-user=_ipsec",
```
q66, line 9: does it make sense to have a "generic" user name for this instead of something like `_strongswan`?

(i don't know, please explain your thoughts here)

rbgarga: No reason at all.  I'm going to change it to `_strongswan`

contrib/strongswan/files/sysusers.conf
```diff
@@ -0,0 +1,5 @@
+# Create _strongswan system user
+
+u _strongswan - "ipsec user" /var/empty /usr/bin/nologin
+g _strongswan - "ipsec group"
+m _strongswan _strongswan
```
triallax, line 5: i don't think these two lines are needed, the first one should automatically create a _strongswan group and add the _strongswan user to it

triallax (changes requested):

contrib/strongswan/files/sysusers.conf
rbgarga: You are right.  Fixed.

PR #2520, opened 2024-07-17
contrib/helix: specify runtime installation dir explicitly

tranzystorekk:
cc @wezm

Feels like the dir itself should be `/usr/lib/helix/runtime` since there is no binary there anymore, but I didn't want to do preemptive changes.

### DIFF
--- a/contrib/helix/template.py
+++ b/contrib/helix/template.py
@@ -1,6 +1,6 @@
 pkgname = "helix"
 pkgver = "24.07"
-pkgrel = 0
+pkgrel = 1
 build_style = "cargo"
 hostmakedepends = ["cargo-auditable", "git"]
 makedepends = ["rust-std"]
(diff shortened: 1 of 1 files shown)


triallax:
i think moving it to /usr/lib/helix/runtime should be fine, but maybe others can chime in as well

wezm:
> i think moving it to /usr/lib/helix/runtime should be fine, but maybe others can chime in as well

There's still at least one binary in the runtime dir (`tutor`) so maybe it should stay as libexec?

wezm (approved):

triallax:
i don't see any tutor binary there? there's /usr/libexec/helix/runtime/tutor but it's just a text file

tranzystorekk:
yeah tutor is just tutorial text that is opened as a temporary copy

wezm:
> yeah tutor is just tutorial text that is opened as a temporary buffer

Oh my mistake 🤦‍♂️

PR #2524, opened 2024-07-18
contrib/openvpn, contrib/networkmanager-openvpn: new packages (2.6.12, 1.12.0)

Gnarwhal:


contrib/networkmanager-openvpn/template.py
```diff
@@ -0,0 +1,39 @@
+pkgname = "networkmanager-openvpn"
+pkgver = "1.12.0"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--disable-static",
+    "--enable-lto=yes",
```
nekopsykose, line 7:
```suggestion
```

this doesn't do anything except maybe replace -flto=thin with -flto (didn't check final order) which is not what we want

contrib/openvpn/template.py
```diff
@@ -0,0 +1,46 @@
...
+maintainer = "Gnarwhal <git.aspect893@passmail.net>"
+license = "GPL-2.0-only WITH openvpn-openssl-exception"
+url = "https://openvpn.net"
+source = f"https://github.com/OpenVPN/openvpn/releases/download/v{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "1c610fddeb686e34f1367c347e027e418e07523a10f4d8ce4a2c2af2f61a1929"
+
+
+def post_install(self):
+    self.chmod(self.files_path / "update-resolv-conf", 0o744)
+    self.install_file(self.files_path / "update-resolv-conf", "etc/openvpn")
```
nekopsykose, line 44:
```suggestion
    self.install_file(self.files_path / "update-resolv-conf", "etc/openvpn", mode=0o744)
```

contrib/openvpn/template.py
```diff
@@ -0,0 +1,46 @@
...
+    "lzo-devel",
+    "lz4-devel",
+    "libnl-devel",
+    "libcap-ng-devel",
+    "linux-headers",
+    "iproute2",
+]
+checkdepends = [
+    "cmocka-devel",
+]
```
nekopsykose, line 32: this is makedepends

contrib/openvpn/template.py
```diff
@@ -0,0 +1,46 @@
...
+depends = ["iproute2"]
+pkgdesc = "Open source VPN daemon"
+maintainer = "Gnarwhal <git.aspect893@passmail.net>"
+license = "GPL-2.0-only WITH openvpn-openssl-exception"
+url = "https://openvpn.net"
+source = f"https://github.com/OpenVPN/openvpn/releases/download/v{pkgver}/{pkgname}-{pkgver}.tar.gz"
```
nekopsykose, line 38:
```suggestion
source = f"https://github.com/OpenVPN/openvpn/releases/download/v{pkgver}/openvpn-{pkgver}.tar.gz"
```

contrib/openvpn/files/update-resolv-conf
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bash
```
nekopsykose, line 1: this is bash but you didn't add depends=bash

you could port the script too

contrib/openvpn/template.py
```diff
@@ -0,0 +1,46 @@
...
+    "lzo-devel",
+    "lz4-devel",
+    "libnl-devel",
+    "libcap-ng-devel",
+    "linux-headers",
+    "linux-pam-devel",
+    "openssl-devel",
+    "pcre2-devel",
+]
+checkdepends = []
```
nekopsykose, line 31:
```suggestion
```

contrib/networkmanager-openvpn/template.py
```diff
@@ -0,0 +1,38 @@
...
+maintainer = "Gnarwhal <git.aspect893@passmail.net>"
+license = "GPL-2.0-or-later"
+url = "https://github.com/NetworkManager/NetworkManager-openvpn/tree/main"
+source = f"https://github.com/NetworkManager/NetworkManager-openvpn/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "def0fe3f8b118fe44bb61fce15da784f6b1560bfc03d1ac609aa57b3618ac1b7"
+options = ["linkundefver"]
+
+
+def post_install(self):
+    self.install_license("COPYING")
```
nekopsykose, line 37:
```suggestion
```

gpl doesn't have to be installed

contrib/openvpn/template.py
```diff
@@ -0,0 +1,46 @@
...
+make_dir = "."
+hostmakedepends = [
+    "automake",
+    "pkgconf",
+    "libtool",
+    "python",
+]
+makedepends = [
+    "cmocka-devel",
+    "iproute2",
```
nekopsykose, line 21: this looks like hostmakedepends if it checks for it

contrib/openvpn/files/update-resolv-conf
q66: that script could be reasonably easily rewritten without using bash and probably shorter and cleaner (e.g. it doesn't need to check where resolvconf is etc.)

nekopsykose: > When you say port it you mean to sh?

yeah, and you have to actually make some changes for that since it's not valid sh as-is

contrib/networkmanager-openvpn/files/sysusers.conf
```diff
@@ -0,0 +1,3 @@
+# Create networkmanager-openvpn system user
+
+u _nm_openvpn - "networkmanager-openvpn user" /var/lib/nm_openvpn /usr/bin/nologin
```
triallax, line 3: doesn't the `/var/lib/nm_openvpn` bit need a corresponding tmpfiles entry to create it

contrib/openvpn/files/update-resolv-conf
```diff
@@ -0,0 +1,71 @@
...
+down)
+  $RESOLVCONF -d "${dev}.inet"
+  ;;
+esac
+
+# Workaround / jm@epiclabs.io 
+# force exit with no errors. Due to an apparent conflict with the Network Manager
+# $RESOLVCONF sometimes exits with error code 6 even though it has performed the
+# action correctly and OpenVPN shuts down.
+exit 0
```
triallax, line 71: a bit of a digression but this workaround confuses me? at this point in the script isn't it going to exit with 0 regardless?

contrib/openvpn/files/update-resolv-conf
```diff
@@ -0,0 +1,70 @@
...
+## PATH below. You might need to directly set the path to 'resolvconf'
+## manually if it still doesn't work, i.e.
+RESOLVCONF=/usr/bin/resolvconf
+## export PATH=$PATH:/sbin:/usr/sbin:/bin:/usr/bin
+## RESOLVCONF=$(type -p resolvconf)
+
+case $script_type in
+
+up)
+  for option in $(env | grep "foreign_option_" | sed "s/\w\+=//" | tr "\n" " ") ; do
```
nekopsykose, line 30: this doesn't work because changing all newlines to ' ' means instead of looping lines you now loop 1 value at a time, i.e. for foreign_option_3='dhcp-option DOMAIN be.bnc.ch' instead of `option` being `dhcp-option DOMAIN be.bnc.ch` it loops three times with each of those space separated values

you probably want to remove the `tr` entirely and set `IFS=$'\n'` instead

nekopsykose: or if you don't want to use the new posix extension,

```
IFS='
'
```

nekopsykose: i trust that you actually tested it because i definitely did not

nekopsykose:
this also needs an openvpn service but that looks like a pain in the ass so have fun :D

PR #2528, opened 2024-07-20
main/mesa: update to 24.2

nekopsykose:
[ci skip]


JamiKettunen:
couldn't we now set `_have_llvm = True` for `riscv64` as well after https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/26018 got merged?

https://gitlab.freedesktop.org/mesa/mesa/-/commit/010b2f94 may also be interesting to handle here already

nekopsykose:
should be done

NovaAndrom3da:
Build fails on patching step, the patch in question is rusticl-bindgen.patch. A fixed version of the patch is at https://bpa.st/I7WA

PR #2536, opened 2024-07-22
contrib/rrdtool: library package

joshdorsey:


q66:
what's the reason behind this? it does not save much space (the binaries are small and don't add any dependencies) or anything else

also, this will fail CI (the formatting is wrong) and it adds several other problems (e.g. the makedepends in depends for -devel does not seem correct at all, we don't install GPL licenses, the package names don't follow convention, the commit message is incorrect, etc.)

joshdorsey:
The reasoning is that there wasn't an rrdtool package in contrib when I started working on a dependent package (collectd), and I wanted to reconcile my changes.  I notice there are other `libpackage/package` packages in the repo (e.g. libcurl/curl), and librrd is a separate artifact in the rrdtool build so I assumed it should be mirrored in the same way.  I'm not sure what you mean by the package names not following convention.

I'll remove the license and change the `self.depends += makedepends` if you want to move forward with this, but the rest of your feedback isn't actionable.  The formatting is wrong?  It linted fine, I'm not sure what you mean by that.  I read Usage.md and Packaging.md initially, since they're referenced in the README, so I didn't see the guidance in CONTRIBUTING.md.  I guess this change would be considered a new package?

joshdorsey:
fair enough, the collectd package is building fine without that separation, I just thought it might save some memory.  you're right, not worth it 😄

PR #2541, opened 2024-07-22
*/*: re-enable debug for go packages

triallax:
objcopy failing on ppc64le is not an issue anymore

can also just postpone this to the next go bump global rebuild if you want @nekopsykose


nekopsykose:
last time i tried this it failed on like aarch64 or something instead, but it was last go release

i guess let's see if it works :D

PR #2543, opened 2024-07-22
 main/python-setuptools: add sysconfig overrides for build_ext 

z-erica:
hopefully this isn't too sketch


q66:
i think it's okay

PR #2545, opened 2024-07-22
contrib/hugo: switch to extended build

ttyyls:
https://github.com/gohugoio/hugo#build-from-source


ttyyls:
im not sure how much if at all this would affect cross though. But without extended mode most the hugo templates out there (using sass) are pretty much useless.

This would also bring it inline with how everyone else packages it.

triallax:
can use go_build_tags instead

triallax:
> im not sure how much if at all this would affect cross though

the template is !cross'd anyway, or am i misreading this?

PR #2553, opened 2024-07-23
contrib/obs-pipewire-audio-capture: new package (1.1.5)

Gnarwhal:


contrib/obs-pipewire-audio-capture/template.py
```diff
@@ -0,0 +1,23 @@
...
+    "pkgconf",
+]
+makedepends = [
+    "obs-studio-devel",
+    "pipewire-devel",
+]
+depends = [
+    "obs-studio",
+    "wireplumber",
+]
```
nekopsykose, line 17:
```suggestion
```

obs is already linked and don't need the other one

Gnarwhal: Does having xxx-devel in makedepends automatically put xxx in depends?

nekopsykose: nope

PR #2556, opened 2024-07-24
contrib/rocksdb: new package (9.4.0)

Gnarwhal:
Still a WIP, but there are a couple parts I'm not sure about.
1. rocksdb provides both `make static_lib` and `make shared_lib`. It recommends using `static_lib`, so that is what I have done. But perhaps it would make sense to provide both (shared in `rocksdb` and static in `rocksdb-static`?). If providing both makes sense, is there a cbuild sanctioned method for providing multiple `make_build_target`s?
2. So far as I can tell, the `Makefile` expects `*FLAGS` to be passed indirectly via `EXTRA_*FLAGS`. Is what I've done for this acceptable, or is there a cleaner way to go about this?

### DIFF
--- a/contrib/rocksdb/template.py
+++ b/contrib/rocksdb/template.py
@@ -1,0 +1,65 @@
+pkgname = "rocksdb"
+pkgver = "9.4.0"
+pkgrel = 0
+build_style = "makefile"
+make_cmd = "gmake"
+make_env = {
+    "PORTABLE": "1",
+}
+make_build_target = ""
+make_build_args = [
+    "shared_lib",
+    "static_lib",
+]
+make_build_env = {
+    "DEBUG_LEVEL": "0",
+}
+make_use_env = True
+hostmakedepends = [
+    "bash",
+    "gmake",
+    "perl",
+    "pkgconf",
+]
+makedepends = [
+    "bzip2-devel",
+    "gflags-devel-static",
+    "liburing-devel",
+    "linux-headers",
+    "lz4-devel",
+    "snappy-devel",
+    "zstd-devel",
+    "zlib-ng-compat-devel",
+]
+pkgdesc = "Embeddable, persistent key-value store for fast storage"
+maintainer = "Gnarwhal <git.aspect893@passmail.net>"
+license = "(GPL-2.0-or-later OR Apache-2.0) AND BSD-3-Clause"
+url = "https://rocksdb.org"
+source = f"https://github.com/facebook/rocksdb/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "1f829976aa24b8ba432e156f52c9e0f0bd89c46dc0cc5a9a628ea70571c1551c"
+tool_flags = {
+    "CXXFLAGS": [
+        "-D_GNU_SOURCE",
+        "-fPIC",
+        "-frtti",
+    ],
+}
+# idk
+# options = ["!check"]
+
+
+def init_build(self):
+    self.make_build_args += [
+        "EXTRA_CFLAGS=" + self.get_cflags(shell=True),
+        "EXTRA_LDFLAGS=" + self.get_ldflags(shell=True),
+        "EXTRA_CXXFLAGS=" + self.get_cxxflags(shell=True),
+    ]
+
+
+def post_install(self):
+    self.install_license("LICENSE.leveldb")
+
+
+@subpackage(f"{pkgname}-devel")
+def _devel(self):
+    return self.default_devel()
--- a/contrib/rocksdb-devel
+++ b/contrib/rocksdb-devel
@@ -1,0 +1,1 @@
+rocksdb


nekopsykose:
>  But perhaps it would make sense to provide both 

you can just build both at once by calling both targets (`make_build_target = shared_lib, make_build_args = static_lib`)

contrib/rocksdb/template.py
```diff
@@ -0,0 +1,58 @@
...
+        "NM=" + self.get_tool("NM"),
+        "EXTRA_CFLAGS=" + self.get_cflags(shell=True),
+        "EXTRA_LDFLAGS=" + self.get_ldflags(shell=True),
+        "EXTRA_CXXFLAGS=" + self.get_cxxflags(shell=True),
+    ]
+
+    if self.stage > 0:
+        tool_args.append("OBJDUMP=" + self.get_tool("OBJDUMP"))
+
+    self.make.build(tool_args)
```
nekopsykose, line 54: delete this entire thing and set make_use_env = True and then set `self.make_build_args += extra_` inside init_build

contrib/rocksdb/template.py
```diff
@@ -0,0 +1,58 @@
...
+
+def post_install(self):
+    self.install_license("LICENSE.Apache")
```
nekopsykose, line 57: apache doesn't need to be installed

contrib/rocksdb/template.py
```diff
@@ -0,0 +1,58 @@
...
+    "gflags-devel-static",
+    "linux-headers",
+    "lz4-devel",
+    "snappy-devel",
+    "zstd-devel",
+    "zlib-ng-compat-devel",
+]
+pkgdesc = "Embeddable, persistent key-value store for fast storage"
+maintainer = "Gnarwhal <git.aspect893@passmail.net>"
+license = "GPL-2.0-or-later OR Apache-2.0 OR custom:leveldb"
```
nekopsykose, line 24:
```suggestion
license = "(GPL-2.0-or-later OR Apache-2.0) AND BSD-3-Clause"
```

the leveldb licence is just a normal bsd-3-clause

contrib/rocksdb/template.py
```diff
@@ -0,0 +1,58 @@
...
+url = "https://rocksdb.org"
+source = f"https://github.com/facebook/rocksdb/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "1f829976aa24b8ba432e156f52c9e0f0bd89c46dc0cc5a9a628ea70571c1551c"
+hardening = ["!pie"]
```
nekopsykose, line 28: not sure what this was there for since it doesn't affect anything, but for the shared build to work you'll need

```suggestion
tool_flags = {"CXXFLAGS": ["-fPIC"]}
```

contrib/rocksdb/template.py
```diff
@@ -0,0 +1,58 @@
...
+    ]
+
+    if self.stage > 0:
+        tool_args.append("OBJDUMP=" + self.get_tool("OBJDUMP"))
+
+    self.make.build(tool_args)
+
+def post_install(self):
+    self.install_license("LICENSE.Apache")
+    self.install_license("LICENSE.leveldb")
```
nekopsykose, line 58: you forgot the -devel package

contrib/rocksdb/template.py
nekopsykose: (as in, `self.make_build_args += [f"EXTRA_CFLAGS={self.get_cflags(shell=True)}", ...]`, etc)

nekopsykose:
you'll also need `make_env = {"PORTABLE": "1"}` so it doesn't use -march=native

nekopsykose:
also, put liburing-devel in makedepends and also put `-D_GNU_SOURCE` into the cxxflags tool_flags along with the other thing

nekopsykose:
> you'll also need make_env = {"PORTABLE": "1"} so it doesn't use -march=native

and `"DEBUG_LEVEL": "0"`

Gnarwhal:
Wow! Cheers! Thank you for the help <3

Gnarwhal:
> > But perhaps it would make sense to provide both
> 
> you can just build both at once by calling both targets (`make_build_target = "shared_lib", make_build_args = ["static_lib"]`)

Would putting them both in `make_build_args` be semantically equivalent? It feels slightly strange to me specifying them separately when they are the same type of thing.

nekopsykose:
> Would putting them both in make_build_args be semantically equivalent? It feels slightly strange to me specifying them separately when they are the same type of thing.

yea, you can do target = "" and both in build_args if you want

Gnarwhal:
Building is working, but I'm getting a couple of errors during check that I'm not 100% sure how to resolve. 
1. Running check through cbuild is throwing many `undefined reference due to --no-allow-shlib-undefined`s. chrooting into the build environment and running `gmake -j8 check` with no other args fixes this. Do you have any idea what would cause this?
2. When running check in the chroot, it's throwing errors that I'm guessing are a result of it expecting `/bin/sh` to be symlinked to `bash`. Would `exec_wrappers[("/bin/bash", "sh")]` be a way to solve this?

contrib/rocksdb/template.py
```diff
@@ -0,0 +1,65 @@
...
+    "PORTABLE": "1",
+}
+make_build_target = ""
+make_build_args = [
+    "shared_lib",
+    "static_lib",
+]
+make_build_env = {
+    "DEBUG_LEVEL": "0",
+}
```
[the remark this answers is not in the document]
Gnarwhal: DEBUG_LEVEL has to be 1 when building tests, so it cannot be set for the whole build.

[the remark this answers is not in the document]
Gnarwhal: Should I set DEBUG_LEVEL: 0 in make_env and then set DEBUG_LEVEL: 1 in make_check_env?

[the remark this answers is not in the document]
nekopsykose: ah yea

sure

PR #2576, opened 2024-07-27
cbuild: implement service linter

nekopsykose:


nekopsykose:
not doable rn since needs dinitcheck support for (something i forgot please fill me in)

PR #2581, opened 2024-07-28
experimental work on a -dbg-src split

z-erica:
[snipped outdated notes]

### DIFF
--- a/src/cbuild/hooks/post_install/007_strip_debug.py
+++ b/src/cbuild/hooks/post_install/007_strip_debug.py
@@ -207,6 +207,8 @@ def invoke(pkg):
         srcs = [
             # relative to srcdir
             pkg.rparent.srcdir / source_file,
+            # relative to cbuild cache
+            paths.cbuild_cache() / source_file,
             # go standard library
             pkg.rparent.bldroot_path / "usr/lib/go/src" / source_file,
             # downloaded go source
--- a/src/cbuild/core/profile.py
+++ b/src/cbuild/core/profile.py
@@ -234,7 +234,11 @@ def _get_ldflags(self, tmpl, name, extra_flags, debug, hardening, shell):
 
 
 def _get_rustflags(self, tmpl, name, extra_flags, debug, hardening, shell):
-    bflags = [f"--remap-path-prefix={tmpl.chroot_srcdir}=."]
+    bflags = [
+        f"--remap-path-prefix={tmpl.chroot_srcdir}=.",
+        # cargo places some sources here
+        f"--remap-path-prefix={paths.cbuild_cache()}=.",
+    ]
 
     if self.cross:
         bflags += [


nekopsykose:
> packaging mentions "-fdebug-prefix-map=/builddir/{wrksrc}=. to improve ccache behavior for CFLAGS and CXXFLAGS." which would be wrong

it's intended that it does not remap to /usr or similar, so you shouldn't touch it. 

afaik generating the source list from the root of the builddir without touching the prefix-map would still give you the same source list, and then you can move it to /usr/src/wherever, and that would still work for debuginfod (i think). and if it didn't, it would be just a small debuginfod feature to make it work probably

the side effect is -dbg-src becomes not usable outside of debuginfod, but that's fine. i'd rather not change the prefix at all from current (and i'm pretty sure it does break ccache on every pkgrel increase since the -f arg becomes different)

src/cbuild/hooks/post_install/007_collect_debug_sources.py
```diff
@@ -0,0 +1,64 @@
...
+    if not pkg.rparent.options["debug"] or not pkg.rparent.build_dbg:
+        return
+
+    # find referenced debug sources
+    if (pkg.rparent.bldroot_path / "usr/bin/list-dwarf-srcfiles").exists():
+        # don't do unnecessary work
+        processed = set()
+
+        for dir in ["bin", "lib", "libexec"]:
+            for v in pkg.destdir.glob(f"usr/{dir}/**/*"):
```
nekopsykose, line 63: you could probably make this rglob instead of /**/*

nekopsykose:
tl;dr don't touch any of the prefix mapping at all. and the resulting -dbg-src is probably still usable in lldb with a startup script that uses:

```
(lldb) set list target.source-map
  target.source-map -- Source path remappings apply substitutions to the paths
                       of source files, typically needed to debug from a
                       different host than the one that built the target.  The
                       source-map property consists of an array of pairs, the
                       first element is a path prefix, and the second is its
                       replacement.  The syntax is `prefix1 replacement1
                       prefix2 replacement2...`.  The pairs are checked in
                       order, the first prefix that matches is used, and that
                       prefix is substituted with the replacement.  A common
                       pattern is to use source-map in conjunction with the
                       clang -fdebug-prefix-map flag.  In the build, use
                       `-fdebug-prefix-map=/path/to/build_dir=.` to rewrite the
                       host specific build directory to `.`.  Then for
                       debugging, use `settings set target.source-map .
                       /path/to/local_dir` to convert `.` to a valid local path.
```

so looping on /usr/src/debug/* and adding those paths should hopefully work, though i'm not sure how it would treat '. /usr/1 . /usr/2 . /usr/3' etc (same prefix) in practice for an executable that has a bunch of .so's that might conflict in filename from the root of any given dir.

nekopsykose:
> but notably, the rust stdlib does not get this treatment (since there's nowhere to collect it in from)

this afaik should somehow work with rust-src- which is already the std sources. i remember it working years ago with the rustup toolchains, and that's what they read.

in any case, not something that matters for this handling :)

src/cbuild/hooks/post_install/007_collect_debug_sources.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        .stdout.strip()
+        .splitlines()
+    ):
+        source_file = pathlib.Path(source_file.strip().decode())
+
+        # don't have access to these!
+        if source_file.name in ["<stdin>"]:
+            continue
+
```
nekopsykose, line 26: after removing the prefix changes:

```suggestion
        if source_file.name in ["<stdin>"]:
            continue

		# /usr sourcefiles like /usr/include headers are already provided by -devel packages
		if source_file.parent.is_relative_to("/usr"):
			continue
```

nekopsykose:
> i imagine an edge case here would be two packages that were built against different versions of the same inlined header, but maybe its not critical enough to want to avoid at all costs

it might also just not be supported at all; i didn't check :p  

> yeah, thats what i figured. i wanted to (ideally) make all the paths Just Work but it makes sense if it isnt worth sacrificing ccache for that sake

yeah, i'd say touching the prefix-map is a nonstarter. at least it does not matter at all for debuginfod eventually :)  

> then the hard part becomes how to map the existing source paths to rust-src  

i don't think it needs any mapping; it should work ootb and we don't have to do any -src changes. i'm not sure why it doesn't- maybe the rust-lldb/-gdb plugins are broken? (you are required to use them)

src/cbuild/hooks/post_install/007_collect_debug_sources.py
```diff
@@ -0,0 +1,67 @@
...
+            else:
+                dst = (
+                    pkg.rparent.destdir_base
+                    / f"{pkg.rparent.pkgname}-dbg-src-{pkg.rparent.pkgver}"
+                    / source_file
+                )
+                dst.parent.mkdir(parents=True, exist_ok=True)
+                shutil.copy(src, dst)
+        else:
+            pkg.log_warn(f"unknown debug source file: {source_file}")
```
nekopsykose, line 52: as a guarantee, every single build is going to have (picked from a random exe):

```
./build/../crt/Scrt1.c
./build/../crt/crt1.c
./build/runtimes/builtins-bins/lib/builtins/./compiler-rt/lib/builtins/crtbegin.c
```

this is because these are referenced by the linked `crt` objects (there is another path for static-linked executables, and maybe another for dsos(?)). they will always be missing except in the build that makes the actual crt (i.e. clang-rt-dbg-src(?) for compiler-rt and musl-dbg-src for libc crt). having a warning printed in every single build is not super great so perhaps there should be some ignored filepaths, but only for ignoring this warning (since ignoring the path whole will prevent the -dbg-src from having it)

nekopsykose:
> rust src source paths now look like vendor/tempfile/library/core/src/error.rs. hopefully that works and maps to rust-src somehow

`./vendor/tempfile` maps to the actual `./vendor` folder we have from `cargo vendor` for the `tempfile` crate. that would be placed in `-dbg-src` already and work :)

rust-src is for the stdlib, and those look like

```
./library/core/src/fmt/mod.rs
./library/core/src/ptr/mod.rs
```

^ that is the part that (perhaps doesn't) work currently, and is obviously ignored (or shouldn't warn at least), but fixing it is unrelated to all this as it should already be fine

nekopsykose:
> ^ that is the part that (perhaps doesn't) work currently, and is obviously ignored (or shouldn't warn at least), but fixing it is unrelated to all this as it should already be fine

if i had to guess:

```
    bflags = [f"--remap-path-prefix={tmpl.chroot_srcdir}=."]
```

breaks the paths for stdlib .rlibs, since we apply that for the rust build itself. i think we have to exclude it (or change it) for main/rust so rust .rlibs point to the rust-src paths (since at least there that would be intended)

> no, but, look at the suffix of that path. that's in core

sure. but i think that happens because each crate emits an rlib (fancy .a), and that .a links the rust std rlib, and so they somehow end up saying ./vendor/crate/<stdpath>

maybe fixing just the above thing for rust-std remap (try some hacks like unsetting or whatnot first in main/rust) will meaningfully fix those paths, maybe not

src/cbuild/hooks/post_install/007_collect_debug_sources.py
nekopsykose: probably like

```
elif source_file.name in ["Scrt1.c", "crt1.c", .....]:
    # doesn't exist, but we don't warn these because they come from the toolchain
```

src/cbuild/hooks/post_install/007_collect_debug_sources.py
```diff
@@ -0,0 +1,76 @@
...
+    ):
+        source_file = pathlib.Path(source_file.strip().decode())
+
+        # don't have access to these!
+        if source_file.name in ["<stdin>"]:
+            continue
+
+        # /usr sourcefiles like /usr/include headers are already provided by -devel packages
+        if source_file.is_relative_to("/usr"):
+            continue
```
nekopsykose, line 29: actually i just realised these are already ignored- the below !source_file.is_absolute check will exclude anything starting with /, because of course, so this doesn't do anything

src/cbuild/hooks/post_install/007_collect_debug_sources.py
```diff
@@ -0,0 +1,76 @@
...
+                ]:
+                    pkg.log_warn(f"missing debug source file: {source_file}")
+            else:
+                dst = (
+                    pkg.rparent.destdir_base
+                    / f"{pkg.rparent.pkgname}-dbg-src-{pkg.rparent.pkgver}"
+                    / source_file
+                )
+                dst.parent.mkdir(parents=True, exist_ok=True)
+                shutil.copy(src, dst)
```
nekopsykose, line 59: happy path first (if src.exists():) and use .name and just the stem instead of full paths as suggested (these are not names anyone but the toolchain is using, but even if they are they will just exist and be picked up already... so it only risks silently not-warning for missing which is fine

src/cbuild/hooks/post_install/007_collect_debug_sources.py
nekopsykose: er, i'm dumb, it skips the warning which was the whole point

...sleepy

main/rust/template.py
```diff
@@ -288,7 +288,13 @@ def do_build(self):
     benv["CARGO_HOME"] = str(self.chroot_cwd / ".cargo")
     # we don't want the default cross sysroot here
     with self.profile("target:native"):
-        benv["RUSTFLAGS"] = self.get_rustflags(shell=True)
+        benv["RUSTFLAGS"] = self.get_rustflags(
+            shell=True,
+            # remap debug source paths to what will become rust-src
+            extra_flags=[
+                "--remap-path-prefix=./library=/usr/lib/rustlib/src/rust/library"
+            ],
```
nekopsykose, line 296: i don't think this in itself will work (but i mean, test it lol) because the reason they say ./library in the first place is because of the previous remap. try /only/ removing this entirely first and seeing what it outputs into the std .rlibs and how that changes rebuilding something that links them so we can compare

src/cbuild/hooks/post_install/007_collect_debug_sources.py
```diff
@@ -0,0 +1,71 @@
+from cbuild.core import chroot
+
+import pathlib
+import shutil
+
+
+def _scan_file(pkg, v, processed):
+    # silently ignore invalid files
+    for source_file in (
+        chroot.enter(
```
q66, line 10:
```
chroot.enter(
    "llvm-dwarfdump",
    "--show-sources",
    ...
)

and then we don't need libdwarf/whatever anymore

Comment (nekopsykose) [reply src/cbuild/hooks/post_install/007_collect_debug_sources.py:10]: now that is convenient

**Comment by q66** [comment]
idea: how about we collect source filenames from the split debuginfo rather than pre-split stuff

**Comment by q66** [comment]
what's the problem with that? these are provided by like, pretty much one package, and i don't see much of a reason to scan them in the first place?

**Comment by q66** [comment]
these are

```
$ llvm-dwarfdump --show-sources /usr/lib/*.o|sort
./build/../crt/../ldso/dlstart.c
./build/../crt/Scrt1.c
./build/../crt/crt1.c
./build/../crt/crt1.c
./build/../crt/rcrt1.c
./build/../include/elf.h
./build/../src/internal/dynlink.h
./build/obj/include/bits/alltypes.h
./crt/x86_64/crti.s
./crt/x86_64/crtn.s
```

i don't see these ever contributing anything useful when debugging

**Comment by q66** [comment]
i mean, the whole point of this is to provide sources for stuff we provide debug info for, right

so is there ever any reason to deal with sources that are *not* referenced by debuginfo we ship

**Comment by q66** [comment]
if we're having -dbg depend on -dbg-src is there any point in having -dbg-src in the first place instead of just including stuff in -dbg?

File: src/cbuild/hooks/post_install/007_strip_debug.py [hunk 8]
```diff
@@ -159,5 +161,59 @@ def invoke(pkg):
     else:
         (pkg.destdir / "usr/lib").rmdir()
 
+    # now collect the referenced debug sources
+    for v in ddest.rglob("*"):
+        if v.is_dir():
+            continue
```
nekopsykose, line 167: thought of this while on the train-

the strip currently was made some amount faster by being able to do the intermediate strip with just `llvm-strip <every file>` instead of once per file. llvm-dwarfdump also supports multiple inputs, instead of only one at a time.

you should check what happens with

```
llvm-dwarfdump --show-sources /package/file/1
llvm-dwarfdump --show-sources /package/file/2
llvm-dwarfdump --show-sources /package/file/1 /package/file/2
```

does it work (and just return all the source lists combined)? or, even better, perhaps automatically dedupe them so we have fewer shutil copies below for the same referenced files (since usually the same headers are copied over and over again, etc)?

but as long as it works correctly (we copy the same source files; the return of the command is the same total output), doing this in one call is just better; so use this `for v` loop to build a list of sources with the v.relative_to transform, then pass it at once with `*mylist` instead, of doing it one at a time.

src/cbuild/hooks/post_install/007_strip_debug.py
```diff
@@ -159,5 +161,59 @@ def invoke(pkg):
...
+                    dst = (
+                        ddest
+                        / f"usr/src/debug/{pkg.pkgname}-{pkg.pkgver}-r{pkg.pkgrel}"
+                        / source_file
+                    )
+                    dst.parent.mkdir(parents=True, exist_ok=True)
+                    shutil.copy(src, dst)
+            else:
+                pkg.log_warn(f"unknown debug source file: {source_file}")
+                continue
```
nekopsykose, line 216:
```suggestion
            if source_file.is_absolute():
                pkg.log_warn(f"unknown debug source file: {source_file}")
                continue

            src = pkg.rparent.srcdir / source_file

            if not src.exists():
                # silence warning for crt object
                if source_file.name not in [
                    "Scrt1.c",
                    "crt1.c",
                    "crtbegin.c",
                ]:
                    pkg.log_warn(
                        f"missing debug source file: {source_file}"
                    )
                continue
             
            dst = (
                ddest
                / f"usr/src/debug/{pkg.pkgname}-{pkg.pkgver}-r{pkg.pkgrel}"
                / source_file
            )
            dst.parent.mkdir(parents=True, exist_ok=True)
            shutil.copy(src, dst)
```

el flatto (easier to read this way)

src/cbuild/hooks/post_install/007_strip_debug.py
```diff
@@ -159,5 +161,59 @@ def invoke(pkg):
...
+                else:
+                    dst = (
+                        ddest
+                        / f"usr/src/debug/{pkg.pkgname}-{pkg.pkgver}-r{pkg.pkgrel}"
+                        / source_file
+                    )
+                    dst.parent.mkdir(parents=True, exist_ok=True)
+                    shutil.copy(src, dst)
+            else:
+                pkg.log_warn(f"unknown debug source file: {source_file}")
```
nekopsykose, line 215: it being absolute doesn't make it unknown, perhaps some other message is more interesting

nekopsykose: also forgot; how common is /usr/whatever in referenced debug files? we have a ton of potential warning spam here overall including the above (perhaps build like all the libs that gnome depends on and see what gets output now, etc), and it's the kind of warning spam that you can do nothing about (i.e. maybe is more useful to actually print nothing in these cases unless it is ever an accident that something can reference a /source path, and it was a legit thing that someone might want to fix; but i can't think of any, being silent wouldn't affect anything in that case)

nekopsykose: ok, seems the cargo case uses / for cache

```

0:00:16.947 => WARNING: libdovi-3.3.1-r0: unknown debug source file: /cbuild_cache/cargo/registry/src/index.crates.io-6f17d22bba15001f/bitvec-1.0.1/src/ptr/span.rs
0:00:16.947 => WARNING: libdovi-3.3.1-r0: unknown debug source file: /cbuild_cache/cargo/registry/src/index.crates.io-6f17d22bba15001f/bitvec-1.0.1/src/access.rs
0:00:16.948 => WARNING: libdovi-3.3.1-r0: missing debug source file: vendor/gimli-0.29.0/src/read/str.rs
0:00:16.949 => WARNING: libdovi-3.3.1-r0: unknown debug source file: /cbuild_cache/cargo/registry/src/index.crates.io-6f17d22bba15001f/bitvec-1.0.1/src/vec/ops.rs
0:00:16.949 => WARNING: libdovi-3.3.1-r0: unknown debug source file: /cbuild_cache/cargo/registry/src/index.crates.io-6f17d22bba15001f/crc-3.2.1/src/crc32/slice16.rs
0:00:16.949 => WARNING: libdovi-3.3.1-r0: missing debug source file: vendor/addr2line-0.22.0/src/lazy.rs
0:00:16.950 => WARNING: libdovi-3.3.1-r0: missing debug source file: vendor/gimli-0.29.0/src/common.rs
0:00:16.950 => WARNING: libdovi-3.3.1-r0: missing debug source file: vendor/object-0.36.0/src/pod.rs
0:00:16.950 => WARNING: libdovi-3.3.1-r0: missing debug source file: vendor/gimli-0.29.0/src/leb128.rs
0:00:16.950 => WARNING: libdovi-3.3.1-r0: unknown debug source file: /cbuild_cache/cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.86/src/wrapper.rs
0:00:16.951 => WARNING: libdovi-3.3.1-r0: unknown debug source file: /cbuild_cache/cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.86/src/ensure.rs
0:00:16.951 => WARNING: libdovi-3.3.1-r0: unknown debug source file: /cbuild_cache/cargo/registry/src/index.crates.io-6f17d22bba15001f/bitvec-1.0.1/src/index.rs
0:00:16.951 => WARNING: libdovi-3.3.1-r0: missing debug source file: vendor/object-0.36.0/src/read/read_ref.rs
0:00:16.952 => WARNING: libdovi-3.3.1-r0: missing debug source file: vendor/gimli-0.29.0/src/read/rnglists.rs
0:00:16.952 => WARNING: libdovi-3.3.1-r0: unknown debug source file: /cbuild_cache/cargo/registry/src/index.crates.io-6f17d22bba15001f/bitvec_helpers-3.1.5/src/bitstream_io_impl/bitstream_io_reader.rs
0:00:16.952 => WARNING: libdovi-3.3.1-r0: unknown debug source file: /cbuild_cache/cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.86/src/backtrace.rs
0:00:16.953 => WARNING: libdovi-3.3.1-r0: missing debug source file: vendor/object-0.36.0/src/read/elf/section.rs
0:00:16.953 => WARNING: libdovi-3.3.1-r0: missing debug source file: vendor/object-0.36.0/src/read/mod.rs
```

nekopsykose: and answering my own question; a ton of builds have stuff like  

```
0:00:21.438 => WARNING: libnl-3.10.0-r0: missing debug source file: build/lib/route/cls/ematch_syntax.y
0:00:21.438 => WARNING: libnl-3.10.0-r0: missing debug source file: build/lib/route/cls/ematch_grammar.l
0:00:21.443 => WARNING: libnl-3.10.0-r0: missing debug source file: build/lib/route/pktloc_grammar.l
0:00:21.444 => WARNING: libnl-3.10.0-r0: missing debug source file: build/lib/route/pktloc_syntax.y
```

so just silence these

src/cbuild/hooks/post_install/007_strip_debug.py
nekopsykose: answering my own question; it works, and the lists are concatenated together with no dedup

src/cbuild/hooks/post_install/007_strip_debug.py
nekopsykose: > does that warrant another prefix remap?

a remap of what where how

nekopsykose: those paths exist and justwork if you didn't ignore-/ so i don't think anything has to be remapped

nekopsykose: hm actually no cause it would fail at runtime later

nekopsykose: hm yeah sure sounds good

Issue #290, opened 2023-08-08
git: use INSTALL_SYMLINKS=1


nekopsykose:
currently, the binaries are duplicated twice:  

```
$ l /usr/libexec/git-core/git
-rwxr-xr-x  1 root  root   4.7M Jun  1 20:23 /usr/libexec/git-core/git
$ l /usr/bin/git
-rwxr-xr-x  1 root  root   4.7M Jun  1 20:23 /usr/bin/git
```

obviously, this takes a bunch of space

PR #2589, opened 2024-07-30
user/codeberg-pages: new package

Gnarwhal:


contrib/codeberg-pages/template.py
```diff
@@ -0,0 +1,44 @@
...
+
+    golang.Golang(self).mod_download()
+
+    # Replace the go dep
+    self.do(
+        "go",
+        "mod",
+        "edit",
+        "-replace",
+        f"github.com/mattn/go-sqlite3@v1.14.16={self.chroot_srcdir / 'go-sqlite3-v1.14.22'}",
```
triallax, line 39: i presume you copied this from the forgejo template, there we also apply a patch (`contrib/forgejo/patches/sqlite3-lib.patch`) and pass `libsqlite3` as a go build tag to make the thing use the system sqlite (plus of course adding `sqlite-devel` to `makedepends`) instead of a vendored one, so please do that

triallax (changes requested):

contrib/codeberg-pages/template.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def init_build(self):
+    from cbuild.util import golang
+
+    self.make_env.update(golang.get_go_env(self))
+    self.make_env["EXTRA_GOFLAGS"] = f"{self.get_goflags(shell=True)} -trimpath"
+    # https://github.com/golang/go/issues/64875
+    self.make_env["CGO_ENABLED"] = "1"
+    self.make_env["TAGS"] = "libsqlite3 sqlite sqlite_unlock_notify"
```
triallax, line 53: this stuff is specific to the forgejo template (as it uses a make file for building instead of the go build style), what you need instead is setting
`go_build_tags = ["libsqlite3", "sqlite", "sqlite_unlock_notify"]` and also `env = {"CGO_ENABLED": "1"}`

contrib/codeberg-pages/template.py
```diff
@@ -0,0 +1,55 @@
+pkgname = "codeberg-pages"
+pkgver = "5.1"
+pkgrel = 0
+build_style = "go"
+make_env = {
+    "CGO_ENABLED": "1",
+}
```
triallax, line 7: please use `env` instead of `make_env`, i intentionally specified `env` instead of `make_env` in my review comment, as the go build style doesn't use the latter at all

Gnarwhal:
v5.1 does not include (#290)[https://codeberg.org/Codeberg/pages-server/pulls/290] which fixes (#235)[https://codeberg.org/Codeberg/pages-server/issues/235]. I have it pulling commit past v5.1 to include these changes, but perhaps there is a less uncouth method of getting that PR.

triallax:
>  I tried using patches, but naturally that didn't work because patch happens after prepare (where mod_download is run :/)

that's what prepare_after_patch is for

Gnarwhal:
> that's what prepare_after_patch is for
Good to know, thanks!
I have since found other packages using `_git...` for this use case, so I assume that is the correct way to do it

nekopsykose:
yes, though i'd rather they just made a new release (i don't see anything that would block it)

user/codeberg-pages/template.py
```diff
@@ -0,0 +1,61 @@
+pkgname = "codeberg-pages"
+pkgver = "6.0"
+pkgrel = 0
+build_style = "go"
+make_check_args = [
+    "-skip",
+    # Requires network connection
+    "TestHandlerPerformance",
+]
+hostmakedepends = [
```
q66, line 10: this should be one line, ditto with makedepends

user/codeberg-pages/template.py
```diff
@@ -0,0 +1,61 @@
...
+    f"https://codeberg.org/Codeberg/pages-server/archive/v{pkgver}.tar.gz",
+    "https://github.com/mattn/go-sqlite3/archive/refs/tags/v1.14.24.tar.gz",
+]
+source_paths = [".", "go-sqlite3-patched"]
+sha256 = [
+    "2bdb2943db86338fdfcb452fbb6d84e7a4ad05712492dd564d4e5c84a4c8d750",
+    "8fa3b0b66914ae2dd4ddef9a954f614c5b3eb6ac9d80ee61ae2d08e3178507ec",
+]
+env = {
+    # https://github.com/golang/go/issues/64875
```
q66, line 35: should probably not be set unconditionally

user/codeberg-pages/template.py
```diff
@@ -0,0 +1,61 @@
...
+        "-replace",
+        f"github.com/mattn/go-sqlite3@v1.14.16={self.chroot_srcdir / 'go-sqlite3-patched'}",
+    )
+
+
+def post_install(self):
+    self.install_bin("build/pages", name="codeberg-pages")
+    self.install_license("LICENSE")
+
+    self.install_file(self.files_path / "config.env", "etc/codeberg-pages")
```
q66, line 59: installing stuff to /etc is discouraged now, as is using ad-hoc paths like that

if we really need an env-file (i guess it cannot really be avoided for all the go-adjacent junk that has universally adopted env vars for configuration) it should be set up via a tmpfiles copy from /usr (there are a few other things that do that) and the correct path would be in `/etc/default`

PR #2592, opened 2024-07-30
contrib/tree-sitter-c: update to 0.21.4

flukeeey:


triallax:
0.21.3 is specifically required by neovim at the moment: https://github.com/neovim/neovim/blob/v0.10.1/cmake.deps/deps.txt#L44

flukeeey:
Ah, thanks for the link. I'll close this for now. :+1:

PR #2598, opened 2024-07-31
main/rust: remap debug paths to rust-src

z-erica:


main/rust/template.py
```diff
@@ -76,6 +76,13 @@ def post_patch(self):
     cargo.clear_vendor_checksums(self, "libc-0.2.155")
 
 
+def init_configure(self):
```
q66, line 79: doesn't need to be a hook

nekopsykose: mfs be like 'doesnt need to be x' without saying 'do y'

nekopsykose: tool_flags=

q66: just put it in toplevel tool_flags as i said on irc

main/rust/template.py
```diff
@@ -31,6 +31,12 @@
 url = "https://rust-lang.org"
 source = f"https://static.rust-lang.org/dist/rustc-{pkgver}-src.tar.xz"
 sha256 = "0b9ca1e2e45b8a5f0b58db140af0dc92f8311faeb0ad883c5b71a72c02dc6e80"
+self.tool_flags = {
```
q66, line 34: s/self.//

nekopsykose:
because is updated but not in repo

q66:
because there's 2 billion things queued (incl curl) while the builders are building chromium

nekopsykose:
mfs be like 2 billion when the number is twelve

nekopsykose:
yo can i get 2 billion units of food on my table rn i'm hungry

q66:
nevermind, the libreoffice build is already done :^)

nekopsykose:
where food

PR #2605, opened 2024-08-01
contrib/ddclient: disable automake treating warnings as error

Gnarwhal:
Not sure when this broke, but packaging ddclient started giving
```
automake: warnings are treated as errors
Makefile.am:20: warning: escaping \# comment markers is not portable
autoreconf: error: automake failed with exit status: 1
```


nekopsykose:
i remember what probably broke this and this is prolly the wrong fix but lemme think about it first

nekopsykose:
ah, ok i see, it's just new automake that has more warnings

remove the -Werror instead

PR #2616, opened 2024-08-04
RFC: external allocator for musl

q66:
This splits the scudo patching and stuff into a completely external library. It's probably not good enough yet in this form but it works where I tested it.


main/musl-allocator-scudo/template.py
```diff
@@ -0,0 +1,63 @@
...
+        "report_linux",
+        "string_utils",
+        "timing",
+    ]
+
+    for inp in inps:
+        cxx.invoke([f"{inp}.cpp"], f"{inp}.o", True)
+
+    self.do("ar", "rcs", "musl-scudo.a", *map(lambda v: f"{v}.o", inps))
+
```
nekopsykose, line 60: manual makefile + build_style=makefile instead (if only so it's not serial.. but it would be just as long as all of this to write anyway)

main/musl-allocator-scudo/template.py
```diff
@@ -0,0 +1,63 @@
...
+        "musl-wrappers",
+        "release",
+        "report",
+        "report_linux",
+        "string_utils",
+        "timing",
+    ]
+
+    for inp in inps:
+        cxx.invoke([f"{inp}.cpp"], f"{inp}.o", True)
```
nekopsykose, line 57: this is missing -mcrc32 on x86_64 passed to crc32_hw.cpp like was set before (but only there; on aarch64 it raises the baseline as armv8-a+crc is above the baseline..)

q66:
since it's looking like we'll go with mimalloc instead where i mostly already incorporate this i'll just close it

PR #2617, opened 2024-08-04
user/unrar-free: new package (0.3.0)

ttyyls:


nekopsykose:
is this just for script compat for `unrar`? from the description this is just a libarchive frontend, i.e. it does the same thing `tar xf` does to .rars

nekopsykose:
(it being for script compat is totally fine to be clear, and you can even put it in contrib if you want, but perhaps people didn't know this is identical to bsdtar already)

PR #2618, opened 2024-08-04
contrib/skopeo: new package

yoctozepto:


contrib/skopeo/patches/unphony-bin-skopeo.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/Makefile b/Makefile
```
nekopsykose, line 1: this patch doesn't do anything

contrib/skopeo/template.py
```diff
@@ -0,0 +1,49 @@
...
+build_style = "go"
+# for compatibility with Makefile targets
+make_dir = "bin"
+make_build_args = [
+    "-mod",
+    "vendor",
+    "./cmd/skopeo",
+]
+hostmakedepends = [
+    "bash",
```
nekopsykose, line 13:
```suggestion
```

not used

contrib/skopeo/template.py
```diff
@@ -0,0 +1,49 @@
+pkgname = "skopeo"
+pkgver = "1.6.0"
+pkgrel = 0
+build_style = "go"
+# for compatibility with Makefile targets
+make_dir = "bin"
```
nekopsykose, line 6:
```suggestion
```

not used

contrib/skopeo/template.py
```diff
@@ -0,0 +1,49 @@
...
+    "containers-common",
+]
+pkgdesc = "OCI image and repo manipulation tool"
+maintainer = "Radosław Piliszek <radek@piliszek.it>"
+license = "Apache-2.0"
+url = "https://github.com/containers/skopeo"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "95d63d786e7efda7711fbbad6e37edf53ac767eaa47a63dc56c19bbd95add2cd"
+# NOTE(yoctozepto): assumes awk is gawk; requires docker; probably more...
+options = ["!check"]
```
nekopsykose, line 35: with bash added back though you can do

```suggestion
def do_check(self):
    self.do("gmake", "test-unit-local")
```

contrib/skopeo/template.py
yoctozepto: Sadly not true - their Makefile calls it.

contrib/skopeo/template.py
yoctozepto: Used to create `bin/skopeo` which Makefile expects.

contrib/skopeo/template.py
nekopsykose: it does not call it for docs, install-docs, or install-completions

contrib/skopeo/patches/unphony-bin-skopeo.patch
yoctozepto: It does. It prevents the `bin/skopeo` from being marked `.PHONY` which causes it to rebuild on install actions.

I should have added a comment on top of the patch though!

contrib/skopeo/template.py
yoctozepto: It calls it for the Makefile itself.

contrib/skopeo/patches/unphony-bin-skopeo.patch
nekopsykose: it does not cause it to be rebuilt at all though; the patch does not do anything.

you only call install-docs and install-completions, none of which depend on bin/skopeo, so this doesn't do anything. you are wrong.

yoctozepto:
I will add more comments, fix the version and repeat.

contrib/skopeo/patches/unphony-bin-skopeo.patch
nekopsykose: technically it gets called with the addition of the tests below, but before that this truly does nothing at all

contrib/skopeo/template.py
nekopsykose: yeah, and what do you think that does for the build?

(the answer is nothing)

contrib/skopeo/template.py
nekopsykose: nothing that you wrote here makes the makefile 'expect' this. the docs or install- targets you call don't depend on that.

nekopsykose: in podman where you copied this from this is actually used because the podman install: target is called, which is not the case here

nekopsykose: (and with the addition of the tests below, this does get used, but it was not the case before that)

contrib/skopeo/patches/unphony-bin-skopeo.patch
yoctozepto: > it does not cause it to be rebuilt at all though; the patch does not do anything.
> 
> you only call install-docs and install-completions, none of which depend on bin/skopeo, so this doesn't do anything. you are wrong.

I am not wrong. Please don't accuse others of being wrong without checking. `install-completions` depend on `completions` which depend on `bin/skopeo`....

contrib/skopeo/template.py
yoctozepto: > yeah, and what do you think that does for the build?
> 
> (the answer is nothing, because you don't build any binaries with the makefile)

It kills the error message that comes from gmake trying to call bash uselessly.

contrib/skopeo/template.py
yoctozepto: They do expect this. See the `completions` comment above. Please don't say I'm wrong when you clearly haven't checked for yourself...

contrib/skopeo/patches/unphony-bin-skopeo.patch
nekopsykose: conveniently, this whole time i was reading 1.6.0

contrib/skopeo/template.py
nekopsykose: it is true in 1.16 indeed

contrib/skopeo/patches/unphony-bin-skopeo.patch
yoctozepto: It was the case with 1.6.0 already. 1.16.0 only allowed me to drop the device mapper dependency.

yoctozepto: Ah, I see what you mean with 1.6.0. Indeed, they did not use that with completions.

yoctozepto: Unlucky typo for all this confusion. :S

nekopsykose: in 1.6.0 there is no completions: target and install-completions/install-docs definitely don't depend on this https://github.com/containers/skopeo/blob/e4b79d7741c40434838ece2790be537083fc2a04/Makefile#L169

i admit i'm completely wrong for 1.16 (so all the changes here are fine; don't worry about it), but on 1.6 this was incorrect

nekopsykose: >Unlucky typo for all this confusion. :S

it's my bad too, i should've double checked..

sorry :/

yoctozepto: > > Unlucky typo for all this confusion. :S
> 
> it's my bad too, i should've double checked..
> 
> sorry :/

No worries. You've been very diligent with the review!

contrib/skopeo/template.py
nekopsykose: yeah just throw this in for the few unit tests and this is ready

yoctozepto: I'm looking into it now. Thanks.

yoctozepto: Done. I have checked the other tests once again and adapted the comment accordinly..

yoctozepto:
> I will add more comments, fix the version and repeat.

Done that.

contrib/skopeo/template.py
nekopsykose: thanks! i put it in the correct ordering (the pre/do/post are ordered based on the actual step order) for you

yoctozepto:
I do wonder if these unit tests are useful enough to warrant extra CPU time tbh.

nekopsykose:
sorry again for being confused before, you got everything right from the beginning and it's embarassing looking back at it >_<

yoctozepto:
> sorry again for being confused before, you got everything right from the beginning and it's embarassing looking back at it >_<

FWIW, I am embarassed by that typo which caused all this confusion so I guess we are even. ;-)

nekopsykose:
> I do wonder if these unit tests are useful enough to warrant extra CPU time tbh.

generally i lean towards "yea" since it doesn't really take super long (a small additional fraction; basically rebuilds all of ./cmd but not the deps + fast running) and you never know when they catch something useful

probably over half the testsuites that get ran never detect anything (in cports)... but it's impossible to know ahead of time if perhaps at some point they will, and then might point to some very interesting/important-to-fix/etc breakage that would point to the thing being actually broken that you wouldn't know otherwise.

contrib/skopeo/template.py
yoctozepto: Thanks.

PR #2619, opened 2024-08-04
contrib/*: drop `-mod vendor` from build args

yoctozepto:
With the current package go versions this is the default.

No version bump as there is no change in how build happens.

### DIFF
--- a/contrib/podman/template.py
+++ b/contrib/podman/template.py
@@ -5,8 +5,6 @@ build_style = "go"
 # for install.bin compat
 make_dir = "bin"
 make_build_args = [
-    "-mod",
-    "vendor",
     "./cmd/podman",
     "./cmd/rootlessport",
 ]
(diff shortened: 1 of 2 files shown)


nekopsykose:
ah right, yeah looks good

the commit title should be `contrib/*: drop -mod vendor from build args` instead (we don't have separate tagging or whatnot for things, though maybe people feel like changing the general commit style, but that would be separate later)

yoctozepto:
> ah right, yeah looks good
> 
> the commit title should be `contrib/*: drop -mod vendor from build args` instead (we don't have separate tagging or whatnot for things, though maybe people feel like changing the general commit style, but that would be separate later)

Ok, fixing. I made it up because my initial idea of `contrib/{podman,skopeo}` was an overkill hah.

nekopsykose:
yeah we just use `*` instead of {explicit} usually, nice and short :)

PR #2620, opened 2024-08-04
contributing: commit title stuff

nekopsykose:


nekopsykose:
mostly i just find the new package version supremely pointless, so we could just get rid of it

the second change is just writing down what we do already

q66:
sure

can you change `bsdutils` to some actually relevant (alternatively: generic) example while at it

nekopsykose:
i hope coolproject is generic enough

q66:
yup, perfect

PR #2621, opened 2024-08-04
contrib/skopeo: link against system libsqlite3

yoctozepto:
This is to be on the same page as other container tools from the
same family: buildah and podman. There does not seem to exist
a behavioural change in my (and probably any typical) use.
It seems Debian does it this way too, but Arch does not, so
it is strictly a matter of builder's preference.


nekopsykose:
oh yeah, i missed this myself, thanks :3

yoctozepto:
> oh yeah, i missed this myself, thanks :3

The pleasure is all mine! :-)

PR #2622, opened 2024-08-04
contrib/gonic: new package (0.16.4)

saagraa:


q66:
please format your template with black or ruff

contrib/gonic/template.py
```diff
@@ -0,0 +1,15 @@
+pkgname = "gonic"
+pkgver = "0.16.4"
+pkgrel = 0
+build_style = "go"
+make_build_args = [ "cmd/gonic/gonic.go" ]
+hostmakedepends = [ "go" ]
+makedepends = [ "sqlite-devel", "taglib-devel", "pkgconf" ]
```
nekopsykose, line 7: pkgconf is hostmakedepends

nekopsykose:
this needs a dinit service for it too pretty sure

contrib/gonic/template.py
nekopsykose: you also need go_build_tags = ["libsqlite3"] for sqlite-devel to be used

contrib/gonic/template.py
```diff
@@ -0,0 +1,15 @@
...
+pkgdesc = "Static site generator"
```
nekopsykose, line 8:
```suggestion
pkgdesc = "Subsonic-compatible music streaming server"
```

saagraa:
> this needs a dinit service for it too pretty sure

Should I match the path upstream uses for config (/etc/gonic/config) and leave that file/path nonexistent? A default config doesn't make too much sense for this I think.

nekopsykose:
that sounds like it would make sense, sure

nekopsykose:
> Edit: upstream has a default config, I'll install that

it does, and you could make the `command =` pass -config-path to it, but that default config doesn't work ootb so idk if that makes sense to just put in /etc/ or not

nekopsykose:
for tmpfiles/sysusers, you can also use the upstream contrib one as reference, just name the user `_gonic` instead

contrib/gonic/files/gonic
```diff
@@ -0,0 +1,5 @@
+type = process
+command = /usr/bin/gonic -config-path /etc/gonic/config
+depends-on = network.target
```
nekopsykose, line 3: network.target is only supplementary; you want to also add depends-on = local.target

contrib/gonic/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "gonic"
+pkgver = "0.16.4"
+pkgrel = 0
+build_style = "go"
+make_build_args = ["cmd/gonic/gonic.go"]
```
nekopsykose, line 5:
```suggestion
make_build_args = ["./cmd/gonic"]
```

contrib/gonic/template.py
```diff
@@ -0,0 +1,23 @@
...
+makedepends = ["sqlite-devel", "taglib-devel"]
+go_build_tags = ["libsqlite3"]
+pkgdesc = "Music streaming server / subsonic server API implementation"
+maintainer = "samg <samg@fastmail.com>"
+license = "GPL-3.0-only"
+url = "https://github.com/sentriz/gonic"
+source = f"https://github.com/sentriz/gonic/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "ba690a861a075bdf3f1f7e8131e45a5ca430ec90902a97faf955ec9f36799461"
+# tests require network access, transcoding test requires ffmpeg
+options = ["!check", "!cross"]
```
nekopsykose, line 16:
```suggestion
# taglib gomod can't cross
options = ["!cross"]
```

nekopsykose: (the default go test ./... seems to work fine)

contrib/gonic/template.py
```diff
@@ -0,0 +1,23 @@
...
+build_style = "go"
+make_build_args = ["cmd/gonic/gonic.go"]
+hostmakedepends = ["go", "pkgconf"]
+makedepends = ["sqlite-devel", "taglib-devel"]
+go_build_tags = ["libsqlite3"]
+pkgdesc = "Music streaming server / subsonic server API implementation"
+maintainer = "samg <samg@fastmail.com>"
+license = "GPL-3.0-only"
+url = "https://github.com/sentriz/gonic"
+source = f"https://github.com/sentriz/gonic/archive/refs/tags/v{pkgver}.tar.gz"
```
nekopsykose, line 13:
```suggestion
source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
```

contrib/gonic/template.py
saagraa: I'm getting a test failure in when building in the sandbox but not outside of it; still trying to sort it out

nekopsykose: i'm not sure how you're testing that but just running a regular `pkg` with this works fine for me, just push it and see what CI says

nekopsykose: hm ok seems to fail, why did it work for me...

nekopsykose: and it definitely runs the same failing test for me locally

```
ok  	go.senan.xyz/gonic/server/ctrlsubsonic	0.101s
```

nothing in the files seem to e.g. do something different based on CI in env or whatnot so i'm very confused..

nekopsykose: aand now it works

so i guess that's actually a flake that only sometimes happens, oh well

PR #2628, opened 2024-08-05
main/bubblewrap: restore all tests

yoctozepto:
No version bump as the pkg does not change.


yoctozepto:
On these notes:

1) The portability patch could be upstreamed by omitting the disablement of seccomp test.
2) I could not get the python binding for seccomp to work. If anyone wants to pick this effort up, the binding is part of libseccomp itself, enabled by the `--enable-python` switch after having installed `python-cython` and `python-devel`. I believe I lack cython and/or legacy packaging experience to fix the resulting build.

yoctozepto:
cc @q66

nekopsykose:
> The only test remaining disabled is the one for seccomp because
chimera lacks python bindings for libseccomp.

added

yoctozepto:
Oh, lovely, @nekopsykose! TYVM

Let me try it.

yoctozepto:
It works!

PR #2630, opened 2024-08-05
main/less: reenable tests

triallax:
submitting as a pr because of this line that comes from running the tests:

```
sh: lessecho: not found
```

the tests still pass, but i'm not sure where this error is coming from

### DIFF
--- a/main/less/template.py
+++ b/main/less/template.py
@@ -1,6 +1,6 @@
 pkgname = "less"
 pkgver = "661"
-pkgrel = 0
+pkgrel = 1
 build_style = "gnu_configure"
 configure_args = ["--with-regex=posix"]
 make_cmd = "gmake"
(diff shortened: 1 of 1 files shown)


nekopsykose:
the test that ''tries invoke it'' is

```
/builddir/less-661/lesstest/lesstest  -e  -s '/builddir/less-661/lesstest/lt_screen' -t '/builddir/less-661/lesstest/lt/utf8-2.txt.lt' '/builddir/less-661/less'
```

and some part of less itself under some conditions that internally call `lglob` use lessecho (which is invoked via a shell wrapper from less, i.e. `sh -c 'lessecho ...'`), so the spawned shell itself then reports that since it's not in path.

there's no way to work around this since `lesstest` drops the whole env, even PATH before executing the actual test... so it's impossible for the spawned shell to find anything. so this is.. intentional? nothing to do :)

nekopsykose:
this code is nigh unreadable, so please send your prayers

triallax:
your prayers

PR #2635, opened 2024-08-06
user/tarsnap: new package

na-sa-do:
Not sure it's proper to put the sample config in /etc.

### DIFF
--- a/user/tarsnap/template.py
+++ b/user/tarsnap/template.py
@@ -1,0 +1,20 @@
+pkgname = "tarsnap"
+pkgver = "1.0.40"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_gen = []
+makedepends = [
+    "e2fsprogs-devel",
+    "openssl-devel",
+    "zlib-ng-compat-devel",
+]
+pkgdesc = "Online backups for the truly paranoid"
+maintainer = "Nasado <hi@nasado.name>"
+license = "custom:tarsnap"
+url = "https://www.tarsnap.com"
+source = f"{url}/download/tarsnap-autoconf-{pkgver}.tgz"
+sha256 = "bccae5380c1c1d6be25dccfb7c2eaa8364ba3401aafaee61e3c5574203c27fd5"
+
+
+def post_install(self):
+    self.install_license("COPYING")
--- a/user/tarsnap/patches/tarsnap-1.0.40-strict-aliasing-fix.patch
+++ b/user/tarsnap/patches/tarsnap-1.0.40-strict-aliasing-fix.patch
@@ -1,0 +1,158 @@
+https://github.com/Tarsnap/tarsnap/commit/ca40c06f290fb8298dc2e583303d45b58878f37b
+https://github.com/Tarsnap/tarsnap/commit/4af6d8350ab53d0f1f3104ce3d9072c2d5f9ef7a
+
+From 4af6d8350ab53d0f1f3104ce3d9072c2d5f9ef7a Mon Sep 17 00:00:00 2001
+From: Graham Percival <gperciva@tarsnap.com>
+Date: Fri, 1 Apr 2022 16:58:43 -0700
+Subject: [PATCH] scrypt: Fix strict aliasing
+
+The original scrypt code treated its data as blobs of bytes, accessing
+them in whatever manner was convenient from time to time:
+
+* as 32-bit words or vectors thereof for the purpose of the Salsa20/8 core
+* in machine-word-sized chunks (aka. size_t) for block copy and xor operations
+* as 32-bit words for the Integerify function.
+
+This worked fine at the time, but newer compilers apply strict aliasing rules
+which allow them to assume that e.g. data accessed as a uint32_t is not the
+same as data accessed as a size_t, resulting in miscompilation.
+
+Note that in recent versions of scrypt (after 2015-07-18; versions 1.2.0 and
+later) such miscompilation should be detected by the built-in runtime testing.
+
+To avoid aliasing problems, the generic scrypt code now operates on uint32_t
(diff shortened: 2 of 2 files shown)


user/tarsnap/template.py
```diff
@@ -0,0 +1,27 @@
+pkgname = "tarsnap"
+pkgver = "1.0.40"
+pkgrel = 0
+build_style = "configure"
```
nekopsykose, line 4: this is gnu_configure

user/tarsnap/template.py
```diff
@@ -0,0 +1,27 @@
...
+source = f"{url}/download/tarsnap-autoconf-{pkgver}.tgz"
+sha256 = "bccae5380c1c1d6be25dccfb7c2eaa8364ba3401aafaee61e3c5574203c27fd5"
+
+
+def post_install(self):
+    self.mkdir(self.destdir / "etc")
+    self.mv(
+        self.destdir / "usr/etc/tarsnap.conf.sample",
+        self.destdir / "etc/tarsnap.conf.sample",
+    )
```
nekopsykose, line 26: self.rename

nekopsykose: though this should be in usr/share/doc or something

nekopsykose:
pretty sure this needs restricted = "non-redistributable" since

```
Redistribution and use in source and binary forms, without modification,
is permitted for the sole purpose of using the "tarsnap" backup service
provided by Tarsnap Backup Inc.
```

we would be 'redistributing' and we cannot guarantee that at all

user/tarsnap/template.py
```diff
@@ -0,0 +1,27 @@
...
+pkgver = "1.0.40"
+pkgrel = 0
+build_style = "configure"
+configure_args = [
+    "--prefix=/usr",
+]
+makedepends = [
+    "openssl-devel",
+    "zlib-ng-compat-devel",
+    "e2fsprogs-devel",
```
nekopsykose, line 11: sort

user/tarsnap/template.py
```diff
@@ -0,0 +1,27 @@
...
+pkgdesc = "Online backups for the truly paranoid"
+maintainer = "Nasado <hi@nasado.name>"
+license = "custom:tarsnap"
+url = "https://www.tarsnap.com"
+source = f"{url}/download/tarsnap-autoconf-{pkgver}.tgz"
+sha256 = "bccae5380c1c1d6be25dccfb7c2eaa8364ba3401aafaee61e3c5574203c27fd5"
+
+
+def post_install(self):
+    self.mkdir(self.destdir / "etc")
```
nekopsykose, line 22: self.install_dir (in general)

user/tarsnap/template.py
nekopsykose: although that would also make it not installed in main package so not sure either..

user/tarsnap/template.py
na-sa-do: are you sure? it builds fine as plain configure, but gnu_configure makes it fail with "configure.ac is required"

nekopsykose: that happens because the default of gnu_configure is to also run autoreconf (which gives you that error), and that requires the actual autotools inputs. since this is a very proprietary project, they removed even those (even though their actual source tree most definitely has the originals)

so what you want to also add is `configure_gen = []` to skip the reconf step- but the ./configure that is there is a gnu_configure script so you don't need to pass the --prefix=/usr. sorry if that wasn't clear :)

na-sa-do:
re: the license, I think you're being a little paranoid, honestly. if you insist I'll add it, but tarsnap is already in a bunch of other distros and [officially acknowledges them on the website](https://www.tarsnap.com/download.html) which seems to mean they're fine with it.

nekopsykose:
> I think you're being a little paranoid, honestly. if you insist I'll add it, but tarsnap is already in a bunch of other distros and [officially acknowledges them on the website](https://www.tarsnap.com/download.html)

> Debian-based (including Ubuntu and Mint)

this lists their own third-party repo (it's not in upstream debian)

> Redhat-based Linux (including Fedora and CentOS)

this lists a third-party copr created by random users; what they link to is actually a search url on the copr page and not a package, even

> Gentoo Linux

gentoo compiles from source on the user machine, so there aren't any distribution concerns.

for the official gentoo bindist, this is not present on it

> SUSE and openSUSE Linux

i don't know anything about suse packaging; it looks like they actually have a package for this, but i have no idea how it works (i.e. if it's actually an official distro package or not), since it's under experimental/community

> Arch Linux
> nixos
> Slackware Linux
> *bsd

yep, actually shipped here

> void linux

marked as restricted

so.. overall a wash :) go ahead and mark it as restricted

na-sa-do:
alright alright. but how official they are to the other distros isn't really relevant -- the important thing is that they mention them on their website, which tells me at least that tarsnap doesn't mind them existing.

nekopsykose:
sure; i guess that's fair enough and makes sense actually and i didn't think of it for some reason..

user/tarsnap/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "tarsnap"
+pkgver = "1.0.40"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_gen = []
+hostmakedepends = [
+    "autoconf",
+]
```
nekopsykose, line 8: i think you don't need this here either (i.e. without reconfing you don't need any hostdeps)

na-sa-do: oh true

nekopsykose:
and i think you'll want https://gitweb.gentoo.org/repo/gentoo.git/tree/app-backup/tarsnap/files/tarsnap-1.0.40-strict-aliasing-fix.patch too

PR #2636, opened 2024-08-06
cbuild: update some autoconf cache

nekopsykose:
some sorted, and strcasestr/c++ ns skipped

a few more values imported from OE

also, move the bulk of common-linux gl_* to musl-linux so it skips most gl garbage even in native instead of only cross. we source musl for both


q66:
the files just come directly from OE (iirc i haven't made any changes to them? or maybe there is something minor), we should probably not be modifying them

nekopsykose:
they seem somewhat different and not the same (though i only looked for like 1 year of changes, maybe they got changed in like the past 4 years)...

i can double check later, if they're actually unmodified then i'd update them and source common- in native too for the same effect then (?) and still add the strcasestr thing as an extra (that one isn't upstream)

at least i don't see any downside to setting these since it's always gonna be correct

q66:
i think i may have removed some shell constructs to shrimplify them and make them ez to parse by cbuild? don't rember now

PR #2639, opened 2024-08-06
contrib/crane: new package

yoctozepto:


contrib/crane/template.py
```diff
@@ -0,0 +1,43 @@
+pkgname = "crane"
+pkgver = "0.20.1"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
```
q66, line 5: `make_build_args = ["./cmd/crane"]`

contrib/crane/template.py
```diff
@@ -0,0 +1,43 @@
...
+    "./cmd/crane",
+]
+hostmakedepends = [
```
q66, line 8: `hostmakedepends = ["go"]`

contrib/crane/template.py
```diff
@@ -0,0 +1,43 @@
...
+]
+pkgdesc = "Container image and registry manipulation tool"
+maintainer = "Radosław Piliszek <radek@piliszek.it>"
+license = "Apache-2.0"
+url = "https://github.com/google/go-containerregistry"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "4bae53f34011e35ef874a60123b8ae70a5e992d804decb030479dbb888afe6d1"
+
+
+_supported_shells = ["bash", "zsh", "fish"]
```
q66, line 19: drop and just put in the for loops

contrib/crane/template.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+# NOTE(yoctozepto): docs are present but they do not render properly
```
q66, line 22: overly verbose comment, also we don't do the `NOTE(whatever)` stuff since it's practically useless

contrib/crane/template.py
```diff
@@ -0,0 +1,43 @@
...
+# as manpages after using go-md2man, so skipping them. Built-in
+# command help has the same content as is typical in cobra-based go programs.
+def post_build(self):
+    for shell in _supported_shells:
+        self.do(
```
q66, line 27: you don't need sh, just `.do()` the `crane` binary

contrib/crane/template.py
```diff
@@ -0,0 +1,43 @@
...
+            "sh",
+            "-c",
+            f"{self.make_dir}/crane completion {shell} > crane.{shell}comp",
+        )
+
+
+def pre_check(self):
+    # NOTE(yoctozepto): This particular test is marked broken on arm64 and darwin.
```
q66, line 35: ditto

contrib/crane/template.py
q66: e.g.

```
for shell in ["bash", "zsh", "fish"]:
    with open(self.cwd / f"crane.{shell}", "w") as cf:
        self.do(
            "./crane",
            "completion",
            shell,
            stdout=cf,
        )

contrib/crane/template.py
yoctozepto: ok, but note `black` does not care and the current version is better for diffing changes

contrib/crane/template.py
yoctozepto: ok, but note `black` does not care and the current version is better for diffing changes

contrib/crane/template.py
yoctozepto: why? it would not be DRY then

contrib/crane/template.py
q66: it looks worse, and the "diffing changes" is so inconsequential and unlikely to ever matter in any way that it might as well not be a thing

contrib/crane/template.py
yoctozepto: ok, note it is useful after refactorings when `git blame` stops being that useful; I will also make it more succinct

contrib/crane/template.py
q66: because it increases verbosity and adds no value

contrib/crane/template.py
yoctozepto: will do

contrib/crane/template.py
yoctozepto: ack

contrib/crane/template.py
q66: the chance of that happening and mattering is zero

contrib/crane/template.py
yoctozepto: agree on "looks worse" and "is so inconsequential"; not having a strong opinion here

contrib/crane/template.py
yoctozepto: I'm not entirely convinced on the happening probability but mattering is probably low; I have just explained why I am trained to leave this kind of extras with NOTEs, agreed to remove

contrib/crane/template.py
yoctozepto: duplicated code increases verbosity more but I don't mind that much in this particular situtation

q66: it does not; it results in 3 lines being gone and people not having to look away for context, and when adding to it updating both places is impossible to miss, defeating the entire purpose of DRY

PR #2640, opened 2024-08-06
contrib/autofs: new package (5.1.9)

z-erica:
pending testing

### DIFF
--- a/contrib/autofs/files/automount
+++ b/contrib/autofs/files/automount
@@ -1,0 +1,3 @@
+type = process
+command = /usr/bin/automount -f
+depends-on = early-fs-local.target
--- a/contrib/autofs/patches/musl.patch
+++ b/contrib/autofs/patches/musl.patch
@@ -1,0 +1,54 @@
+fixes:
+```
+../include/hash.h:27:2: error: Wordsize not 32 or 64                                                                                                                 
+   27 | #error Wordsize not 32 or 64                                                                                                                                 
+      |  ^                                                                        
+```
+
+--- a/include/hash.h
++++ b/include/hash.h
+@@ -6,7 +6,7 @@
+ #ifdef __GLIBC__
+ #include <bits/wordsize.h>
+ #else
+-#include <bits/reg.h>
++#include <sys/reg.h>
+ #endif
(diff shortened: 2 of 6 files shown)


contrib/autofs/template.py
```diff
@@ -0,0 +1,45 @@
...
+    "automake",
+    "bison",
+    "flex",
+    "gmake",
+    "heimdal-devel",
+    "libsasl-devel",
+    "libtirpc-devel",
+    "libxml2-devel",
+    "openldap-devel",
+    "openssl-devel",
```
nekopsykose, line 21: this is all makedepends

contrib/autofs/files/automount
```diff
@@ -0,0 +1,4 @@
+# automount service
+
```
nekopsykose, line 2:
```suggestion
```

contrib/autofs/files/automount
```diff
@@ -0,0 +1,4 @@
...
+type = process
+command = /usr/bin/automount -f
```
nekopsykose, line 4: pretty sure this is missing at least some dependencies to work

contrib/autofs/patches/musl.patch
```diff
@@ -0,0 +1,39 @@
...
+      |                                     ^     ~~~~~~~~~~~~~~
+```
+
+--- a/include/automount.h
++++ b/include/automount.h
+@@ -20,6 +20,7 @@
+ #include <unistd.h>
+ #include <fcntl.h>
+ #include <mntent.h>
++#include <libgen.h>
```
q66, line 36: this is not correct, autofs uses basename that takes a constant string while the posix libgen.h version potentially changes its input

note the build warning:

```
/builddir/autofs-5.1.9/daemon/automount.c:2541:31: warning: passing 'const char *' to parameter of type 'char *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
```

contrib/autofs/files/automount
q66: i'd probably put it at early-fs-local.target here since this can run pretty early (at least after everything else is mounted)

PR #2642, opened 2024-08-06
contrib/podman-tui: new package (1.2.0)

ttyyls:


contrib/podman-tui/template.py
```diff
@@ -0,0 +1,20 @@
...
+pkgver = "1.2.0"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go"]
+makedepends = ["linux-headers", "libbtrfs-devel"]
+checkdepends = ["pkgconf", "gpgme-devel"]
+go_build_tags = [
+    "exclude_graphdriver_devicemapper",
+    "exclude_graphdriver_btrfs",
+    "btrfs_noversion",
```
nekopsykose, line 11: these are definitely not meant to be set

nekopsykose: hm well they're set by default, though in alpine they're not..

nekopsykose: let's assume it works

ttyyls: i got them from the project's makefile.

PR #2645, opened 2024-08-07
cbuild: add option to run update checks for a specific maintainer

Gnarwhal:
I have read #2013, but I wasn't sure that it would be high priority and thought this might still be useful in the meantime. If filtering based on who touched in git is still being considered I could also attempt that. Or I can do nothing if it is entirely undesirable ¯\\\_(ツ)\_/¯

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -51,6 +51,7 @@ opt_allowcat = "main contrib user"
 opt_restricted = False
 opt_updatecheck = False
 opt_acceptsum = False
+opt_maintby = None
 opt_maint = "unknown <cports@local>"
 opt_tdata = {}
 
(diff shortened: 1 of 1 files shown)


q66:
considering the maintainer field is practically unused and still due for removal i dunno if i want a new option that is very particular to that field

nekopsykose:
locally i do

```
rg -l -g template.py 'somestringinthetemplatedoesntevenhavetobethemaintainerfield' main contrib user | sort -u | cut -d/ -f1-2 | parallel -j10 ./cbuild update-check
```

i guess it's 'nicer' to technically have a cbuild feature that kinda does it, but this always worked for me (bonus: it's parallel)

and for maintainer-specific there is https://repo.chimera-linux.org/cports-updates/cports-updates.txt , so all this does is let you kinda run it on demand i guess

yoctozepto:
This is limited in usefulness because the current maintainer field is single-entity. One can be interested in keeping arbitrary software up-to-date so an external solution would adapt better.

PR #2646, opened 2024-08-07
main/libseccomp: switch to slibtool

yoctozepto:


main/python-libseccomp/template.py
```diff
@@ -1,7 +1,7 @@
 # update libseccomp alongside this
 pkgname = "python-libseccomp"
 pkgver = "2.5.5"
-pkgrel = 0
+pkgrel = 1
```
nekopsykose, line 4: this doesn't need bumping

yoctozepto: Well, it links again the lib rebuilt in `libseccomp` and the comment reminds to keep the two versions in sync so I guess it is actually better to bump `pkgrel` in sync as well.

q66: it's not

nekopsykose: the version didn't change so there is nothing to keep in sync

q66:
is there any reason to switch this specifically or

main/python-libseccomp/template.py
yoctozepto: ok, it's functionally a noop anyway so I won't be arguing over it

yoctozepto:
> is there any reason to switch this specifically or

Mostly "because we can". libseccomp does not abuse libtool so the speed gain is negligible.

yoctozepto:
> > is there any reason to switch this specifically or
> 
> Mostly "because we can". libseccomp does not abuse libtool so the speed gain is negligible.

And IMHO the more we cover with slibtool, the better we can push it as a better alternative to libtool.

q66:
still doesn't really make sense, are you gonna open a PR switching every single random thing

because i'm not sure if i want that

yoctozepto:
> still doesn't really make sense, are you gonna open a PR switching every single random thing
> 
> because i'm not sure if i want that

Not going for that either, would be bad. I switched what I knew how to validate.

We could have a switching strategy; i.e., try to switch as many as we can without breaking the build, document the failing, let users test whether there aren't any runtime surprises...

nekopsykose:
> We could have a switching strategy; i.e., try to switch as many as we can without breaking the build,

it's something for me personally to work on as i feel like it and i don't need help with it currently (there are some amount of things that fail with it and i have to get each one fixed one at a time, but that's for me); not something that needs any proactive changing anywhere else or whatnot   

overall it doesn't fix/change things, only in a few places (and i already changed it in places where it fixed something where i knew it would). the rest is just incremental for me to do as i work on maintenance myself :)

and, of course, for new things added may as well start with slibtool instead.

yoctozepto:
> > We could have a switching strategy; i.e., try to switch as many as we can without breaking the build,
> 
> it's something for me personally to work on as i feel like it and i don't need help with it currently (there are some amount of things that fail with it and i have to get each one fixed one at a time, but that's for me); not something that needs any proactive changing anywhere else or whatnot
> 
> overall it doesn't fix/change things, only in a few places (and i already changed it in places where it fixed something where i knew it would). the rest is just incremental for me to do as i work on maintenance myself :)

Ack, I get that.

> and, of course, for new things added may as well start with slibtool instead.

Roger that.

q66:
as i see it it's fine to switch things alongside updates but "switch to slibtool and rebuild" is rarely useful on its own because it almost never actually changes anything

nekopsykose:
ah and that yea, when touching it for other reasons can also change it ofc :)

yoctozepto:
yeah, in this case I could have also given up on bumping the pkgrel entirely too because it produces the same happy lib

PR #2647, opened 2024-08-07
packaging: clarify slibtool

yoctozepto:


nekopsykose:
hm yeah sounds good, thanks :3  

i would just not format it with the `**` since it's not that interesting/emphasised (the document uses `**` in like two places for actual text)

nekopsykose:
thank you :3

PR #2649, opened 2024-08-07
contrib/lazydocker: new package (0.23.3)

ttyyls:


contrib/lazydocker/template.py
```diff
@@ -0,0 +1,15 @@
+pkgname = "lazydocker"
+pkgver = "0.23.3"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go"]
+pkgdesc = "Lazier way to manage docker"
```
q66, line 6: that's an undescriptive pkgdesc if i've ever seen one

ttyyls: that's just the repo description though...

q66: there is no rule or precedent saying pkgdesc should be the same as upstream repo description

ttyyls: i wouldn't go so far as to argue from precedants but fwiw a pkgdesc acts simply as a tertiary data point to quickly confirm wether the package one is inspecting from the package manager is the same as the one in mind alongside the name, and version with the url pointing to the homepage if available or docs or upstream repo.

All to say that pkgdesc that get too creative might harm that purpose as i understand it at least, but i digress.

contrib/lazydocker/template.py
```diff
@@ -0,0 +1,15 @@
+pkgname = "lazydocker"
+pkgver = "0.23.3"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go"]
+pkgdesc = "TUI for docker resource visualization and management"
+maintainer = "ttyyls <contact@behri.org>"
+license = "MIT"
+url = "https://github.com/jesseduffield/lazydocker"
+source = f"{url}/archive/refs/tags/v0.23.3.tar.gz"
```
nekopsykose, line 10:
```suggestion
source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
```

PR #2652, opened 2024-08-07
 contrib/debootstrap: new package (1.0.137)

z-erica:
doesnt actually need dpkg lol

PR #2653, opened 2024-08-07
 user/dpkg: new package (1.22.6) 

z-erica:
originally for #2652 before i found out it doesn't need it, but i do have a working template so here goes


q66:
how does this even use gpatch

q66:
also => user

q66:
looks like it's all for the dpkg perl scripts (and dpkg-dev) which happen to be built

but we don't need those (and probably shouldn't ship those), as they're just in general not going to work or be anyhow useful (only the non-dev tools are useful at all in this kind of system)

q66:
e.g. look in https://packages.debian.org/sid/all/dpkg-dev/filelist and https://packages.debian.org/sid/all/libdpkg-perl/filelist , all this stuff is probably useless

q66:
otherwise having gpatch is probably fine (doesn't even need to go in user) but it probably isn't actually used anyhow in this build

q66:
feel free depending on how much you want to deal with this, personally i don't care much

PR #2671, opened 2024-08-09
user/taskwarrior: make patches more upstreamable

jcgruenhage:
Not sure whether it makes sense to merge this into cports, but I've had it lying around anyway to test my patches while working on getting some of them upstream, so I thought I might as well open a PR. If stuff like this shouldn't be PR'd against cports, please let me know.


jcgruenhage:
First patch was already merged upstream so that'll get into the next release, same seems to be happening to the ring update. For the bundling, a CMake option will probably be added as well, but that'll take a bit more time as they are currently changing the way they integrate the rust lib and want to wait until they're done with that bit.

nekopsykose:
seems good :D

PR #2673, opened 2024-08-10
[RFC] main/musl: switch from scudo to mimalloc

q66:
This is a proof of concept integration of mimalloc into musl.

It should be functional (i.e. you should be able to run your desktop with it etc.) but there is currently no guarantee of things not being buggy.

I'd like to reduce the patch if possible, ideally in cooperation with upstream. This would require upstream support for tagging internal APIs, which make up a bulk of the patch.

We also need to tweak the arena size and other parameters to be more suitable for a system allocator. Right now it allocates 1024MB arenas by default which is a bit too much virtual memory for us. The difference between different values needs to be benchmarked.

We also make a hardened build here, which is a bit slower than a regular build but should be worth it to get at least similar level of security to our current setup.


q66:
first issue found: videos with audio fail to play in firefox (e.g. from youtube)

```
[Child 11948, MediaDecoderStateMachine #1] WARNING: Decoder=20efb798f00 Decode error: NS_ERROR_DOM_MEDIA_REMOTE_DECODER_CRASHED_UTILITY_ERR (0x806e0011): file dom/media/MediaDecoderStateMachineBase.cpp:167
```

dkwo:
Just a thought: why not go for https://github.com/GrapheneOS/hardened_malloc ?

q66:
because we aim to make the performance better and integration cleaner rather than the other way around

Drsheppard01:
I'm not sure it's helped but Linux World has experience with change memory allocator for Musl https://docs.bell-sw.com/alpaquita-linux/latest/how-to/malloc/

q66:
> I'm not sure it's helps but in Linux World has experience with change memory allocator for Musl https://docs.bell-sw.com/alpaquita-linux/latest/how-to/malloc/
> 
> Also i found patched Musl from alpaquita https://github.com/bell-sw/alpaquita-aports/tree/stream/core/musl-perf

this is not really interesting for us (don't really need help with that anyway), it's just `LD_PRELOAD` shenanigans

q66:
this is now fully working and near done; there are some more option tweaks incoming, but nothing too signficant, so this is very close to landing

main/musl/patches/0001-plumb-in-support-for-externally-provided-allocator-l.patch
```diff
@@ -0,0 +1,292 @@
...
+ 	$(AS_CMD)
+ 
+@@ -158,11 +168,11 @@ obj/%.lo: $(srcdir)/%.S
+ obj/%.lo: $(srcdir)/%.c $(GENH) $(IMPH)
+ 	$(CC_CMD)
+ 
+-lib/libc.so: $(LOBJS) $(LDSO_OBJS)
++lib/libc.so: $(LOBJS) $(LDSO_OBJS) $(EXTRA_OBJ)
+ 	$(CC) $(CFLAGS_ALL) $(LDFLAGS_ALL) -nostdlib -shared \
+ 	-Wl,-e,_dlstart -o $@ $(LOBJS) $(LDSO_OBJS) $(LIBCC)
```
sidkshatriya, line 89: @q66 shouldn't we have something like 
```
-Wl,-e,_dlstart -o $@ $(LOBJS) $(EXTRA_OBJ) $(LDSO_OBJS) $(LIBCC)
```

here ? You seem to be depending on the lazy and repeated evaluation of `$(LOBJS)`. The `$(LOBJS)` in the rule prerequisite will be different from the `$(LOBJS)` in the rule recipe...

main/musl/patches/0001-plumb-in-support-for-externally-provided-allocator-l.patch
```diff
@@ -0,0 +1,292 @@
...
+ 
+-lib/libc.a: $(AOBJS)
++lib/libc.a: $(AOBJS) $(EXTRA_OBJ)
+ 	rm -f $@
+ 	$(AR) rc $@ $(AOBJS)
```
sidkshatriya, line 94: @q66 Same question here as above comment -- why not something like:
```
$(AR) rc $@ $(AOBJS) $(EXTRA_OBJ)
```

PR #2684, opened 2024-08-13
contrib/worker: update to 5.1.0

triallax:


contrib/worker/template.py
```diff
@@ -1,9 +1,8 @@
 pkgname = "worker"
-pkgver = "5.0.2"
+pkgver = "5.1.0"
 pkgrel = 0
-build_style = "configure"
-configure_args = ["--prefix=/usr"]
-hostmakedepends = ["pkgconf"]
+build_style = "gnu_configure"
+hostmakedepends = ["automake", "libtool", "pkgconf"]
```
triallax, line 5: note: build fails with slibtool

PR #2689, opened 2024-08-13
contrib/openvpn: add service files

Gnarwhal:


contrib/openvpn/files/openvpn-server
```diff
@@ -0,0 +1,7 @@
+# openvpn server service
+
+type = process
+command = /usr/bin/openvpn --status /var/log/openvpn-server-status.log --status-version 2 --suppress-timestamps --log /var/log/openvpn-server-service.log --config ${OPENVPN_SERVER:-default}.conf
+working-dir = /etc/openvpn/server
```
nekopsykose, line 5: having a working-dir in /etc seems weird, is this actually required or just a convenience for --config to be relative?

if the latter then pass full path

contrib/openvpn/files/openvpn-server
```diff
@@ -0,0 +1,7 @@
+# openvpn server service
+
+type = process
+command = /usr/bin/openvpn --status /var/log/openvpn-server-status.log --status-version 2 --suppress-timestamps --log /var/log/openvpn-server-service.log --config ${OPENVPN_SERVER:-default}.conf
```
nekopsykose, line 4: ${} is substituted from env, but there is no way to set an env for a service. you'd probably want env-file= too with that (maybe /etc/default/openssl-{client,server}) and set OPENVPN_{CLIENT,SERVER}_CONFIG=/etc/openvpn/.... in it

nekopsykose: also, is the server-status useful for anything, and why version =2 (there's 3 too, defaults to 1, etc)

contrib/openvpn/files/openvpn-client
```diff
@@ -0,0 +1,7 @@
+# openvpn client service
+
+type = process
+command = /usr/bin/openvpn --suppress-timestamps --nobind --log /var/log/openvpn-client-service.log --config ${OPENVPN_CLIENT:-default}.conf
```
nekopsykose, line 4: --log erases the file every start which is probably not what you want. there's --log-append instead

but you probably want --syslog openvpn-client instead

contrib/openvpn/files/openvpn-server
Gnarwhal: > there is no way to set an env for a service

Does `dinitctl setenv` not work for variable substitution? Or is it a bad idea to rely on it?

Also, the sample services for systemd used `%i` as a method of selecting a VPN tunnel, but I didn't see a convenient analogue for this in dinit. If you think an `env-file` is the best way, then so shall it be. 

> is the server-status useful for anything

Admittedly I have no idea. I've never administrated an OpenVPN server. 

> why version =2

The example service files provided by OpenVPN use version 2, and I assume they know more than I do.

contrib/openvpn/files/openvpn-client
Gnarwhal: As somebody who knows nothing and is only asking out of curiosity, when would one prefer a log file vs the system log?

contrib/openvpn/files/openvpn-server
nekopsykose: > Does dinitctl setenv not work for variable substitution? Or is it a bad idea to rely on it?

that sets it for every service which isn't what you want

>  If you think an env-file is the best way, then so shall it be.

i just don't see any other way really

contrib/openvpn/files/openvpn-client
nekopsykose: we just have no builtin log handling yet so using a log file means it never gets rotated. syslog goes to /var/log/messages and can then be configured to do anything via syslog-ng conf

so generally you'd prefer to send to syslog in this case for now

Gnarwhal: Ah understood thank you for explaining

rbgarga:
https://github.com/davmac314/dinit/pull/386 will help openvpn service

PR #2691, opened 2024-08-14
contrib/bazel: new package

Gnarwhal:


contrib/bazel/patches/cbuild_flags.patch
```diff
@@ -0,0 +1,53 @@
...
+     )
+ 
++    cbuild_flags_feature = feature(
++        name = "cbuild_flags",
++        enabled = True,
++        flag_sets = [
++            flag_set(
++                actions = all_compile_actions,
++                flag_groups = [flag_group(flags = [
++                    "-D_LARGEFILE64_SOURCE",
```
triallax, line 15: we're very unlikely to merge this with the problematic copying of the flags from cbuild  (will easily go out of sync, etc.)

i briefly looked up making the build respect our env vars and found https://github.com/bazelbuild/bazel/issues/1258, which was opened in 2016 and closed by a stalebot, so not very encouraging

Gnarwhal: Fair enough

q66: it could probably be done dynamically but not sure what that would look like

triallax: feel free to give it a try anyway :)

PR #2693, opened 2024-08-14
contrib/whois: avoid conflict with main/expect

ttyyls:


nekopsykose:
if anything the mkpasswd in expect is the worse one

nekopsykose:
(and you can't remove that one either because i'm pretty sure it's relied on by things that use expect)

i would either rename the whois mkpasswd (you'll also have to rename the manpage stuff) or just do nothing

PR #2697, opened 2024-08-14
contrib/openjdk*: use custom targets for bootstrap

triallax:
[ci skip]

### DIFF
--- a/contrib/openjdk17/template.py
+++ b/contrib/openjdk17/template.py
@@ -81,9 +81,6 @@ env = {
     "CBUILD_BYPASS_STRIP_WRAPPER": "1",
 }
 
-# set to True to generate a bootstrap tarball
-_bootstrap = False
-
 # we want this on BE too, and on LE the buildsystem skips it for clang
 # skipping it means generating code for ELFv1 ABI and that does not work
 if self.profile().arch == "ppc64" or self.profile().arch == "ppc64le":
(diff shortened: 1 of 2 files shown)


contrib/openjdk17/template.py
```diff
@@ -126,26 +123,27 @@ def do_configure(self):
...
+    for f in (self.cwd / "build/images/jdk").iterdir():
+        self.cp(f, bdirn, recursive=True)
+    # remove src, we don't need it
+    self.rm(self.cwd / bdirn / "lib/src.zip")
+    # strip libs
+    for f in (self.cwd / bdirn).rglob("*.so"):
+        print("STRIP", f.relative_to(self.cwd))
+        self.do("llvm-strip", f.relative_to(self.cwd))
+    # make an archive
+    self.do("tar", "cvJf", f"{bdirn}.tar.xz", bdirn)
```
nekopsykose, line 140: make this use the zst thing from go/template.py too

nekopsykose: or since that is just really slow for no reason

```
self.do("tar", "cvf", f"{bdirn}.tar", bdirn)
self.do("zstd", f"-T{self.make_jobs}", "-19", "--rm", f"{bdirn}.tar")
```

nekopsykose: hm nvm now is the same, so yeah just copy the same thing

triallax: works?

nekopsykose: yea, tested that locally too

PR #2698, opened 2024-08-14
cbuild: correct check for empty soname in scanelf

triallax:
the original condition evaluates to false for empty strings because they are falsy in python

### DIFF
--- a/src/cbuild/core/scanelf.py
+++ b/src/cbuild/core/scanelf.py
@@ -239,7 +239,7 @@ def _scan_one(fpath):
     inf.close()
 
     # sanitize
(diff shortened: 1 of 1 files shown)


src/cbuild/core/scanelf.py
```diff
@@ -239,7 +239,7 @@ def _scan_one(fpath):
     inf.close()
 
     # sanitize
-    if soname and len(soname) == 0:
+    if soname is not None and len(soname) == 0:
```
triallax, line 242: the change should be correct as long as soname is either `None` or a string, if that's not true and i missed something my bad

PR #2703, opened 2024-08-15
main/clang-rt-crt-cross: add missing gmake dependency

catgirlconspiracy:
not sure how that would end up missing? a transitive dependency that got changed?

### DIFF
--- a/main/clang-rt-crt-cross/template.py
+++ b/main/clang-rt-crt-cross/template.py
@@ -1,7 +1,7 @@
 pkgname = "clang-rt-crt-cross"
 _musl_ver = "1.2.5"
 pkgver = "18.1.8"
-pkgrel = 0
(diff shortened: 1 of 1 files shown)


q66:
hm? it's not needed

q66:
ah, no, it is needed for the musl headers setup

@nekopsykose got a bit overeager dropping stuff in 1565ff08114814c9ff8a27d6fac72ccbe1a0c862 and probably did not notice

q66:
(for next time it is not necessary to increase pkgrel for this kind of changes, it only occured to me now)

catgirlconspiracy:
so i guess that's why it didn't get rebuilt?

q66:
it never needed rebuilding, it's just a build tool switch

PR #2704, opened 2024-08-15
user/gamescope: update to 3.14.29

triallax:


triallax:
couldn't seem to figure out how to test this properly, so would appreciate that

nekopsykose:
the reason i didn't update it is because it doesn't work :P  

from .25 the pointer is inverted in some applications like sauerbraten (without force-grab) and there's the double effect that if you move the cursor out of the window while it's inverted it just segfaults instantly

nekopsykose:
well it works in "not sauerbraten" but is sufficiently scuffed i'd rather just wait until the next round of refactoring fixes that happen to fix it (i test every release)

triallax:
noted

PR #2707, opened 2024-08-15
main/libedit: fix -devel-man conflict with readline-devel-man

triallax:
[update skip]

### DIFF
--- a/main/libedit/template.py
+++ b/main/libedit/template.py
@@ -1,6 +1,6 @@
 pkgname = "libedit"
 pkgver = "20240711"
-pkgrel = 0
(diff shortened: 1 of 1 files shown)


main/libedit/template.py
```diff
@@ -17,6 +17,8 @@
 
 def post_install(self):
     self.install_license("COPYING")
+    # conflicts with readline-devel-man
+    self.uninstall("usr/share/man/man3/history.3")
```
triallax, line 21: could otherwise rename it but idk, there are a ton of manpages and not renaming them also would be inconsistent in that case

PR #2708, opened 2024-08-15
main/eventlog: drop

triallax:
i could also probably make syslog-ng link against system eventlog, but idk


nekopsykose:
i think if things link a vendored copy they should at least do it statically. otherwise in the future eventlog might be readded for some reason and then it just conflicts again

so, either make it link system, or make it link the vendored one without installing a /usr/lib/whatever.so

q66:
the vendoring should be fixed, but this technically does not change anything from current status other than dropping the other package so it's fine for now

nekopsykose:
sure, we can forget about it forever and not fix it

q66:
there

nekopsykose:
ty vm

PR #2711, opened 2024-08-16
contrib/nyacme: fix checksum

z-erica:


triallax:
was this a retag or

triallax:
makes sense, knowing ptrc i wouldn't expect a retag anyway :D

nekopsykose:
is the same forgejo version, just moved from an x86_64 host on alpine to aarch64 on chimera

it doesn't use system zlib or something for zlib-ng so might just be architecture-nondeterministic too in the go compression library or something

nekopsykose:
though rather than confirming 'the files are the same' you should see the actual byte diff

nekopsykose:
ok looked into it- the cached tarballs are gone (obviously), forgejo generates new ones with `git archive` which itself does use system zlib, so zlib-ng has diff checksums (expected)

PR #2714, opened 2024-08-16
contrib/onscripter-en: new package

z-erica:


contrib/onscripter-en/patches/smpeg_register.patch
```diff
@@ -0,0 +1,728 @@
+iso c++17 does not allow register storage class specifier, so yank it indiscriminately
```
nekopsykose, line 1: just link the system one unless this is patched

otherwise -std=c++11 the subproject in buildsystem and you can drop the entire patch; clang just defaults to 17 when unspecified, this is still valid in 11

contrib/onscripter-en/template.py
```diff
@@ -0,0 +1,34 @@
...
+]
+makedepends = [
+    "bzip2-devel",
+    "freetype-devel",
+    "libogg-devel",
+    "libvorbis-devel",
+    "libx11-devel",
+    "sdl1.2_image-devel",
+    # note: refuses to use system SDL_mixer if it's not linked with smpeg
+    # "sdl1.2_mixer-devel",
```
nekopsykose, line 21: link it with smpeg then (i don't see a reason to not do it; i just forgot to ask back when it was added)

nekopsykose: seems to only need patching configure.in to pkg-config smpeg and it would link ok

nekopsykose: well, and also needing smpeg-not-2

PR #2718, opened 2024-08-17
user/puzzles: update to 20240817

EarldridgeJazzedPineda:


triallax:
for future reference, we would prefer that both commits in this pr be only one

nekopsykose:
it's fine as two as well

PR #2719, opened 2024-08-17
UnityX desktop environment

EarldridgeJazzedPineda:
### Please note that I still have more packages to do and that I haven't packaged and compiled the shell itself yet!

List of packages:
* contrib/plotinus (0.2.0)
* contrib/nux (4.0.8)
* contrib/xfce4-windowck-plugin-unityx (0.5.1)

### DIFF
--- a/contrib/plotinus/template.py
+++ b/contrib/plotinus/template.py
@@ -1,0 +1,17 @@
+pkgname = "plotinus"
+pkgver = "0.2.0"
+pkgrel = 0
+build_style = "cmake"
+# Vendored dependency
+cmake_dir = "unityx/plotinus"
+hostmakedepends = ["cmake", "ninja", "pkgconf", "vala"]
+makedepends = ["gtk+3-devel"]
+pkgdesc = "Searchable command palette in every modern GTK+ application"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/p-e-w/plotinus"
+source = "https://gitlab.com/ubuntu-unity/unity-x/unityx/-/archive/main/unityx-main.tar.gz?path=unityx/plotinus>plotinus.tgz"
+sha256 = "22c131b1dce41e5b648a8b9263e15453515ba2141393ebc321a9d1ad1feda3f4"
+hardening = ["vis", "cfi"]
+# No tests
+options = ["!check"]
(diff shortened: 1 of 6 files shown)


contrib/plotinus/template.py
```diff
@@ -0,0 +1,17 @@
...
+build_style = "cmake"
+# Vendored dependency
+cmake_dir = "unityx/plotinus"
+hostmakedepends = ["cmake", "ninja", "pkgconf", "vala"]
+makedepends = ["gtk+3-devel"]
+pkgdesc = "Searchable command palette in every modern GTK+ application"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/p-e-w/plotinus"
+source = "https://gitlab.com/ubuntu-unity/unity-x/unityx/-/archive/main/unityx-main.tar.gz?path=unityx/plotinus>plotinus.tgz"
```
nekopsykose, line 13: you can't package the 'main' branch of things and then pretend it's '0.2.0'; you would at least need a git ref and then to make the version something like 0.2.0_git<somedate>

the `>plotinus.tgz` rename is also wrong, since on any upgrade it would still have the old tarball with the same name cached

contrib/xfce4-windowck-plugin-unityx/template.py
```diff
@@ -0,0 +1,35 @@
...
+    "libxfce4ui-devel",
+    "libxfce4util-devel",
+    "xfce4-panel-devel",
+]
+pkgdesc = "Xfce window controls/title bar panel plugin"
+subdesc = "UnityX version"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-3.0-or-later"
+url = "https://docs.xfce.org/panel-plugins/xfce4-windowck-plugin/start"
+source = "https://gitlab.com/ubuntu-unity/unity-x/unityx/-/archive//main/unityx-main.tar.gz?path=unityx/windowck-plugin>windowck-plugin.tgz"
```
nekopsykose, line 34: same here, and same in nux

if this has no versions at all and you're meant to just use random git commits then it doesn't seem very good to packages

contrib/plotinus/template.py
EarldridgeJazzedPineda: There aren't any version numbers in the source code, all I can do is take the original author's word for it.

contrib/xfce4-windowck-plugin-unityx/template.py
EarldridgeJazzedPineda: There aren't any stable tarballs for UnityX, it didn't even make its way to the Ubuntu repositories yet.

EarldridgeJazzedPineda:
It isn't worth packaging, as it's not yet production-ready.

PR #2722, opened 2024-08-17
contrib/znc: new package

jabuxas:
https://wiki.znc.in/ZNC


contrib/znc/template.py
```diff
@@ -0,0 +1,37 @@
...
+    "argon2-devel",
+    "icu-devel",
+    "icu-libs",
+    "libsasl-devel",
+    "openssl-devel",
+    "python-devel-static",
+    "tcl-devel",
+    "zlib-ng-devel",
+]
+pkgdesc = "Advanced IRC bouncer to disconnect without losing the chat session"
```
nekopsykose, line 28:
```suggestion
pkgdesc = "IRC bouncer with module support"
```

contrib/znc/template.py
```diff
@@ -0,0 +1,37 @@
...
+make_cmd = "gmake"
+hostmakedepends = [
+    "cmake",
+    "gmake",
+    "perl",
+    "pkgconf",
+    "python",
+]
+makedepends = [
+    "argon2-devel",
```
nekopsykose, line 19: also needs boost-devel here and gettext in hostmake

contrib/znc/template.py
```diff
@@ -0,0 +1,37 @@
...
+    "icu-devel",
+    "icu-libs",
+    "libsasl-devel",
+    "openssl-devel",
+    "python-devel-static",
```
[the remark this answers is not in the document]
jabuxas: i didnt try python-devel lol, my only testing was removing it entirely. for some reason i thought python-devel was being pulled automatically by just python

contrib/znc/files/znc.user
```diff
@@ -0,0 +1,4 @@
+type = process
```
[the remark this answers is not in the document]
jabuxas: what is the package for documentation on tmpfiles? it says to look at tmpfiles.d(5) but man 5 tmpfiles.d didnt work

contrib/znc/template.py
jabuxas: i didnt try installing with localization because afaik musl didnt have any locales per se? i thought of the boost-locale something like that package, but that was compiling some other stuff during the compile of  znc itself so i just scrapped the idea

nekopsykose: it does, this is application locales

contrib/znc/files/znc.user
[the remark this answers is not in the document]
nekopsykose: no package holds that because @q66 forgot to add it to sd-utils manpages prolly, but merely googling it gives you the fdo html page for it

its not very useful in this case, since all you need is already shown in the tmpfiles/sysusers files in cports, and most of the features are not used so it would just be confusing

contrib/znc/files/sysusers.conf
```diff
@@ -0,0 +1,3 @@
+# create _znc user
+
+u _znc - "User for ZNC bouncer" /var/empty /bin/sh
```
nekopsykose, line 3: /var/empty should be /var/lib/znc

and /bin/sh should be /usr/bin/nologin unless the service itself uses the user via su scripts or similar; for actual administration you don't need a shall ehere

contrib/znc/files/tmpfiles.conf
```diff
@@ -0,0 +1,3 @@
+# create znc datadir
+
+d /var/lib/znc 0755 znc znc -
```
nekopsykose, line 3: you call this user/group `znc` but you make `_znc` above so this doesn't exist, should be _znc _znc

nekopsykose: and probably 0750

contrib/znc/files/znc
```diff
@@ -0,0 +1,5 @@
+type = process
+command = /usr/bin/znc
+log-type = buffer
```
nekopsykose, line 3: what does it normally log? buffer is super small and overflows really fast, so if it logs any informational non-error messages this should be log-file instead most likely

contrib/znc/files/sysusers.conf
jabuxas: yup i didnt know `su -s`  was a thing

contrib/znc/files/znc
jabuxas: it doesnt log a lot.

```
Checking for list of available modules... 
Opening config [/jbx/.znc/configs/znc.conf]... 
Loading global module [corecaps]... 
Loading global module [webadmin]... 
Binding to port [1040]... 
Loading user [jaba]
Loading user module [chansaver]... 
Loading user module [controlpanel]... 
Forking into the background... [pid: 24658]
ZNC 1.9.1 - https://znc.in
```

contrib/znc/template.py
nekopsykose: the reason it 'builds boost' is because your mr is hundreds of commits behind, and in that state the main/boost is older than the repo so it starts building the old one

jabuxas:
will the service complain for running as _znc even though _znc is /usr/bin/nologin?

contrib/znc/files/znc
nekopsykose: this logs enough to show that it doesn't work at all; the 'fork into background' makes dinit lose track of it instantly and spawn a new one:  

```
    26258 doas dinitctl start znc
      26259 /usr/bin/unix_chkpwd demon chkexpiry
      26260 dinitctl start znc
26261 /usr/bin/znc -d /var/lib/znc
26264 /usr/bin/znc -d /var/lib/znc
26267 /usr/bin/znc -d /var/lib/znc
26269 /usr/bin/znc -d /var/lib/znc
26270 /usr/bin/znc -d /var/lib/znc
26271 /usr/bin/znc -d /var/lib/znc
```

and the final state is stopped because all of them 'died', even if it's running in the end

nekopsykose:
it will not

contrib/znc/files/znc
nekopsykose: you need to pass --foreground to znc

nekopsykose: after that, yeah seems to work

contrib/znc/files/znc
```diff
@@ -0,0 +1,6 @@
+type = process
+command = /usr/bin/znc -d /var/lib/znc
```
nekopsykose, line 2:
```suggestion
command = /usr/bin/znc --datadir /var/lib/znc --foreground
```

jabuxas: so change buffer to log-file?

nekopsykose: if it logs a bunch of stuff as it runs, probably yea

jabuxas: im not sure how much it logs on the foreground and i cant test it right now

nekopsykose: the same as in what you already saw, but idk if it does more as things connect or something

jabuxas:
seems to log quite a bit when the service has been started for a long time, so, switching to log-file

contrib/znc/files/znc
```diff
@@ -0,0 +1,6 @@
+type = process
+command = /usr/bin/znc --datadir /var/lib/znc --foreground
+run-as = _znc
+log-type = log-file
```
nekopsykose, line 4:
```suggestion
log-type = log-file
```
```suggestion
log-file = /var/log/znc.log
```

PR #2724, opened 2024-08-18
main/libsasl: don't uninstall plugins

z-erica:
kmail needs these for imap auth, or else it dies with:
    SASL(-4): no mechanism available: No worthy mechs found


nekopsykose:
so that's what the test failures were

PR #2728, opened 2024-08-18
contrib/knot: new package (0.6.1)

jcgruenhage:
- **contrib/fstrm: new package (0.6.1)**
- **contrib/knot: new package (0.6.1)**


contrib/knot/template.py
```diff
@@ -0,0 +1,54 @@
...
+    self.install_service(self.files_path / "knotd")
+
+@subpackage("knot-devel")
+def _devel(self):
+    return self.default_devel()
+
+
+@subpackage("knot-libs")
+def _libs(self):
+    return ["usr/lib/*.so.*"]
```
nekopsykose, line 45: afaict self.default_libs() does the same for this, so can use that

contrib/knot/template.py
```diff
@@ -0,0 +1,54 @@
...
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--enable-dnstap",
+    "--enable-fastparser",
+    "--with-rundir=/run/knot",
+]
+make_cmd = "gmake"
+hostmakedepends = [
+    "autoconf",
```
nekopsykose, line 12: autoconf is pulled by automake so can remove it here

contrib/knot/template.py
```diff
@@ -0,0 +1,54 @@
+pkgname = "knot"
+pkgver = "3.3.8"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--enable-dnstap",
+    "--enable-fastparser",
```
nekopsykose, line 7: you also want --enable-quic, and then nghttp2+ngtcp2 in makedeps (i'll add the latter right now)

contrib/knot/files/knotd
```diff
@@ -0,0 +1,8 @@
+# knotd service
+
```
nekopsykose, line 2:
```suggestion
```

don't need the top comments anymore for something so simple

contrib/knot/files/knotd
```diff
@@ -0,0 +1,8 @@
...
+type = process
+command = knotd
+depends-on = local.target
+depends-on = network.target
+smooth-recovery = yes
+
```
nekopsykose, line 8: there's an extra trailing newline here

contrib/fstrm/template.py
```diff
@@ -0,0 +1,28 @@
...
+    "gmake"
+]
+makedepends = [
+    "libevent-devel",
+]
+pkgdesc = "C implementation of the Frame Streams data transport protocol"
+maintainer = "Jan Christian Grünhage <jan.christian@gruenhage.xyz>"
+license = "MIT"
+url = "https://github.com/farsightsec/fstrm"
+source = f"https://github.com/farsightsec/fstrm/archive/v{pkgver}.tar.gz"
```
nekopsykose, line 19:
```suggestion
source = f"{url}/archive/v{pkgver}.tar.gz"
```

contrib/knot/files/knotd
```diff
@@ -0,0 +1,8 @@
+# knotd service
+
+type = process
+command = knotd
```
q66, line 4: services always need a full path to executable

contrib/knot/template.py
jcgruenhage: considering that ngtcp2 isn't packaged yet, I'd like to push that out for a follow-up

nekopsykose: it is

nekopsykose: (i said 'right now' and added it 10 minutes later :) )

contrib/fstrm/template.py
```diff
@@ -0,0 +1,24 @@
+pkgname = "fstrm"
+pkgver = "0.6.1"
+pkgrel = 0
+build_style = "gnu_configure"
+make_cmd = "gmake"
+hostmakedepends = ["automake", "libtool", "pkgconf", "gmake"]
```
nekopsykose, line 6:
```suggestion
hostmakedepends = ["automake", "slibtool", "pkgconf", "gmake"]
```

where it works we prefer to start with this now instead, seems to work here

(and doesn't for knot below; i checked)

contrib/knot/template.py
jcgruenhage: indeed.

jcgruenhage: I apparently missed reading the bit in brackets...

nekopsykose:
without a tmpfiles.conf

```
# Create knotd state directories

d /var/lib/knot 0750 _knot _knot -
d /run/knot 0750 _knot _knot -
```

it doesn't even start, since without the latter it fails to write the pidfile (even though it's foregrounded; though i guess one could break it with knotc later), and without the former it errors on the default db dir being missing

with that added it works ok

jcgruenhage:
@nekopsykose /var/lib/knot should not be a temporary file though, right? with https://github.com/chimera-linux/cports/pull/2728/files#diff-982dbc5adbdf10faef2f3c7701c8f6e61b1010c20fced13621ff5ed5ab0683efR3 saying that `/var/lib/knot` is the home dir of the user, I'd have assumed that the home dir would be created?

nekopsykose:
`d /var/lib/knot` in tmpfiles is permanent and not temporary if there is no `age` arg :)  

there's nothing that makes homedirs from sysusers spec, that's why the two are always together (unless tmpfiles is not needed at all and the homedir is /var/empty, but in that case something else is making the varempty)

PR #2734, opened 2024-08-19
main/harfbuzz: document !cross

z-erica:


q66:
that doesn't need documenting, all g-i packages are universally uncrossable

triallax:
> dunno, i feel like that fact should be documented somehow (either individually per g-i package like im doing, or somewhere else globally)

Packaging.md would be nice imo

q66:
i mean in the sense of needing a comment in every single thing using gobject-introspection, we definitely don't want that

if you want to drop a note somewhere else go ahead, but putting it in the template makes no sense unless it's also everywhere else and putting it everywhere is kinda meh

q66:
but tbh i'd rather have a mechanism to make this automatic (i.e. gobject-introspection exists in hostdeps, you have implicit !cross)

but i thought about this before and haven't come up with anything good yet

PR #2739, opened 2024-08-20
 user/protontricks: new package 

z-erica:


q66:
imo user/

user/protontricks/template.py
```diff
@@ -0,0 +1,32 @@
...
+depends = [
+    "virtual:cmd:winetricks!winetricks",
+    "virtual:cmd:zenity!zenity",
+    *_pydeps,
+]
+checkdepends = [
+    "bash",
+    "python-pytest",
+    *_pydeps,
+]
```
nekopsykose, line 26:
```suggestion
depends = [
    "winetricks",
    "zenity",
    <the pydeps i am too lazy to type>
]
checkdepends = [
    "bash",
    "python-pytest",
    *depends,
]
```

a little extra hostdep install via check is ok, saves on needing the whole variable

for the virtual, imo it doesn't make sense here because it's very clearly 'winetricks and zenity' as the deps wholesale

PR #2741, opened 2024-08-20
contrib/ansible: new package

catgirlconspiracy:
(+ dependencies)

### DIFF
--- a/contrib/ansible-core/template.py
+++ b/contrib/ansible-core/template.py
@@ -1,0 +1,38 @@
+pkgname = "ansible-core"
+pkgver = "2.17.3"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+]
+depends = [
+    "python-cryptography",
+    "python-jinja2",
+    "python-packaging",
+    "python-pyyaml",
+    "python-resolvelib",
+]
+checkdepends = [
+    "git",
+    "openssh",
+    "python-pytest",
+    "python-pytest-mock",
+    "python-pytest-xdist",
+    *depends,
+]
+pkgdesc = "Configuration management and multinode orchestration framework"
+subdesc = "core components"
+maintainer = "Mara <177581589+catgirlconspiracy@users.noreply.github.com>"
+license = "GPL-3.0-or-later"
+url = "https://ansible.com"
+source = f"$(PYPI_SITE)/a/ansible-core/ansible_core-{pkgver}.tar.gz"
+sha256 = "917557065339fe36e7078e9bea47eefab6d6877f3bd435fa5f0d766d04c58485"
+
+
+def do_check(self):
+    self.do(
+        "./bin/ansible-test",
+        "units",
+    )
--- a/contrib/ansible/template.py
+++ b/contrib/ansible/template.py
@@ -1,0 +1,37 @@
+pkgname = "ansible"
+pkgver = "10.3.0"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+]
+depends = ["ansible-core"]
+checkdepends = [
+    "python-pytest",
+    "python-pytest-mock",
+    "python-pytest-xdist",
+    *depends,
+]
+pkgdesc = "Configuration management and multinode orchestration framework"
+maintainer = "Mara <177581589+catgirlconspiracy@users.noreply.github.com>"
+license = "GPL-3.0-or-later"
+url = "https://ansible.com"
+source = f"$(PYPI_SITE)/a/ansible/ansible-{pkgver}.tar.gz"
+sha256 = "6144fb4bc785f917f86b1b0b6eadc9b894e9751ff9e9a7875afcfa2f74581ffd"
+# many collections either require additional Python modules, or the tests
+# require dependencies that aren't in the tarball, or they're just broken
+# (possibly due to Python version compat issues)
+options = ["!check"]
+
+
+def do_check(self):
+    for collection_dir in self.find("ansible_collections", "*/*/tests/unit"):
+        collection_dir = collection_dir.parent.parent
+        print(f"checking {collection_dir}")
+        self.do(
+            "ansible-test",
+            "units",
+            wrksrc=self.chroot_cwd / collection_dir,
+        )
--- a/contrib/python-commentjson/template.py
+++ b/contrib/python-commentjson/template.py
@@ -1,0 +1,22 @@
+pkgname = "python-commentjson"
+pkgver = "0.9.0"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+]
+depends = ["python-lark-parser"]
+pkgdesc = "Python parser for JSON with Python/Javascript style inline comments"
+maintainer = "Mara <177581589+catgirlconspiracy@users.noreply.github.com>"
+license = "MIT"
+url = "https://github.com/vaidik/commentjson"
+source = f"$(PYPI_SITE)/c/commentjson/commentjson-{pkgver}.tar.gz"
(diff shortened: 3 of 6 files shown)


contrib/ansible-core/template.py
```diff
@@ -0,0 +1,39 @@
...
+]
+checkdepends = [
+    "git",
+    "openssh",
+    "python-pytest",
+    "python-pytest-mock",
+    "python-pytest-xdist",
+    *_runtime_deps,
+]
+depends = ["python", *_runtime_deps]
```
nekopsykose, line 25:
```suggestion
depends = [
    "python-cryptography",
    "python-jinja2",
    "python-packaging",
    "python-resolvelib",
    "python-pyyaml",
]
checkdepends = [
    "git",
    "openssh",
    "python-pytest",
    "python-pytest-mock",
    "python-pytest-xdist",
    *depends,
]
```

contrib/ansible-core/template.py
```diff
@@ -0,0 +1,39 @@
...
+url = "https://ansible.com"
+source = f"$(PYPI_SITE)/a/ansible-core/ansible_core-{pkgver}.tar.gz"
+sha256 = "917557065339fe36e7078e9bea47eefab6d6877f3bd435fa5f0d766d04c58485"
+
+
+def do_check(self):
+    self.do(
+        self.chroot_cwd / "bin/ansible-test",
+        "units",
+        "--verbose",
```
nekopsykose, line 38:
```suggestion
```

the verbose just makes it extra spammy but actual errors are printed regardless

contrib/ansible-core/template.py
```diff
@@ -0,0 +1,39 @@
...
+maintainer = "Mara <177581589+catgirlconspiracy@users.noreply.github.com>"
+license = "GPL-3.0-or-later"
+url = "https://ansible.com"
+source = f"$(PYPI_SITE)/a/ansible-core/ansible_core-{pkgver}.tar.gz"
+sha256 = "917557065339fe36e7078e9bea47eefab6d6877f3bd435fa5f0d766d04c58485"
+
+
+def do_check(self):
+    self.do(
+        self.chroot_cwd / "bin/ansible-test",
```
nekopsykose, line 36:
```suggestion
        "./bin/ansible-test",
```

contrib/ansible/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "python-installer",
+    "python-setuptools",
+]
+checkdepends = [
+    "ansible-core",
+    "python-pytest",
+    "python-pytest-mock",
+    "python-pytest-xdist",
+]
+depends = ["ansible-core"]
```
nekopsykose, line 16:
```suggestion
depends = ["ansible-core"]
checkdepends = [
    "python-pytest",
    "python-pytest-mock",
    "python-pytest-xdist",
    *depends,
]
```

contrib/ansible/template.py
```diff
@@ -0,0 +1,38 @@
...
+pkgdesc = "Radically simple IT automation"
+maintainer = "Mara <177581589+catgirlconspiracy@users.noreply.github.com>"
+license = "GPL-3.0-or-later"
+url = "https://ansible.com"
+source = f"$(PYPI_SITE)/a/ansible/ansible-{pkgver}.tar.gz"
+sha256 = "6144fb4bc785f917f86b1b0b6eadc9b894e9751ff9e9a7875afcfa2f74581ffd"
+# many collections either require additional Python modules, or the tests
+# require dependencies that aren't in the tarball, or they're just broken
+# (possibly due to Python version compat issues)
```
nekopsykose, line 25: you could run only the ones that work

contrib/ansible/template.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def do_check(self):
+    for collection_dir in self.find("ansible_collections", "*/*/tests/unit"):
+        collection_dir = collection_dir.parent.parent
+        print(f"checking {collection_dir}")
+        self.do(
+            "ansible-test",
+            "units",
+            "--verbose",
```
nekopsykose, line 36:
```suggestion
```

same as above

contrib/ansible-core/template.py
```diff
@@ -0,0 +1,39 @@
...
+checkdepends = [
+    "git",
+    "openssh",
+    "python-pytest",
+    "python-pytest-mock",
+    "python-pytest-xdist",
+    *_runtime_deps,
+]
+depends = ["python", *_runtime_deps]
+pkgdesc = "Radically simple IT automation"
```
q66, line 26: this should be changed to concisely say what it does

and should be differentiated for core and the other package (i don't use ansible so i have no idea what the difference is, make it obvious)

nekopsykose: `Infrastructure as code software toolkit` + `subdesc = "core libraries"` :)

catgirlconspiracy: I improvised something from Alpine's description that should be a bit better

contrib/ansible/template.py
catgirlconspiracy: I tried excluding all the tests that didn't work

I gave up at ... 15 or 20 or so I think because it seemed like a lot of work and didn't feel worth it

like, I can set it up if you think it makes sense, but it's really quite a lot that don't test

nekopsykose: sure, i guess !check is fine then

contrib/python-resolvelib/template.py
```diff
@@ -0,0 +1,21 @@
...
+    "python-installer",
+    "python-setuptools",
+]
+checkdepends = ["python-packaging", "python-pytest"]
+depends = ["python"]
+pkgdesc = "Resolve abstract dependencies into concrete ones"
+maintainer = "Mara <177581589+catgirlconspiracy@users.noreply.github.com>"
+license = "ISC"
+url = "https://github.com/sarugaku/resolvelib"
+source = f"$(PYPI_SITE)/r/resolvelib/resolvelib-{pkgver}.tar.gz"
```
nekopsykose, line 16: make this `{url}/archive/refs/tags/{pkgver}.tar.gz` and then you can drop the patch above (if pypi tarballs are scuffed and miss data feel free to use the github tarballs)

contrib/ansible-core/template.py
nekopsykose: looks good :)

contrib/python-resolvelib/template.py
catgirlconspiracy: ...so apparently the Github tarball contains even more tests that were missing from the pypi package, which require at least one other external module; do I just package that one as well? (no transitive deps)

nekopsykose: go for it

catgirlconspiracy:
I added the dep for resolvelib's tests (python-commentjson) but its tests also failed and I can't really tell how to make them work, it looks like even the imports don't make sense and I just didn't have the energy for that so I set that to `!check` as well

also I snuck in sshpass which I forgot about - ansible asks you to install it if you use SSH with a password (rather than public key) for connecting to a host

contrib/python-commentjson/template.py
```diff
@@ -0,0 +1,22 @@
+pkgname = "python-commentjson"
+pkgver = "0.9.0"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+]
+depends = ["python", "python-lark-parser"]
```
nekopsykose, line 10:
```suggestion
depends = ["python-lark-parser"]
```

hard python is just for leafs

contrib/sshpass/template.py
```diff
@@ -0,0 +1,11 @@
+pkgname = "sshpass"
+pkgver = "1.10"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = ["autoconf", "automake"]
```
nekopsykose, line 5:
```suggestion
hostmakedepends = ["automake"]
```

implied by latter

nekopsykose:
sure, i'll merge it after i look at the tests and probably conclude the same thing, looks good :)

PR #2743, opened 2024-08-20
packaging: use _ for decorated functions

EarldridgeJazzedPineda:
Decorated functions must be named a single underscore, which is enforced by cbuild.


nekopsykose:
there's another two on around line 3180

PR #2756, opened 2024-08-23
 contrib/godot: new package

z-erica:


contrib/miniupnpc/template.py
```diff
@@ -0,0 +1,26 @@
+pkgname = "miniupnpc"
+pkgver = "2.2.8"
+pkgrel = 0
+build_style = "makefile"
```
nekopsykose, line 4: seems to have cmake

nekopsykose: -D UPNPC_BUILD_SAMPLE=OFF

nekopsykose: well, unless this isn't actually a sample and people use this client

contrib/godot/patches/miniupnpc.patch
```diff
@@ -0,0 +1,29 @@
+API changed in miniupnpc 2.2.8
+
+Patch-Source: https://git.alpinelinux.org/aports/tree/testing/godot/miniupnpc.patch
```
nekopsykose, line 3: these have to point to something stable and not a url that will break in 2 weeks, like https://gitlab.alpinelinux.org/alpine/aports/-/blob/8dd3d8861eff2b4b5171d861e9cde50a92adb93b/testing/godot/miniupnpc.patch

contrib/miniupnpc/template.py
```diff
@@ -0,0 +1,24 @@
...
+build_style = "cmake"
+configure_args = ["-DUPNPC_BUILD_SAMPLE=OFF"]
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+pkgdesc = "UPnP client library"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "BSD-3-Clause"
+url = "http://miniupnp.free.fr"
+source = f"{url}/files/miniupnpc-{pkgver}.tar.gz"
+sha256 = "05b929679091b9921b6b6c1f25e39e4c8d1f4d46c8feb55a412aa697aee03a93"
+tool_flags = {"CFLAGS": ["-fPIC"]}
```
nekopsykose, line 13: use the cmake CMAKE_POSITION_INDEPENDENT_CODE=ON instead

also what error do you get without this? obs also has this issue but is a bit too big to debug, i wonder why cmake is not just passing this by default

contrib/godot/patches/miniupnpc.patch
nekopsykose: cgit also has a way to get a permalink (https://git.alpinelinux.org/aports/tree/testing/godot/miniupnpc.patch?id=44b273d76190945ed980cad6b16276b88017a859; note the `?id=commit`) but i don't really see a point in linking to it so i didn't use it in the example :P

nekopsykose: there's also another one to fix above and below

PR #2762, opened 2024-08-24
contrib/godot-export-templates-linux: new package

z-erica:
these are intended to run in non-chimera systems by virtue of being statically linked, keeping all dependencies vendored (matching the upstream templates) and dlopen'ing the system libraries that aren't vendored

### DIFF
--- a/contrib/godot-export-templates-linux-arm32
+++ b/contrib/godot-export-templates-linux-arm32
@@ -1,0 +1,1 @@
+godot-export-templates-linux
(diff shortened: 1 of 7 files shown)


nekopsykose:
you can't dlopen from static binaries

nekopsykose:
so what you're trying to do is actually not possible  

i think having a statically linked libc++ and then trying to dlopen anything that ends up pulling libstdc++ will also not work on top of that (i'm assuming one of said libs are c++)

nekopsykose:
and then on top of that even if the dlopen worked, dlopening a glibc into the same address space (just to be able to open graphics drivers as an example; where mesa drivers also depend on the c++ std) is also broken too

tl;dr you're not the first and this can't work at all unless you're trying to do something different than what it looks like

PR #2763, opened 2024-08-24
user/tailspin: new package

sonata-chen:


user/tailspin/template.py
```diff
@@ -0,0 +1,22 @@
...
+make_check_args = ["--bins"]  # disable integration tests
+hostmakedepends = ["cargo-auditable"]
+makedepends = ["rust-std"]
+pkgdesc = "Log file highlighter"
+maintainer = "sonata-chen <sonatachen39@gmail.com>"
+license = "MIT"
+url = "https://github.com/bensadeh/tailspin"
+source = [
+    f"{url}/archive/refs/tags/{pkgver}.tar.gz",
+]
```
nekopsykose, line 14: don't need the [] here

triallax:
any particular reason why this is in user instead of contrib?

sonata-chen:
> any particular reason why this is in user instead of contrib?
no

q66:
honestly i think for random shit that's to be used by 1 person we should use user more

PR #2767, opened 2024-08-24
cbuild: check that completions have corresponding commands

triallax:
on multiple occasions, completions have been installed with the wrong names (whether due to `install_completion` default to the pkgname as the command name, or due to typos), so this should guard against that

### DIFF
--- a/src/cbuild/hooks/post_install/010_completions.py
+++ b/src/cbuild/hooks/post_install/010_completions.py
@@ -1,0 +1,46 @@
+def invoke(pkg):
+    comps = []
+
+    if (
+        compspath := pkg.destdir / "usr/share/fish/vendor_completions.d"
+    ).is_dir():
+        comps.extend(
+            (f, f.name.removesuffix(".fish")) for f in compspath.iterdir()
+        )
+
+    if (compspath := pkg.destdir / "usr/share/zsh/site-functions").is_dir():
+        comps.extend(
+            (f, f.name[1:])
+            for f in compspath.iterdir()
+            if f.name.startswith("_")
+        )
+
+    if (
+        pkg.pkgname != "bash-completion"
+        and (
+            compspath := pkg.destdir / "usr/share/bash-completion/completions"
+        ).is_dir()
+    ):
+        comps.extend((f, f.name) for f in compspath.iterdir())
+
+    # we skip checking nushell comps because the dir to which they get installed
+    # is for auto-loaded functions in general and not just comps, and there's no
+    # robust way to distinguish them (zsh does the same but zsh completions have
+    # a _ prefix so it's ok)
+
+    if (binpath := pkg.destdir / "usr/bin").is_dir():
+        commands = {f.name for f in binpath.iterdir()}
+    else:
+        commands = set()
+
+    fail = False
+
+    for comppath, compname in comps:
+        if compname not in commands:
+            pkg.log_red(
+                f"completion file {comppath.relative_to(pkg.destdir)} has no matching usr/bin binary"
+            )
+            fail = True
+
+    if fail:
+        pkg.error("completions check failed")


src/cbuild/hooks/post_install/010_completions.py
```diff
@@ -0,0 +1,53 @@
...
+        comps.extend(
+            (f, f.name.removesuffix(".fish")) for f in compspath.iterdir()
+        )
+
+    if (
+        pkg.pkgname != "fish-shell"
+        and (
+            compspath := pkg.destdir / "usr/share/fish/vendor_completions.d"
+        ).is_dir()
+    ):
```
nekopsykose, line 17: fish-shell only ships the top one, so the first part of this check doesn't do anything (when it's fish-shell, this is also empty)

for the top one, we have actual packages linted out from putting stuff in the top path too, don't we?

so i think this should just be

```
if pkg.pkgname in ["fish-shell", "bash-completion", "zsh"]:
  return
```

or, better- add an options= !lintcompletions and use that instead

triallax: > fish-shell only ships the top one, so the first part of this check doesn't do anything (when it's fish-shell, this is also empty)

good point yes, missed that

> for the top one, we have actual packages linted out from putting stuff in the top path too, don't we?

i couldn't find any lint for that, where is it?

nekopsykose: nvm we don't

```
def _split_fishcomp(pkg):
    pkg.take("usr/share/fish/completions", missing_ok=True)
    pkg.take("usr/share/fish/vendor_completions.d", missing_ok=True)
```

seems it just takes it. i guess that's fine :)  

in that case forget what i said for that

triallax: actually, according to https://fishshell.com/docs/current/completions.html#where-to-put-completions:

> The completions shipped with fish, usually installed in /usr/share/fish/completions

so custom completions should go in the vendor completions dir, and we don't need to consider `usr/share/fish/completions` here either

nekopsykose: i already know the docs, but my previous experience was random apps put it in fish/completions anyway.

we could lint that too and do stuff but in general i don't think that will have much benefit. at best, it avoids a later file conflict when fish itself ships a completion for the same name, and it happens to clash. if not for that though, it makes zero runtime difference and just involves moving files around, and hopefully upstreaming it.

but from a quick check nothing breaks this currently, so i guess it's ok, feel free to add that in a separate mr

src/cbuild/hooks/post_install/010_completions.py
```diff
@@ -0,0 +1,42 @@
...
+            compspath := pkg.destdir / "usr/share/bash-completion/completions"
+        ).is_dir()
+    ):
+        comps.extend((f, f.name) for f in compspath.iterdir())
+
+    if (
+        compspath := pkg.destdir / "usr/share/nushell/vendor/autoload"
+    ).is_dir():
+        comps.extend(
+            (f, f.name.removesuffix(".nu")) for f in compspath.iterdir()
```
triallax, line 30: technically this directory is not only for completions, but also for any kind of auto-loaded functions

this is also the case for zsh but in zsh's case it's mitigated by the convention of the underscore prefix for completion files i guess

nekopsykose: if we know it's broken and can be used for autoload then i think we should skip this dir entirely

nekopsykose: (for zsh, i think the underscore isn't only a convention but it also refuses to load another type of completion when pressing tab? i.e. if you have `foo` and name the completion anything other than `_foo`, it will not work, so it's ok)

triallax: skipping nushell comps here sounds good to me, but this exact same issue crops up in the context of autosplitting `-nucomp` and `-zshcomp` packages; these packages currently take the functions dirs wholesale

nekopsykose: taking that is fine, the only issue is the name might be misleading. but it is indeed shell-specific data that is taken with the shell so the contents are correct and work fine

do you feel like renaming them to -nushelldata -zshdata or something? :)

src/cbuild/hooks/post_install/010_completions.py
triallax: opened #2768 for that

PR #2769, opened 2024-08-25
cbuild: add rust to binaries section in meson crossfile

triallax:
fixes cross builds for meson packages that have `rust` as a language passed to `project`:

meson.build:1:0: ERROR: 'rust' compiler binary not defined in cross file [binaries] section

### DIFF
--- a/src/cbuild/util/meson.py
+++ b/src/cbuild/util/meson.py
@@ -33,6 +33,7 @@ readelf = '{pkg.get_tool("READELF")}'
 objcopy = '{pkg.get_tool("OBJCOPY")}'
 pkgconfig = '{pkg.get_tool("PKG_CONFIG")}'
 llvm-config = '/usr/bin/llvm-config'
+rust = ['rustc', '--target', '{pkg.profile().triplet}', '--sysroot', '{pkg.profile().sysroot / 'usr'}']
 
 [properties]
 needs_exe_wrapper = true


src/cbuild/util/meson.py
```diff
@@ -33,6 +33,7 @@ def _make_crossfile(pkg, build_dir):
 objcopy = '{pkg.get_tool("OBJCOPY")}'
 pkgconfig = '{pkg.get_tool("PKG_CONFIG")}'
 llvm-config = '/usr/bin/llvm-config'
+rust = ['rustc', '--target', '{pkg.profile().triplet}', '--sysroot', '{pkg.profile().sysroot / 'usr'}']
```
triallax, line 36: do i want to use `get_tool_flags` here? not sure

nekopsykose: rustflags read by rustc or cargo? i forgot

if the latter only, then probably you do..

triallax: reading https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads i understand that it's cargo

but also given what i said in my edit it should be fine(?)

nekopsykose: > edit: actually, nothing else in the binaries section has any flags, so it probably doesn't matter

ah yea; the rustc will be set, but meson will pass through whatever flags anyway i guess like it normally would. (maybe it doesn't handle rustflags at all, what happens in main/mesa?)

sure

PR #2770, opened 2024-08-25
Cinnamon desktop environment and X-Apps

EarldridgeJazzedPineda:
Cinnamon should be able to run by itself, without depending on `gnome`, and even without a display manager. This pull request is good to go!

New packages:
* user/adwaita-icon-theme-legacy (46.2)
* user/bulky (3.4)
* user/cinnamon (6.4.0)
* user/cinnamon-control-center (6.4.0)
* user/cinnamon-desktop (6.4.0)
* user/cinnamon-menus (6.4.0)
* user/cinnamon-schemas (6.4.0)
* user/cinnamon-screensaver (6.4.0)
* user/cinnamon-session (6.4.0)
* user/cinnamon-settings-daemon (6.4.0)
* user/cjs (6.4.0)
* user/libgnomekbd (3.28.1)
* user/libsass-python (0.23.0)
* user/libtimezonemap (0.4.6)
* user/lightdm (1.32.0)
* user/mint-cursor-themes (1.0.2)
* user/mint-themes (2.2.0)
* user/mint-x-icons (1.7.2)
* user/mint-y-icons (1.7.9)
* user/muffin (6.4.0)
* user/nemo (6.4.0)
* user/pix (3.4.3)
* user/python-pam (2.0.2)
* user/python-xapp (2.4.2)
* user/slick-greeter (2.0.7)
* user/timeshift (24.06.3) (restricted until further notice, due to timeshift-gtk causing a segfault at strlen())
* user/touchegg (2.0.17)
* user/xapp (2.8.6)
* user/xapp-apps (1)
* user/xdg-desktop-portal-xapp (1.1.0)
* user/xed (3.6.7)
* user/xreader (4.2.2)
* user/xviewer (3.4.6)


user/nemo/template.py
```diff
@@ -0,0 +1,39 @@
...
+    "tracker-devel",
+    "xapp-devel",
+]
+pkgdesc = "File browser for Cinnamon"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-2.0-or-later AND GFDL-1.1-or-later AND LGPL-2.0-or-later"
+url = "https://projects.linuxmint.com/cinnamon"
+source = f"https://github.com/linuxmint/nemo/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "68d829614beeb2596aa706db5ff070fcfc0e96ab7e3d2561772b52bed6c8ca38"
+# Tests require a graphical environment
```
nekopsykose, line 33: this exists, see make_check_wrapper wlheadless-run use in the repo

since this is gtk3, you will probably need `"wlheadless-run", "-c", "cage", "--"` as the invocation

EarldridgeJazzedPineda: Tests require its own GSettings schemas to be installed, I will not be enabling the "check" phase.

nekopsykose: there's env vars for that, but the bigger issue is that the nemo tests are broken entirely for other reasons (known upstream). i just mean to show you that "need graphical env" is not actually the blocker :)

user/cinnamon-desktop/template.py
```diff
@@ -0,0 +1,33 @@
...
+]
+depends = ["chimera-artwork", "hwdata-pnp"]
+pkgdesc = "Cinnamon desktop library and common settings schemas"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-2.0-or-later AND LGPL-2.0-or-later"
+url = "https://projects.linuxmint.com/cinnamon"
+source = f"https://github.com/linuxmint/cinnamon-desktop/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "95bb6fc16597601b1febde4bd1317ba9c3ab662119b6b8e602b49e832d3f3eb7"
+# No tests
+options = ["!check"]
```
nekopsykose, line 28: when there's no tests but the default check also passes (like meson just says 'no tests' and continues), you don't have to set this, and just doing nothing is ok

user/cinnamon/template.py
```diff
@@ -0,0 +1,55 @@
+pkgname = "cinnamon"
+pkgver = "6.2.9"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = ["gobject-introspection", "intltool", "meson", "pkgconf"]
```
nekopsykose, line 5: multiline these too

user/cinnamon/template.py
```diff
@@ -0,0 +1,55 @@
...
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_dir("etc/xdg/menus/applications-merged", empty=True)
+
+
+@subpackage("cinnamon-static")
+def _(self):
+    return ["usr/lib/cinnamon/*.a"]
```
nekopsykose, line 48: pass `-Ddefault_library=shared` to configure_args instead; getting this means they used `library()` for plugins instead of a plugin target, and so the default =both generates statics (that you delete here). you should pass the arg and remove this subpackage instead

user/cinnamon/template.py
```diff
@@ -0,0 +1,55 @@
...
+source = (
+    f"https://github.com/linuxmint/cinnamon/archive/refs/tags/{pkgver}.tar.gz"
+)
+sha256 = "a0b15b98a899532d531689ea11ff6805a87fcf5994e238322b6c916bf8c29919"
+# meson.build: tests are not currently functional
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_dir("etc/xdg/menus/applications-merged", empty=True)
```
nekopsykose, line 43: what happens if this directory does not exist?

user/cinnamon/template.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@subpackage("cinnamon-static")
+def _(self):
+    return ["usr/lib/cinnamon/*.a"]
+
+
+@subpackage("cinnamon-python")
+def _(self):
+    self.subdesc = "Python bindings"
```
nekopsykose, line 53: self.depends += ["python"] too

nekopsykose: actually probably want self.depends = [self.parent] instead or similar

user/caribou/template.py
```diff
@@ -0,0 +1,47 @@
...
+    "libxml2-devel",
+    "libxtst-devel",
+    "python-gobject-devel",
+]
+depends = ["clutter"]
+pkgdesc = "Text entry and UI navigation application"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "LGPL-2.1-or-later"
+url = "https://wiki.gnome.org/Projects/Caribou"
+source = f"https://download.gnome.org/sources/caribou/{pkgver[:-3]}/caribou-{pkgver}.tar.xz"
```
nekopsykose, line 32:
```suggestion
source = f"https://download.gnome.org/sources/caribou/{".".join(pkgver.split(".")[0:2)}/caribou-{pkgver}.tar.xz"
```

deleting the last three chars would just break when it becomes 0.5.3 or similar

user/caribou/template.py
```diff
@@ -0,0 +1,47 @@
...
+sha256 = "9c43d9f4bd30f4fea7f780d4e8b14f7589107c52e9cb6bd202bd0d1c2064de55"
+tool_flags = {"CFLAGS": ["-Wno-incompatible-function-pointer-types"]}
```
nekopsykose, line 34:
```suggestion
# vala
tool_flags = {"CFLAGS": ["-Wno-incompatible-function-pointer-types"]}
```

user/libgnomekbd/template.py
```diff
@@ -0,0 +1,26 @@
...
+    "pkgconf",
+    "glib-devel",
+    "gobject-introspection",
+]
+makedepends = ["gtk+3-devel", "libxklavier-devel"]
+pkgdesc = "Gnome keyboard configuration library"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://gitlab.gnome.org/GNOME/libgnomekbd"
+source = f"$(GNOME_SITE)/{pkgname}/{pkgver[:-2]}/{pkgname}-{pkgver}.tar.xz"
```
nekopsykose, line 18:
```suggestion
source = f"$(GNOME_SITE)/libgnomekbd/{pkgver[:-2]}/libgnomekbd-{pkgver}.tar.xz"
```

user/libgnomekbd/template.py
```diff
@@ -0,0 +1,26 @@
...
+pkgver = "3.28.1"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Dintrospection=false"]
+hostmakedepends = [
+    "meson",
+    "gettext",
+    "pkgconf",
+    "glib-devel",
+    "gobject-introspection",
```
nekopsykose, line 11:
```suggestion
    "gettext",
    "glib-devel",
    "gobject-introspection",
    "meson",
    "pkgconf",
```

user/cinnamon/template.py
EarldridgeJazzedPineda: cbuild will output an "unknown provider" error when scanning for symlink dependencies.

user/cinnamon/template.py
EarldridgeJazzedPineda: Why would you want to remove static library subpackages?

user/caribou/template.py
EarldridgeJazzedPineda: Caribou hasn't been updated in a while, but I will add this change in there just in case it gets updated.

user/caribou/template.py
EarldridgeJazzedPineda: What is that comment for?

user/libgnomekbd/template.py
EarldridgeJazzedPineda: For your information, I used the template from an attempt to package the Budgie desktop environment (#172), and modified it to make it compilable once again.

user/cinnamon/template.py
nekopsykose: because these are not actual static libraries that are meant to be linked to, they are internal cinnamon libraries. they only exist because of default_library=both, as i said. so if you pass -Ddefault_library=shared, then they don't get (wrongly) installed, and then you don't need the subpackage

this is just repeating the same thing i already said; did i word it poorly?

user/caribou/template.py
nekopsykose: Wno-incompatible-function-pointer-types is an error that should be fixed

..except in the case of vala where the code it generates is broken in that way, so the comment says that this is intended since we know vala does that. otherwise it just looks like silencing something that has to be fixed.

user/libgnomekbd/template.py
nekopsykose: sure, that's okay :)

user/caribou/template.py
EarldridgeJazzedPineda: Caribou is indeed written in Vala.

user/cinnamon/template.py
EarldridgeJazzedPineda: That directory doesn't need to be present, because by default it scans for .desktop files in /usr/share/applications.

user/libgnomekbd/template.py
```diff
@@ -0,0 +1,26 @@
+pkgname = "libgnomekbd"
+pkgver = "3.28.1"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Dintrospection=false"]
```
triallax, line 5: why no introspection?

user/libgnomekbd/template.py
```diff
@@ -0,0 +1,26 @@
...
+]
+makedepends = ["gtk+3-devel", "libxklavier-devel"]
+pkgdesc = "Gnome keyboard configuration library"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://gitlab.gnome.org/GNOME/libgnomekbd"
+source = f"$(GNOME_SITE)/libgnomekbd/{pkgver[:-2]}/libgnomekbd-{pkgver}.tar.xz"
+sha256 = "22dc59566d73c0065350f5a97340e62ecc7b08c4df19183804bb8be24c8fe870"
+# needs graphical environment
+options = ["!check"]
```
triallax, line 21: for the checks that need graphical environment, `xserver-xorg-xvfb` in `checkdepends` and `make_check_wrapper = ["xvfb-run"]` should do the job (we usually prefer `wlheadless-run`  but gtk+3 has trouble with headless wayland)

user/muffin/template.py
```diff
@@ -0,0 +1,85 @@
...
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://projects.linuxmint.com/cinnamon"
+source = (
+    f"https://github.com/linuxmint/muffin/archive/refs/tags/{pkgver}.tar.gz"
+)
+sha256 = "889200ceb45334899e7568f1735ec2b15e1ae042d176821a0d45c524960bee0a"
+hardening = ["vis"]
+# Tests require a graphical environment
+options = ["!check"]
```
triallax, line 69: same xvfb-run thing here

user/caribou/template.py
```diff
@@ -0,0 +1,48 @@
+pkgname = "caribou"
+pkgver = "0.4.21"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--disable-gtk2-module"]
```
triallax, line 5:
```suggestion
configure_args = ["--disable-gtk2-module", "--disable-static"]
```
and drop the -static subpkg

user/cinnamon-settings-daemon/template.py
```diff
@@ -0,0 +1,50 @@
+pkgname = "cinnamon-settings-daemon"
+pkgver = "6.2.0"
+pkgrel = 0
+build_style = "meson"
```
triallax, line 4:
```suggestion
build_style = "meson"
configure_args = ["-Ddefault_library=shared"]
```
and drop -static subpkg

user/cinnamon-desktop/template.py
```diff
@@ -0,0 +1,37 @@
+pkgname = "cinnamon-desktop"
+pkgver = "6.2.0"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Dalsa=true", "-Dpnp_ids=/usr/share/hwdata/pnp.ids"]
```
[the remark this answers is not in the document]
EarldridgeJazzedPineda: Why you want ALSA support disabled?

user/libgnomekbd/template.py
EarldridgeJazzedPineda: I copied the libgnomekbd template from #172, with a few modifications. I didn't touch this variable.

user/cinnamon-desktop/template.py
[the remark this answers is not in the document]
triallax: no reason to have it enabled really, libpulse is already supported

user/libgnomekbd/template.py
EarldridgeJazzedPineda: libgnomekbd doesn't have tests, it just says "No tests defined."

user/muffin/template.py
EarldridgeJazzedPineda: GSettings schemas have to be installed during tests. I will not be enabling this.

user/libgnomekbd/template.py
EarldridgeJazzedPineda: Meson does compile tests, but it doesn't run them.

nekopsykose: if the check phases passes without issues then remove the !check even if no tests are ran

triallax: yeah, and i'm asking you to run them if possible, as i already explained

user/libgnomekbd/template.py
triallax: yeah just remove this line, there's no reason to have introspection disabled as far as i can see

user/python-tinycss2/template.py
```diff
@@ -0,0 +1,24 @@
...
+hostmakedepends = ["python-build", "python-flit_core", "python-installer"]
+depends = ["python-webencodings"]
+checkdepends = ["python-pytest", *depends]
+pkgdesc = "Tiny CSS parser"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "BSD-3-Clause"
+url = "https://github.com/Kozea/tinycss2"
+source = [
+    f"{url}/archive/refs/tags/v{pkgver}.tar.gz",
+    "https://github.com/CourtBouillon/css-parsing-tests/archive/43e65b244133f17eb8a4d4404d5774672b94824f.zip",
```
nekopsykose, line 14:
```suggestion
    "https://github.com/CourtBouillon/css-parsing-tests/archive/43e65b244133f17eb8a4d4404d5774672b94824f.tar.gz",
```

this goes for all github tarballs, like mint-cursor-themes too

user/xed/template.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@subpackage("xed-devel")
+def _(self):
+    return self.default_devel()
+
+
+@subpackage("xed-static")
+def _(self):
+    return ["usr/lib/xed/*.a", "usr/lib/xed/plugins/*.a"]
```
triallax, line 42: do the `-Ddefault_library=shared` bit here too

user/xviewer/template.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@subpackage("xviewer-devel")
+def _(self):
+    return self.default_devel()
+
+
+@subpackage("xviewer-static")
+def _(self):
+    return ["usr/lib/xviewer/*.a"]
```
triallax, line 46: ditto, `-Ddefault_library=shared` and remove -static subpkg

triallax (changes requested):

user/bulky/template.py
```diff
@@ -0,0 +1,18 @@
...
+build_style = "makefile"
+hostmakedepends = ["gettext"]
+depends = ["base-files-doc", "bash", "python-magic"]
+pkgdesc = "Bulk renamer"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-3.0-or-later"
+url = "https://projects.linuxmint.com/xapps"
+source = f"https://github.com/linuxmint/bulky/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "dd3350ccb98ce1b1f3df67151ca091430df67e41f19162a415550ef8ebd67c45"
+# No tests
```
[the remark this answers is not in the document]
EarldridgeJazzedPineda: The script will not work for two reasons:
1. It requires that Bulky be installed
2. The cbuild container's root is read-only

user/bulky/patches/set-version-number-in-about-dialog.patch
```diff
@@ -0,0 +1,12 @@
+diff -ru a/usr/lib/bulky/bulky.py b/usr/lib/bulky/bulky.py
+--- a/usr/lib/bulky/bulky.py	2024-07-21 17:26:01.000000000 +0800
++++ b/usr/lib/bulky/bulky.py	2024-09-06 20:08:30.391435713 +0800
+@@ -403,7 +403,7 @@
+         except Exception as e:
+             print (e)
+ 
+-        dlg.set_version("__DEB_VERSION__")
++        dlg.set_version("3.4")
```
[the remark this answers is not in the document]
EarldridgeJazzedPineda: I'm not sure if that's possible using the available cbuild functions. In other distros, this is handled by using `sed` to replace `__DEB_VERSION__` with the version number, but sed is forbidden in templates.

user/cinnamon/template.py
```diff
@@ -0,0 +1,82 @@
...
+
+@subpackage("cinnamon-meta")
+def _(self):
+    self.subdesc = "recommends package"
+    self.install_if = [self.parent]
+    self.depends = [
+        "gnome-terminal",
+        # Core icons and themes
+        "mint-cursor-themes",
+        "mint-themes",
```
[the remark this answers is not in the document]
EarldridgeJazzedPineda: I will split the themes into a separate subpackage so that other DEs can use them

user/bulky/template.py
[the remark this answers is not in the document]
triallax: i didn't say to run the script, i said to correct the comment because there _is_ a test

user/cinnamon-desktop/template.py
```diff
@@ -0,0 +1,41 @@
...
+    "libx11-devel",
+    "libxext-devel",
+    "libxkbfile-devel",
+    "libxrandr-devel",
+    "xkeyboard-config",
+]
+depends = ["chimera-artwork", "hwdata-pnp"]
+pkgdesc = "Cinnamon desktop library and common settings schemas"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-2.0-or-later AND LGPL-2.0-or-later"
```
triallax, line 22: some files (e.g. `libcinnamon-desktop/gnome-installer.c`) are under `LGPL-2.1-or-later`, so i'd do this
```suggestion
license = "GPL-2.0-or-later AND LGPL-2.1-or-later"
```

user/cinnamon-menus/template.py
```diff
@@ -0,0 +1,24 @@
...
+hostmakedepends = ["gobject-introspection", "gtk-doc-tools", "meson", "pkgconf"]
+makedepends = ["glib-devel"]
+pkgdesc = "Cinnamon menu library"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-2.0-or-later AND LGPL-2.0-or-later"
+url = "https://projects.linuxmint.com/cinnamon"
+source = f"https://github.com/linuxmint/cinnamon-menus/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "95170602c5291db6f5031dfc245ebf61872ff93860fb1fc18ec433bec638cf16"
+# No tests
+options = ["!check", "!cross"]
```
triallax, line 19: meson can handle no tests just fine, no need for !check here
```suggestion
options = ["!cross"]
```

user/cinnamon-screensaver/template.py
```diff
@@ -0,0 +1,32 @@
...
+# Including cinnamon will create a circular dependency
+depends = ["caribou", "cinnamon-desktop", "python-xapp"]
+pkgdesc = "Cinnamon screen locker and screensaver program"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-2.0-or-later AND LGPL-2.0-or-later"
+url = "https://projects.linuxmint.com/cinnamon"
+source = f"https://github.com/linuxmint/cinnamon-screensaver/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "8fdc012b630b2b1236fc57fa5f006c7c757028ae2ef4528fd05e4d68f4902eda"
+# No tests
+options = ["!check", "!cross"]
```
triallax, line 27:
```suggestion
options = ["!cross"]
```

user/cinnamon-session/template.py
```diff
@@ -0,0 +1,33 @@
...
+]
+pkgdesc = "Cinnamon session manager"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://projects.linuxmint.com/cinnamon"
+source = f"https://github.com/linuxmint/cinnamon-session/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "69d39494201f2cd7e30d880813df69ddeef143befc660ed953683d5ddac6153f"
+hardening = ["vis"]
+# No tests
+options = ["!check"]
```
triallax, line 33:
```suggestion
```

user/cinnamon-settings-daemon/patches/XyG2.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/install-scripts/meson_mk_symlink.py b/install-scripts/meson_mk_symlink.py
```
triallax, line 1: this patch shouldn't be necessary since c9d75988b64479395b86ebacafbda031c8aedd8c (though the presence of the script at all isn't ideal, but i'll address that upstream)

user/cinnamon-settings-daemon/template.py
```diff
@@ -0,0 +1,46 @@
...
+    "xorgproto",
+]
+pkgdesc = "Settings daemon for the Cinnamon desktop"
+maintainer = "Earldridge Jazzed Pineda <earldridgejazzedpineda@gmail.com>"
+license = "GPL-2.0-or-later AND LGPL-2.1-or-later"
+url = "https://projects.linuxmint.com/cinnamon"
+source = f"https://github.com/linuxmint/cinnamon-settings-daemon/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "ecb360017284cdf8a5776a64175ac69f730790d3031f1f4cabc5f8c52660a498"
+# No tests
+options = ["!check"]
```
triallax, line 41:
```suggestion
```

user/cinnamon/template.py
```diff
@@ -0,0 +1,89 @@
...
+broken_symlinks = ["etc/xdg/menus/cinnamon-applications-merged"]
+# meson.build: tests are not currently functional
+options = ["!check", "!cross"]
+
+
+@subpackage("cinnamon-apps")
+def _(self):
+    self.subdesc = "apps"
+    self.install_if = [self.parent]
+    self.depends = ["gnome-terminal", "xapp-apps"]
```
triallax, line 63: is pulling in gnome-terminal something that upstream encourages or suggests? if not, then i'm not quite sure we should do it

user/cinnamon/template.py
```diff
@@ -0,0 +1,89 @@
...
+    "mint-themes",
+    "nemo",
+    "network-manager-applet",
+    "python-pam",
+    "python-pexpect",
+    "python-pillow",
+    "python-pytz",
+    "python-requests",
+    "python-tinycss2",
+    "touchegg",
```
triallax, line 44: add `accountsservice` and `gsound` here, cinnamon-settings doesn't launch without them for me

user/cinnamon/template.py
EarldridgeJazzedPineda: Yes.
1. By default, the GNOME Terminal application is pinned to the taskbar
2. mint-y-icons has an icon for GNOME Terminal but not GNOME Console.
3. Linux Mint uses GNOME Terminal because it uses GTK+3, while GNOME Console uses GTK4 and libadwaita (which it avoids because libadwaita doesn't respect themes, see https://blog.linuxmint.com/?p=4675 ).

EarldridgeJazzedPineda:
I've been working on this pull request for almost 40 days, keeping up with changes to the cbuild infrastructure, and I can finally say (for real this time) it's good to go.

Here is the final screenshot for the Cinnamon desktop environment:
VirtualBox_Chimera Linux_02_10_2024_10_11_02

EarldridgeJazzedPineda:
@triallax I have pretty much finalized all 33 templates for this pull request, but I will make additional edits as necessary in case you do a review (but not for anything major such as a new package). Good luck!

EarldridgeJazzedPineda:
bump

q66:
i reviewed + fixed + pushed a portion of these to make it easier to review, you can rebase it and i'll check the rest later

q66:
stale

PR #2771, opened 2024-08-25
contrib/godot: build export templates

z-erica:


contrib/godot/template.py
```diff
@@ -151,3 +153,19 @@ def do_install(self):
...
+        f"bin/godot.linuxbsd.template_debug.{_godot_arch}.llvm",
+        name="godot-template-debug",
+    )
+    self.install_bin(
+        f"bin/godot.linuxbsd.template_release.{_godot_arch}.llvm",
+        name="godot-template-release",
+    )
+
+
+@subpackage("godot-export-templates")
```
q66, line 167: should this depend on the parent package?

nekopsykose: yeah, these should work to just run a distributed godot game by essentially replacing the engine it ships with (so you can run it natively without a chroot). i dunno if that actually works in practice since games might use other native libs(?)

the base package is the IDE and unrelated

PR #2774, opened 2024-08-25
contrib/foot: include user service

denisstrizhkin:


contrib/foot/template.py
```diff
@@ -37,6 +37,7 @@ def post_install(self):
         f"usr/share/licenses/{pkgname}/LICENSE",
         relative=False,
     )
+    self.install_service(self.files_path / "foot.user", enable=False)
```
nekopsykose, line 40:
```suggestion
    self.install_service(self.files_path / "foot.user")
```

is the default

PR #2776, opened 2024-08-25
user/pulsemixer: new package

denisstrizhkin:
could maybe go to contrib

### DIFF
--- a/user/pulsemixer/template.py
+++ b/user/pulsemixer/template.py
@@ -1,0 +1,18 @@
+pkgname = "pulsemixer"
+pkgver = "1.5.1"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = ["python-build", "python-installer", "python-setuptools"]
+depends = ["libpulse"]
+pkgdesc = "CLI and curses mixer for PulseAudio"
+maintainer = "Denis Strizhkin <strdenis02@gmail.com>"
+license = "MIT"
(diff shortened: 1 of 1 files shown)


q66:
this does not seem to use hatch? uses setuptools

denisstrizhkin:
fails if i remove it

nekopsykose:
because hatch-vcs pulls in setuptools

it's still wrong and you just need setuptools

q66:
doesn't fail if you actually add the buildsystem it uses, which is setuptools

user/pulsemixer/template.py
```diff
@@ -0,0 +1,18 @@
...
+pkgver = "1.5.1"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = ["python-build", "python-installer", "python-setuptools"]
+depends = ["libpulse"]
+pkgdesc = "CLI and curses mixer for PulseAudio"
+maintainer = "Denis Strizhkin <strdenis02@gmail.com>"
+license = "MIT"
+url = "https://github.com/GeorgeFilipkin/pulsemixer"
+source = f"https://github.com/GeorgeFilipkin/{pkgname}/archive/{pkgver}.tar.gz"
```
nekopsykose, line 11:
```suggestion
source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
```

PR #2778, opened 2024-08-25
user/mu: new package

tjheeta:


user/mu/template.py
```diff
@@ -0,0 +1,12 @@
+pkgname = "mu"
+pkgver = "1.12.6"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = ["meson", "pkgconf", "cmake", "emacs-console"]
```
nekopsykose, line 5:
```suggestion
hostmakedepends = ["meson", "pkgconf", "emacs-console"]
```

not used

user/mu/template.py
```diff
@@ -0,0 +1,12 @@
...
+makedepends = ["xapian-core-devel", "gmime-devel", "glib-devel"]
```
nekopsykose, line 6: i think you also want guile-devel here

nekopsykose:
the commit messages should be `user/mu: new package` not ` - `

user/mu/template.py
nekopsykose: you can then split usr/share/guile + usr/lib/guile to a subpackage

tjheeta: hint on the correct code block to do it?

user/mu/template.py
```diff
@@ -0,0 +1,28 @@
...
+    "pkgconf",
+]
+makedepends = [
+    "glib-devel",
+    "gmime-devel",
+    "xapian-core-devel",
+]
+pkgdesc = "Maildir-utils indexer/searcher and mu4e extension for emacs"
+maintainer = "tj <tjheeta@gmail.com>"
+license = "GPL-3.0-only"
```
nekopsykose, line 17:
```suggestion
license = "GPL-3.0-or-later"
```

PR #2780, opened 2024-08-25
user/incus: new package

tjheeta:


user/lxc/template.py
```diff
@@ -0,0 +1,46 @@
...
+license = "GPL-2.0-only AND LGPL-2.1-only"
+url = "https://linuxcontainers.org"
+source = f"{url}/downloads/{pkgname}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "d8195423bb1e206f8521d24b6cde4789f043960c7cf065990a9cf741dcfd4222"
+file_modes = {"usr/libexec/lxc/lxc-user-nic": ("root", "root", 0o4755)}
+
+def post_extract(self):
+    # https://github.com/void-linux/void-packages/pull/38493/files
+    # workaround https://github.com/lxc/lxc/issues/4178
+    self.mkdir("/etc/default", parents = True)
```
nekopsykose, line 42: this is fixed in lxc 6.0 already and you should use the latest version

user/lxc/template.py
```diff
@@ -0,0 +1,46 @@
...
+]
+depends = [
+    "xz",
+    "wget2"
+]
+pkgdesc = "Linux containers"
+maintainer = "tj <tjheeta@gmail.com>"
+license = "GPL-2.0-only AND LGPL-2.1-only"
+url = "https://linuxcontainers.org"
+source = f"{url}/downloads/{pkgname}/{pkgname}-{pkgver}.tar.gz"
```
nekopsykose, line 35:
```suggestion
source = f"{url}/downloads/lxc/lxc-{pkgver}.tar.gz"
```

user/lxc/template.py
```diff
@@ -0,0 +1,46 @@
...
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+    "git"
+]
+makedepends = [
+    "libcap-devel",
+    "libcap-devel-static",
+    #"libapparmor-devel",
+    "libseccomp-devel",
```
nekopsykose, line 19: sort these lists

user/lxc/template.py
```diff
@@ -0,0 +1,46 @@
+pkgname = "lxc"
+pkgver = "5.0.1"
+pkgrel = 1
```
nekopsykose, line 3: pkgrels start at 0

user/incus/files/sysusers.conf
```diff
@@ -0,0 +1,4 @@
+# Create incus system users
+
+u incus - "incus user" /var/lib/incus /usr/bin/nologin
+g incus-admin -
```
nekopsykose, line 4: if the group is easily usable with an underscore, name this _incus / _incus_admin (since the above daemon should then be runnable with --group _incus_admin)

user/incus/files/incus
```diff
@@ -0,0 +1,5 @@
+# incus service
+
+type = process
+command = /usr/bin/incusd --group incus-admin --syslog
+depends-on = network.target
```
nekopsykose, line 5: network.target is supplementary, you also need local.target

user/incus/files/incus
```diff
@@ -0,0 +1,5 @@
+# incus service
+
```
nekopsykose, line 2:
```suggestion
```

user/incus/template.py
```diff
@@ -0,0 +1,81 @@
...
+    "util-linux"
+]
+go_build_tags=[ "libsqlite3" ]
+pkgdesc = "Powerful system container and virtual machine manager"
+maintainer = "tj <tjheeta@gmail.com>"
+license = "Apache-2.0"
+url = "https://github.com/lxc/incus"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "c6ad8260a4954b0e4f77de66220d5450290b57858f8c16ee83f12996eea86db4"
+# void ignores TestConvertNetworkConfig (https://github.com/void-linux/void-packages/blob/master/srcpkgs/incus/template)
```
nekopsykose, line 48: linking to unstable urls that break in a week is not useful, either use a permalink or actually pass this in make_check_args and enable the test

user/incus/template.py
```diff
@@ -0,0 +1,81 @@
...
+# go test -v -tags libsqlite3 -skip TestConvertNetworkConfig ./...
+options = ["!check"]
+
+
+def post_build(self):
+    # Build the agent statically
+    self.env["CGO_ENABLED"] = "0"
+    self.env["CGO_LDFLAGS"] = "-extldflags=-static"
+    self.go_build_tags = ["agent", "netgo", "osusergo"]
+    self.make_build_args = ["-buildmode=default", "./..."]
```
nekopsykose, line 58: buildmode should be pie still

user/squashfs-tools/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "squashfs-tools"
```
nekopsykose, line 1: does incus not work with the slightly differently named squashfs-tools-ng?

nekopsykose:
see ci for formatting

user/squashfs-tools/template.py
tjheeta: No, it used the binary unsquashfs which is apparently different than rdsquashfs provided by squashfs-tools-ng.

nekopsykose: ok, makes sense. at first i thought it conflicted but all the bin names are unique so it's ok :)

user/lxc/template.py
tjheeta: Good catch. Will do.

user/incus/files/sysusers.conf
nekopsykose: (the reason for this is we just have _-prefixed system groups/users where possible, but it isn't always. if it then tries to use a hardcoded group name from the code then sometimes it's patchable but if it's too much effort what you put already is ok)

user/lxc/template.py
tjheeta: done

user/lxc/template.py
tjheeta: should be pkgver 6.01 and pkgrel 0 now

user/incus/template.py
```diff
@@ -0,0 +1,81 @@
...
+    self.golang.build(wrksrc="cmd/incus-agent")
+
+def do_install(self):
+    self.install_file(f"{self.make_dir}/incusd", "usr/bin")
```
nekopsykose, line 62: the default do_install for go buildstyle installs all bins from make_dir to usr/bin, does that not work here? it should do the same thing

tjheeta: There were a bunch of extra binaries that were built, so I only wanted to take the ones that mattered.

tjheeta: Not sure if there's an easier way to filter them out.

nekopsykose: how many and what kind? might make more sense to self.uninstall a few instead

user/incus/template.py
nekopsykose: hm so

- the extldflags don't do something without also -linkmode=external
- go itself with its own linker seems to not be able to output a static pie executable
- so, either you need what you already put (and can remove the ldflags), or you can cgo_enabled=1 cgo_ldflags=-static-pie and remove the -buildmode. i don't know which is better, but that at least gives a static pie instead

user/incus/template.py
tjheeta: yah, there were only a few. i'll make the change.

user/incus/files/incus
tjheeta: done

user/incus/files/incus
tjheeta: done

user/incus/template.py
tjheeta: used permalink for now

user/incus/template.py
tjheeta: Just tried

```
    self.env["CGO_ENABLED"] = "1"
    self.env["CGO_LDFLAGS"] = "-static-pie"
    self.go_build_tags = ["agent", "netgo", "osusergo"]
    self.make_build_args = ["./..."]
    self.golang.build(wrksrc="cmd/incus-agent")
```

Gives the error:
```
/usr/lib/go/pkg/tool/linux_amd64/link: running clang failed: exit status 1
/usr/bin/clang -m64 -Wl,-z,relro -pie -o $WORK/b001/exe/a.out -Wl,--export-dynamic-symbol=_cgo_panic -Wl,--export-dynamic-symbol=_cgo_topofstack -Wl,--export-dynamic-symbol=crosscall2 -Qunused-arguments -Wl,--compress-debug-sections=zlib /tmp/go-link-1580340833/go.o /tmp/go-link-1580340833/000000.o /tmp/go-link-1580340833/000001.o /tmp/go-link-1580340833/000002.o /tmp/go-link-1580340833/000003.o /tmp/go-link-1580340833/000004.o /tmp/go-link-1580340833/000005.o /tmp/go-link-1580340833/000006.o /tmp/go-link-1580340833/000007.o /tmp/go-link-1580340833/000008.o /tmp/go-link-1580340833/000009.o /tmp/go-link-1580340833/000010.o /tmp/go-link-1580340833/000011.o /tmp/go-link-1580340833/000012.o /tmp/go-link-1580340833/000013.o /tmp/go-link-1580340833/000014.o /tmp/go-link-1580340833/000015.o /tmp/go-link-1580340833/000016.o /tmp/go-link-1580340833/000017.o -static-pie -lutil -lpthread -static-pie -lpthread
ld: error: unable to find library -l:libunwind.a
ld: error: unable to find library -latomic
ld: error: unable to find library -lc
clang: error: linker command failed with exit code 1 (use -v to see invocation)

```

Looks like it's still trying to link to musl and others. Going to try to put the static versions as makedepends.

tjheeta: Hmmm...I'm not sure if this worked or not.
```
$ ldd /usr/bin/incus-agent 
        ldd (0x70c2db22a000)

```

tjheeta: From irc, have to use file to verify that pie binaries are built statically. 

```
file /usr/bin/incus-agent
/usr/bin/incus-agent: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), static-pie linked, Go BuildID=01tUkAstqv9ow_1cyjgs/NIM2E9c867DsBpZK4JBg/eTRB6Ypl35xXoqgWjeSM/_tNAfH8EX9KYB8VIpqvp, BuildID[xxHash]=a0f2727ea2699ca7, stripped
```

user/cowsql/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "cowsql"
+pkgver = "1.15.6"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_gen = ["autoreconf", "-i"]
```
nekopsykose, line 5: does this have to be specified? the default is already `autoreconf -fi` so it should be the same unless the forceful overwrite of uptodate breaks it (what error do you get?)

user/cowsql/template.py
```diff
@@ -0,0 +1,23 @@
...
+hostmakedepends = [
+    "automake",
+    "slibtool",
+    "bash",
+    "pkgconf",
```
nekopsykose, line 10: sort

user/incus/template.py
```diff
@@ -0,0 +1,82 @@
+pkgname = "incus"
+pkgver = "6.4.0"
+pkgrel = 15
```
nekopsykose, line 3:
```suggestion
pkgrel = 0
```

user/cowsql/template.py
nekopsykose: this goes for the other *depends lists too when they're vertical

user/lxc/patches/meson.patch
```diff
@@ -0,0 +1,23 @@
+--- lxc-5.0.1/meson.build	2022-07-27 21:20:41.000000000 -0700
++++ lxc-5.0.1.bk/meson.build	2024-08-24 09:19:18.731887904 -0700
+@@ -316,7 +316,8 @@
+ if time_epoch == ''
+     time_epoch = run_command(date, '+%s', check: true).stdout()
+ endif
+-generate_date = run_command(date, '--utc', '--date=@' + time_epoch, '+%Y-%m-%d', check: true).stdout().strip()
++#generate_date = run_command(date, '--utc', '--date=@' + time_epoch, '+%Y-%m-%d', check: true).stdout().strip()
++generate_date = run_command(date, '+%Y-%m-%d', check: true).stdout().strip()
```
nekopsykose, line 9: this doesn't seem correct since the previous code uses --utc and stuff. i dunno how would would inject source_date_epoch into it so i'll let someone else think about it

user/lxc/template.py
```diff
@@ -0,0 +1,40 @@
+pkgname = "lxc"
+pkgver = "6.0.1"
+pkgrel = 0
+build_style = "meson"
+configure_args = [
+    "-Dapparmor=false",
+    "-Dman=false",
+    "-Dinit-script=sysvinit",
```
nekopsykose, line 8: this installs a etc/init.d/ script but that is not used since no openrc

user/cowsql/template.py
nekopsykose: same goes below in raft

user/lxc/patches/meson.patch
tjheeta: Both commands come up with the same result.

user/incus/template.py
tjheeta: done

user/cowsql/template.py
tjheeta: done

user/lxc/template.py
tjheeta: Not sure what to do about that, the two options are systemd and and sysinit.

nekopsykose: you can `self.uninstall` the etc/init.d folder

user/lxc/patches/meson.patch
nekopsykose: only if /etc/localtime happens to be utc. for bsd date, `-z utc` is equivalent

user/lxc/template.py
tjheeta:
```
Configuring lxc-apparmor-load using configuration
Run-time dependency systemd found: NO (tried pkgconfig)

config/init/systemd/meson.build:13:18: ERROR: Dependency "systemd" not found, tried pkgconfig

From meson configure, looks like we need something. 
```
  init-script          [systemd]              [systemd, sysvinit]    init script                               
```
```

user/cowsql/template.py
tjheeta: done

user/lxc/template.py
nekopsykose: yea, i mean you can leave it as is but just delete the file after

user/lxc/patches/meson.patch
tjheeta: fixed

user/incus/template.py
```diff
@@ -0,0 +1,81 @@
...
+pkgdesc = "Powerful system container and virtual machine manager"
+maintainer = "tj <tjheeta@gmail.com>"
+license = "Apache-2.0"
+url = "https://github.com/lxc/incus"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "c6ad8260a4954b0e4f77de66220d5450290b57858f8c16ee83f12996eea86db4"
+# void ignores TestConvertNetworkConfig
+# https://github.com/void-linux/void-packages/blob/7dd18621834b9bdb2d4e5b081c820d24fa3180dd/srcpkgs/incus/template#L40
+# go test -v -tags libsqlite3 -skip TestConvertNetworkConfig ./...
+options = ["!check"]
```
triallax, line 53: you could just skip the test instead of !check'ing the whole thing

user/incus/template.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def post_build(self):
+    # Build the agent statically
+    self.env["CGO_ENABLED"] = "1"
+    self.env["CGO_LDFLAGS"] = "-static-pie"
+    self.go_build_tags = ["agent", "netgo", "osusergo"]
+    self.make_build_args = ["./..."]
+    self.golang.build(wrksrc="cmd/incus-agent")
```
triallax, line 62: i don't think you're supposed to modify template attributes inside template functions except for `init_*`, but i'm not quite sure how to do this better

user/incus/template.py
```diff
@@ -0,0 +1,81 @@
...
+    "libuv-devel",
+    "libunwind-devel-static",
+    "lxc-devel",
+    "musl-devel-static",
+    "raft-devel",
+    "sqlite-devel",
+    "udev-devel",
+]
+depends = [
+    "acl",
```
triallax, line 31: `acl` gets pulled in automatically as something in the package links against it, so this shouldn't be needed?

user/lxc/template.py
```diff
@@ -0,0 +1,44 @@
...
+    "-Ddistrosysconfdir=default",
+]
+hostmakedepends = ["git", "meson", "pkgconf"]
```
triallax, line 11: git doesn't seem to be needed
```suggestion
hostmakedepends = ["meson", "pkgconf"]
```

user/raft/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "raft"
+pkgver = "0.22.1"
+pkgrel = 0
+build_style = "gnu_configure"
+make_dir = "."
+hostmakedepends = [
+    "automake",
+    "slibtool",
+    "bash",
```
triallax, line 9: bash doesn't seem to be necessary, or?

user/squashfs-tools/template.py
```diff
@@ -0,0 +1,34 @@
...
+    "LZMA_XZ_SUPPORT=1",
+    "LZ4_SUPPORT=1",
+    "ZSTD_SUPPORT=1",
+]
+make_install_args = [
+    "INSTALL_PREFIX=${DESTDIR}/usr",
+    "INSTALL_MANPAGES_DIR=${DESTDIR}/usr/share/man/man1",
+]
+make_use_env = True
+hostmakedepends = []
```
triallax, line 18:
```suggestion
```

user/incus/template.py
tjheeta: Yah, I wasn't quite sure how.

user/incus/template.py
```diff
@@ -0,0 +1,81 @@
...
+    self.uninstall("usr/bin/generate")
+    self.uninstall("usr/bin/mini-oidc")
+    self.uninstall("usr/bin/sysinfo")
+    self.uninstall("usr/bin/tls2jwt")
+
+
+@subpackage("incus-client")
+def _(self):
+    self.subdesc = "client only"
+    self.depends = [self.parent]
```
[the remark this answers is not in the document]
tjheeta: Hmmm... @nekopsykose suggested to add self.parent, but perhaps I misunderstood the request. 

I think
- The client should be able to be installed independently from the server. 
- If convenient, the server should also install the client. 

What are the correct tags for that behaviour?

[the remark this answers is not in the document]
nekopsykose: depends self.with_pkgver incus-client in toplevel package depends

the client shouldnt depend on parent, mightve said that by mistake but dont remember saying it for thay

user/incus/template.py
triallax: can pass the -skip thing via make_check_args probably, or barring that, skip the test in a patch

nekopsykose:
@triallax not sure how you want to fix this one:  

```
.rwxr-xr-x  46k demon demon 1 minute init.lxc*
.rwxr-xr-x 189k demon demon 1 minute init.lxc.static*
.rwxr-xr-x  73k demon demon 1 minute lxc-attach*
.rwxr-xr-x  57k demon demon 1 minute lxc-autostart*
.rwxr-xr-x  52k demon demon 1 minute lxc-cgroup*
.rwxr-xr-x 8.8k demon demon 1 minute lxc-checkconfig*
.rwxr-xr-x  58k demon demon 1 minute lxc-checkpoint*
.rwxr-xr-x  52k demon demon 1 minute lxc-console*
.rwxr-xr-x  63k demon demon 1 minute lxc-copy*
.rwxr-xr-x  63k demon demon 1 minute lxc-create*
.rwxr-xr-x  55k demon demon 1 minute lxc-destroy*
.rwxr-xr-x  61k demon demon 1 minute lxc-device*
.rwxr-xr-x  65k demon demon 1 minute lxc-execute*
.rwxr-xr-x  51k demon demon 1 minute lxc-freeze*
.rwxr-xr-x  57k demon demon 1 minute lxc-info*
.rwxr-xr-x  74k demon demon 1 minute lxc-ls*
.rwxr-xr-x  68k demon demon 1 minute lxc-monitor*
.rwxr-xr-x  54k demon demon 1 minute lxc-snapshot*
.rwxr-xr-x  66k demon demon 1 minute lxc-start*
.rwxr-xr-x  53k demon demon 1 minute lxc-stop*
.rwxr-xr-x  69k demon demon 1 minute lxc-top*
.rwxr-xr-x  51k demon demon 1 minute lxc-unfreeze*
.rwxr-xr-x  78k demon demon 1 minute lxc-unshare*
.rwxr-xr-x  80k demon demon 1 minute lxc-usernsexec*
.rwxr-xr-x  52k demon demon 1 minute lxc-wait*
$ l /tmp/bldroot/destdir/lxc-bashcomp-6.0.1/usr/share/bash-completion/completions/
.rw-r--r-- 33k demon demon 57 seconds _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-attach -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-autostart -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-cgroup -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-checkpoint -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-config -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-console -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-copy -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-create -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-destroy -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-device -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-execute -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-freeze -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-info -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-ls -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-monitor -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-snapshot -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-start -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-stop -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-top -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-unfreeze -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-unshare -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-usernsexec -> _lxc
lrwxrwxrwx   - demon demon 57 seconds lxc-wait -> _lxc
```

user/incus/template.py
tjheeta: Is there anything for me to do on this? Shall I resolve?

tjheeta:
The completion package was working before I believe. Is there something more I can do here or is that due to a recent change in cbuild?

PR #2785, opened 2024-08-26
contrib/emptty: update to 0.13.0

tvrzna:


nekopsykose:
for future reference, if you find a bug in a release make a new one (0.13.1) instead of retagging, even if you find it really quickly :)  

never know when someone has the tarballs cached

tvrzna:
Noted, however I hope there won't be anything like a discovered bug just after release.
Thank you

PR #2788, opened 2024-08-26
main/linux-*: unselect FB_HYPERV

yoctozepto:
The old framebuffer device of Hyper-V graphics seems to be picked
up first (compared to the new DRM one) but fails to work for either
Wayland or Xorg which load the DRM. At that point, however, DRM is
unable to find the graphics device (as it is managed by hyperv_fb).

Even the help for DRM_HYPERV suggests deselecting FB_HYPERV so that
is what this change does.

The alternative, supposing someone would like to play with the old
FB_HYPERV, would be to blacklist hyperv_fb by default.

[ci skip]


nekopsykose:
thanks, ill roll it into next update

PR #2789, opened 2024-08-26
contrib/python-websockets: update to 0.13

c7skasku:
I just needed this locally for something, I'm not sure what yt-dlp uses it for but it's the only thing that depends on it, so I haven't tested that beyond just checking yt-dlp still runs.


triallax:
please append `, adopt` to the commit message

PR #2791, opened 2024-08-26
user/v2ray: new package

denisstrizhkin:


user/v2ray/template.py
```diff
@@ -0,0 +1,30 @@
...
+url = "https://v2fly.org"
+source = (
+    f"https://github.com/v2fly/v2ray-core/archive/refs/tags/v{pkgver}.tar.gz"
+)
+sha256 = "e6798d1a29f6a52a3c0cc7176803b73e292427bc7858d534d0529a278936b8b0"
+# check: needs network access
+options = ["!check"]
+
+
+def do_build(self):
```
nekopsykose, line 18: you can pass this in make_build_args cant you

user/v2ray/template.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def do_install(self):
+    self.install_bin("build/main", name="v2ray")
+    self.install_license("LICENSE")
+    self.install_service(self.files_path / "v2ray")
+    self.install_file(
+        "release/config/config.json",
+        dest="etc/v2ray",
+        name="config.json.example",
```
triallax, line 26: might want to install this in e.g. /usr/share/examples/v2ray/ instead

PR #2796, opened 2024-08-27
contrib/lua5.4-luasystem: fix missing symbols

ttyyls:


nekopsykose:
is this reported upstream somewhere, or

ttyyls:
no, just something I noticed when executing lua-busted

nekopsykose:
could you send it upstream then? looks like a real issue that should be fixed :)

ttyyls:
fixed in 0.4.4
#2850

PR #2798, opened 2024-08-27
contrib/inotify-tools: new package

c7skasku:


contrib/inotify-tools/template.py
```diff
@@ -0,0 +1,20 @@
+pkgname = "inotify-tools"
+pkgver = "4.23.9.0"
+pkgrel = 0
+build_style = "gnu_configure"
+makedepends = ["automake", "libtool"]
```
nekopsykose, line 5: does it work if you put slibtool here?

contrib/inotify-tools/template.py
```diff
@@ -0,0 +1,20 @@
...
+pkgdesc = "Command-line interface to inotify"
+maintainer = "c7s <c7s@kasku.net>"
+license = "GPL-2.0-only"
+url = "https://github.com/inotify-tools/inotify-tools"
+source = f"https://github.com/inotify-tools/inotify-tools/archive/refs/tags/{pkgver}.tar.gz"
```
nekopsykose, line 10:
```suggestion
source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
```

contrib/inotify-tools/template.py
nekopsykose: yeah

nekopsykose:
> btw I don't know how the symlinks are meant to look, does that all look fine?

you only have to run `./cbuild relink-subpkgs`

contrib/inotify-tools/inotify-tools-devel-static
```diff
@@ -0,0 +1 @@
+inotify-tools-devel-static
```
nekopsykose, line 1: this and the symlink above are inside the inotify-tools folder and so don't actually do anything (and are merely ignored by all the tooling, so there's no errors and also nothing happens, but they shouldn't be there). a funny failure i haven't seen before i guess :D  

in general you never have to make these, just running cbuild relink-subpkgs does everything for you

contrib/inotify-tools/template.py
```diff
@@ -0,0 +1,20 @@
...
+pkgdesc = "Command-line interface to inotify"
+maintainer = "c7s <c7s@kasku.net>"
+license = "GPL-2.0-only"
+url = "https://github.com/inotify-tools/inotify-tools"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "1dfa33f80b6797ce2f6c01f454fd486d30be4dca1b0c5c2ea9ba3c30a5c39855"
+
+
+def post_install(self):
+    self.install_license("COPYING")
```
JamiKettunen, line 15: This wasn't needed for `GPL-2.0-only` btw, it fails to pkg when one isn't installed but should be

PR #2799, opened 2024-08-27
contrib/tesseract: new package

z-erica:


contrib/leptonica/template.py
```diff
@@ -0,0 +1,30 @@
+pkgname = "leptonica"
+pkgver = "1.84.1"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DBUILD_SHARED_LIBS=ON", "-DCMAKE_BUILD_TYPE=Release"]
```
nekopsykose, line 5:
```suggestion
configure_args = ["-DBUILD_SHARED_LIBS=ON"]
```

doesn't need to be set here seemingly

contrib/tesseract/template.py
```diff
@@ -0,0 +1,43 @@
...
+]
+pkgdesc = "OCR engine"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "Apache-2.0"
+url = "https://tesseract-ocr.github.io"
+source = f"https://github.com/tesseract-ocr/tesseract/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "c4bc2a81c12a472f445b7c2fb4705a08bd643ef467f51ec84f0e148bd368051b"
+
+
+@subpackage("libtesseract")
```
nekopsykose, line 30: is there some reason it has to be called libtesseract instead of tesseract-libs?

nekopsykose: sure

contrib/tesseract/template.py
```diff
@@ -0,0 +1,43 @@
...
+def _(self):
+    return self.default_libs(
+        extra=[
+            "usr/lib/libcommon_training.so",
```
nekopsykose, line 34: looking at the makefile used in e.g. alpine, these are never installed at all

i assume they are maybe built as .a and statically linked into libtesseract.so.5 instead?

nekopsykose: hm no libtesseract doesn't use those, so maybe the cli tools

nekopsykose: yeah, just wondering which of the things is linking them wrong

in any case they are just default `add_library`, so with shared_libs=on they are all shared, and without it libtesseract is static. i would use autotools intead here which does all this correctly

nekopsykose: in this case is broke :^)

contrib/tesseract/template.py
```diff
@@ -0,0 +1,78 @@
...
+
+def post_build(self):
+    # also build training tools
+    self.make.build(["training"])
+
+
+def post_install(self):
+    # also install training tools
+    self.make.install(["training-install"])
+
```
nekopsykose, line 41: you can technically put this in make_*_args and they'll be appended and work the same

contrib/tesseract/template.py
nekopsykose: though thinking about it this doesn't matter that much so i'd rather use tesseract-libs/tesseract-devel here anyway

Issue #2800, opened 2024-08-27
take() a binary alongside its manual file if it exists


z-erica:
i feel like this kind of logic could be useful to abstract away somehow:

```py
@subpackage("tesseract-training")
def _(self):
    self.subdesc = "training tools"

    def install():
        for tool in [
            "ambiguous_words",
            "classifier_tester",
            "cntraining",
            "combine_lang_model",
            "combine_tessdata",
            "dawg2wordlist",
            "lstmeval",
            "lstmtraining",
            "merge_unicharsets",
            "mftraining",
            "set_unicharset_properties",
            "shapeclustering",
            "text2image",
            "unicharset_extractor",
            "wordlist2dawg",
        ]:
            self.take(f"usr/bin/{tool}")
            self.take(f"usr/share/man/man1/{tool}.1")

    return install
```

nekopsykose:
making it implicitly sounds bad (take takes a file and any more magic is very not obvious), but a take_bin() that checks if the manpage exists and then also takes it sounds like it would do what you want (even skipping the manpage when it doesn't exist, and checking both .1 and .8 etc)

PR #2802, opened 2024-08-27
 cbuild: add take_bin 

z-erica:
implements #2800


q66:
i wonder if there is any point in doing this considering it only affects a handful of packages and in those where it does it does not make things that much nicer?

PR #2808, opened 2024-08-28
contrib/rrdtool: add rrdcached service

joshdorsey:
Add an initial service file for rrdcached.


joshdorsey:
I looked into the failing test a little bit, and I'm not an expert in RRD's RPN feature so I'm not sure why it's breaking.  On my machine it doesn't tend to break if you rebuild with `./cbuild pkg -f contrib/rrdtool`, so I think it's just inconsistent.

PR #2814, opened 2024-08-29
user/niri: add libEGL as runtime-dependency

tulilirockz:
Whenever running Niri within a virtual machine with OpenGL capabilities (virgl) or a host with an OpenGL-supported CPU Smithay tries to load `libEGL.so` to provide support for OpenGL functionality. LibEGL is also listed as a runtime dependency in most of niri's official packages. Trying to run it within that context just crashes the program, thus it would be nice to have this as a requirement for the package

### DIFF
--- a/user/niri/template.py
+++ b/user/niri/template.py
@@ -1,6 +1,6 @@
 pkgname = "niri"
 pkgver = "0.1.8"
-pkgrel = 0
+pkgrel = 1
 build_style = "cargo"
 make_build_args = [
     "--no-default-features",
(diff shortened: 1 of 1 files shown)


nekopsykose:
> Smithay tries to load libEGL.so to provide support for OpenGL functionality.  

it actually loads libEGL.so.1 and if it loaded the bare .so this dep would not work

classic meme where things don't link to libraries and use dlopen instead..

user/niri/template.py
```diff
@@ -25,6 +25,9 @@
     "rust-std",
     "udev-devel",
 ]
+depends = [
+    "libegl",
```
nekopsykose, line 29:
```suggestion
    "so:libEGL.so.1!libegl",
```

nekopsykose:
you also have to bump the pkgrel for the changes to have an effect on the package

nekopsykose:
and squish the commits into one (git rebase, etc)

thanks! :)

tulilirockz:
There you go, should be good to go now

PR #2816, opened 2024-08-30
Build manpages with asciidoctor

triallax:
why is it so hard to come up with a good title for this pr

### DIFF
--- a/main/asciidoctor/template.py
+++ b/main/asciidoctor/template.py
@@ -1,6 +1,6 @@
 pkgname = "asciidoctor"
 pkgver = "2.0.23"
-pkgrel = 0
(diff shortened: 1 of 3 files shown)


nekopsykose:
you can't link ccache to fmt because it's a base-cbuild dependency so when fmt changes soname (every release) it cannot install the new fmt to build against

triallax:
epic, i'll revert that

nekopsykose:
(it would be possible if we split fmtXX versions so fmtold and fmtnew+fmtnew-devel could exist but also i think there's no point to do that since fmt requires new code changes every release and it's annoying

also making ccache as low overhead as possible is nice just because it's ccache)

PR #2817, opened 2024-08-30
user/anyrun: new package

tulilirockz:
Anyrun is a gtk-based krunner-like application launcher. Everything seems to be building fine, tests are not meaningful (0 tests, 0 succeded, 0 failed). The only issue seems to be with the releases that they have, none! I dont know exactly what to do here, as a workaround I just put the package release as 0.0.0 and pulled off of master

### DIFF
--- a/user/anyrun/template.py
+++ b/user/anyrun/template.py
@@ -1,5 +1,5 @@
 pkgname = "anyrun"
-pkgver = "0.0.0"
+pkgver = "0_git20240716"
 pkgrel = 0
 build_style = "cargo"
 hostmakedepends = ["cargo-auditable", "pkgconf"]
--- a/user/anyrun/template.py
+++ b/user/anyrun/template.py
@@ -15,8 +15,8 @@ pkgdesc = "Wayland native krunner-like runner"
 maintainer = "tulilirockz <tulilirockz@outlook.com>"
 license = "GPL-3.0-or-later"
 url = "https://github.com/anyrun-org/anyrun"
-source = f"{url}/archive/refs/heads/master.tar.gz"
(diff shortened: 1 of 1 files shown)


user/anyrun/template.py
```diff
@@ -0,0 +1,26 @@
...
+    "gtk+3-devel",
+    "gtk-layer-shell-devel",
+    "pango-devel",
+    "rust-std",
+]
+pkgdesc = "Wayland native krunner-like runner"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/anyrun-org/anyrun"
+source = f"{url}/archive/refs/heads/master.tar.gz"
```
EarldridgeJazzedPineda, line 18: Don't use "master", because the repository can change at any time. If you're using Git as the source, then use a specific commit.

user/anyrun/template.py
```diff
@@ -0,0 +1,26 @@
+pkgname = "anyrun"
+pkgver = "0.0.0"
```
EarldridgeJazzedPineda, line 2:
```suggestion
pkgver = "0_git20240716"
```

tulilirockz: Oh! So thats how you do it, thank you very much!

user/anyrun/template.py
tulilirockz: Ill be changing it then! Thanks!

tulilirockz:
Package compiles fine, installs fine, but sadly doesnt work by default without any plugins such as `/etc/anyrun/plugins/libapplications.so`, will add it in a bit

q66:
i'm reluctant to accept this right now considering there are no releases, there is currently barely any development, and pretty much no distro packages this

tulilirockz:
Honestly, completely fair, probably would be better not to merge it at all

PR #2818, opened 2024-08-30
contrib/static-web-server: new package (2.32.2)

jcgruenhage:
Still missing a dinit service, but I'd like to add that later. Just using it for local testing at the moment.

### DIFF
--- a/user/static-web-server/patches/0001-remove-jemalloc.patch
+++ b/user/static-web-server/patches/0001-remove-jemalloc.patch
@@ -1,0 +1,83 @@
+From 8090649470800b5059e9b7cfa3086c7a69d12da2 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Jan=20Christian=20Gr=C3=BCnhage?=
+ <jan.christian@gruenhage.xyz>
+Date: Fri, 30 Aug 2024 08:44:42 +0200
+Subject: [PATCH 1/2] remove jemalloc
+
+---
+ Cargo.lock        | 21 ---------------------
+ Cargo.toml        |  3 ---
+ src/bin/server.rs |  4 ----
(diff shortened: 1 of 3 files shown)


contrib/static-web-server/template.py
```diff
@@ -0,0 +1,34 @@
...
+def post_build(self):
+    self.do(
+        f"target/{self.profile().triplet}/release/static-web-server",
+        env={"SWS_GENERATE": "generated"},
+    )
+
+
+def post_install(self):
+    self.install_license("LICENSE-MIT")
+    self.install_license("LICENSE-APACHE")
```
nekopsykose, line 28:
```suggestion
```

contrib/static-web-server/patches/0003-generate-man-page-and-completions.patch
```diff
@@ -0,0 +1,219 @@
+From 6d3747bb733837686df0f0f655bf0cd7ae774346 Mon Sep 17 00:00:00 2001
```
nekopsykose, line 1: send this upstream and only backport after it's actually merged

nekopsykose:
also put this in user/

jcgruenhage:
Any specific reason for putting this into `user/`?

contrib/static-web-server/patches/0003-generate-man-page-and-completions.patch
jcgruenhage: It's not merged *yet*, but I'm happy with just keeping this open until it is merged upstream. Should happen soon.

jcgruenhage: Patch is merged upstream now.

jcgruenhage:
Well with contrib gone and main definitely not being the right place, user it is.

PR #2827, opened 2024-08-30
user/g4music: new package

tulilirockz:
I really wanted to use this music player as my default in my setup, so heres the package! Apparently everything seems to install properly, the binary and the desktop files :p

### DIFF
--- a/user/g4music/template.py
+++ b/user/g4music/template.py
@@ -1,0 +1,24 @@
+pkgname = "g4music"
+pkgver = "3.9"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "desktop-file-utils",
+    "gettext",
+    "gobject-introspection",
+    "meson",
+    "pkgconf",
+    "vala",
+]
+makedepends = [
+    "gst-plugins-base-devel",
+    "gstreamer-devel",
+    "gtk4-devel",
+    "libadwaita-devel",
+]
+pkgdesc = "GTK4 music player with a fluent adaptive user interface"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "GPL-3.0-or-later"
+url = "https://gitlab.gnome.org/neithern/g4music"
+source = f"{url}/-/archive/v{pkgver}/g4music-v{pkgver}.tar.gz"
+sha256 = "cdc1a79ca579cc211d5da596db4b5791d0899ceafc19dd9028f63b3d25232a36"


user/g4music/template.py
```diff
@@ -0,0 +1,29 @@
...
+    "gettext-devel",
+    "gobject-introspection",
+    "gst-plugins-bad",
+    "gst-plugins-base-devel",
+    "gst-plugins-good",
+    "gst-plugins-ugly",
+    "gstreamer-devel",
+    "gtk4-devel",
+    "libadwaita-devel",
+    "vala-devel"
```
nekopsykose, line 22: this is hostmakedepends (and probably doesn't have to be -devel)

user/g4music/template.py
```diff
@@ -0,0 +1,29 @@
...
+# Only checks the validity of the desktop files
+configure_args = ["-Dunit_tests=disabled"]
+hostmakedepends = [
+    "meson",
+    "pkgconf"
+]
+makedepends = [
+    "desktop-file-utils",
+    "gettext-devel",
+    "gobject-introspection",
```
nekopsykose, line 11: these are hostmakedepends

user/g4music/template.py
```diff
@@ -0,0 +1,29 @@
+pkgname = "g4music"
+pkgver = "3.8.2"
+pkgrel = 0
+build_style = "meson"
+# Only checks the validity of the desktop files
+configure_args = ["-Dunit_tests=disabled"]
```
nekopsykose, line 6: but they run instantly and take zero time, so it's fine to leave them enabled and not touch it

user/g4music/template.py
```diff
@@ -0,0 +1,29 @@
...
+    "pkgconf"
+]
+makedepends = [
+    "desktop-file-utils",
+    "gettext-devel",
+    "gobject-introspection",
+    "gst-plugins-bad",
+    "gst-plugins-base-devel",
+    "gst-plugins-good",
+    "gst-plugins-ugly",
```
nekopsykose, line 15: bad/good/ugly here are not makedepends or checked afaict, at best they're regular runtime depends

but it's fine to leave them out and have them be manually installed later

user/g4music/template.py
nekopsykose: i didn't mean every single dependency is hostmakedepends, just the lines mentioned

nekopsykose: fixed for ya

PR #2832, opened 2024-08-31
user/juce: new package (8.0.1)

z-erica:


nekopsykose:
i'm not sure how this is meant to be used, things that use juce just vendor random commits of it

nekopsykose:
ah, i guess it's actually for projucer?

user/juce/template.py
```diff
@@ -0,0 +1,60 @@
+pkgname = "juce"
+pkgver = "8.0.1"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DCMAKE_BUILD_TYPE=Release",
+    "-DJUCE_BUILD_EXTRAS=ON",
+]
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+_devendored_deps = [
```
nekopsykose, line 10: just inline these, no point in separating it

user/juce/template.py
```diff
@@ -0,0 +1,60 @@
+pkgname = "juce"
+pkgver = "8.0.1"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DCMAKE_BUILD_TYPE=Release",
```
nekopsykose, line 6:
```suggestion
```

doesn't do anything here

user/juce/template.py
```diff
@@ -0,0 +1,60 @@
...
+        self.rm(path, recursive=True)
+
+
+def post_install(self):
+    self.install_bin(
+        "build/extras/Projucer/Projucer_artefacts/Release/Projucer"
+    )
+
+
+@subpackage("projucer")
```
nekopsykose, line 55: you want a -dev too if only for the cmake part

user/juce/template.py
```diff
@@ -0,0 +1,60 @@
...
+source = (
+    f"https://github.com/juce-framework/JUCE/archive/refs/tags/{pkgver}.tar.gz"
+)
+sha256 = "1a5aae997075ae2df045793fd47a4dc16234b2a7939d445fea17075219397b05"
+
+
+def post_extract(self):
+    # ensure these are never vendored
+    for path in [
+        "modules/juce_audio_formats/codecs/flac/libFLAC",
```
nekopsykose, line 40: there's also apparently a whole harfbuzz in there, but dunno how hard that is

user/juce/template.py
nekopsykose: o rite tru

nekopsykose:
if its possible then its ok

PR #2833, opened 2024-08-31
contrib/tcsh: new package

tulilirockz:
This PR adds the TCSH shell to Chimera! It seems to be building completely fine and running fine, the only issue is the test suite that requires all the locales and tcsh to be installed (not possible due to /usr/* being ro during check step)

### DIFF
--- a/contrib/tcsh/template.py
+++ b/contrib/tcsh/template.py
@@ -1,0 +1,18 @@
+pkgname = "tcsh"
+pkgver = "6.24.13"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--disable-nls"]
+hostmakedepends = ["automake"]
+makedepends = ["ncurses-devel"]
+pkgdesc = "Enhanced version of the Berkeley UNIX C shell"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "BSD-2-Clause"
+url = "https://astron.com/pub/tcsh"
+source = f"{url}/tcsh-{pkgver}.tar.gz"
+sha256 = "1e927d52e9c85d162bf985f24d13c6ccede9beb880d86fec492ed15480a5c71a"
+
+
+def post_install(self):
+    self.install_shell("/usr/bin/tcsh")
+    self.install_license("Copyright")


contrib/tcsh/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "tcsh"
+pkgver = "6.24.13"
+pkgrel = 0
+build_style = "configure"
```
nekopsykose, line 4: this looks like gnu_configure and not configure

contrib/tcsh/template.py
```diff
@@ -0,0 +1,23 @@
...
+hostmakedepends = ["autoconf", "automake"]
```
nekopsykose, line 5:
```suggestion
hostmakedepends = ["automake"]
```

nekopsykose:
> the only issue is the test suite that requires all the locales and tcsh to be installed

i don't think that's the case? just the default `make check` seems to run them fine, and alpine runs them like that

contrib/tcsh/template.py
```diff
@@ -0,0 +1,23 @@
...
+license = "BSD-2-Clause"
+url = "https://astron.com/pub/tcsh"
+source = f"{url}/tcsh-{pkgver}.tar.gz"
+sha256 = "1e927d52e9c85d162bf985f24d13c6ccede9beb880d86fec492ed15480a5c71a"
+# check: requires both the tcsh binary and locales to be installed in /usr/* (read only at check step)
+options = ["!check"]
+
+
+def install(self):
+    self.do("make", "-C", "nls", f"prefix={self.chroot_destdir}/usr", "install")
```
nekopsykose, line 18: it has a normal `make install` so i'm not sure why you're calling this manually

prefix=destdir/usr is also incorrect

you can make this post_install and remove this .do and it becomes correct since the makefile already does this, and installs the things below too, so all you actuall yneed to install is the Copyright and install_shell

nekopsykose: i.e.

```
def post_install(self):
    self.install_license("Copyright")
    self.install_shell("/usr/bin/tcsh")
```

nekopsykose:
> i don't think that's the case? just the default make check seems to run them fine, and alpine runs them like that

or to be specific, they segfault, but that's because the code is broken:

```
#ifdef NL_CAT_LOCALE /* POSIX-compliant. */
    /*
     * Check if LC_MESSAGES is set in the environment and use it, if so.
     * If not, fall back to the setting of LANG.
     */
    catd = catopen(catalog, tgetenv(STRLC_MESSAGES) ? NL_CAT_LOCALE : 0);
#else /* pre-POSIX */
# ifndef MCLoadBySet
#  define MCLoadBySet 0
#  endif
    catd = catopen(catalog, MCLoadBySet);
#endif
    if (catalog != default_catalog)
	xfree(catalog);
#if defined(HAVE_ICONV) && defined(HAVE_NL_LANGINFO)
    /* xcatgets (), not CGETS, the charset name should be in ASCII anyway. */
    catgets_iconv = iconv_open (nl_langinfo (CODESET),
				xcatgets(catd, 255, 1, "UTF-8"));
```

the `catopen` fails but they still pass the `catd`  to `catgets` so it segfaults.

this means the shell itself is broken:  

```
$ tcsh
fish: Job 1, 'tcsh' terminated by signal SIGSEGV (Address boundary error)
```

which doesn't make much sense and should be fixed...

interestingly this fails even with the locales for it installed. does it work for you?

nekopsykose:
the only way i can get it to work at all is with LANG=C

q66:
just add `configure_args = ["--disable-nls"]`

q66:
still not sure why anybody would actually want to use csh though :p

tulilirockz:
> just add configure_args = ["--disable-nls"]
oh my god that just fixed everthing

contrib/tcsh/template.py
tulilirockz: Makes sense! Changing it.

tulilirockz:
> > the only issue is the test suite that requires all the locales and tcsh to be installed
> 
> i don't think that's the case? just the default `make check` seems to run them fine, and alpine runs them like that

I have really no idea why, but I was getting so many issues when doing it without adding the locales to the system? TCSH just refused to open with the LANG=C + no locales installed environment

contrib/tcsh/template.py
tulilirockz: > it has a normal `make install` so i'm not sure why you're calling this manually
> 
> prefix=destdir/usr is also incorrect
> 
> you can make this post_install and remove this .do and it becomes correct since the makefile already does this, and installs the things below too, so all you actuall yneed to install is the Copyright and install_shell

For some reason it just refused to work in my environment, I really could not get the thing installed without that

tulilirockz:
Well, apparently just changing the build_style to gnu_configure fixed literally just everything LOL god damn, here is the proper package now then

PR #2815, opened 2024-08-30
user/wpaperd: new package

tulilirockz:
Wpaperd is a wayland-only daemon for managing wallpapers in wayland compositors through wlr_layer_shell. This package should work fine, the wpaperd and wpaperctl binaries seem to be fine, I just dont currently know how to add the scdoc-generated documentation, what could I do here?

PR #2835, opened 2024-09-01
user/wpaperd: add completions for bash/zsh/fish

tulilirockz:
[As @triallax mentioned](https://github.com/chimera-linux/cports/pull/2815#issuecomment-2319953690), there are no installed completions in the initial wpaperd package version, this fixes it!

### DIFF
--- a/user/wpaperd/template.py
+++ b/user/wpaperd/template.py
@@ -1,6 +1,6 @@
 pkgname = "wpaperd"
 pkgver = "1.0.1"
-pkgrel = 0
+pkgrel = 1
 build_style = "cargo"
 hostmakedepends = [
     "cargo-auditable",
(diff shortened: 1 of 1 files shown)


user/wpaperd/template.py
```diff
@@ -25,7 +25,15 @@ def post_build(self):
...
+    with self.pushd(f"target/{self.profile().triplet}/release"):
+        self.install_bin("wpaperd")
+        self.install_bin("wpaperctl")
+        with self.pushd("completions"):
+            self.install_completion("wpaperd.bash", "bash")
+            self.install_completion("wpaperctl.bash", "bash")
+            self.install_completion("_wpaperd", "zsh")
+            self.install_completion("_wpaperctl", "zsh")
+            self.install_completion("wpaperd.fish", "fish")
+            self.install_completion("wpaperctl.fish", "fish")
```
nekopsykose, line 39: you have to pass "wpaperctl" as the third arg for those since the default name is the pkgname, which means this currently only installs a wpaperd-named completion

tulilirockz: oh, fair

PR #2840, opened 2024-09-02
main/openssh: enable kerberos5 support

yoctozepto:
Enables the GSSAPI with krb5 using Heimdal.


yoctozepto:
The reasoning being: Heimdal is available in main and all major distros support GSSAPI with krb5.

PR #2843, opened 2024-09-03
user/fragments: new package (3.0.1)

tulilirockz:
The Fragments GTK4 Bittorrent Client! Had to use a workaround to make the vendoring work with the weird cargo/meson hybrid system but everything seems to be working fine. This requires transmission-daemon, which is not yet packaged, so: WIP!

### DIFF
--- a/user/fragments/template.py
+++ b/user/fragments/template.py
@@ -1,0 +1,46 @@
+pkgname = "fragments"
+pkgver = "3.0.1"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "cargo-auditable",
+    "desktop-file-utils",
+    "gettext",
+    "git",
+    "glib-devel",
+    "gtk-update-icon-cache",
+    "meson",
+    "pkgconf",
+]
+makedepends = [
+    "dbus-devel",
+    "libadwaita-devel",
+    "openssl-devel",
+    "rust-std",
+]
+depends = ["transmission-daemon"]
+pkgdesc = "BitTorrent client for the GNOME desktop environment"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "GPL-3.0-or-later"
+url = "https://gitlab.gnome.org/World/Fragments"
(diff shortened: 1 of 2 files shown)


nekopsykose:
see contrib/helvum for how vendoring works with this setup, you don't need the mkdirs with that

user/fragments/template.py
```diff
@@ -0,0 +1,31 @@
...
+hostmakedepends = [
+    "cargo-auditable",
+    "meson",
+    "pkgconf",
+    "git",
+    "gettext",
+    "desktop-file-utils",
+]
+makedepends = [
+    "dbus-devel",
```
nekopsykose, line 14: also need rust-std here

user/fragments/template.py
```diff
@@ -0,0 +1,31 @@
...
+    "openssl-devel",
+    "glib-devel",
+    "gtk4-devel",
```
nekopsykose, line 17: libadwaita already pulls gtk4 and glib for you

user/fragments/template.py
```diff
@@ -0,0 +1,31 @@
+pkgname = "fragments"
+pkgver = "3.0.1"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "cargo-auditable",
+    "meson",
+    "pkgconf",
+    "git",
```
nekopsykose, line 9: sort this list

nekopsykose:
> This requires transmission-daemon, which is not yet packaged, so: WIP!

i'll probably package this later if you don't mind waiting a day or two

tulilirockz:
Im packaging it right now! It requires libnatpmp, so, another WIP.

user/fragments/template.py
tulilirockz: aight

nekopsykose:
> It requires libnatpmp

that's already in the tarball it comes with, it's a submodule and not a separate package

nekopsykose:
you want the init_build part too

PR #2853, opened 2024-09-04
contrib/yt-dlp: add mutagen as runtime dependency

tulilirockz:
This MR adds the mutagen audio processing module to yt-dlp as a runtime package, it is necessary for the `--audio-format` (or any other audio processing) argument to work

### DIFF
--- a/contrib/yt-dlp/template.py
+++ b/contrib/yt-dlp/template.py
@@ -9,7 +9,7 @@ hostmakedepends = [
     "python-installer",
     "python-wheel",
 ]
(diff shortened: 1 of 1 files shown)


nekopsykose:
that and like 50 other deps are also required for various --args, but they are all optional, that's why they're not there

tulilirockz:
Fair!

nekopsykose:
it might be nice to have a uhh `yt-dlp-something` subpackage that is `install_if = [self.parent]` though, that just pulls default useful deps

PR #2854, opened 2024-09-04
contrib/virt-manager: add virglrenderer as runtime dependency

tulilirockz:
VirglRenderer is a renderer for OpenGL in virtual machines, providing hardware-accelerated GPU processing to the virtual machines through OpenGL using the SPICE protocol, although it is not strictly required for virt-manager, it is required for the "use OpenGL" functionality in SPICE for the virtual machines.

### DIFF
--- a/contrib/virt-manager/template.py
+++ b/contrib/virt-manager/template.py
@@ -52,6 +52,7 @@ depends = [
     "gtksourceview4",
     "libvirt-glib",
     "spice-gtk",
+    "virglrenderer",
     "vte-gtk3",
 ]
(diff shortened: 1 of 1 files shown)


nekopsykose:
you have to bump the pkgrel too

nekopsykose:
actually i'm not sure when this changes anything, since all of qemu depends on virglrenderer already, so it's always pulled. how do you end up without it?

tulilirockz:
i literally just found out i didnt even have qemu on my system, this PR is entirely just a fluke LOL im sorry

PR #2855, opened 2024-09-05
user/newsflash: new package (3.3.4)

tulilirockz:
Had to do a workaround because the last part of the build tried to copy a non-existent file over to usr/bin, making a stub then installing it properly seems to work fine!

### DIFF
--- a/user/newsflash/template.py
+++ b/user/newsflash/template.py
@@ -1,0 +1,46 @@
+pkgname = "newsflash"
+pkgver = "3.3.4"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "blueprint-compiler",
+    "cargo-auditable",
+    "desktop-file-utils",
+    "gettext",
+    "meson",
+    "pkgconf",
+]
+makedepends = [
+    "clapper-devel",
+    "libadwaita-devel",
+    "libxml2-devel",
+    "openssl-devel",
+    "rust-std",
+    "webkitgtk4-devel",
+]
+pkgdesc = "Feed reader designed for the GNOME desktop"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "GPL-3.0-or-later"
+url = "https://gitlab.com/news-flash/news_flash_gtk"
+source = f"{url}/-/archive/v.{pkgver}/news_flash_gtk-v.{pkgver}.tar.gz"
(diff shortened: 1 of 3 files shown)


user/newsflash/template.py
```diff
@@ -0,0 +1,53 @@
...
+    self.mkdir(self.cwd / "build/src/release", parents=True)
+    (self.cwd / "build/src/release/news_flash_gtk").touch(0o755)
+
+
+def post_install(self):
+    self.uninstall("usr/bin/io.gitlab.news_flash.NewsFlash")
+    self.install_bin(
+        f"build/src/{self.profile().triplet}/release/news_flash_gtk",
+        name="io.gitlab.news_flash.NewsFlash",
+    )
```
nekopsykose, line 53: look at how loupe does it

user/newsflash/template.py
```diff
@@ -0,0 +1,53 @@
...
+pkgver = "3.3.4"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "blueprint-compiler",
+    "cargo-auditable",
+    "desktop-file-utils",
+    "gettext",
+    "meson",
+    "ninja",
```
nekopsykose, line 11:
```suggestion
```

user/newsflash/template.py
tulilirockz: oh. my. god? thats why i didnt understand what they were doing differently before! they literally just patch the god damn meson.build! sorry for the hacks, thats much cleaner

nekopsykose: yea, and don't forget to remove the CARGO_HOME part of the env thing in src/meson.build too (pretty much all the cargo+meson projects are the same)

nekopsykose:
for the update check, you'll need to an overriding update.py with perhaps a

```
url = "https://gitlab.com/news-flash/news_flash_gtk/-/tags"
pattern = r"news_flash_gtk-v.([\d.]+).tar"
```

the versioning scheme has a `.` in it so the default patterns don't match

nekopsykose:
(the update check gets ran daily and thrown into https://repo.chimera-linux.org/cports-updates/cports-updates.txt)

tulilirockz:
The update.py thing you suggested seems to be working fine!
```
❯ ./cbuild update-check user/newsflash
user/newsflash: 3.3.3 -> 3.3.4
```

PR #2859, opened 2024-09-05
contrib/minetest: update to 5.9.0

ttyyls:


contrib/minetest/template.py
```diff
@@ -34,17 +42,23 @@
...
+sha256 = "070bc292a0b7fc60d7ff0a14b364c8229c5cbe38296a80f948ea2c2591545a5c"
 hardening = ["!int"]
+# fails to resolve the vendored irrlicht engine symlink
+options = ["brokenlinks"]
+
+
+def init_configure(self):
+    if self.has_lto():
+        self.configure_args += ["-DENABLE_LTO=Thin"]
+
```
nekopsykose, line 57: you don't need this because we already set lto via cflags

contrib/minetest/template.py
```diff
@@ -34,17 +42,23 @@
...
-    "2fde8e27144988210b9c0ff1e202905834d9d25aaa63ce452763fd7171096adc",
-]
-tool_flags = {"CXXFLAGS": ["-Wno-deprecated-declarations"]}
+source = (
+    f"https://github.com/minetest/minetest/archive/refs/tags/{pkgver}.tar.gz"
+)
+sha256 = "070bc292a0b7fc60d7ff0a14b364c8229c5cbe38296a80f948ea2c2591545a5c"
 hardening = ["!int"]
+# fails to resolve the vendored irrlicht engine symlink
+options = ["brokenlinks"]
```
nekopsykose, line 51: this looks like a symlink that should be deleted instead of ignored

if it has to be there, use broken_symlinks instead

contrib/minetest/template.py
nekopsykose:
```suggestion
```

contrib/minetest/template.py
```diff
@@ -21,11 +29,12 @@
     "libogg-devel",
     "libpng-devel",
     "libvorbis-devel",
-    "libxi-devel",
```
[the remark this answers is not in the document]
nekopsykose: so does mesa-devel but that's not relevant because they are unrelated to eachother and checked for directly

contrib/minetest/template.py
```diff
@@ -1,14 +1,21 @@
 pkgname = "minetest"
-pkgver = "5.8.0"
-_irrlichtver = "1.9.0mt13"
-pkgrel = 1
+pkgver = "5.9.0"
+pkgrel = 0
 build_style = "cmake"
 configure_args = [
     "-DBUILD_SERVER=TRUE",
+    "-DCMAKE_BUILD_TYPE=Release",
```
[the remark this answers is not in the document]
nekopsykose: actually, don't need the patch in this case so can leave that out

contrib/minetest/template.py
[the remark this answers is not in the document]
ttyyls: fair enough

contrib/minetest/template.py
```diff
@@ -1,14 +1,23 @@
...
+    "-DDEVELOPMENT_BUILD=FALSE",
```
nekopsykose, line 8: this is already off by default

contrib/minetest/template.py
```diff
@@ -34,32 +42,33 @@
...
+    self.do("bin/minetestserver", "--run-unittests")
+
+
 def post_install(self):
     self.install_file(
         "minetest.conf.example",
         "etc/minetest",
         name="minetest.conf",
     )
+    self.rm("usr/share/minetest/client/shaders/Irrlicht")
```
nekopsykose, line 64:
```suggestion
    # dead symlink
    self.uninstall("usr/share/minetest/client/shaders/Irrlicht")
```

(the rm will fail since there's no destdir prefix)

PR #2865, opened 2024-09-07
contrib/nextcloud-client: new package

jabuxas:
working on it

### DIFF
--- a/contrib/nextcloud-client/template.py
+++ b/contrib/nextcloud-client/template.py
@@ -1,0 +1,33 @@
+pkgname = "nextcloud-client"
+pkgver = "3.14.0_rc1"
+pkgrel = 0
+build_style = "cmake"
+make_check_env = {"QT_QPA_PLATFORM": "offscreen"}
+hostmakedepends = [
+    "cmake",
+    "extra-cmake-modules",
+    "inkscape",
+    "ninja",
+    "openssl-devel",
+    "pkgconf",
+]
+makedepends = [
+    "cmocka-devel",
+    "karchive-devel",
+    "kguiaddons-devel",
+    "kio-devel",
+    "qt6-qt5compat-devel",
+    "qt6-qtsvg-devel",
+    "qt6-qttools-devel",
+    "qt6-qtwebengine-devel",
+    "qt6-qtwebsockets-devel",
+    "qtkeychain-devel",
+    "sqlite-devel",
+    "zlib-ng-compat-devel",
+]
+pkgdesc = "Desktop sync client for Nextcloud"
+maintainer = "jabuxas <jabuxas@proton.me>"
+license = "GPL-2.0-only"
+url = "https://github.com/nextcloud/desktop"
+source = f"{url}/archive/refs/tags/v{pkgver.replace('_', '-')}.tar.gz"
+sha256 = "21089b8408f3a7a40ad5db3bf9096b437e2e96460e32e541063454ba52115351"
--- a/contrib/nextcloud-client/patches/statvfs64fix.patch
+++ b/contrib/nextcloud-client/patches/statvfs64fix.patch
@@ -1,0 +1,13 @@
+--- a/src/common/utility.cpp    2024-03-06 03:37:52.000000000 -0600
++++ b/src/common/utility.cpp    2024-03-11 13:54:42.266466151 -0500
+@@ -192,8 +192,8 @@ qint64 Utility::freeDiskSpace(const QStr
+         return (qint64)stat.f_bavail * stat.f_frsize;
+     }
+ #elif defined(Q_OS_UNIX)
+-    struct statvfs64 stat{};
+-    if (statvfs64(path.toLocal8Bit().data(), &stat) == 0) {
++    struct statvfs stat{};
++    if (statvfs(path.toLocal8Bit().data(), &stat) == 0) {
+         return (qint64)stat.f_bavail * stat.f_frsize;
+     }
+ #elif defined(Q_OS_WIN)


jabuxas:
what is subprocess aborted in the build ci

contrib/nextcloud-sync/template.py
```diff
@@ -0,0 +1,34 @@
+pkgname = "nextcloud-sync"
+pkgver = "3.14.0"
```
nekopsykose, line 2: it's not 3.14.0 if it's actually some tag with rc1 in the name, so this pkgver is wrong

contrib/nextcloud-sync/template.py
```diff
@@ -0,0 +1,34 @@
...
+pkgdesc = "Desktop sync client for Nextcloud"
+maintainer = "jabuxas <jabuxas@proton.me>"
+license = "GPL-2.0-only"
+url = "https://github.com/nextcloud/desktop"
+source = f"{url}/archive/refs/tags/v{_pkgver}.tar.gz"
+sha256 = "21089b8408f3a7a40ad5db3bf9096b437e2e96460e32e541063454ba52115351"
+
+
+def post_install(self):
+    self.install_license("COPYING")
```
nekopsykose, line 34: gpl licences don't have to be installed

contrib/nextcloud-sync/template.py
```diff
@@ -0,0 +1,34 @@
...
+build_style = "cmake"
+hostmakedepends = [
+    "cmake",
+    "inkscape",
+    "ninja",
+    "openssl-devel",
+    "pkgconf",
+]
+makedepends = [
+    "karchive-devel",
```
nekopsykose, line 14: missing kio and cmocka

nekopsykose:
> what is subprocess aborted in the build ci

it's the exact same error you get when you build this locally too and not related to the ci. you skipped the tests :)

for this, it's because qt wants a display (or aborts otherwise), so you can use `make_check_env = {"QT_QPA_PLATFORM": "offscreen"}`

contrib/nextcloud-sync/template.py
nekopsykose: (yes, i know before 3.14.0_rc1 there is no qt6 support, but put that version instead of a made up one)

but even better, wait for the actual 3.14.0

contrib/nextcloud-sync/template.py
```diff
@@ -0,0 +1,34 @@
+pkgname = "nextcloud-sync"
```
nekopsykose, line 1:
```suggestion
pkgname = "nextcloud-client"
```

contrib/nextcloud-sync/template.py
jabuxas: well it's because it has a license and a copying but sure

contrib/nextcloud-sync/template.py
jabuxas: sure but it still doesn't fix the need of having a _pkgver since it is -rc1 and not _rc1

nekopsykose: yes it does, source= can have {pkgver.replace("_", "-")}

nekopsykose: (i also don't know what that has to do with the pkgver being wrong)

contrib/nextcloud-sync/template.py
jabuxas: oh yea true, i thought copying was something else, didnt see that it was gpl2 inside of it

contrib/nextcloud-sync/template.py
jabuxas: > yes it does, source= can have {pkgver.replace("_", "-")}

oh yea true

jabuxas: > (i also don't know what that has to do with the pkgver being wrong)

i dont get what u mean, u saying 3.14 is not a version?

nekopsykose: > i dont get what u mean, u saying 3.14 is not a version?

the version is `3.14.0-rc1`. this is not `3.14.0`, in the same way it is not `921840128412o1i2o4u214`

the version has to be remapped in some way that apk would accept (so, - -> _), but if you call it `3.14.0`, it gets merged as `3.14.0`, then it a) looks like an actual 3.14.0 is out (it isn't), and b) later when the actual 3.14.0 comes out you would ""upgrade"" it into the same version (which doesn't make sense)

contrib/nextcloud-client/template.py
```diff
@@ -0,0 +1,33 @@
...
+hostmakedepends = [
+    "cmake",
+    "inkscape",
+    "ninja",
+    "openssl-devel",
+    "pkgconf",
+]
+makedepends = [
+    "cmocka-devel",
+    "extra-cmake-modules",
```
nekopsykose, line 15:
```suggestion
    "extra-cmake-modules",
```

hostmakedepends

contrib/nextcloud-sync/template.py
jabuxas: i see, then it is fixed, 3.14.0_rc1

contrib/nextcloud-client/template.py
jabuxas: ok

jabuxas:
is your divine intuition saying something?

nekopsykose:
it's saying 'but wait for the real non-rc release anyway', even rc2 is out now so any day now :)

jabuxas:
ok i might try nextcloud-server maybe not sure

doskel:
Far's I can tell, the RC and some point releases are out by now; do we want to try and get this updated and merged?

jabuxas:
i haven't been using chimera lately so I'll close the pull request so someone can open a new one

PR #2866, opened 2024-09-07
user/waydroid: new package

tulilirockz:
This library + its dev variant is required for libgbinder to be built

### DIFF
--- a/user/python-gbinder/patches/cython.patch
+++ b/user/python-gbinder/patches/cython.patch
@@ -1,0 +1,11 @@
+--- a/setup.py	2023-09-16 14:58:25.000000000 -0300
++++ b/setup.py	2024-09-08 13:48:40.503581594 -0300
+@@ -10,7 +10,7 @@
+         kw.setdefault(flag_map.get(token[:2]), []).append(token[2:])
+     return kw
+ 
+-USE_CYTHON = False
++USE_CYTHON = True
+ if "--cython" in sys.argv:
+     sys.argv.remove("--cython")
+     USE_CYTHON = True
--- a/user/waydroid/patches/apparmor-config.patch
+++ b/user/waydroid/patches/apparmor-config.patch
@@ -1,0 +1,9 @@
+--- a/data/configs/config_3	2024-08-07 18:01:14.000000000 -0300
++++ b/data/configs/config_3	2024-09-07 14:21:28.466488580 -0300
+@@ -1,6 +1,5 @@
+ lxc.uts.name = waydroid
+ 
+-lxc.apparmor.profile = unconfined
+ lxc.seccomp.profile = /var/lib/waydroid/lxc/waydroid/waydroid.seccomp
+ 
+ lxc.no_new_privs = 1
--- a/user/waydroid/patches/sed-chimera.patch
+++ b/user/waydroid/patches/sed-chimera.patch
@@ -1,0 +1,17 @@
+--- a/tools/helpers/lxc.py	2024-09-07 13:39:38.419147354 -0300
++++ b/tools/helpers/lxc.py	2024-09-07 14:10:55.719543931 -0300
+@@ -157,12 +157,12 @@
+     tools.helpers.run.user(args, command)
+     command = ["sh", "-c", "cat {} > \"{}\"".format(' '.join('"{0}"'.format(w) for w in config_snippets), lxc_path + "/config")]
+     tools.helpers.run.user(args, command)
+-    command = ["sed", "-i", "s/LXCARCH/{}/".format(platform.machine()), lxc_path + "/config"]
++    command = ["sed", "-i", "\'\'","s/LXCARCH/{}/".format(platform.machine()), lxc_path + "/config"]
+     tools.helpers.run.user(args, command)
+     command = ["cp", "-fpr", seccomp_profile, lxc_path + "/waydroid.seccomp"]
+     tools.helpers.run.user(args, command)
+     if get_apparmor_status(args):
+-        command = ["sed", "-i", "-E", "/lxc.aa_profile|lxc.apparmor.profile/ s/unconfined/{}/g".format(LXC_APPARMOR_PROFILE), lxc_path + "/config"]
++        command = ["sed", "-i", "\'\'", "-E", "/lxc.aa_profile|lxc.apparmor.profile/ s/unconfined/{}/g".format(LXC_APPARMOR_PROFILE), lxc_path + "/config"]
+         tools.helpers.run.user(args, command)
+ 
+     nodes = generate_nodes_lxc_config(args)
--- a/user/waydroid/template.py
+++ b/user/waydroid/template.py
@@ -1,0 +1,35 @@
+pkgname = "waydroid"
+pkgver = "1.4.3"
+pkgrel = 0
+build_style = "makefile"
+make_install_args = ["USE_SYSTEMD=0", "USE_NFTABLES=1"]
+depends = [
+    "dnsmasq",
+    "lxc",
+    "python-dbus",
+    "python-gbinder",
+    "python-gobject",
+    "python-pyclip",
+]
+pkgdesc = "Container-based approach to boot a full Android system"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/waydroid/waydroid"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "6557c6fed6a0a7417503eaaab3602efd67c6ced2026725ac24ec8c809fc672e4"
+# check: no tests
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_service(self.files_path / "waydroid-container")
+    self.install_file(
+        self.files_path / "51_waydroid.nft",
+        "etc/nftables.d",
+        name="51_waydroid.nft",
+    )
+    self.install_file(
+        self.files_path / "modules-load.conf",
+        "usr/lib/modules-load.d",
+        name="waydroid.conf",
+    )
--- a/user/libgbinder-devel
+++ b/user/libgbinder-devel
@@ -1,0 +1,1 @@
+libgbinder
--- a/user/libgbinder/template.py
+++ b/user/libgbinder/template.py
@@ -1,0 +1,24 @@
+pkgname = "libgbinder"
+pkgver = "1.1.40"
+pkgrel = 0
+build_style = "makefile"
+make_install_target = "install-dev"
+make_check_target = "test"
+make_use_env = True
+hostmakedepends = ["pkgconf"]
+makedepends = ["libglibutil-devel", "linux-headers"]
+pkgdesc = "GLib-style interface to binder"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "BSD-3-Clause"
+url = "https://github.com/mer-hybris/libgbinder"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "9e86243df6502ffd0a68ee8384c5c36b9cd4093733ea620313f1947f312abbd1"
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+
+
+@subpackage("libgbinder-devel")
+def _(self):
+    return self.default_devel()
--- a/user/libglibutil-devel
+++ b/user/libglibutil-devel
@@ -1,0 +1,1 @@
+libglibutil
--- a/user/libglibutil/template.py
+++ b/user/libglibutil/template.py
@@ -1,0 +1,24 @@
+pkgname = "libglibutil"
+pkgver = "1.0.79"
+pkgrel = 0
+build_style = "makefile"
+make_install_target = "install-dev"
+make_check_target = "test"
+make_use_env = True
+hostmakedepends = ["pkgconf"]
+makedepends = ["glib-devel"]
+pkgdesc = "Library of glib utilities"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "BSD-3-Clause"
+url = "https://github.com/sailfishos/libglibutil"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "4d689cb79f8ea061e46b89008370dc771b07164ee496850d9d56d9d85f1be1c3"
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+
+
+@subpackage("libglibutil-devel")
+def _(self):
+    return self.default_devel()
--- a/user/python-gbinder/template.py
+++ b/user/python-gbinder/template.py
@@ -1,0 +1,23 @@
+pkgname = "python-gbinder"
+pkgver = "1.1.2"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "pkgconf",
+    "python-build",
+    "python-cython",
+    "python-installer",
+    "python-setuptools",
+]
+makedepends = [
+    "libgbinder-devel",
+    "python-devel",
+]
+pkgdesc = "Python bindings for libgbinder"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/erfanoabdi/gbinder-python"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "2dc424d5c2594146612e4bd752964f8928a62eec7c5ce6046f4c582079d0b537"
+# check: needs manually running script in .py with manual setup
+options = ["!check"]
--- a/user/waydroid/files/51_waydroid.nft
+++ b/user/waydroid/files/51_waydroid.nft
@@ -1,0 +1,10 @@
+#!/usr/sbin/nft -f
+table inet filter {
+    chain input {
+        iifname "waydroid0" accept comment "Allow incoming network traffic from WayDroid"
+    }
+    chain forward {
+        iifname "waydroid0" accept comment "Allow outgoing network traffic from WayDroid"
+        ct state {established, related} counter accept comment "accept established connections"
+    }
+}
--- a/user/waydroid/files/modules-load.conf
+++ b/user/waydroid/files/modules-load.conf
@@ -1,0 +1,2 @@
+fuse
+loop
--- a/user/waydroid/files/waydroid-container
+++ b/user/waydroid/files/waydroid-container
@@ -1,0 +1,6 @@
+type = process
+command = /usr/bin/waydroid -w container start
+stop-command = /usr/bin/waydroid -w container stop
+depends-on = network.target
+depends-on = local.target
+logfile = /var/log/waydroid-container.log
--- a/user/waydroid/patches/nft-install.patch
+++ b/user/waydroid/patches/nft-install.patch
@@ -1,0 +1,11 @@
+--- Makefile	2024-08-07 18:01:14.000000000 -0300
++++ Makefile	2024-09-08 13:55:52.827469906 -0300
+@@ -54,7 +54,7 @@
+ 		cp systemd/waydroid-container.service $(INSTALL_SYSD_DIR); \
+ 	fi
+ 	if [ $(USE_NFTABLES) = 1 ]; then \
+-		sed '/LXC_USE_NFT=/ s/false/true/' -i $(INSTALL_WAYDROID_DIR)/data/scripts/waydroid-net.sh; \
++		sed -i '' '/LXC_USE_NFT=/ s/false/true/' $(INSTALL_WAYDROID_DIR)/data/scripts/waydroid-net.sh; \
+ 	fi
+ 
+ install_apparmor:


nekopsykose:
> This library + its dev variant is required for libgbinder to be built

then package along with 'libgbinder'. or more specifically, with what actually uses these libraries

contrib/libglibutil/template.py
```diff
@@ -0,0 +1,27 @@
...
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "4d689cb79f8ea061e46b89008370dc771b07164ee496850d9d56d9d85f1be1c3"
+# check: no tests
+options = ["!check"]
+
+
+def install(self):
+    self.do(
+        "make", "STRIP=true", f"DESTDIR={self.chroot_destdir}", "install-dev"
+    )
```
nekopsykose, line 21:
```suggestion
def post_install(self):
```

make_install_target = "install-dev"

tulilirockz:
Oh! Alright! I wanted to package waydroid, and it required a cascade of dependencies -> waydroid -> python-gbinder -> libgbinder -> libglibutil. Didnt know I could do that :p

contrib/libglibutil/template.py
tulilirockz: thx!

contrib/libglibutil/template.py
```diff
@@ -0,0 +1,27 @@
...
+hostmakedepends = ["pkgconf"]
+makedepends = ["glib-devel"]
+pkgdesc = "Library of glib utilities"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "BSD-3-Clause"
+url = "https://github.com/sailfishos/libglibutil"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "4d689cb79f8ea061e46b89008370dc771b07164ee496850d9d56d9d85f1be1c3"
+# check: no tests
+options = ["!check"]
```
nekopsykose, line 15:
```suggestion
```

make_check_target = "test"

tulilirockz:
Ill be opening another PR for waydroid and its dependencies, kay?

nekopsykose:
just put them all in this pr

tulilirockz:
honestly this race condition im running into in libgbinder is the weirdest god damn bug ever. apparently after N tests, make just craps out and cant include `../common/Makefile` again and segfaults

contrib/libgbinder/template.py
```diff
@@ -0,0 +1,29 @@
+pkgname = "libgbinder"
+pkgver = "1.1.14"
+pkgrel = 0
+build_style = "makefile"
+make_install_target = "install-dev"
+make_check_target = "test"
+make_use_env = True
+hostmakedepends = [
+    "clang",
```
nekopsykose, line 9: clang is always installed

contrib/libgbinder/template.py
```diff
@@ -0,0 +1,29 @@
...
+    "pkgconf",
+]
+makedepends = ["libglibutil-devel", "linux-headers"]
+pkgdesc = "GLib-style interface to binder"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "BSD-3-Clause"
+url = "https://github.com/mer-hybris/libgbinder"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "106ab4c6605d518f6d0e969ce9f2676da3073ac7b2ed42bd89a5895833a1ae05"
+# check: race condition with include in make that makes it so it segfaults randomly when testing (trying to strace fixes the race condition)
```
nekopsykose, line 19: i can't reproduce this at all but can you reproduce it with make_check_args = ["-j1"]

tulilirockz: latest update fixes it... i literally just got the wrong version

contrib/libgbinder/template.py
tulilirockz: oh! fair

tulilirockz:
ill just go to sleep rn but i was trying to make the python-gbinder package and my container is just refusing to update python-build :p gotta try tomorrow
```
❯ ./cbuild pkg contrib/python-gbinder                                                                                                                                                          
0:00:00.087 => python-gbinder-1.1.2-r0: cleaning build directory...                                                                                                                            
0:00:01.384 => python-gbinder-1.1.2-r0: start build (target: pkg), available hardening:                                                                                                        
0:00:01.384 =>   +format +int +pie +scp +ssp +var-init                                                                                                                                         
0:00:01.384 =>   -bti -cet -cfi -cfi-genptr -cfi-icall -pac -sst -vis                                                                                                                          
0:00:01.385 => python-gbinder-1.1.2-r0: running fetch hook: 000_sources...                                                                                                                     
0:00:01.391 => cbuild: updating software in /home/tulili/opt/tulilirockz/cports/bldroot container...                                                                                           
(1/2) Installing python-build-pycache (1.2.1-r0)                                                                                                                                               
ERROR: python-build-pycache-1.2.1-r0: package mentioned in index not found (try 'apk update')                                                                                                  
(2/2) Installing python-build (1.2.1-r0)                                                                                                                                                       
ERROR: python-build-1.2.1-r0: package mentioned in index not found (try 'apk update')                                                                                                          
2 errors; 596 MiB in 72 packages                                                                                                                                                               
0:00:04.418 => A failure has occurred!                                                                                                                                                         
0:00:04.418 => Stack trace:                                                                                                                                                                    
0:00:04.422 =>   /home/tulili/opt/tulilirockz/cports/src/cbuild/core/build.py:24: in function 'build'                                                                                          
0:00:04.422 =>   /home/tulili/opt/tulilirockz/cports/src/cbuild/core/build.py:159: in function '_build'                                                                                        
0:00:04.422 =>   /home/tulili/opt/tulilirockz/cports/src/cbuild/core/chroot.py:575: in function 'update'                                                                                       
0:00:04.422 =>   /home/tulili/opt/tulilirockz/cports/src/cbuild/apk/cli.py:192: in function 'call_chroot'                                                                                      
0:00:04.422 =>   /home/tulili/opt/tulilirockz/cports/src/cbuild/core/chroot.py:839: in function 'enter'                                                                                        
0:00:04.422 => Raised exception:                                                                                                                                                               
0:00:04.422 =>   CalledProcessError: command '/usr/bin/bwrap --unshare-all --hostname cbuild --bind /home/tulili/opt/tulilirockz/cports/bldroot / --bind /home/tulili/opt/tulilirockz/cports/bl
droot/builddir /builddir --ro-bind /home/tulili/opt/tulilirockz/cports/bldroot/destdir /destdir --ro-bind /home/tulili/opt/tulilirockz/cports/sources /sources --dev /dev --proc /proc --tmpfs 
/run --tmpfs /tmp --tmpfs /var/tmp --new-session --die-with-parent --ro-bind /home/tulili/opt/tulilirockz/cports/packages /binpkgs --ro-bind /home/tulili/opt/tulilirockz/cports/pkgstage /stag
epkgs --bind /home/tulili/opt/tulilirockz/cports/cbuild_cache /cbuild_cache --uid 1337 --gid 1337 --share-net --setenv FAKEROOTDONTTRYCHOWN 1 -- linux64 sh /.cbuild_fakeroot.sh apk upgrade --
repositories-file /dev/null --no-interactive --clean-protected --repository /binpkgs/contrib --repository /binpkgs/contrib/debug --repository https://repo.chimera-linux.org/current/contrib --
repository https://repo.chimera-linux.org/current/main --repository https://repo.chimera-linux.org/current/contrib/debug --repository https://repo.chimera-linux.org/current/main/debug --cache
-dir /cbuild_cache/apk/x86_64 --available --cache-max-age 0' failed with exit code 2                                                                                                           
0:00:04.423 => Phase 'setup' failed for package 'python-gbinder'.           
```

nekopsykose:
that one is what seems like an apk bug that i'm not sure how to work around, the local downloaded apkindex has 'python-build-1.2.1' in it, but the remote repo has 1.2.2 (because it was updated), so it doesn't find 1.2.1.

the same happened even before when `apk update + apk upgrade` was ran separately; sometimes it seems to just not download a new apkindex (which would say 1.2.2 is the existing version), uses the old one locally (which says 1.2.1), then fails to find that

manually doing `./cbuild chroot` and running `apk upgrade -Ua` in it or something fixes it (since then it downloads the newer index), but that is already what `cbuild: updating software in` is doing (literally)- it just doesn't actually update it for some reason even when you rerun `pkg` or something multiple times that all calls the same command

contrib/python-gbinder/template.py
```diff
@@ -0,0 +1,24 @@
+pkgname = "python-gbinder"
+pkgver = "1.1.2"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "pkgconf",
+    "python-build",
+    "python-cython",
+    "python-installer",
+    "python-pytest",
```
nekopsykose, line 10:
```suggestion
    "python-pytest",
```
checkdepends

nekopsykose: or just removing it since it's not used

contrib/python-gbinder/template.py
```diff
@@ -0,0 +1,24 @@
...
+    "python-setuptools",
+]
+makedepends = [
+    "libgbinder-devel",
+    "python-devel",
+]
+pkgdesc = "Python bindings for libgbinder"
```
nekopsykose, line 17: you'll need

```
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@
         kw.setdefault(flag_map.get(token[:2]), []).append(token[2:])
     return kw
 
-USE_CYTHON = False
+USE_CYTHON = True
 if "--cython" in sys.argv:
     sys.argv.remove("--cython")
     USE_CYTHON = True
```

to build it since there's no way to pass an argv with pep517

contrib/python-gbinder/template.py
```diff
@@ -0,0 +1,24 @@
...
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/erfanoabdi/gbinder-python"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "2dc424d5c2594146612e4bd752964f8928a62eec7c5ce6046f4c582079d0b537"
+# check: no tests
```
nekopsykose, line 23:
```suggestion
# check: needs manually running script in .py with manual setup
```

tulilirockz:
> that one is what seems like an apk bug that i'm not sure how to work around, the local downloaded apkindex has 'python-build-1.2.1' in it, but the remote repo has 1.2.2 (because it was updated), so it doesn't find 1.2.1.
> 
> the same happened even before when `apk update + apk upgrade` was ran separately; sometimes it seems to just not download a new apkindex (which would say 1.2.2 is the existing version), uses the old one locally (which says 1.2.1), then fails to find that
> 
> manually doing `./cbuild chroot` and running `apk upgrade -Ua` in it or something fixes it (since then it downloads the newer index), but that is already what `cbuild: updating software in` is doing (literally)- it just doesn't actually update it for some reason even when you rerun `pkg` or something multiple times that all calls the same command

i cleared the env today and it just fixed itself :p probably that would have worked tho

tulilirockz:
this thing depends on god damn LXC :(

nekopsykose:
i don't know if it needs an actual 'lxc service' since there's no way to really do that currently, but there you go lxc is there

tulilirockz:
Had to remove the _lxc bash completions because this package does not contain the `lxc` binary, just the `lxc-*` stuff

nekopsykose:
lxc is already in the repo so i'm not sure why you added a copy of it here

tulilirockz:
i had outdated repos on my machine so i literally did not know there was lxc in the repos LOL

tulilirockz:
only thing left for full compatibility is just the binder kernel modules, that might be best in a separate PR or something

nekopsykose:
those are already in the kernel, unless you mean something else?

nekopsykose:
well

```
CONFIG_ANDROID_BINDER_DEVICES=""
```

it might be nice to make that `"binder,hwbinder,vndbinder"`, but afaict that is just a default for `modprobe binder-linux` and you can still `modprobe binder-linux device=binder,hwbinder,vndbinder` anyway and everything will work, since

```
CONFIG_ANDROID_BINDER_IPC=y
CONFIG_ANDROID_BINDERFS=y
```

is there

tulilirockz:
> those are already in the kernel, unless you mean something else?

for some reason, waydroid runs with the /dev/anbox-binder thing in my system, when in other distros it used to use /dev/binder... i really have like no idea why tho :p

user/waydroid/patches/apparmor-config.patch
```diff
@@ -0,0 +1,9 @@
+--- a/data/configs/config_3	2024-08-07 18:01:14.000000000 -0300
++++ b/data/configs/config_3	2024-09-07 14:21:28.466488580 -0300
+@@ -1,6 +1,5 @@
+ lxc.uts.name = waydroid
+ 
+-lxc.apparmor.profile = unconfined
```
nekopsykose, line 6: just out of curiosity, what error does lxc give when it's built without apparmor support but this config is there?

tulilirockz: it just says that LXC was build without apparmor support and exits. https://github.com/lxc/lxc/blob/bad6d6f180da45b2d3ae06d44201197df38504d7/src/lxc/confile.c#L1651

nekopsykose: very annoying, thanks

tulilirockz:
trying to make the `waydroid-container` dinit service work properly :P

tulilirockz:
> well
> 
> ```
> CONFIG_ANDROID_BINDER_DEVICES=""
> ```
> 
> it might be nice to make that `"binder,hwbinder,vndbinder"`, but afaict that is just a default for `modprobe binder-linux` and you can still `modprobe binder-linux device=binder,hwbinder,vndbinder` anyway and everything will work, since
> 
> ```
> CONFIG_ANDROID_BINDER_IPC=y
> CONFIG_ANDROID_BINDERFS=y
> ```
> 
> is there
For some reason this shows up on the linux-stable config
```
~ 
❯ doas modprobe binder-linux device=binder,hwbinder,vndbinder
doas (tulili@studio) password: 
modprobe: FATAL: Module binder-linux not found in directory /lib/modules/6.10.8-0-generic
```

nekopsykose:
right, cause that's the old external module

for the new thing it's built in so you'll need i think maybe `binder.devices=binder,hwbinder,vndbinder` in kernel cmdline (and reboot), not sure how to do it at runtime. if that works then i'll update the kconfig later

tulilirockz:
```
~ 
❯ doas dinitctl start waydroid-container
doas (tulili@studio) password: 
Service 'waydroid-container' started.

~ 
❯ cat /proc/cmdline
initrd=\initrd.img-6.10.8-0-generic root=PARTUUID=7a9e28b3-6bdc-43b5-814a-daf42b973453 ro quiet splash binder.devices=binder,hwbinder,vndbinder

~ 
❯ waydroid show-full-ui
[17:09:56] Starting waydroid session
[17:10:08] Android with user 0 is ready
^C

~ took 15s 
❯ ls /dev/binder -la
╭───┬─────────────┬─────────┬────────────────┬──────────┬───────────┬───────────┬───────┬─────╮
│ # │    name     │  type   │     target     │ readonly │   mode    │ num_links │ inode │ ... │
├───┼─────────────┼─────────┼────────────────┼──────────┼───────────┼───────────┼───────┼─────┤
│ 0 │ /dev/binder │ symlink │ /dev/binderfs/ │ false    │ rwxrwxrwx │         1 │   639 │ ... │
│   │             │         │ binder         │          │           │           │       │     │
╰───┴─────────────┴─────────┴────────────────┴──────────┴───────────┴───────────┴───────┴─────╯
```

nekopsykose:
oh i see, empty string is the new binderfs, makes sense

i guess everything just works then even with no cmdline changes?

tulilirockz:
yup! everything seems to be working perfectly! the service works fine, and the waydroid session too! im just installing a ton of stuff with waydroid_script to check out if something blows up or smth

nekopsykose:
makes sense :)  

looking at the alpine init.d there is a conditional modprobe for `fuse` and `loop` modules. if those are required for waydroid (and maybe others), i guess you can make a modules-load.conf (see main/flatpak) so they're auto loaded

nekopsykose:
and put the deps into user/ too

user/waydroid/template.py
```diff
@@ -0,0 +1,41 @@
+pkgname = "waydroid"
+pkgver = "1.4.3"
+pkgrel = 0
+build_style = "makefile"
+make_build_args = ["USE_SYSTEMD=0", "USE_NFTABLES=1"]
+depends = [
+    "dbus",
+    "dnsmasq",
+    "lxc",
+    "python",
```
nekopsykose, line 10:
```suggestion
```

only need hard python for leafs

user/waydroid/template.py
```diff
@@ -0,0 +1,41 @@
+pkgname = "waydroid"
+pkgver = "1.4.3"
+pkgrel = 0
+build_style = "makefile"
+make_build_args = ["USE_SYSTEMD=0", "USE_NFTABLES=1"]
+depends = [
+    "dbus",
```
nekopsykose, line 7:
```suggestion
```

user/waydroid/template.py
```diff
@@ -0,0 +1,41 @@
...
+        name="51_waydroid.nft",
+    )
+
+
+@subpackage("waydroid-nftables")
+def _(self):
+    self.depends = ["nftables"]
+    self.subdesc = ""
+    self.install_if = [self.parent, "nftables"]
+    return ["etc/nftables.d/51_waydroid.nft"]
```
nekopsykose, line 41: considering this is like a 4kb config file, is there a need to subpackage it at all?

tulilirockz: i just saw alpine and they had split those. probably would be better to just be into the base instead

nekopsykose: in alpine they subpackage everything even when it kinda doesn't make sense, in this case it was added like that because of a start to add nft rules for various things that a person wanted (to help with showing how to expose various services), so overall it might make sense one way or another there but here you might as well just throw in the whole file tbh

tulilirockz:
Should be good to go now

user/waydroid/template.py
```diff
@@ -0,0 +1,36 @@
+pkgname = "waydroid"
+pkgver = "1.4.3"
+pkgrel = 0
+build_style = "makefile"
+make_build_args = ["USE_SYSTEMD=0", "USE_NFTABLES=1"]
```
nekopsykose, line 5: this has to be make_install_args, currently it's just ignored

then you don't need the uninstall below for systemd either

user/python-gbinder/patches/cython.patch
```diff
@@ -0,0 +1,35 @@
...
+ from distutils.core import setup, Extension
++from Cython.Build import cythonize
+ 
+ 
+ def pkgconfig(package, kw):
+@@ -10,23 +11,12 @@
+         kw.setdefault(flag_map.get(token[:2]), []).append(token[2:])
+     return kw
+ 
+-USE_CYTHON = False
```
nekopsykose, line 14: this patch is one line if you change this False to True and make literally no other changes

tulilirockz:
I believe this actually should be good now 👍

nekopsykose:
it is, nice work :)  


@JamiKettunen you were looking at this before, so have another look over now to see if it works as you'd expect :)

user/waydroid/patches/sed-chimera.patch
```diff
@@ -0,0 +1,17 @@
+--- a/tools/helpers/lxc.py	2024-09-07 13:39:38.419147354 -0300
++++ b/tools/helpers/lxc.py	2024-09-07 14:10:55.719543931 -0300
+@@ -157,12 +157,12 @@
+     tools.helpers.run.user(args, command)
+     command = ["sh", "-c", "cat {} > \"{}\"".format(' '.join('"{0}"'.format(w) for w in config_snippets), lxc_path + "/config")]
+     tools.helpers.run.user(args, command)
+-    command = ["sed", "-i", "s/LXCARCH/{}/".format(platform.machine()), lxc_path + "/config"]
++    command = ["sed", "-i", "\'\'","s/LXCARCH/{}/".format(platform.machine()), lxc_path + "/config"]
```
q66, line 8: @tulilirockz is this actually correct? seems to me like it should be more like `"sed", "-i", "", "..."` (otherwise you will get backup files that are called `''`) unless there is something i'm missing?

q66: (i'd test this myself but i have no idea how this is supposed to work, so just something i spotted while looking over the diff - so if this is intentional, nevermind me)

tulilirockz: Yup! It works fine. Apparent bsdsed needs to have an argument specifiying the empty extension. If you just gave it an empty arg it would thing the `s/*` command is the backup extension, like:

```
sed -i s/thing/thong/g thing.txt
sed -i "" -> sed -i (empty space) s/thing/thong #...
# where as '' gets expanded to actual single quotes
sed -i '' -> sed -i (no extension) s/thing/thong
```

q66: i think you misunderstand:

```
q66@chimera: /home/q66$ cat test.txt
foo
q66@chimera: /home/q66$ sed -i '' 's/foo/bar/' test.txt
q66@chimera: /home/q66$ cat test.txt                   
bar
q66@chimera: /home/q66$ sed -i "" 's/bar/baz/' test.txt
q66@chimera: /home/q66$ cat test.txt                   
baz
q66@chimera: /home/q66$ sed -i "''" 's/baz/foo/' test.txt
q66@chimera: /home/q66$ cat test.txt*
foo
baz
q66@chimera: /home/q66$ cat test.txt
test.txt      test.txt\'\'
```

yes, you need to pass an empty argument to `-i` to get no backup, but you are already doing so if you pass `""`

if you pass `"\'\'"`, you are passing two single quotes as a backup extension, which will result in files with extensions being run

you can see this if you do it from python:

```
>>> import subprocess
>>> subprocess.run(["sed", "-i", "", "s/foo/bar/", "test.txt"])
CompletedProcess(args=['sed', '-i', '', 's/foo/bar/', 'test.txt'], returncode=0)
>>> subprocess.run(["sh", "-c", "ls -l test.txt*"])
-rw-r--r--  1 q66 q66 4 Sep  8 21:39 test.txt
CompletedProcess(args=['sh', '-c', 'ls -l test.txt*'], returncode=0)
>>> subprocess.run(["sed", "-i", "\'\'", "s/foo/bar/", "test.txt"])
CompletedProcess(args=['sed', '-i', "''", 's/foo/bar/', 'test.txt'], returncode=0)
>>> subprocess.run(["sh", "-c", "ls -l test.txt*"])
-rw-r--r--  1 q66 q66 4 Sep  8 21:40 test.txt
-rw-r--r--  1 q66 q66 4 Sep  8 21:39 test.txt''
CompletedProcess(args=['sh', '-c', 'ls -l test.txt*'], returncode=0)
```

that is unless the `tools.helpers.run.user` has different semantics here, which is what i am asking (as it does not seem to be the case)

q66: as far as i can tell, the command is just passed to `subprocess.Popen` which has the same semantics as `subprocess.run`, so i should be correct here

tulilirockz: oh my god you are right, it literally just generated the `''` files as you said :O, i didnt see them earlier

tulilirockz:
```
~ 
❯ ls /var/lib/waydroid/lxc/waydroid/
╭───┬─────────────────────────────────────────────────┬──────┬─────────┬─────────────╮
│ # │                      name                       │ type │  size   │  modified   │
├───┼─────────────────────────────────────────────────┼──────┼─────────┼─────────────┤
│ 0 │ /var/lib/waydroid/lxc/waydroid/config''         │ file │   955 B │ 2 hours ago │
│ 1 │ /var/lib/waydroid/lxc/waydroid/config           │ file │   954 B │ 2 hours ago │
│ 2 │ /var/lib/waydroid/lxc/waydroid/waydroid.seccomp │ file │   346 B │ 3 hours ago │
│ 3 │ /var/lib/waydroid/lxc/waydroid/config_nodes     │ file │ 1.5 KiB │ 2 hours ago │
│ 4 │ /var/lib/waydroid/lxc/waydroid/config_session   │ file │   313 B │ an hour ago │
╰───┴─────────────────────────────────────────────────┴──────┴─────────┴─────────────╯
```

q66: ok :) you can make another PR to fix that then

tulilirockz: alright!

PR #2867, opened 2024-09-07
gnome 47 libs

nekopsykose:
[update skip]


nekopsykose:
they're broken on rebuild too (i disabled), i think it's new weston that broke some display stuff, since some kde testsuites also broke. maybe the way wlheadless-run starts it isn't valid, or maybe it's a weston regression; someone else can look into it tho

nekopsykose:
everything still works on 46 so i'll merge this after someone else also acks that everything still works ok, then we wait for rest of 47 parts :-)

nekopsykose:
> i think it's new weston that broke some display stuff

if i had to guess, maybe new version has some connection limit and the parallel test connections hit a limit and they get disconnected

and the reason the test failed but passed for triallax is because wlheadless-run might not forward the exit code of meson test inside it in such a case but just exit 0 somehow

PR #2872, opened 2024-09-07
user/ani-cli: new package

denisstrizhkin:


user/ani-cli/template.py
```diff
@@ -0,0 +1,17 @@
...
+pkgrel = 0
+depends = ["aria2", "chimerautils", "curl", "ffmpeg", "fzf", "mpv"]
+pkgdesc = "CLI to browse and watch anime"
+maintainer = "Denis Strizhkin <strdenis02@gmail.com>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/pystardust/ani-cli"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "bdd5e3c264ab67760b13d34174ec86c3da3aaaaacda3ba529d8b2648bce2ef08"
+# there is no tests
+options = ["!check"]
```
nekopsykose, line 12:
```suggestion
```

doesn't fail without that set so don't have to set it

user/ani-cli/template.py
```diff
@@ -0,0 +1,17 @@
+pkgname = "ani-cli"
+pkgver = "4.9"
+pkgrel = 0
+depends = ["aria2", "chimerautils", "curl", "ffmpeg", "fzf", "mpv"]
```
nekopsykose, line 4:
```suggestion
depends = ["aria2", "curl", "ffmpeg", "fzf", "mpv"]
```

always there

also isn't yt-dlp prefered to 'ffmpeg' given the readme or nah

denisstrizhkin: Good catch, never noticed it. Should be yt-dtp yeah.

PR #2874, opened 2024-09-07
user/vivid: new package

jabuxas:


contrib/vivid/template.py
```diff
@@ -0,0 +1,17 @@
...
+pkgdesc = "Themeable LS_COLORS generator"
+maintainer = "jabuxas <jabuxas@proton.me>"
+license = "MIT AND Apache-2.0"
+url = "https://github.com/sharkdp/vivid"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "88db6158dad60aba66ae16f2cd1b09f515625940a33bada65da5562a03538e49"
+
+
+def post_install(self):
+    self.install_license("LICENSE-APACHE")
```
nekopsykose, line 16:
```suggestion
```

jabuxas: uhmm why? main/rust installs both licenses

nekopsykose: because apache doesn't have to be installed

jabuxas: ok so why does rust do it

nekopsykose: because someone put it there when they didn't have to

jabuxas: sure

PR #2877, opened 2024-09-07
user/tofi: new package

jabuxas:


contrib/tofi/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "tofi"
+pkgver = "0.9.1"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = ["meson", "pkgconf", "scdoc"]
+makedepends = [
+    "wayland-devel",
+    "musl-fts-devel",
+    "freetype-devel",
+    "cairo-devel",
```
nekopsykose, line 10: sort

contrib/tofi/template.py
```diff
@@ -0,0 +1,21 @@
...
+    "pango-devel",
+    "wayland-protocols",
+    "libxkbcommon-devel",
+    "linux-headers",
+]
+pkgdesc = "Tiny wayland dynamic menu"
```
nekopsykose, line 16:
```suggestion
pkgdesc = "Wayland dynamic menu"
```

jabuxas:
for some reason post_install fails when installing license..

nekopsykose:
> for some reason post_install fails when installing license..

because the meson.build already installs it to the same place :)

so when you do it it says it already exists

jabuxas:
yea makes sense, thats what i thought as well. though this is the first package i do that already does that automatically.. makes u wonder why other dont

nekopsykose:
some do but not to the exact same folder 😆

you should also be putting these random things in user/ instead

jabuxas:
on the premise that im a new contributor or is there a deeper reasoning? im not against it at all, in fact now that i see, #2874  should probably go to user

nekopsykose:
on the divine premise of 'seems correct'

jabuxas:
lol haha

jabuxas:
id imagine this would be a contrib since fuzzel is contrib as well but doesnt really change much

nekopsykose:
fuzzel is fluffy and very pettable though

jabuxas:
lol

PR #2878, opened 2024-09-07
contrib/flare: new package (0.15.0)

jcgruenhage:


contrib/flare/template.py
```diff
@@ -0,0 +1,51 @@
...
+    self.make_dir = "build"
+
+
+def init_build(self):
+    from cbuild.util import cargo
+
+    self.make_dir = ""
+    renv = cargo.get_environment(self)
+    self.make_env.update(renv)
+    self.make_dir = "build"
```
jcgruenhage, line 46: This is horrible, but if I don't get rid of the `make_dir` for these steps, it fails to build. If I remove the `make_dir` for the whole template, it also fails to build. Any ideas?

contrib/flare/patches/0002-don-t-attempt-copying-the-binary.patch
```diff
@@ -0,0 +1,41 @@
+From aa73eae3f886eb6efc26dec420626906342549c8 Mon Sep 17 00:00:00 2001
```
nekopsykose, line 1: make these 1 patch

contrib/flare/template.py
```diff
@@ -0,0 +1,51 @@
+pkgname = "flare"
+pkgver = "0.15.0"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "blueprint-compiler",
+    "cargo-auditable",
+    "pkgconf",
```
nekopsykose, line 8: sort

contrib/flare/template.py
```diff
@@ -0,0 +1,51 @@
...
+    "protoc",
+    "meson",
+    "gettext",
+    "desktop-file-utils",
+]
+makedepends = [
+    "cairo-devel",
+    "gdk-pixbuf-devel",
+    "glib-devel",
+    "gtk4-devel",
```
nekopsykose, line 18: sort

contrib/flare/template.py
nekopsykose, line 46:
```suggestion
    cargo.Cargo(self, wrksrc=".").vendor()


def init_build(self):
    from cbuild.util import cargo

    renv = cargo.get_environment(self)
    self.make_env.update(renv)
```

contrib/flare/template.py
```diff
@@ -0,0 +1,51 @@
...
+    "gtksourceview-devel",
+    "libadwaita-devel",
+    "pango-devel",
+    "libspelling-devel",
+]
+pkgdesc = "Unofficial Signal Desktop app"
+maintainer = "Jan Christian Grünhage <jan.christian@gruenhage.xyz>"
+license = "AGPL-3.0-only"
+url = "https://mobile.schmidhuberj.de/flare"
+source = f"https://gitlab.com/schmiddi-on-mobile/flare/-/archive/{pkgver}/flare-{pkgver}.tar.gz"
```
jcgruenhage, line 28: It would be possible to get a tarball that already includes vendored rust libs, but the URL is https://gitlab.com/schmiddi-on-mobile/flare/-/package_files/139350251/download, which obviously doesn't work as it would need a new url on each new version.

If there's anything I'm missing wrt getting the "package" from https://gitlab.com/schmiddi-on-mobile/flare/-/releases/0.15.0 which includes the vendored source code, I'd appreciate a pointer there as well.

nekopsykose: there's no issue with running cargo vendor so there's not much point

jcgruenhage: yep, with your suggestion it works as it should and without the hacky make_dir changes. Thanks!

PR #2879, opened 2024-09-07
main/gnome-tour: make rust dependencies auditable

jcgruenhage:


nekopsykose:
something i forgot about in the other place is that cargo-auditable is broken on riscv64 so you can't actually hardcode it like this (for build_style=cargo it skips itself on riscv64)

nekopsykose:
the error for reference is something like

```
_pixbuf-2.0" "-lcairo-gobject" "-lcairo" "-lharfbuzz" "-lvulkan" "-lgraphene-1.0" "-lgio-2.0" "-lgobject-2.0" "-lglib-2.0" "-lgraphene-1.0" "-lgobject-2.0" "-lglib-2.0" "-lgtk-4" "-lpangocairo-1.0" "-lpango-1.0" "-lgdk_pixbuf-2.0" "-lcairo-gobject" "-lcairo" "-lharfbuzz" "-lvulkan" "-lgraphene-1.0" "-lgio-2.0" "-lgobject-2.0" "-lglib-2.0" "-lpango-1.0" "-lgobject-2.0" "-lharfbuzz" "-lglib-2.0" "-lgdk_pixbuf-2.0" "-lgobject-2.0" "-lglib-2.0" "-lgobject-2.0" "-lgio-2.0" "-lgobject-2.0" "-lglib-2.0" "-lgobject-2.0" "-lglib-2.0" "-lcairo-gobject" "-lcairo" "-lgobject-2.0" "-lglib-2.0" "-lgobject-2.0" "-lglib-2.0" "-lunwind" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-L" "/usr/lib/rustlib/riscv64-chimera-linux-musl/lib" "-o" "/builddir/flare-0.15.0/build/target/riscv64-chimera-linux-musl/release/deps/flare-037da9acabb18a59" "-Wl,--gc-sections" "-pie" "-Wl,-z,relro,-z,now" "-Wl,-O1" "-Wl,-z,pack-relative-relocs" "/builddir/flare-0.15.0/build/target/riscv64-chimera-linux-musl/release/deps/flare_audit_data.o" "-Wl,--undefined=AUDITABLE_VERSION_INFO"
= note: ld: error: /builddir/flare-0.15.0/build/target/riscv64-chimera-linux-musl/release/deps/flare_audit_data.o: cannot link object files with different floating-point ABI from /lib/Scrt1.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

it probably generates the .o in some weird way

jcgruenhage:
@nekopsykose do you know whether this is a Chimera Linux problem or upstream in https://github.com/rust-secure-code/cargo-auditable? If it's upstream, I wonder if @Shnatsel knows about this, considering I couldn't find an issue about this?

nekopsykose:
it's an upstream problem because regular cargo works and it's the generated .o that is invalid

jcgruenhage:
Okay, so we'll have to wait for a solution from upstream before being able to unconditionally use it like this. For the time being, what would you suggest? Not using `cargo auditable` unless we're using the `cargo` build style?

jcgruenhage:
Because patching in meson options and the like for using cargo-auditable conditionally seems a bit much effort.

nekopsykose:
the issue is most likely that:  

```
fn riscv_features(target_triple: &str) -> String {
    let arch = target_triple.split('-').next().unwrap();
    assert_eq!(&arch[..5], "riscv");
    let mut extensions = arch[7..].to_owned();
    if extensions.contains('g') {
        extensions.push_str("imadf");
    }
    extensions
}
```

which is tested:  

```
    #[test]
    fn test_riscv_abi_detection() {
        // real-world target with double floats
        let features = riscv_features("riscv64gc-unknown-linux-gnu");
        assert!(features.contains('c'));
        assert!(features.contains('d'));
        assert!(features.contains('f'));
        // real-world target without floats
        let features = riscv_features("riscv32imac-unknown-none-elf");
        assert!(features.contains('c'));
        assert!(!features.contains('d'));
        assert!(!features.contains('f'));
        // real-world target without floats or compression
        let features = riscv_features("riscv32i-unknown-none-elf");
        assert!(!features.contains('c'));
        assert!(!features.contains('d'));
        assert!(!features.contains('f'));
        // made-up target without compression and with single floats
        let features = riscv_features("riscv32if-unknown-none-elf");
        assert!(!features.contains('c'));
        assert!(!features.contains('d'));
        assert!(features.contains('f'));
    }
```

in our case, like every distro we use gc:  

```
--- /dev/null
+++ b/compiler/rustc_target/src/spec/targets/riscv64_chimera_linux_musl.rs
@@ -0,0 +1,10 @@
+use crate::spec::Target;
+
+pub fn target() -> Target {
+    let mut base = super::riscv64gc_unknown_linux_musl::target();
+
+    base.llvm_target = "riscv64-chimera-linux-musl".into();
+    base.options.vendor = "chimera".into();
+
+    base
+}
```

but we don't put it in the triple, because uhh nobody puts their riscv feature set into the arch of the triple. so it misses every feature because it relies on the triple string:

```
            // Select the appropriate floating-point ABI
            if features.contains('d') {
                e_flags |= elf::EF_RISCV_FLOAT_ABI_DOUBLE;
            } else if features.contains('f') {
                e_flags |= elf::EF_RISCV_FLOAT_ABI_SINGLE;
            } else {
                e_flags |= elf::EF_RISCV_FLOAT_ABI_SOFT;
            }
            e_flags
```

and outputs soft float instead of double

nekopsykose:
as for 'how to get actual rust triple feature set without parsing the text string of the triple and guessing' i have no idea though

nekopsykose:
> because uhh nobody puts their riscv feature set into the arch of the triple

(except the names in rustup where gc and imac exist, but in general this isn't a reliable way to detect it. in C land for instance one can always ask the compiler and look at pp defines or test a default .o output, ...., but in rust i haven't seen anyone do conftests like this so dunno)

nekopsykose:
i'm not sure how this works on alpine since they also use auditable and have riscv64 with the same bare arch, but i'm going to guess that ld.bfd simply Doesn't Give A Fuck and outputs some code anyway while ld.lld tells you you can't mix the abi's, or at least i have a vague memory of that from two years ago...

nekopsykose:
> For the time being, what would you suggest? Not using cargo auditable unless we're using the cargo build style?

pretty much yeah, though honestly i think we should get a cbuild wrapper (like the link wrapper and whatnot) for `cargo` so it transparently invokes `cargo auditable` itself via shell with a `CBUILD_NO_CARGO_AUDITABLE` optout and one doesn't have to patch meson to call something else

jcgruenhage:
Okay, so if we patch cargo-auditable to special case `riscv64-chimera-linux-musl` and tell it that it's `gc`, then it should also just work, right?

Getting a cbuild wrapper still sounds reasonable so that we don't have to patch anything anywhere.

nekopsykose:
> Okay, so if we patch cargo-auditable to special case riscv64-chimera-linux-musl and tell it that it's gc, then it should also just work, right?

it would work yeah, but since this isn't suuper important (we added it but also we do nothing with the info distro-side in the meantime) i'd rather see it fixed upstream somehow in a way that is not downstream specialcasing, since there has to be a way to do it normally...

unless there actually is no way to detect this which would be very funny i guess. not sure about then, i guess then we will patch it :)

jcgruenhage:
I have done and still occasionally do stuff with the info. From some initial testing, it looks to me that this does the trick:

```
diff --git a/main/cargo-auditable-bootstrap/template.py b/main/cargo-auditable-bootstrap/template.py
index ff07f048c..a816aaabf 100644
--- a/main/cargo-auditable-bootstrap/template.py
+++ b/main/cargo-auditable-bootstrap/template.py
@@ -1,7 +1,7 @@
 # Keep in sync with cargo-auditable
 pkgname = "cargo-auditable-bootstrap"
 pkgver = "0.6.4"
-pkgrel = 0
+pkgrel = 1
 build_style = "cargo"
 make_build_args = ["-p", "cargo-auditable"]
 make_check_args = [
diff --git a/main/cargo-auditable/patches/0001-define-riscv64-feature-set-for-riscv64-chimera-linux.patch b/main/cargo-auditable/patches/0001-define-riscv64-feature-set-for-riscv64-chimera-linux.patch
new file mode 100644
index 000000000..0ae02fda4
--- /dev/null
+++ b/main/cargo-auditable/patches/0001-define-riscv64-feature-set-for-riscv64-chimera-linux.patch
@@ -0,0 +1,34 @@
+From f2ca2f425e1b7aa68cbc87e249d022fe78fa737f Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Jan=20Christian=20Gr=C3=BCnhage?=
+ <jan.christian@gruenhage.xyz>
+Date: Sat, 7 Sep 2024 21:46:28 +0200
+Subject: [PATCH] define riscv64 feature set for riscv64-chimera-linux-musl
+
+---
+ cargo-auditable/src/object_file.rs | 10 +++++++---
+ 1 file changed, 7 insertions(+), 3 deletions(-)
+
+diff --git a/cargo-auditable/src/object_file.rs b/cargo-auditable/src/object_file.rs
+index 6dcb9bc..5688d26 100644
+--- a/cargo-auditable/src/object_file.rs
++++ b/cargo-auditable/src/object_file.rs
+@@ -188,9 +188,13 @@ fn create_object_file(
+ // `sess.target.options.features`
+ // We do not have access to compiler internals, so we have to reimplement this function.
+ fn riscv_features(target_triple: &str) -> String {
+-    let arch = target_triple.split('-').next().unwrap();
+-    assert_eq!(&arch[..5], "riscv");
+-    let mut extensions = arch[7..].to_owned();
++    let mut extensions = if target_triple == "riscv64-chimera-linux-musl" {
++        String::from("gc")
++    } else {
++        let arch = target_triple.split('-').next().unwrap();
++        assert_eq!(&arch[..5], "riscv");
++        arch[7..].to_owned()
++    };
+     if extensions.contains('g') {
+         extensions.push_str("imadf");
+     }
+-- 
+2.46.0
+
diff --git a/main/cargo-auditable/template.py b/main/cargo-auditable/template.py
index 06a9361b0..6cba66757 100644
--- a/main/cargo-auditable/template.py
+++ b/main/cargo-auditable/template.py
@@ -1,7 +1,7 @@
 # Keep in sync with cargo-auditable-bootstrap
 pkgname = "cargo-auditable"
 pkgver = "0.6.4"
-pkgrel = 0
+pkgrel = 1
 build_style = "cargo"
 make_build_args = ["-p", "cargo-auditable"]
 make_check_args = [
diff --git a/src/cbuild/util/cargo.py b/src/cbuild/util/cargo.py
index 83256b3ed..623f8c3d1 100644
--- a/src/cbuild/util/cargo.py
+++ b/src/cbuild/util/cargo.py
@@ -136,9 +136,9 @@ class Cargo:
             self.template.bldroot_path / "usr/bin/cargo-auditable"
         ).exists()
 
-        # fails to link
-        if tmpl.profile().arch == "riscv64":
-            auditable = False
+        # # fails to link
+        # if tmpl.profile().arch == "riscv64":
+        #     auditable = False
 
         cargo = ["cargo", "auditable"] if auditable else ["cargo"]
         return self.template.do(
```

I'd say the patch is small enough that it's better than having the special casing for riscv64? 

https://github.com/rust-secure-code/cargo-auditable/blob/master/cargo-auditable/src/object_file.rs#L187-L189 sounds like they can't do it "properly" because those are compiler internals that aren't properly exposed in any way?

Shnatsel:
Author of `cargo-auditable` here. Thanks for mentioning me!

Unfortunately I cannot implement a proper, clean fix without requiring a nightly compiler because the `--print=target-spec-json` option for `rustc` is unstable.

I double-checked the upstream target list, and the current detection also misses the `riscv64-linux-android` target. I'm going to change `cargo auditable` to treat all Linux-capable riscv CPUs as `gc`, which would fix both the Android target and your custom target.

Shnatsel:
I've fixed the issue upstream in https://github.com/rust-secure-code/cargo-auditable/pull/162

I'm not sure when I'll cut the next release, so you'll have to cherry-pick the patch until then. Sorry about that!

jcgruenhage:
Thank you @Shnatsel! I've picked your patch up in #2881. I think this PR itself isn't needed anymore, but thanks everyone for the discussion :)

PR #2881, opened 2024-09-07
cargo auditable wrapper and riscv64 fix

jcgruenhage:
- **main/cargo-auditable: fix usage on riscv64**: This problem came up in https://github.com/chimera-linux/cports/pull/2879. The patch is already merged upstream, but we're picking it here because it's not clear yet when a new release will be cut.
- **cbuild: reenable cargo-auditable on riscv64**: With the problem fixed above, it should be possible to reenable cargo-auditable on riscv64.
- **cbuild: use wrapper to enable cargo-auditable**: As suggested in https://github.com/chimera-linux/cports/pull/2879#issuecomment-2336414310, using a wrapper improves this more because we don't have to patch meson builds to get auditable builds.

### DIFF
--- a/src/cbuild/wrappers/cargo.sh
+++ b/src/cbuild/wrappers/cargo.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+if [ -n "$CBUILD_NO_CARGO_AUDITABLE" ]; then
+    exec /usr/bin/cargo "$@"
+elif command -v cargo-auditable >/dev/null; then
+    exec /usr/bin/cargo auditable "$@"
+else
+    exec /usr/bin/cargo "$@"
+fi
--- a/contrib/amberol/template.py
+++ b/contrib/amberol/template.py
@@ -1,9 +1,9 @@
 pkgname = "amberol"
 pkgver = "0.10.3"
-pkgrel = 0
+pkgrel = 1
 build_style = "meson"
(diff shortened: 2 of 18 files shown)


src/cbuild/wrappers/cargo.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+echo "Wrapper called with $@" > cargo-wrapper-invocation-$(date -Iseconds).log
+
+if [ -n "$CBUILD_NO_CARGO_AUDITABLE" ]; then
+    exec /usr/bin/cargo "$@"
+elif command -v cargo-auditable>/dev/null; then
+    exec /usr/bin/cargo auditable "$@"
```
nekopsykose, line 8:
```suggestion
if [ -n "$CBUILD_NO_CARGO_AUDITABLE" ]; then
    exec /usr/bin/cargo "$@"
elif command -v cargo-auditable >/dev/null; then
    exec /usr/bin/cargo auditable "$@"
```

jcgruenhage: eh, had that in for debugging purposes and forgot to remove it. Thanks!

nekopsykose:
thanks, i like this way much more :)

jcgruenhage:
It's a lot more elegant, thanks for pointing it out!

PR #2883, opened 2024-09-08
user/polari: new package

tulilirockz:
The GNOME Polari IRC client and all of its dependencies


EarldridgeJazzedPineda:
Please don't re-add dbus-glib. That package is deprecated and was removed from Chimera Linux a few months ago. From what I can see, it's needed by Telepathy components. Please tell upstream Telepathy to remove the dbus-glib dependency, or tell upstream Polari to remove the Telepathy dependency.

You can create an issue for it on Telepathy or Polari's GitLab:
https://gitlab.freedesktop.org/groups/telepathy/-/issues
https://gitlab.gnome.org/GNOME/polari/-/issues

user/dbus-glib/template.py
```diff
@@ -0,0 +1,26 @@
...
+build_style = "gnu_configure"
+hostmakedepends = ["automake", "libtool", "gtk-doc-tools", "pkgconf"]
+makedepends = [
+    "dbus-devel",
+    "glib-devel",
+]
+checkdepends = ["dbus"]
+pkgdesc = "Deprecated library for D-Bus IPC"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "AFL-2.1"
```
EarldridgeJazzedPineda, line 13:
```suggestion
license = "AFL-2.1 OR GPL-2.0-or-later"
```

user/telepathy-glib/template.py
```diff
@@ -0,0 +1,28 @@
+pkgname = "telepathy-glib"
+pkgver = "0.24.2"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--enable-vala-bindings"]
+hostmakedepends = ["automake", "pkgconf", "libtool", "gtk-doc-tools", "vala"]
```
EarldridgeJazzedPineda, line 6:
```suggestion
hostmakedepends = ["automake", "gtk-doc-tools", "libtool", "pkgconf", "vala"]
```

nekopsykose:
since this needs dbus-glib it won't be added

but you can use the flatpak for polari just fine so that's ok

tulilirockz:
Thats fair! Sorry guys didnt know you had removed the dbus-glib thing before!

q66:
the whole telepathy framework seems to be very dead, polari should probably remove it upstream (though as i understand it it may be difficult, i wonder why there is no issue open for it)

PR #2891, opened 2024-09-08
user/waydroid: fix backup arg in sed

tulilirockz:
This PR should fix a weird sed -i bug that waydroid has when using chimerautils, as me and @q66  talked about in https://github.com/chimera-linux/cports/pull/2866#discussion_r1749347399 - in this case it just fixes a misunderstanding i had when trying to fix it


q66:
bump pkgrel please

tulilirockz:
oh true

PR #2898, opened 2024-09-09
user/f3-utils: new package

jabuxas:
not cute enough


user/f3-utils/template.py
```diff
@@ -0,0 +1,20 @@
+pkgname = "f3-utils"
+pkgver = "8.0"
+pkgrel = 0
+build_style = "makefile"
+make_build_args = ["extra"]
+make_install_args = ["install-extra"]
+hostmakedepends = []
+makedepends = ["argp-standalone", "parted-devel", "linux-headers", "udev-devel"]
+depends = []
```
nekopsykose, line 9:
```suggestion
makedepends = ["argp-standalone", "parted-devel", "linux-headers", "udev-devel"]
```

user/f3-utils/template.py
```diff
@@ -0,0 +1,20 @@
...
+pkgdesc = "Flash devices tester"
+maintainer = "jabuxas <jabuxas@proton.me>"
+license = "GPL-3.0-only"
+url = "https://github.com/AltraMayor/f3"
+source = f"{url}/archive/v{pkgver}.tar.gz"
```
nekopsykose, line 14: you should patch the `install: all` rule to be `install:` so it doesn't rebuild everything in the install step

nekopsykose:
```
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@
 all: $(TARGETS)
 extra: $(EXTRA_TARGETS)
 
-install: all
+install:
 	$(INSTALL) -d $(DESTDIR)$(PREFIX)/bin
 	$(INSTALL) -m755 $(TARGETS) $(DESTDIR)$(PREFIX)/bin
 	$(INSTALL) -d $(DESTDIR)$(PREFIX)/share/man/man1
```

jabuxas: im not sure if i did it right, does make_install_args append to default or does it substitute it? it doesnt make sense to me that only install-extra also calls in regular install

nekopsykose: > does make_install_args append to default

yes

PR #2899, opened 2024-09-10
user/gdu: new package (5.29.0)

tulilirockz:


user/gdu/template.py
```diff
@@ -0,0 +1,18 @@
+pkgname = "gdu"
+pkgver = "5.29.0"
+pkgrel = 0
+build_style = "go"
+make_build_args = ["./cmd/gdu"]
+make_check_args = ["./cmd/gdu"]
```
nekopsykose, line 6: all the actual tests are in other folders too so leaving this as default is probably more correct

nekopsykose: (because this runs zero tests)

PR #2903, opened 2024-09-10
contrib/thermald: fix group

triallax:
not sure if this is better or having a power group in base-files or something


nekopsykose:
the point of the group is to allow managing thermald as a user after adding to group (i guess), so an entire service group feels weird

i would guess wheel makes more sense but idk

triallax:
not sure co-opting wheel for something like this is a nice idea, so maybe this pr is fine as is

PR #2906, opened 2024-09-10
*/*: use cmd:/man:

triallax:
[ci skip]


triallax:
i'll just close this, there's a lot of subtle things that are easy to get wrong here (and that I've already gotten wrong); it's better to migrate the templates piecemeal

PR #2908, opened 2024-09-10
user/topgrade: new package (15.0.0)

tulilirockz:


user/topgrade/template.py
```diff
@@ -0,0 +1,31 @@
+pkgname = "topgrade"
+pkgver = "15.0.0"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo-auditable"]
+pkgdesc = "System updater written in Rust"
```
triallax, line 6: like i said in irc, no "written in <lang>"

user/topgrade/template.py
```diff
@@ -0,0 +1,31 @@
...
+def post_build(self):
+    TOPGRADE_BIN = f"target/{self.profile().triplet}/release/topgrade"
+    with open(f"{self.cwd}/topgrade.bash", "w") as o:
+        self.do(TOPGRADE_BIN, "--gen-completion", "bash", stdout=o)
+    with open(f"{self.cwd}/topgrade.fish", "w") as o:
+        self.do(TOPGRADE_BIN, "--gen-completion", "fish", stdout=o)
+    with open(f"{self.cwd}/_topgrade", "w") as o:
+        self.do(TOPGRADE_BIN, "--gen-completion", "zsh", stdout=o)
+    with open(f"{self.cwd}/topgrade.8", "w") as o:
+        self.do(TOPGRADE_BIN, "--gen-manpage", stdout=o)
```
triallax, line 23:
```suggestion
    for shell in ["bash", "fish", "zsh"]:
        with open(f"{self.cwd}/topgrade.{shell}", "w") as o:
            self.do(f"target/{self.profile().triplet}/release/topgrade", "--gen-completion", "bash", stdout=o)
    with open(f"{self.cwd}/topgrade.8", "w") as o:
        self.do(f"target/{self.profile().triplet}/release/topgrade", "--gen-manpage", stdout=o)
```
also side note, we don't typically use screaming case in cports

user/topgrade/template.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def post_install(self):
+    self.install_completion("topgrade.bash", "bash")
+    self.install_completion("topgrade.fish", "fish")
+    self.install_completion("_topgrade", "zsh")
```
triallax, line 29:
```suggestion
    self.install_completion("topgrade.zsh", "zsh")
```
per above

user/topgrade/template.py
```diff
@@ -0,0 +1,31 @@
...
+    self.install_man("topgrade.8")
+    self.install_file("config.example.toml", "usr/share/doc/topgrade")
```
triallax, line 31: install to `/usr/share/examples/topgrade/config.toml` instead

triallax (changes requested):

user/topgrade/template.py
triallax: or could even do a loop here as well

user/topgrade/template.py
tulilirockz: oh! i thought that ZSH *had* to have the `_` prefix! sorry for the suboptimal implementation

user/topgrade/template.py
tulilirockz: Fair!

user/topgrade/template.py
nekopsykose: it does, but install_completion renames the file

tulilirockz:
Should be good!

PR #2909, opened 2024-09-11
user/papers: new package

tulilirockz:
GNOME papers and its deps!


user/papers/template.py
```diff
@@ -0,0 +1,63 @@
+pkgname = "papers"
+pkgver = "46.2"
+pkgrel = 0
+build_style = "meson"
+configure_args = [
+    "-Dcomics=enabled",
```
nekopsykose, line 6: you can documentation=false and then don't need gi-docgen at all

tulilirockz: LOL no way

user/papers/patches/tests.patch
```diff
@@ -0,0 +1,18 @@
...
+   install_dir: join_paths(pps_datadir, 'metainfo'),
+ )
+-
+-# Validate Appdata
+-if get_option('tests')
+-  appstreamcli = find_program('appstreamcli', required: true)
+-  if (appstreamcli.found())
+-    test('validate-appdata',
+-         appstreamcli,
+-         args: ['validate', '--strict', '--pedantic', '--explain', metainfo_file],
```
nekopsykose, line 14: you can probably add only `--nonet` to these args and it will work :)

user/papers/template.py
```diff
@@ -0,0 +1,59 @@
...
+    cargo.Cargo(self, wrksrc="shell-rs").vendor()
+
+
+def init_build(self):
+    from cbuild.util import cargo
+
+    renv = cargo.get_environment(self)
+    self.make_env.update(renv)
+
+
```
nekopsykose, line 55: i think you also need a -devel package given it has libraries and a usr/include

also maybe a -nautilus (see font-manager subpackage and copy that)

user/papers/patches/tests.patch
tulilirockz: omg true

PR #2912, opened 2024-09-11
contrib/snort: new package (3.3.5.0)

rbgarga:
- **contrib/daq: new package (3.0.16)**
- **contrib/libdnet: new package (1.18.0)**
- **contrib/pcre: new package (8.45)**
- **contrib/snort: new package (3.3.5.0)**


contrib/pcre/template.py
```diff
@@ -0,0 +1,38 @@
+pkgname = "pcre"
```
q66, line 1: we definitely don't want to bring back this one (it was previously removed, is long unmaintained/obsolete, and contains security issues)

upstreams should migrate to pcre2

contrib/snort/template.py
```diff
@@ -0,0 +1,38 @@
...
+pkgver = "3.3.5.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DCMAKE_BUILD_TYPE:STRING=RelWithDebInfo",
+    "-DDNET_INCLUDE_DIR=/usr/include/dnet",
+    "-DENABLE_HARDENED_BUILD:BOOL=true",
+    "-DENABLE_SHELL:BOOL=true",
+    "-DENABLE_LARGE_PCAP:BOOL=true",
+    "-DSTATIC_LUAJIT:BOOL=true",
```
nekopsykose, line 11:
```suggestion
    "-DENABLE_SHELL=ON",
    "-DENABLE_LARGE_PCAP=ON",
```

the hardened build passes like 4 cflags that are already the default so that doesn't do anything

contrib/libdnet/template.py
```diff
@@ -0,0 +1,22 @@
+pkgname = "libdnet"
+pkgver = "1.18.0"
+pkgrel = 0
+build_style = "cmake"
+hostmakedepends = ["cmake", "linux-headers", "ninja"]
+pkgdesc = (
+    "Simplified, portable interface to several low-level networking routines"
```
nekopsykose, line 7:
```suggestion
    "Interface to several low-level networking routines"
```

contrib/pcre/template.py
rbgarga: I agree.  There is a PR on snort3 project from 2022

https://github.com/snort3/snort3/issues/254

contrib/snort/template.py
```diff
@@ -0,0 +1,38 @@
...
+makedepends = [
+    "daq-devel-static",
+    "hwloc-devel",
+    "libdnet",
+    "libdnet-static",
+    "libnuma-devel",
+    "libpcap-devel",
+    "libtirpc-devel",
+    "libuuid-devel",
+    "luajit-devel-static",
```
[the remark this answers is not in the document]
nekopsykose: i guess for daq the .a's are actually most of the library and it won't be shared, but that doesn't apply to the other two

PR #2914, opened 2024-09-11
user/valent: new package (1.0.0.alpha.46)

tulilirockz:
Everything works completely fine, only problem is that update-check is broken with this release schema. I wonder how that could be fixed?


tulilirockz:
fair

tulilirockz:
ill close it till it gets a full-on release then

PR #2923, opened 2024-09-12
main/curl: update to 8.10.0

fossdd:


nekopsykose:
i had reverted this because it was broken

it's fixed now in master but i'm just waiting a day or two to see if there will be a point release :)

fossdd:
alrighty. makes sense. thanks!

PR #2924, opened 2024-09-13
lxde desktop environment

jabuxas:
latest versions available (that's why mix/match github and sourceforge, some aren't on sourceforge yet), we prio sourceforge since thats the "main" one apparently. 

image

notable stuff:

- libfm depends on dbus-glib and thats no longer a thing, so we apply quite a big patch to change all of that
- i wasnt able to compile obconf without pulling master. even getting last release and [patching it til master](https://paste.jabuxas.xyz/4692155485.gf08R0YqZ5) didnt work, but pulling from master directly does - no clue. same thing kind of happened with pcmanfm as well, where one source doenst play nice but another does
- we cant have obconf-qt yet to substitute it because qt6 was merged just 2 weeks ago and no new release
- not all of the suite is ported yet, not sure if we should do all of it, maybe bundle in a different lxde-apps pkg?


q66:
isn't this thing like dead af

i'm not sure if i feel like having more dead desktops for a dead display server

if you want a scuffed x11 desktop what does this offer over xfce

nekopsykose:
> isn't this thing like dead af

considering the patches to make it work are from 2020, yea

nekopsykose:
mayb this is an epic side effect of 'kde1' so everyone thinks this is a good idea :^)

q66:
yeah, sorry

i want to actually remove things like intltool someday and not just keep piling on more unmaintained crap that would make it stay in forever

jabuxas:
it had commits like 3 weeks ago and its rather lively for a x11 desktop, should you guys reconsider. pcmanfm is also nice

q66:
maybe they should apply their 2020 patches then

PR #2931, opened 2024-09-15
main/epiphany: update to 47.0

tulilirockz:
Depends on #2867


q66:
i deal with gnome updates myself, prs for individual bits just add me work

tulilirockz:
oh! fair! thats why i didny see any other epiphany PR :p

PR #2934, opened 2024-09-15
user/gamescope: update to 5.19

tulilirockz:


user/gamescope/template.py
```diff
@@ -58,7 +59,7 @@
     f"{url}/archive/refs/tags/{pkgver}.tar.gz",
     "https://github.com/ValveSoftware/openvr/archive/ebd425331229365dc3ec42d1bb8b2cc3c2332f81.tar.gz",
     "https://github.com/Joshua-Ashton/vkroots/archive/5106d8a0df95de66cc58dc1ea37e69c99afc9540.tar.gz",
-    "https://gitlab.freedesktop.org/wlroots/wlroots/-/archive/a5c9826e6d7d8b504b07d1c02425e6f62b020791.tar.gz",
+    "https://gitlab.freedesktop.org/wlroots/wlroots/-/archive/0.18.0.tar.gz",
```
nekopsykose, line 62: this is the wrong commit and not what is used upstream

the reason i haven't updated it yet is because the newer versions are broken, see the other mr that wanted to update it

tulilirockz: oh! makes sense, i just found it weird that the thing wasnt using the latest wlroots version

nekopsykose: the exact things are submodules so you have to follow the submodule git state in th repo to see what to put for all these

nekopsykose: that said it's only that one thing that was broken so i guess i'll update it anyway

tulilirockz: the thing is that the upstream's submodule points to a random outdated fork of wlroots on github tho

nekopsykose: the fork is actually the same as 0.18.0 when you look at the commit

but regardless of what it is, it's what they use in submodule so if you use submodules it's what you have to follow unless there's a reason not to

PR #2937, opened 2024-09-15
user/warp: new package

tulilirockz:


user/warp/template.py
```diff
@@ -0,0 +1,40 @@
...
+hostmakedepends = [
+    "meson",
+    "cargo-auditable",
+    "pkgconf",
+    "gettext",
+    "desktop-file-utils",
+    "itstool",
+]
+makedepends = [
+    "rust-std",
```
nekopsykose, line 15: sort these

tulilirockz: oh  true

PR #2944, opened 2024-09-16
user/snapper: new package

tulilirockz:


user/snapper/patches/btrfs.patch
```diff
@@ -0,0 +1,17 @@
...
+ #include <sys/types.h>
+ #include <sys/stat.h>
+ #include <unistd.h>
+-#ifdef HAVE_LIBBTRFS
+ #include <btrfs/ioctl.h>
+ #include <btrfs/send-utils.h>
+-#else
+-#include <linux/btrfs.h>
+-#endif
+-#ifdef HAVE_LIBBTRFSUTIL
```
nekopsykose, line 13: not sure what this fixes, this is already the path taken if these HAVE_ are set and they should be, if they're not then the build system has to be fixed instead..?

user/snapper/patches/ncurses.patch
```diff
@@ -0,0 +1,10 @@
+diff --git a/client/utils/Makefile.am b/client/utils/Makefile.am
+index e6a9968..aeeea5a 100644
+--- a/client/utils/Makefile.am
++++ b/client/utils/Makefile.am
+@@ -20,4 +20,4 @@ libutils_la_SOURCES =					\
+ 	CsvFormatter.cc	    	CsvFormatter.h		\
+ 	JsonFormatter.cc    	JsonFormatter.h
+ 
+-libutils_la_LIBADD = ../../snapper/libsnapper.la -ltinfo
++libutils_la_LIBADD = ../../snapper/libsnapper.la -lncurses
```
nekopsykose, line 10: this isn't needed, libtinfo compat symlinks exist, they're just in ncurses-libtinfo-devel for some reason, i'll put them in the main one

user/snapper/template.py
```diff
@@ -0,0 +1,64 @@
...
+    "boost-devel",
+    "dbus-devel",
+    "e2fsprogs-devel",
+    "json-c-devel",
+    "libbtrfs-devel",
+    "libbtrfsutil-devel",
+    "libmount-devel",
+    "libxml2-devel",
+    "linux-pam-devel",
+    "ncurses-devel",
```
nekopsykose, line 36: stray newlines

user/snapper/template.py
```diff
@@ -0,0 +1,64 @@
...
+    self.install_file("data/sysconfig.snapper", "etc/snapper", name="snapper")
+    self.install_service(self.files_path / "snapper")
+    self.install_tmpfiles(self.files_path / "tmpfiles.conf")
+    self.uninstall("usr/lib/snapper/testsuite")
+    self.uninstall("usr/lib/snapper/systemd-helper")
+
+
+@subpackage("snapper-libs")
+def _(self):
+    self.install_if = [self.parent]
```
nekopsykose, line 57: the parent links to the libs already so this install_if doesn't do anything

but there's also no reason to split them that i see

user/snapper/template.py
```diff
@@ -0,0 +1,64 @@
...
+    return self.default_libs()
+
+
+@subpackage("pam_snapper")
+def _(self):
+    self.install_if = [self.parent]
```
nekopsykose, line 63: don't need this subpackage either

user/snapper/patches/btrfs.patch
nekopsykose: i guess this is easier than making the build system define HAVE_LIBBTRFS, but the latter is defined ok afaict

user/snapper/template.py
nekopsykose: but you do need a -devel one

user/snapper/patches/btrfs.patch
tulilirockz: apparently these HAVE_s arent being set properly. ive tried doing a ton of stuff and it wouldnt fix this damn thing

nekopsykose: yeah, the former is not checked at all :)

user/snapper/template.py
tulilirockz: from what I know, transactional-update depends on libsnapper but not the snapper CLI... althought that probably would even work on chimera, right? ill just merge it into the base package

tulilirockz:
oh my god
what

tulilirockz:
what happened

nekopsykose:
you rebased something wrong

tulilirockz:
better now 👍

tulilirockz:
patched cron script cuz without that it wont get recognized by run-parts

PR #2947, opened 2024-09-17
user/udiskie: new package

tulilirockz:
I did strip out some dependencies like gtk3 and libappindicator for the tray indicator. Would that be of interest? I just thought that usually people would use this in WMs and stuff like that

### DIFF
--- a/user/udiskie/files/50-udiskie.rules
+++ b/user/udiskie/files/50-udiskie.rules
@@ -1,0 +1,24 @@
+/**
+ * Polkit permissions for udiskie.
+ * Source: Udiskie README (see https://pypi.python.org/pypi/udiskie)
+ */
+polkit.addRule(function(action, subject) {
+  var YES = polkit.Result.YES;
+  // NOTE: there must be a comma at the end of each line except for the last:
+  var permission = {
+    // required for udisks2:
+    "org.freedesktop.udisks2.filesystem-mount": YES,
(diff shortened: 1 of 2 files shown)


triallax:
> I did strip out some dependencies like gtk3 and libappindicator for the tray indicator. Would that be of interest? I just thought that usually people would use this in WMs and stuff like that

wms don't necessarily lack trays (e.g. sway), i'd add those deps

tulilirockz:
> > I did strip out some dependencies like gtk3 and libappindicator for the tray indicator. Would that be of interest? I just thought that usually people would use this in WMs and stuff like that
> 
> wms don't necessarily lack trays (e.g. sway), i'd add those deps

the thing is that python-appindicator is deprecated isnt it?

user/udiskie/files/50-udiskie.rules
```diff
@@ -0,0 +1,24 @@
+/**
+ * Polkit permissions for udiskie.
+ * Source: Udiskie README (see https://pypi.python.org/pypi/udiskie)
```
triallax, line 3: don't see it in the README, seems to be here instead now: https://github.com/coldfix/udiskie/wiki/Permissions

also, i don't really know anything about polkit but this file is missing `org.freedesktop.udisks2.encrypted-unlock-system`  and `org.freedesktop.udisks2.filesystem-unmount-others`, which are in the link above, you might want to have these in as well

PR #2954, opened 2024-09-18
main/fwupd: remove virtual made obsolete by contrib/main merge

triallax:
i don't think this requires a new `provides` but correct me if i'm wrong


nekopsykose:
don't think it does, nobody would install that

PR #2961, opened 2024-09-19
main/protobuf: update to 28.2

tulilirockz:


nekopsykose:
this needs rebuilds every time it's upgraded because the soname changes

PR #2971, opened 2024-09-20
main/libcxx-mingw-w64: disable scp hardening

z-erica:
silences the following warning:

```
clang: warning: argument unused during compilation: '-fstack-clash-protection' [-Wunused-command-line-argument]
```


nekopsykose:
does that warning apply to every tu though? it's probably unused for .dll but there is some normal code in there too i think

PR #2973, opened 2024-09-20
cbuild: reset text formatting after error hints

EarldridgeJazzedPineda:


q66:
why does this matter? the color escapes are per-line and the hint is the last thing ever printed

q66:
i guess if python crashed right after printing the hint it could leave the terminal inconsistent? seems unlikely though

q66:
hm looks like other places already reset it so whatever

PR #2975, opened 2024-09-20
user/transactional-update: new package

tulilirockz:
OpenSuSE MicroOS's approach to immutability and its dependencies, depends on snapper - #2944

### DIFF
--- a/user/libeconf/patches/meson.patch
+++ b/user/libeconf/patches/meson.patch
@@ -1,0 +1,12 @@
+--- a/meson.build	2024-09-20 17:11:49.271033354 -0300
++++ a/meson.build	2024-09-20 17:11:58.804736776 -0300
+@@ -15,8 +15,7 @@
+ inc = include_directories('include')
+ 
+ add_project_arguments(['-D_GNU_SOURCE=1',
+-		       '-DXTSTRINGDEFINES',
+-		       '-D_FORTIFY_SOURCE=2'], language : 'c')
++		       '-DXTSTRINGDEFINES'], language : 'c')
+ 
+ possible_cc_flags = [
+ 		  '-Dwerror=true',
--- a/user/libeconf/template.py
+++ b/user/libeconf/template.py
@@ -1,0 +1,27 @@
+pkgname = "libeconf"
+pkgver = "0.7.4"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Ddefault_library=shared"]
+hostmakedepends = ["meson", "pkgconf", "doxygen"]
+pkgdesc = "Config file parser"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "MIT"
+url = "https://github.com/openSUSE/libeconf"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "4174ca94e958cbb6c8bb4ea9e6909877d6178e00b6a65349eade64a462534da0"
+options = ["linkundefver"]
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+
+
+@subpackage("libeconf-devel")
+def _(self):
+    return self.default_devel()
+
+
+@subpackage("libeconf-progs")
+def _(self):
+    return self.default_progs()
--- a/user/transactional-update/template.py
+++ b/user/transactional-update/template.py
@@ -1,0 +1,48 @@
+pkgname = "transactional-update"
+pkgver = "4.8.3"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_env = {
+    "SYSTEMDDIR": "/usr/lib/systemd/system",
+    "TMPFILESDIR": "/usr/lib/tmpfiles.d",
+    "DRACUTDIR": "/usr/lib/dracut/modules.d",
+}
+hostmakedepends = ["autoconf-archive", "automake", "pkgconf", "slibtool"]
+makedepends = [
+    "dbus-devel",
+    "elogind-devel",
+    "libeconf-devel",
+    "libmount-devel",
+    "udev-devel",
+]
+depends = ["snapper"]
+pkgdesc = "Toolkit for atomic updates"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "LGPL-2.1-or-later AND GPL-2.0-or-later"
+url = "https://github.com/openSUSE/transactional-update"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "2dc4e39b361d4d776ec3e9eef8ab6ca55e710a859b95477a77e5818d046838a7"
+
+
+def post_install(self):
+    self.install_service(self.files_path / "tukitd")
+    self.install_initramfs(self.files_path / "tukit.hook")
+    self.install_initramfs(self.files_path / "tukit.init-bottom", "init-bottom")
+
+
+@subpackage("transactional-update-initramfs-tools")
+def _(self):
+    self.subdesc = "initramfs scripts"
+    self.install_if = []
+    self.depends = ["initramfs-tools", self.parent]
+    return ["usr/share/initramfs-tools"]
+
+
+@subpackage("transactional-update-devel")
+def _(self):
+    return self.default_devel()
+
+
+@subpackage("transactional-update-libs")
+def _(self):
+    return self.default_libs()
--- a/user/libeconf-devel
+++ b/user/libeconf-devel
@@ -1,0 +1,1 @@
+libeconf
--- a/user/libeconf-progs
+++ b/user/libeconf-progs
@@ -1,0 +1,1 @@
+libeconf
--- a/user/libeconf/patches/tests.patch
+++ b/user/libeconf/patches/tests.patch
@@ -1,0 +1,10 @@
+Fails on Musl
+--- a/tests/meson.build	2024-09-18 07:53:14.000000000 -0300
++++ b/tests/meson.build	2024-09-20 07:10:42.449539955 -0300
+@@ -203,6 +203,3 @@
+ test('tst-shells2', tst_shells2_exe)
+ 
+ 
+-test('tst_econftool1', find_program('tst-econftool1.sh'))
+-test('tst_econftool_show1', find_program('tst-econftool_show1.sh'))
+-test('tst_econftool_cat', find_program('tst-econftool_cat.sh'))
--- a/user/transactional-update-devel
+++ b/user/transactional-update-devel
@@ -1,0 +1,1 @@
+transactional-update
--- a/user/transactional-update-initramfs-tools
+++ b/user/transactional-update-initramfs-tools
@@ -1,0 +1,1 @@
+transactional-update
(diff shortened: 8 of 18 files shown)


user/libeconf/template.py
```diff
@@ -0,0 +1,27 @@
+pkgname = "libeconf"
+pkgver = "0.7.3"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = ["meson", "ninja", "pkgconf", "doxygen"]
+makedepends = []
```
nekopsykose, line 6:
```suggestion
hostmakedepends = ["meson", "pkgconf", "doxygen"]
```

user/libeconf/patches/meson.patch
```diff
@@ -0,0 +1,14 @@
+--- a/meson.build	2024-09-18 07:53:14.000000000 -0300
++++ b/meson.build	2024-09-20 06:48:05.316117427 -0300
+@@ -77,11 +77,6 @@
+   version : meson.project_version(),
+ )
+ 
+-libeconf_static = static_library('econf',
```
nekopsykose, line 7: you can delete this patch and do -Ddefault_library=shared in meson args, pretty sure

user/libeconf/template.py
```diff
@@ -0,0 +1,27 @@
...
+pkgdesc = "Config file parser"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "MIT"
+url = "https://github.com/openSUSE/libeconf"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "2e185c86674cdf128b518988cc740c27bdc93adafadf65cb6e3e860538fefefa"
+tool_flags = {"CFLAGS": ["-Wl,--undefined-version", "-D_FORTIFY_SOURCE=2"]}
```
nekopsykose, line 13:
```suggestion
```

the fortify doesn't do anything, and the -Wl is ldflags not cflags, but use options = linkundefver for that

nekopsykose: if you want to silence the fortify warnings then delete https://github.com/openSUSE/libeconf/blob/d78b86e72b5818b93f06c54b8fb8e19669b45844/meson.build#L19 i guess

user/tukit/template.py
```diff
@@ -0,0 +1,48 @@
...
+}
+hostmakedepends = [
+    "automake",
+    "slibtool",
+    "pkgconf",
+]
+makedepends = [
+    "udev-devel",
+    "dbus-devel",
+    "autoconf-archive",
```
nekopsykose, line 18: -archive is hostmakedepends

also sort the lists

user/tukit/template.py
```diff
@@ -0,0 +1,48 @@
...
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "LGPL-2.1-or-later AND GPL-2.0-or-later"
+url = "https://github.com/openSUSE/transactional-update"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "fc67baa3b34723d73cc0366abf74bf7f0a21e4d3ec08d29bc38e846d493c20b0"
+
+
+def pre_install(self):
+    self.install_dir("usr")
+    self.install_link("usr/etc", "../etc")
```
nekopsykose, line 33: fix this in the buildsystem instead unless it's like 100 lines of fixes

user/tukit/template.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def post_install(self):
+    self.uninstall("usr/etc")
+    self.uninstall("usr/lib/pkgconfig")
```
nekopsykose, line 38: why is the pkgconf removed

tulilirockz:
should be good now

tulilirockz:
also, ive noticed that it generates stuff with `/sysroot`. what directory does initramfs-tools generate for the new root? `/newroot`? `/root`? - probably should patch to make that work fine

nekopsykose:
/init says rootmnt=/root and does stuff with that, so probably that, feel free to test

tulilirockz:
yeah i also checked it out rn - probably will work with /root

tulilirockz:
probably should make a initramfs script that mounts `overlayfs /etc` off of the user's fstab (if it is applied) then moves it over to /root/etc

q66:
how does this even work at all? it seems to need dracut

also shouldn't it be called transactional-update

q66:
i'm not really interested in packaging software that will conditionally break because of requiring random stuff depending on how you invoke it (i see references to... dracut, systemctl, ... and considering a lot of this consists of random shell that is presumably not tested on non-gnu userland, there is probably more, even if it "works" on the surface)

so i think this is only fine if you can guarantee that this stuff will not randomly break and that it won't be a net longer term maintenance burden

tulilirockz:
> how does this even work at all? it seems to need dracut
> 
> also shouldn't it be called transactional-update

Transactional-update is the OpenSUSE-specific implementation of this thing, tukit is just a snapper wrapper that makes is to you can do something like `tukit execute apk upgrade` and have the updates be applied on-boot if they worked

tulilirockz:
> i'm not really interested in packaging software that will conditionally break because of requiring random stuff depending on how you invoke it (i see references to... dracut, systemctl, ... and considering a lot of this consists of random shell that is presumably not tested on non-gnu userland, there is probably more, even if it "works" on the surface)
> 
> so i think this is only fine if you can guarantee that this stuff will not randomly break and that it won't be a net longer term maintenance burden

Most of these dependencies only appear on the tukitd thing which is not necessary for the usage of this system, you can totally just go with tukit and ur good to go. Transacional-update is also not necessary due to it just straight up being an OpenSUSE tool, so no need to worry about the shell wrapper

tulilirockz:
The dracut dependency is also just for making it so the initramfs mounts /etc as an overlayfs like this fstab entry `overlayfs /etc overlay defaults,lowerdir=/root/var/lib/overlay/15/etc:/root/sysroot/etc,upperdir=/root/var/lib/overlay/17/etc,workdir=/root/var/lib/overlay/17/work-etc 0 2`.

tulilirockz:
i just added the initramfs hooks! they work completely fine apparently (although there is a warning when running update-initramfs). everything else, like etc cleanup or whatever should be implemented by a higher level tool (like transactional-update on suse) - this should actually totally be good to go now! maybe this isnt a good idea to merge due to the maintenance burden, though. what do you all think?

q66:
i don't really understand why it's the initramfs that needs to mount that, why can't it be done after initramfs by early dinit services?

q66:
also as far as i can tell the package should still be called transactional-update as that's the name of the repo/project

also i don't see any reason to leave out tukitd? it only depends on sd-bus api (you can supply it with `basu` or `elogind`, preferably the former here) and the systemd service seems optional (the dbus service has traditional activation) so you can just not install it

tulilirockz:
> also as far as i can tell the package should still be called transactional-update as that's the name of the repo/project

i mean, at this point it is just a naming thing - i just think that it makes more sense for it to be "tukit" since we arent really messing around wirh the `transactional-update` cli, right?

tulilirockz:
> also i don't see any reason to leave out tukitd? it only depends on sd-bus api (you can supply it with `basu` or `elogind`, preferably the former here) and the systemd service seems optional (the dbus service has traditional activation) so you can just not install it
i just didnr know you could do that! ill be changinf this PR to include that! sweet!

tulilirockz:
> i don't really understand why it's the initramfs that needs to mount that, why can't it be done after initramfs by early dinit services?

honestly it doesnt make that much sense indeed, this probaly could just run on the `early-` time on boot. but from what i understood you arent really supposed to have an /etc folder, it jusr mounts /var/lib/overlay/(current overlay)/etc + (previous overlay etc) to provide that. mounting it on initramfs time makes it so dinit works fine while not having that base `/etc` folder

q66:
dinit should still work fine without it if the overlay mount is done early enough, it does not depend on that in any way

q66:
> > also as far as i can tell the package should still be called transactional-update as that's the name of the repo/project
> 
> i mean, at this point it is just a naming thing - i just think that it makes more sense for it to be "tukit" since we arent really messing around wirh the `transactional-update` cli, right?

https://github.com/chimera-linux/cports/blob/master/Packaging.md#naming

package name is the project/repo name

tulilirockz:
> > > also as far as i can tell the package should still be called transactional-update as that's the name of the repo/project
> > 
> > 
> > i mean, at this point it is just a naming thing - i just think that it makes more sense for it to be "tukit" since we arent really messing around wirh the `transactional-update` cli, right?
> 
> https://github.com/chimera-linux/cports/blob/master/Packaging.md#naming
> 
> package name is the project/repo name

fair! ill just rename it real quick

tulilirockz:
> dinit should still work fine without it if the overlay mount is done early enough, it does not depend on that in any way

i thought it had to have the /etc/dinit.d directory to actually work, but i guess theres /usr/lib/dinit.d, right?

tulilirockz:
Not mounting /etc on the initramfs leads to some really funky behaviour from tukit, like, it just cant do anything outside of `$PWD=/` for some reason

q66:
> > dinit should still work fine without it if the overlay mount is done early enough, it does not depend on that in any way
> 
> i thought it had to have the /etc/dinit.d directory to actually work, but i guess theres /usr/lib/dinit.d, right?

it can get services from there, but it doesn't need it present per se, especially not in early boot (so it's fine if it appears later)

initramfs shouldn't be doing this because one doesn't even necessarily need to have an initramfs, it should never be a hard assumption

tulilirockz:
After some debugging and testing, I managed to make it work without the initramfs things! - This, though, doesnt seem like the best solution... Like, upstream they just expect that you have an initrd generated to mount /etc. Should this package still have the initramfs stuff as an optional subpackage or should it just only support the non-initramfs approach? @q66

tulilirockz:
Been using this on my main system for a bit and it doesnt seem to have broken anything? - I hope this can be merged since its such a wonderful way for atomic updates.

tulilirockz:
oh! thanks for the typo fix! i didnt see it earlier @triallax

lz-coder:
> Been using this on my main system for a bit - without using the initramfs hook - and it doesn't seem to have broken anything? - I hope this can be merged since its such a wonderful way for atomic updates.

Hi, what partition scheme should i use for this to work? and what you used to accomplish the same transactional-update(the openSUSE front-end one with zypper wrappers...) functionality?

user/transactional-update/template.py
```diff
@@ -0,0 +1,48 @@
...
+sha256 = "2dc4e39b361d4d776ec3e9eef8ab6ca55e710a859b95477a77e5818d046838a7"
+
+
+def post_install(self):
+    self.install_service(self.files_path / "tukitd")
+    self.install_initramfs(self.files_path / "tukit.hook")
+    self.install_initramfs(self.files_path / "tukit.init-bottom", "init-bottom")
+
+
+@subpackage("transactional-update-initramfs-tools")
```
q66, line 33: this will split automatically, you don't need to declare it

PR #2978, opened 2024-09-21
user/openttd: new package

breakgimme:
i've disabled cross-compilation for this package because binaries that are compiled are used later in the build process and i don't see a way to do this (or do this sensibly at least) in cbuild with the upstream solution (https://github.com/OpenTTD/OpenTTD/pull/8350)
please correct me if i'm wrong though

### DIFF
--- a/user/openttd/template.py
+++ b/user/openttd/template.py
@@ -1,0 +1,38 @@
+pkgname = "openttd"
+pkgver = "14.1"
+pkgrel = 0
+build_style = "cmake"
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+makedepends = [
+    "fluidsynth-devel",
+    "fontconfig-devel",
+    "freetype-devel",
+    "harfbuzz-devel",
+    "libcurl-devel",
+    "libpng-devel",
+    "lzo-devel",
+    "icu-devel",
+    "sdl-devel",
+    "xz-devel",
+    "zlib-ng-devel",
+]
+pkgdesc = "Simulation game based upon Transport Tycoon Deluxe"
+maintainer = "breakgimme <adam@plock.com>"
+license = "GPL-2.0-or-later"
+url = "https://openttd.org"
+source = f"https://cdn.openttd.org/openttd-releases/{pkgver}/openttd-{pkgver}-source.tar.xz"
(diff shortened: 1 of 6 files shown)


user/openttd/template.py
```diff
@@ -0,0 +1,38 @@
+pkgname = "openttd"
+pkgver = "14.1"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DCMAKE_BUILD_TYPE=Release", "-DCMAKE_INSTALL_BINDIR=bin"]
```
nekopsykose, line 5:
```suggestion
```

neither of these do anything meaningful; the bindir i'll add to cbuild instead

user/openttd/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "harfbuzz-devel",
+    "libcurl-devel",
+    "libpng-devel",
+    "lzo-devel",
+    "icu-devel",
+    "sdl-devel",
+    "xz-devel",
+    "zlib-ng-devel",
+]
+pkgdesc = "Open source simulation game based upon Transport Tycoon Deluxe"
```
nekopsykose, line 20:
```suggestion
pkgdesc = "Simulation game based upon Transport Tycoon Deluxe"
```

user/openttd/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://cdn.openttd.org/openttd-releases/latest.yaml"
+pattern = r"version:\s*(.+)"
```
nekopsykose, line 2:
```suggestion
pattern = r"version:\s*([\d.]+)"
```

user/openttd/template.py
```diff
@@ -0,0 +1,38 @@
...
+options = ["!cross"]
+
+
+def check(self):
+    self.do("./build/openttd_test")
+
+
+@subpackage("openttd-locale")
+def _(self):
+    return ["usr/share/games/openttd/lang/*[!english]*.lng"]
```
nekopsykose, line 38: add self.install_if = [self.parent, "base-locale"]

nekopsykose:
also feel free to patch out the

```
        "$<$<NOT:$<CONFIG:Debug>>:-D_FORTIFY_SOURCE=2>" # FORTIFY_SOURCE should only be used in non-debug builds (requires -O1+)
```

in cmake/compileflags so it doesn't spam redefs

PR #2983, opened 2024-09-22
user/opentabletdriver-udev: new package

tulilirockz:
This adds some Udev rules so that the flatpak version of OpenTabletDriver works fine, just needed to create a group to share the rules to any user, but thats all

### DIFF
--- a/user/opentabletdriver-udev/patches/git.patch
+++ b/user/opentabletdriver-udev/patches/git.patch
@@ -1,0 +1,20 @@
+--- generate-rules.sh	2024-09-22 02:10:25.521129020 -0300
++++ generate-rules.sh	2024-09-22 02:11:43.889844039 -0300
+@@ -8,7 +8,7 @@
+   echo "  -h, --help                              Print this help message"
+ }
+ 
+-for c in git jq tr awk sed; do
++for c in jq tr awk sed; do
+   command -v $c > /dev/null
+   if [[ $? > 0 ]]; then
+     echo "Error: Command $c not found in \$PATH." >&2
+@@ -23,7 +23,7 @@
+ shopt -s globstar
+ set -eu
+ 
+-OTD_CONFIGURATIONS="${OTD_CONFIGURATIONS:="$(git rev-parse --show-toplevel)/OpenTabletDriver.Configurations/Configurations"}"
++OTD_CONFIGURATIONS="./OpenTabletDriver.Configurations/Configurations"
+ 
+ while [ $# -gt 0 ]; do
+   case "$1" in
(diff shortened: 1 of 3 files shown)


nekopsykose:
i don't think an all-accessing group is correct, hidraw devices are already uaccess and you should be able to access them as your user

tulilirockz:
thats really weird tho, i really cant access the `/dev/hidraw*` devices from the flatpak :(

tulilirockz:
after you mentioned that it just started working fine :+1: it works

tulilirockz:
actually no it doesnt??? flatpak is acting so weird what
```
Detect:Debug]	Searching for tablet 'Wacom CTL-472'
[DeviceUnauthorizedAccessException:Debug]	Not permitted to open HID class device at /dev/hidraw0.
   at HidSharp.Platform.Linux.LinuxHidStream.DeviceHandleFromPath(String path, HidDevice hidDevice, oflag oflag)
   at HidSharp.Platform.Linux.LinuxHidDevice.TryParseReportDescriptor(ReportDescriptor& parser, Byte[]& report
```

nekopsykose:
the https://github.com/OpenTabletDriver/OpenTabletDriver/blob/3d887e6cf638031d5ab762f95a24e1e8071ddd2b/generate-rules.sh#L65 should be setting uaccess on all the devices so afaict it should work

also the above line for uinput conflicts with the existing uinput rule we set

tulilirockz:
it just works completely fine, i was getting issues due to the flatpak and me restarting opentabletdriver 1 billion times to see if it worked at all. after a reboot it just works fine

tulilirockz:
> the https://github.com/OpenTabletDriver/OpenTabletDriver/blob/3d887e6cf638031d5ab762f95a24e1e8071ddd2b/generate-rules.sh#L65 should be setting uaccess on all the devices so afaict it should work
> 
> also the above line for uinput conflicts with the existing uinput rule we set

should i patch it out?

nekopsykose:
yea

nekopsykose:
the `rm` rule also looks weird but i don't actually know what it's meant to fix

tulilirockz:
> the `rm` rule also looks weird but i don't actually know what it's meant to fix

probably removes any duplicate virtual opentabletdriver input if somehow its still bound to /dev? - honestly probably not even needed

tulilirockz:
the rm rule is absolutely necessary for artist mode :p it just does not even move the cursor if you dont have that one

tulilirockz:
should be good now

nekopsykose:
yep, looks correct, nice to see it works :) deathmist wanted these before..

PR #2984, opened 2024-09-22
user/forgejo-runner: new package

Gnarwhal:
The setup instructions for forgejo-runner have it run as a separate user, but I don't know how to get forgejo-runner access to the podman socket when not run as root. Would it be possible to have podman create the socket with rw access for an `_podman` (or whatever it should be called) group? Or maybe running it as root is not a concern.

### DIFF
--- a/user/forgejo-runner/files/docker_host.env
+++ b/user/forgejo-runner/files/docker_host.env
@@ -1,0 +1,1 @@
+DOCKER_HOST=unix:///run/podman/podman.sock
--- a/user/forgejo-runner/files/forgejo-runner
+++ b/user/forgejo-runner/files/forgejo-runner
@@ -1,0 +1,7 @@
+type = process
+command = /usr/bin/forgejo-runner daemon
+working-dir = /var/lib/forgejo-runner/$1
+env-file = /usr/share/forgejo-runner/docker_host.env
+depends-on = podman
+logfile = /var/log/forgejo-runner.log
+smooth-recovery = true
--- a/user/forgejo-runner/files/tmpfiles.conf
+++ b/user/forgejo-runner/files/tmpfiles.conf
@@ -1,0 +1,3 @@
+# Create forgejo-runner home directory
+
+d /var/lib/forgejo-runner 0750 root root -
--- a/user/forgejo-runner/template.py
+++ b/user/forgejo-runner/template.py
@@ -1,0 +1,32 @@
+pkgname = "forgejo-runner"
+pkgver = "3.5.1"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    f"-ldflags=-X gitea.com/gitea/act_runner/internal/pkg/ver.version=v{pkgver}"
+]
+make_check_args = [
+    "-skip",
+    "Test_runCreateRunnerFile|Test_ping",
+    "./...",
+]
+hostmakedepends = ["go"]
+depends = ["podman"]
+pkgdesc = "Daemon that runs CI jobs for Forgejo"
+maintainer = "Gnarwhal <git.aspect893@passmail.net>"
+license = "MIT"
+url = "https://code.forgejo.org/forgejo/runner"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "0c53e5ee6ea57eadcc3ebbb4fbe203478500f78b058b13d8ccc3b63b775955c0"
+
+
+def install(self):
+    self.install_bin(
+        f"{self.cwd}/{self.make_dir}/act_runner", name="forgejo-runner"
+    )
+    self.install_file(
+        self.files_path / "docker_host.env", "usr/share/forgejo-runner"
+    )
+    self.install_service(self.files_path / "forgejo-runner")
+    self.install_tmpfiles(self.files_path / "tmpfiles.conf")
+    self.install_license("LICENSE")
--- a/user/forgejo-runner/update.py
+++ b/user/forgejo-runner/update.py
@@ -1,0 +1,2 @@
+url = "https://code.forgejo.org/forgejo/runner/tags"
+pattern = r"/archive/v([\d.]+).tar.gz"


user/forgejo-runner/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://code.forgejo.org/forgejo/runner/tags"
+pattern = "/archive/v([\\d.]+)(?=\\.tar\\.gz)"
```
nekopsykose, line 2:
```suggestion
url = "https://code.forgejo.org/forgejo/runner/tags"
pattern = r"/archive/v([\d.]+).tar.gz"
```

user/forgejo-runner/template.py
```diff
@@ -0,0 +1,51 @@
...
+
+def install(self):
+    self.install_bin("forgejo-runner")
+    self.install_file(
+        self.files_path / "docker_host.env", "usr/share/forgejo-runner"
+    )
+    self.install_service(self.files_path / "forgejo-runner")
+    self.install_tmpfiles(self.files_path / "tmpfiles.conf")
+
+    self.install_license("LICENSE")
```
nekopsykose, line 51:
```suggestion
pkgname = "forgejo-runner"
pkgver = "3.5.1"
pkgrel = 0
build_style = "go"
make_check_args = [
    "-skip",
    "Test_runCreateRunnerFile|Test_ping",
    "./...",
]
hostmakedepends = ["go"]
depends = ["podman"]
pkgdesc = "Daemon that runs CI jobs for Forgejo"
maintainer = "Gnarwhal <git.aspect893@passmail.net>"
license = "MIT"
url = "https://code.forgejo.org/forgejo/runner"
source = f"{url}/archive/v{pkgver}.tar.gz"
sha256 = "0c53e5ee6ea57eadcc3ebbb4fbe203478500f78b058b13d8ccc3b63b775955c0"


def install(self):
    self.install_bin(
        f"{self.cwd}/{self.make_dir}/act_runner", name="forgejo-runner"
    )
    self.install_file(
        self.files_path / "docker_host.env", "usr/share/forgejo-runner"
    )
    self.install_service(self.files_path / "forgejo-runner")
    self.install_tmpfiles(self.files_path / "tmpfiles.conf")
    self.install_license("LICENSE")
```

and delete both the patches

nekopsykose:
> The setup instructions for forgejo-runner have it run as a separate user, but I don't know how to get forgejo-runner access to the podman socket when not run as root. Would it be possible to have podman create the socket with rw access for an _podman (or whatever it should be called) group?

whoever can access the socket has the same permissions as the podman api service, so currently even if the socket was "`_podman`" it would just mean whoever accesses it gets root. i don't think it's easy to run that api service as non-root either, so there's no point in a separate user for this

user/forgejo-runner/files/docker_host.env
```diff
@@ -0,0 +1 @@
+DOCKER_HOST=unix://${XDG_RUNTIME_DIR}/podman/podman.sock
```
nekopsykose, line 1: podman api socket is in /run/podman/podman.sock, and the service won't have a xdg_runtime_dir either, so how did this work

user/forgejo-runner/files/forgejo-runner
```diff
@@ -0,0 +1,9 @@
+# forgejo-runner service
+
+type = process
+command = /usr/bin/forgejo-runner daemon
+working-dir = /var/lib/forgejo-runner
+env-file = /usr/share/forgejo-runner/docker_host.env
+log-type = file
```
nekopsykose, line 7:
```suggestion
```

redundant with logfile

user/forgejo-runner/files/forgejo-runner
```diff
@@ -0,0 +1,9 @@
+# forgejo-runner service
+
```
nekopsykose, line 2:
```suggestion
```

user/forgejo-runner/files/tmpfiles.conf
```diff
@@ -0,0 +1,3 @@
+# Create forgejo-runner home directory
+
+d /var/lib/forgejo-runner 0755 root root -
```
nekopsykose, line 3:
```suggestion
d /var/lib/forgejo-runner 0750 root root -
```

probably

user/forgejo-runner/template.py
nekopsykose: feel free to add make_build_args for the `f"-ldflags=-X gitea.com/gitea/act_runner/internal/pkg/ver.version=v{pkgver}"` too

Gnarwhal: Unreal...

user/forgejo-runner/files/docker_host.env
Gnarwhal: Um, well, it did not

user/forgejo-runner/files/forgejo-runner
```diff
@@ -0,0 +1,7 @@
+type = process
+command = /usr/bin/forgejo-runner daemon
+working-dir = /var/lib/forgejo-runner/$1
```
nekopsykose, line 3: a) nothing creates this directory, so how does it work? i'm pretty sure dinit refuses to start a service with a nonexistent working-dir=

b) when do you ever run more than one of these at once with a separate workdir?

c) if you do run more than one, how do they share the same logfile= below? i am pretty sure that can't work

nekopsykose:
and now that i look at it again, d):

the repo has a fat warning on it: `WARNING: this is alpha release quality code and should not be considered secure enough to deploy in production.`, so pretty sure nobody is meant to use this?

user/forgejo-runner/files/forgejo-runner
Gnarwhal: a) It does not work without manual intervention by the user, but this service is only useful if the user has manually setup and registered the runner anyway. Requiring one more `mkdir` doesn't seem unreasonable to me. I don't really know how to provide documentation to a user about how to use a service that requires setup. Just comments in the service file?

b) You can create runners with different capabilities (this affects which runner `runs on` selects). You can also create runners for different contexts (global, organization, user, repo). It seems reasonable to me that someone may want more than one, but I'm often wrong. 
  
c) Fair point

d) I feel like its got a pretty generous amount of installation instructions in the main forgejo docs for something literally nobody should use, but I can understand why it may be undesirable to have in the repos. Maybe add `restricted`?

PR #2987, opened 2024-09-23
main/pipewire: enable roc support

triallax:
probably should figure out the really annoying noise (~~according to psykose udp has a lot of noise with musl's memcpy~~i was trolled)

### DIFF
--- a/main/pipewire/template.py
+++ b/main/pipewire/template.py
@@ -1,6 +1,6 @@
 pkgname = "pipewire"
 pkgver = "1.2.5"
-pkgrel = 1
+pkgrel = 2
 build_style = "meson"
 configure_args = [
     "--auto-features=enabled",
@@ -10,7 +10,6 @@ configure_args = [
     "-Dsdl2=disabled",  # examples
     "-Dsystemd=disabled",
     "-Dlibffado=disabled",
-    "-Droc=disabled",  # TODO
     "-Dselinux=disabled",
     "-Dsnap=disabled",
     "-Dbluez5-codec-lc3plus=disabled",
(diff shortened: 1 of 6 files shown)


nekopsykose:
> (according to psykose udp has a lot of noise with musl's memcpy)

this was an audiophile shitpost lmao, haha

main/pipewire/template.py
```diff
@@ -9,7 +9,6 @@
     "-Dsdl2=disabled",  # examples
     "-Dsystemd=disabled",
     "-Dlibffado=disabled",
-    "-Droc=disabled",  # TODO
```
nekopsykose, line 12: should prolly split the module and the libs below and then install_if = parent, roc-toolkit

triallax:
@nekopsykose classic

main/pipewire/template.py
triallax: should be good now

main/roc-toolkit/template.py
```diff
@@ -0,0 +1,52 @@
...
+hardening = ["vis", "cfi"]
+# check: needs cpputest
+# cross: scons, can't be  bothered
+options = ["!check", "!cross"]
+
+
+_scons_flags = [
+    "--disable-openfec",
+    "--disable-sox",
+    "--disable-alsa",
```
nekopsykose, line 36: what setup do you use that you get crackling in? is it pipewire on both sides using the plugins, or is one side using roc-send/recv?

if you're using the cli utils on either side then you might actually want to have alsa enabled so they are directly outputting and not going through an extra layer, or at least that's a guess

triallax: sender is chimera linux with pipewire-roc, the receiver is android using roc droid, so i think what you're suggesting doesn't apply

q66:
stale, feel free to reopen when you feel like working on it

PR #2990, opened 2024-09-23
user/kind: new package

tulilirockz:


user/kind/template.py
```diff
@@ -0,0 +1,32 @@
...
+            self.do(
+                "build/kind",
+                "completion",
+                completion,
+                stdout=o,
+            )
+
+
+def post_install(self):
+    self.install_license("LICENSE")
```
triallax, line 30: no need to install apache-2.0

triallax: the most common licenses that should be installed are `MIT` and `BSD*` (see `src/cbuild/core/spdx.py` if you're feeling curious)

PR #2991, opened 2024-09-23
replace eog with loupe

Drsheppard01:


q66:
nah, i don't think so yet

Drsheppard01:
Not to be intrusive, but instead of the familiar gnome-terminal and gedit, gnome meta comes with their counterparts with less functionality. The gnome meta package also includes gnome-screenshot, but its function was implemented in gnome-shell

q66:
i don't see how any of that is related to this?

Drsheppard01:
Of course I should have asked for a reason why you're not ready to merge it. But if I understand correctly, the reason is a lack of functional parity

PR #2992, opened 2024-09-24
user/gtk4-layer-shell: new package

valpackett:
[It's like main/gtk-layer-shell but gtk4](https://github.com/wmww/gtk4-layer-shell).

### DIFF
--- a/user/gtk4-layer-shell-devel
+++ b/user/gtk4-layer-shell-devel
@@ -1,0 +1,1 @@
+gtk4-layer-shell
--- a/user/gtk4-layer-shell/template.py
+++ b/user/gtk4-layer-shell/template.py
@@ -1,0 +1,32 @@
+pkgname = "gtk4-layer-shell"
+pkgver = "1.0.3"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "gobject-introspection",
+    "meson",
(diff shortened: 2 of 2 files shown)


nekopsykose:
yeah, but what uses it currently? also does it work? it relies on some libwayland interposition hax to intercept function calls from gtk or something and that didn't look like it was guaranteed to work

also put it in user/

valpackett:
> what uses it currently?

[sysmenu](https://github.com/System64fumo/sysmenu) and [dynisland](https://github.com/cr3eperall/dynisland) at least, but I mostly want it for my own development right now

> that didn't look like it was guaranteed to work

eh, it's cursed but should be very robust as libwayland and xdg-shell themselves are not gonna change much (?)

> also put it in user/

Sure. Definitely fair for something with no users (heh) in tree.

But overall I'm not a fan of this new trend of everything getting pushed away to user/. Why did you put niri there? :disappointed:

nekopsykose:
> but I mostly want it for my own development right now

sure :)

> should be very robust as libwayland and xdg-shell themselves are not gonna change much (?)

i meant more like maybe -fno-semantic-interposition on everything broke it, but seems not, things still work so that's ok

> Why did you put niri there? 

dunno..

PR #3005, opened 2024-09-25
main/fcitx5: add bash dependency

denisstrizhkin:
`fcitx5-diagnose` requires bash to work because it's a bash script


q66:
if it's not mandatory for function it should be split in a separate package because pulling in bash through library dependencies is bad

nekopsykose:
fcitx5 isn't a library dep, it's a whole ime you have to manually install and is never present otherwise

should still be split anyway tho

denisstrizhkin:
As in make fcitx5-diagnose a separate package?

main/fcitx5/template.py
```diff
@@ -56,3 +56,9 @@ def post_extract(self):
 @subpackage("fcitx5-devel")
 def _(self):
     return self.default_devel()
+
+
+@subpackage("fcitx5-diagnose")
+def _(self):
+    self.depends = ["bash"]
```
nekopsykose, line 63:
```suggestion
    self.depends = [self.parent, "bash"]
    self.install_if = [self.parent, "bash"]
```

PR #3006, opened 2024-09-25
main/python-tzdata: update to 2024.2

ttyyls:


triallax:
thank you for the pr, but i'm wondering why this package exists in the first place? according to https://github.com/python/tzdata?tab=readme-ov-file#tzdata-python-package-providing-iana-time-zone-data:
> It is intended to be a fallback for systems that do not have system time zone data installed (or don't have it installed in a standard location)

i checked our python's configure process and here it is:
```
checking for --with-tzpath... "/usr/share/zoneinfo:/usr/lib/zoneinfo:/usr/share/lib/zoneinfo:/etc/zoneinfo"
```
(our tzdata is at `/usr/share/zoneinfo`)

so installing `tzdata` instead of `python-tzdata` (which is just a redistribution of the tzdata database) both allow python programs to access timezone data just fine, so unless i'm missing something this package isn't necessary

(though this isn't directly relevant to this pr, it might make sense to add `tzdata` as a runtime dep of python? it's not a hard dependency however so unsure about it)

nekopsykose:
i thought it might've existed to at least make the 'something does import tzdata' case work without a patch, but no, this has never been used by anything

ttyyls:
it was a checkd dependency of python-jsonickle which was brought by python-pandas which was a dependency of certbot, another maintainer rebased out that dep by skipping the pandas test thus dropping pandas but not it's leaf dep: python-tzdata.

This should've never made it in to begin with, drop it at your convenience.

triallax:
thank you, dropped in c06ef5c

PR #3011, opened 2024-09-26
user/nxengine-evo: new package

tulilirockz:
just had to get the data/ directory from the windows release cuz the source code does not contain the necessary assets to actually boot up the game

### DIFF
--- a/user/nxengine-evo/template.py
+++ b/user/nxengine-evo/template.py
@@ -1,0 +1,37 @@
+pkgname = "nxengine-evo"
+# change source on update
+pkgver = "2.6.5.1"
+pkgrel = 0
+build_style = "cmake"
+hostmakedepends = [
+    "cmake",
+    "ninja",
+]
+makedepends = [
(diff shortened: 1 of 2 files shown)


user/nxengine-evo/template.py
```diff
@@ -0,0 +1,35 @@
+pkgname = "nxengine-evo"
+pkgver = "2.6.5"
```
nekopsykose, line 2:
```suggestion
pkgver = "2.6.5.1"
```

nekopsykose:
this has ancient vendored copies of utfcpp/spdlog in deps/ that will fail to compile sooner or later

tulilirockz:
managed to forcefully link the system libraries and i changed the include_path to have the necessary deps in it. is this fine or is a general patch that changes the "include <json.hpp>"-like declarations to have the proper include path preferrable? - its just that it would be a kind of big patch

nekopsykose:
dw, did it for u

PR #3018, opened 2024-09-28
user/curtail: new package

tulilirockz:
honestly kinda funny to see that this tool uses shelling out to actually compress stuff. but well, it is quite useful!

### DIFF
--- a/user/curtail/template.py
+++ b/user/curtail/template.py
@@ -1,0 +1,27 @@
+pkgname = "curtail"
+pkgver = "1.10.0"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "desktop-file-utils",
+    "gettext",
+    "glib-devel",
+    "gtk-update-icon-cache",
+    "meson",
+    "pkgconf",
+]
+depends = [
+    "jpegoptim",
+    "libadwaita",
+    "libwebp-progs",
+    "oxipng",
+    "pngquant",
+    "python-gobject",
+    "python-scour",
+]
+pkgdesc = "GTK image compressor"
+maintainer = "tulilirockz <tulilirockz@outlook.com>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/Huluti/Curtail"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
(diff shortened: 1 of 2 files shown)


user/oxipng/template.py
```diff
@@ -0,0 +1,15 @@
+pkgname = "oxipng"
+pkgver = "9.1.2"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo-auditable"]
```
EarldridgeJazzedPineda, line 5:
```suggestion
hostmakedepends = ["cargo-auditable"]
makedepends = ["rust-std"]
```

user/curtail/template.py
```diff
@@ -0,0 +1,13 @@
+pkgname = "curtail"
+pkgver = "1.10.0"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = ["desktop-file-utils", "gettext", "meson", "pkgconf"]
+makedepends = ["gtk4-devel"]
+depends = ["jpegoptim", "libwebp-progs", "oxipng", "pngquant", "python-scour"]
+pkgdesc = "GTK Image compressor"
```
EarldridgeJazzedPineda, line 8:
```suggestion
pkgdesc = "GTK image compressor"
```

user/curtail/template.py
```diff
@@ -0,0 +1,13 @@
+pkgname = "curtail"
+pkgver = "1.10.0"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = ["desktop-file-utils", "gettext", "meson", "pkgconf"]
+makedepends = ["gtk4-devel"]
+depends = ["jpegoptim", "libwebp-progs", "oxipng", "pngquant", "python-scour"]
```
nekopsykose, line 7: this is missing libadwaita and python-gobject

user/oxipng/template.py
tulilirockz: good catch! didnt see that earlier

user/curtail/template.py
nekopsykose: i don't mean in makedepends, it's a runtime dependency; this is a python application and opens it via gobject. the -devel's don't do anything here :) 

but fixed it for you, thanks, forgot to add oxipng before...

tulilirockz: oh! didnt know that was a runtime one!

PR #3021, opened 2024-09-29
main/bambootracker: sort these lists

z-erica:
small oopsie


q66:
sure, there is a lot of unsorted stuff in the tree as is though

i wonder if we should just batch-sort them...

PR #3023, opened 2024-09-30
move gnome-console into core

Drsheppard01:
Problem: when i mask gnome-apps, which i think so bloated, it's operation deleted my terminal. For fix this problem me need use tty and install gnome-console (or other terminal) again
Solution: put gnome-console into gnome meta-package
If you for example use something other than gnome you can use gnome-terminal, which is also available in the repository.


q66:
what you feel is bloated is inconsequential, console does not belong in core

if you don't want apps but want console, install console separately

PR #3024, opened 2024-09-30
main/lowdown: actually update to 1.1.2

JamiKettunen:
Only version was updated with 00eccb1 and not the checksum resulting on 1.1.1 being rebuilt: https://build.chimera-linux.org/#/builders/6/builds/1468/steps/5/logs/pkg_main_lowdown_1_1_2-r0


nekopsykose:
it's actually the whole url that is wrong

PR #3029, opened 2024-10-01
main/rustic: update to 0.9.0

jcgruenhage:


main/rustic/template.py
```diff
@@ -13,24 +13,11 @@
...
-        "cargo",
-        "update",
-        "--package",
-        "rustix",
-        "--precise",
-        "0.38.35",
-        allow_network=True,
-    )
-
-
```
jcgruenhage, line 33: rustic 0.9.0 depends on rustix 0.38.37, which has this fixed.

PR #3039, opened 2024-10-03
user/hashcat: new package

z-erica:


user/hashcat/template.py
```diff
@@ -0,0 +1,34 @@
+pkgname = "hashcat"
+pkgver = "6.2.6"
+pkgrel = 0
+build_style = "makefile"
+make_env = {
+    "PREFIX": "/usr",
+    # needs LzmaDec.h
+    # "USE_SYSTEM_LZMA": "1",
```
nekopsykose, line 8: this is the 7zip-sdk, not related to xz-devel, so this would never be system, just remove the comment

user/hashcat/patches/no_gcc.patch
```diff
@@ -0,0 +1,37 @@
...
+ CC                      := clang
+@@ -706,8 +706,8 @@
+ ## Compiler paths
+ ##
+ 
+-CC_LINUX                := gcc
+-CXX_LINUX               := g++
++CC_LINUX                := $(CC)
++CXX_LINUX               := $(CXX)
+ CC_WIN                  := x86_64-w64-mingw32-gcc
```
nekopsykose, line 35: there's also a `LFLAGS                  += -s` line to patch out

user/hashcat/template.py
```diff
@@ -0,0 +1,32 @@
...
+]
+pkgdesc = "Password recovery tool"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "MIT"
+url = "https://hashcat.net/hashcat"
+source = f"https://hashcat.net/files/hashcat-{pkgver}.tar.gz"
+sha256 = "b25e1077bcf34908cc8f18c1a69a2ec98b047b2cbcf0f51144dcf3ba1e0b7b2a"
+# check: no obvious test suite
+options = ["!check"]
+
```
nekopsykose, line 29: think you have to make it broken on endian=big or it'll fail

nekopsykose: no idea

PR #3043, opened 2024-10-03
postgres 17

nekopsykose:
[ci skip]

not done (doesn't install); fixes the libpq dep for older 16 (though didn't test that yet), drops tcl (who even uses that), uses meson

the main thing that needs fixing is the existing template requires the per-subdir makefile install to compute a per-contrib install filelist, for accurate per-subpackage splitting without manual specification of every file; the meson build doesn't have individual subtargets afaict so that would become impossible


nekopsykose:
maybe we could do an epic meme of 'dont even split most of the contribs', since by filesize they add like 1mb total (lol), and postgres also does not automatically load any (you have to manually load them with superuser privileges), so there's probably not much downside

PR #3049, opened 2024-10-05
main/fzf: install fzf-tmux

ttyyls:


main/fzf/template.py
```diff
@@ -17,9 +17,20 @@ def post_install(self):
...
         self.install_completion("completion.zsh", "zsh")
+
         for ext in ["bash", "fish", "zsh"]:
             self.install_file(f"key-bindings.{ext}", "usr/share/fzf")
+
+
+@subpackage("fzf-extras")
+def _(self):
+    return [
+        "usr/bin/fzf-tmux",
```
q66, line 34: just install fzf-tmux into the main package and don't install the other script

q66:
actually that's explicitly a bash script so it needs to be split (as `fzf-tmux`)

and the package needs to depend on bash and tmux (maybe `install_if` with these two too)

and it needs a subdesc

main/fzf/template.py
```diff
@@ -17,9 +17,20 @@ def post_install(self):
...
+
         for ext in ["bash", "fish", "zsh"]:
             self.install_file(f"key-bindings.{ext}", "usr/share/fzf")
+
+
+@subpackage("fzf-tmux")
+def _(self):
+    self.subdesc = "tmux plugin"
+    self.depends = [self.parent, "bash"]
+    self.install_if = [self.parent]
```
q66, line 34:
```
self.depends = [self.parent, "bash", "tmux"]
self.install_if = [self.parent, "bash", "tmux"]
```

main/fzf/template.py
```diff
@@ -17,9 +17,20 @@ def post_install(self):
...
         self.install_completion("completion.bash", "bash")
         self.install_completion("completion.zsh", "zsh")
+
         for ext in ["bash", "fish", "zsh"]:
             self.install_file(f"key-bindings.{ext}", "usr/share/fzf")
+
+
+@subpackage("fzf-tmux")
+def _(self):
+    self.subdesc = "tmux plugin"
```
q66, line 32: it's not a plugin

main/fzf/template.py
```diff
@@ -17,9 +17,21 @@ def post_install(self):
...
             self.install_file(f"key-bindings.{ext}", "usr/share/fzf")
+
+
+@subpackage("fzf-tmux")
+def _(self):
+    self.subdesc = "tmux integration script"
+    self.depends = [self.parent, "bash", "tmux"]
+    self.install_if = [self.parent, "bash", "tmux"]
+
+    return ["usr/bin/fzf-tmux", "usr/share/man/man1/fzf-tmux.1"]
```
triallax, line 37: can just use `cmd:fzf-tmux` here (should be good after this, unless q66 has something to say perhaps)

PR #3058, opened 2024-10-07
user/python-ssh2: new package

z-erica:


user/python-ssh2/template.py
```diff
@@ -0,0 +1,30 @@
...
+    "python-jinja2",
+    "python-pytest",
+]
+pkgdesc = "Python bindings for libssh2"
+maintainer = "Erica Z <zerica@callcc.eu>"
+license = "LGPL-2.1-only"
+url = "https://github.com/ParallelSSH/ssh2-python"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "70c6b6efd8ca9f8de9c2d77e7cb1d5859542588347ea426d6822b0ffd9889af3"
+# check: ModuleNotFoundError: No module named 'ssh2.session'
```
q66, line 29: this is probably worth looking into first

nekopsykose: if there's no tests on pypi then that is a reason to not use it

nekopsykose: as for the error, it's probably the usual compiled module cwd thing

PR #3060, opened 2024-10-08
main/msmtp: update to 1.8.27

z-erica:


main/msmtp/patches/tests.patch
```diff
@@ -0,0 +1,140 @@
+these files are missing from the 1.8.27 dist tarball
```
nekopsykose, line 1: is there a non-dist tarball then? i don't think patching back in every test file is a good idea

nekopsykose: looks ok then

PR #3062, opened 2024-10-08
contrib/plasma-desktop: make most -meta subpkgs easily maskable

JamiKettunen:
Now one can just add !plasma-desktop-meta instead of having to do the same for each of these individually when not wanting all of them around (which is arguably often the case due to just how much things get installed by default when adding plasma-desktop).

@nekopsykose Mostly making this so you're aware and may roll this into the v6.2.0 mega-bump, just wanting to separate this change from https://github.com/chimera-linux/cports/pull/2500 since that'll prob still be stuck for a while as a draft while this change is ready.


nekopsykose:
thanks, i'll roll it in (already mostly done with it)

PR #3064, opened 2024-10-09
main/musl: update to 23ab04a8630225371455d5f4538fd078665bb646

nekopsykose:
works ok, but could use a bit more testing

### DIFF
--- a/main/musl/patches/0001-plumb-in-support-for-externally-provided-allocator-l.patch
+++ b/main/musl/patches/0001-plumb-in-support-for-externally-provided-allocator-l.patch
@@ -169,14 +169,6 @@ diff --git a/src/exit/exit.c b/src/exit/exit.c
 index a6869b3..6a0b78f 100644
 --- a/src/exit/exit.c
 +++ b/src/exit/exit.c
-@@ -1,6 +1,7 @@
- #include <stdlib.h>
- #include <stdint.h>
- #include "libc.h"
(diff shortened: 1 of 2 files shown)


nekopsykose:
llvm-nm seems to segfault when building the kernel:

```
(lldb) bt
* thread #1, name = 'llvm-nm', stop reason = signal SIGSEGV: invalid address
  * frame #0: 0x0000783bffae961a ld-musl-x86_64.so.1`exit [inlined] a_crash at atomic_arch.h:108:2
    frame #1: 0x0000783bffae961a ld-musl-x86_64.so.1`exit(code=74) at exit.c:41:19
    frame #2: 0x0000783bf6ddf4de libLLVM.so.18.1`llvm::sys::DefaultOneShotPipeSignalHandler() at Signals.inc:440:3
    frame #3: 0x0000783bf6ddfe5a libLLVM.so.18.1`::SignalHandler() at Signals.inc:0
    frame #4: 0x0000783bffb27e7d ld-musl-x86_64.so.1`__restore_rt
    frame #5: 0x0000000000000007
```

nekopsykose:
i'm gonna guess in the middle of normal exit it also gets killed via sigpipe gmake memes, which also calls exit, so it crashes "as expected".. which means there's no real fix aside from maybe something in kbuild? removing the exit in the sigpipe handler seems hard

PR #3104, opened 2024-10-22
user/sdl_gfx: new package

LtPeriwinkle:
One thing I wasn't sure on - is this the correct naming convention for sdl2-based libraries?


nekopsykose:
it's correct, yes. but soon sdl3 will be out and then it makes no sense again ('sdl2-compat' will provide the sdl2 libs, and 'sdl' or 'sdl3' will be sdl3..), i wish we just named everything with the version in the name to begin with

q66:
apologies for taking a while, i pushed this with an updated naming scheme as a part of the sdl3 switch

PR #3109, opened 2024-10-24
user/pgp2ssh: new package

as400l:
Derive private ed25519 SSH key from private PGP key. GPG itself only supports exporting public SSH keys and other tools don't work for ed25519 keys.

### DIFF
--- a/user/pgp2ssh/template.py
+++ b/user/pgp2ssh/template.py
@@ -1,0 +1,16 @@
+pkgname = "pgp2ssh"
+pkgver = "0_git20240601"
+pkgrel = 0
+_commit = "815c00db15664512705f7cb7b36469b099fdc3eb"
+build_style = "go"
+hostmakedepends = ["go"]
+pkgdesc = "Convert PGP/GPG private keys to SSH private keys"
+maintainer = "as400 <as400@smail.net.pl>"
+license = "MIT"
+url = "https://github.com/pinpox/pgp2ssh"
+source = f"{url}/archive/{_commit}.tar.gz"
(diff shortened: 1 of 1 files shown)


user/pgp2ssh/template.py
```diff
@@ -0,0 +1,16 @@
+pkgname = "pgp2ssh"
+pkgver = "0.1"
+pkgrel = 0
+archs = ["x86_64"]
```
nekopsykose, line 4: why is this limited to one arch

user/pgp2ssh/template.py
```diff
@@ -0,0 +1,16 @@
...
+build_style = "go"
+hostmakedepends = ["go"]
+pkgdesc = "Convert PGP/GPG private keys to SSH private keys"
+maintainer = "as400 <as400@smail.net.pl>"
+license = "MIT"
+url = "https://github.com/pinpox/pgp2ssh"
+source = f"{url}/archive/refs/heads/main.tar.gz"
```
EarldridgeJazzedPineda, line 11: Please don't use "main", because the repository can change at any time. If you're using Git as the source, then use a specific commit.

as400l:
Removed arch limitation and using a specific git commit right now. Thanks.

EarldridgeJazzedPineda:
Please squash the commits

as400l:
Done.

triallax:
you might want to try fixing update-check, see `user/plan9port/update.py` for instance

q66:
git package, also stale

PR #3113, opened 2024-10-25
main/anthy: new package

denisstrizhkin:
Can't figure out the build error for now.


ttyyls:
update check fails to get any versions, the ci does an update-check before trying to build the package.
I suppose the following update.py will do the trick
```python
url = "https://osdn.net/projects/anthy/releases"
pattern = r"anthy-([\d.]+)[a-z]"
```

PR #3116, opened 2024-10-27
main/cmus: update to 2.12.0

z-erica:


main/cmus/template.py
```diff
@@ -22,8 +22,10 @@
 hostmakedepends = ["pkgconf", "bash"]
 makedepends = [
     "elogind-devel",
+    "faad2-devel",
     "ffmpeg-devel",
     "flac-devel",
+    # user: "libao-devel",
```
nekopsykose, line 28: but does it add anything? it already outputs to pulse or jack

PR #3118, opened 2024-10-27
user/thrift: new package (0.21.0)

a16bitsysop:


a16bitsysop:
build is fixed now, missing file was not in source from apache

main/thrift/template.py
```diff
@@ -0,0 +1,69 @@
...
+source = f"https://github.com/apache/thrift/archive/v{pkgver}.tar.gz"
+sha256 = "31e46de96a7b36b8b8a457cecd2ee8266f81a83f8e238a9d324d8c6f42a717bc"
+
+
+@subpackage("thrift-devel")
+def _(self):
+    return self.default_devel()
+
+
+@subpackage("thrift-glib")
```
q66, line 57: all of these should have the `-libs` suffix, and an appropriate `self.subdesc`

and perhaps `thrift-libs` with `self.default_libs()` as well (that does not need special subdesc)

q66: or actually, maybe no `-libs` package for thrift itself, and the compiler should go in the `-devel` package?

q66:
additionally this should not be `main`

main/thrift/template.py
q66: the CI output says:

```
0:06:36.911 => thrift-0.21.0-r0: changed providers:
0:06:36.911 =>   +cmd:thrift=0.21.0-r0
0:06:36.911 =>   +so:libthrift.so.0.21.0=0.21.0
```

that means the main package contains the command thrift, and library libthrift.so.0.21.0

i guess my question is largely if libthrift.so.0.21.0 makes sense on its own

PR #3122, opened 2024-10-29
user/c2rust: new package

z-erica:


user/python-pytest-cov/template.py
```diff
@@ -0,0 +1,29 @@
+pkgname = "python-pytest-cov"
```
nekopsykose, line 1: we removed this before because it's always trivial to patch out (just remove --cov from the pytest args)

nekopsykose: same goes for the coverage above then transitively

PR #3129, opened 2024-10-31
main/*: declare libusb-bootstrap

rbgarga:
It's not a dependency per se but it's installed indirectly which
will break when real libusb is not yet built.


q66:
hm this does not seem right, it should only be declared in stuff that is a bootstrap path

having to declare it in these probably just means your bulk transaction was not sorted correctly

PR #3137, opened 2024-11-02
user/zrepl-dsh2dsh: new package

robertek:
This is an enhanced fork of the original zrepl ZFS snapshot and replication daemon.

The package builds and works as expected on x86_64 and it contains everything that is needed to properly use on Chimera.

### DIFF
--- a/user/zrepl/files/tmpfiles.conf
+++ b/user/zrepl/files/tmpfiles.conf
@@ -1,0 +1,3 @@
+# Directory for control socket
+
+d /run/zrepl 0755 root root -
--- a/user/zrepl/files/zrepl
+++ b/user/zrepl/files/zrepl
@@ -1,0 +1,4 @@
+type = process
+command = /usr/bin/zrepl --config /etc/zrepl/zrepl.yml daemon
+depends-on = local.target
+logfile = /var/log/zrepl.log
--- a/user/zrepl/template.py
+++ b/user/zrepl/template.py
@@ -1,0 +1,27 @@
+pkgname = "zrepl"
+pkgver = "0.9.2"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go"]
+depends = ["zfs"]
+pkgdesc = "ZFS backup and replication tool - dsh2dsh's enhanced fork"
+maintainer = "Robert David <robert.david@posteo.net>"
+license = "MIT"
+url = "https://github.com/dsh2dsh/zrepl"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "1262c854c32e66cf67dd1d8c2ca6e546b6d42100c9bca9857ba37b6a16a5b1d1"
+# check needs to run zfs command
+options = ["!check"]
+
+
+def install(self):
+    self.install_bin("build/zrepl")
+    self.install_files(
+        "internal/config/samples", "usr/share/examples", name="zrepl"
+    )
+    self.install_file(
+        "dist/freebsd/etc/zrepl/zrepl.yml", "usr/share/examples/zrepl"
+    )
+    self.install_service(self.files_path / "zrepl")
+    self.install_tmpfiles(self.files_path / "tmpfiles.conf")
+    self.install_license("LICENSE")


user/zrepl-dsh2dsh/template.py
```diff
@@ -0,0 +1,24 @@
+pkgname = "zrepl-dsh2dsh"
```
q66, line 1: just call it zrepl

user/zrepl-dsh2dsh/template.py
```diff
@@ -0,0 +1,24 @@
...
+pkgver = "0.9.2"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go"]
+depends = ["zfs"]
+provides = [self.with_pkgver("zrepl")]
```
q66, line 7: drop this

user/zrepl-dsh2dsh/template.py
```diff
@@ -0,0 +1,24 @@
...
+pkgdesc = "One-stop ZFS backup & replication solution - dsh2dsh enhanced fork"
+maintainer = "Robert David <robert.david@posteo.net>"
+license = "MIT"
+url = "https://github.com/dsh2dsh/zrepl"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "1262c854c32e66cf67dd1d8c2ca6e546b6d42100c9bca9857ba37b6a16a5b1d1"
+
```
q66, line 14: this extra newline does not belong here

user/zrepl-dsh2dsh/template.py
```diff
@@ -0,0 +1,24 @@
...
+# check needs to run zfs command
+options = ["!check"]
+
+def install(self):
+    self.install_bin("build/zrepl")
+    self.install_file("dist/freebsd/etc/zrepl/zrepl.yml", "etc/zrepl", name="zrepl.yml.sample")
```
q66, line 20: examples don't go in `/etc`

user/zrepl-dsh2dsh/files/tmpfiles.conf
```diff
@@ -0,0 +1,3 @@
+# Directory for control socket
+
+d /run/zrepl 0755 root root -
```
q66, line 3: it can't create it by itself?

user/zrepl-dsh2dsh/template.py
robertek: Corresponds with the comment above. I may actually provide template for both zrepl and this new fork. Than this would be needed.

user/zrepl-dsh2dsh/template.py
q66: unless there is something in original zrepl that makes it worth preferring over the fork and thus warrants having a separate package for it that would conflict, this should be just zrepl

user/zrepl-dsh2dsh/files/tmpfiles.conf
robertek: No. It fails without that. I'll file that as a bug. The default is /var/run/zrepl/control, but it may be changed in the config.

user/zrepl-dsh2dsh/template.py
robertek: I may also change it to be zrepl.yml as an default config. But I feel it is more like a sample than default config to be used as is.

user/zrepl-dsh2dsh/template.py
```diff
@@ -0,0 +1,24 @@
+pkgname = "zrepl-dsh2dsh"
+pkgver = "0.9.2"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go"]
+depends = ["zfs"]
+provides = [self.with_pkgver("zrepl")]
+pkgdesc = "One-stop ZFS backup & replication solution - dsh2dsh enhanced fork"
```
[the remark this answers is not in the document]
robertek: The fork info is probably needed here. This is not the original zrepl. This message is copied directly from FreeBSD port.

robertek:
I have fixed several of the review suggestions.

q66:
i guess this is fine enough for now, i'll just merge it

PR #3144, opened 2024-11-05
user/terraform: new package

dmdutra:


q66:
what's the point of having both terraform and opentofu

dmdutra:
> what's the point of having both terraform and opentofu

Are different projects maintain by different organizations.  Many people uses Terraform instead of OpenTofu.

q66:
that doesn't answer anything

one is literally a fork of the other and one is non-free, so unless there is something meaningful the non-free offers over the free one, i don't see the point in including it

dmdutra:
> that doesn't answer anything
> 
> one is literally a fork of the other and one is non-free, so unless there is something meaningful the non-free offers over the free one, i don't see the point in including it

It's ok for me. I added Terraform just as an alternative to Opentofu. Terraform has more usage than opentofu. But, I will close this PR :)

PR #3148, opened 2024-11-05
main/nyagetty: fix syntax error in agetty-service.sh

mathijs28:
When starting a serial getty by specifying it in `/etc/default/agetty` or running `dinitctl start agetty-service@ttyS0`, the service would crash because of a syntax error in the helper script `/usr/lib/agetty-service`.


mathijs28:
I am not sure if I had to increment the `pkgrel` for such a small bugfix, so please let me know if I shouldn't.

q66:
it changes the installed script which was previously broken so pkgrel is correct to increment

it probably accidentally made it in during the last round of refactor...

PR #3156, opened 2024-11-07
user/j4-dmenu-desktop: new package

whypet:
Also I tried to cross compile for other architectures on an amd64 host but it fails because pkg-config couldn't find spdlog, however it was able to find fmt, even though the pc files for both libraries were present under `bldroot/usr/lib/pkgconfig`. It works when not cross compiling.

I tried to run `PKG_CONFIG_PATH=bldroot/usr/lib/pkgconfig bldroot/builddir/.cbuild-j4-dmenu-desktop/wrappers/aarch64-chimera-linux-musl-pkg-config spdlog` manually (and with fmt too), and got a successful exit but only when specifying the PKG_CONFIG_PATH variable.

```
C++ compiler for the host machine: aarch64-chimera-linux-musl-clang++ (clang 18.1.8 "clang version 18.1.8")
C++ linker for the host machine: aarch64-chimera-linux-musl-clang++ ld.lld 18.1.8
C++ compiler for the build machine: clang++ (clang 18.1.8 "clang version 18.1.8")
C++ linker for the build machine: clang++ ld.lld 18.1.8
Build machine cpu family: x86_64
Build machine cpu: x86_64
Host machine cpu family: aarch64
Host machine cpu: aarch64
Target machine cpu family: aarch64
Target machine cpu: aarch64
Check usable header "sys/inotify.h" : YES 
Found pkg-config: YES (/builddir/.cbuild-j4-dmenu-desktop/wrappers/aarch64-chimera-linux-musl-pkg-config) 2.3.0
Run-time dependency fmt found: YES 11.0.2
Found CMake: NO
Run-time dependency spdlog found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency spdlog

src/meson.build:63:9: ERROR: Automatic wrap-based subproject downloading is disabled
```


user/j4-dmenu-desktop/template.py
```diff
@@ -0,0 +1,20 @@
...
+license = "GPL-3.0-or-later"
+url = "https://github.com/enkore/j4-dmenu-desktop"
+source = f"{url}/archive/r{pkgver}.tar.gz"
+sha256 = "0e6cf25663cc220e3e3e2bc013fe957c3e4a44f900b5ee6a7609cd501021652d"
+hardening = ["vis", "!cfi", "!sst"]
+
+
+def init_configure(self):
+    if self.run_check:
+        self.configure_args.append("-Denable-tests=true")
```
nekopsykose, line 20: just put -Denable-tests=true at the top, why put it here

user/j4-dmenu-desktop/template.py
```diff
@@ -0,0 +1,20 @@
...
+hostmakedepends = ["meson", "pkgconf", "catch2-devel", "spdlog-devel"]
+makedepends = ["spdlog", "fmt-devel"]
+checkdepends = ["catch2", "python-pytest"]
+pkgdesc = "Fast desktop menu using dmenu"
+maintainer = "wipet <47226783+whypet@users.noreply.github.com>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/enkore/j4-dmenu-desktop"
+source = f"{url}/archive/r{pkgver}.tar.gz"
+sha256 = "0e6cf25663cc220e3e3e2bc013fe957c3e4a44f900b5ee6a7609cd501021652d"
+hardening = ["vis", "!cfi", "!sst"]
```
nekopsykose, line 15:
```suggestion
```

those 2 are already disabled by default so this doesn't do anything

user/j4-dmenu-desktop/template.py
```diff
@@ -0,0 +1,20 @@
+pkgname = "j4-dmenu-desktop"
+pkgver = "3.1"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Denable-tests=false"]
+hostmakedepends = ["meson", "pkgconf", "catch2-devel", "spdlog-devel"]
+makedepends = ["spdlog", "fmt-devel"]
+checkdepends = ["catch2", "python-pytest"]
```
nekopsykose, line 8:
```suggestion
hostmakedepends = ["meson", "pkgconf"]
makedepends = ["spdlog-devel", "fmt-devel", "catch2-devel"]
checkdepends = ["python-pytest"]
```

nekopsykose:
with all these changes it builds fine in cross and non-cross and the tests work normally

PR #3160, opened 2024-11-09
user/k6: new package

dtrsh:


user/k6/template.py
```diff
@@ -0,0 +1,17 @@
+pkgname = "k6"
+pkgver = "0.54.0"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go"]
+pkgdesc = "Modern load testing tool, using Go and Javascript"
```
q66, line 6: Load testing tool

user/k6/template.py
```diff
@@ -0,0 +1,17 @@
...
+license = "AGPL-3.0-only"
+url = "https://github.com/grafana/k6"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "706089f6dd65b8e50cee71ac30d32b44be8ddb6c65acf8ff54b6a0729027148e"
+# k6 tests requires network connection
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_license("LICENSE.md")
```
triallax, line 17: agpl doesn't need install

dtrsh: Seriously? I got an error during the local build

```
ERROR: license installation necessary but no license installed
```

q66: agpl does need license installation, the lint is correct as is the install_license

PR #3163, opened 2024-11-10
user/gmid: new package

ttyyls:


user/gmid/files/gmid
```diff
@@ -0,0 +1,10 @@
+# gmid system service
```
q66, line 1: drop comment

user/gmid/files/gmid
```diff
@@ -0,0 +1,10 @@
...
+
+type = process
+command = /usr/bin/gmid -f
+working-dir = /var/lib/gmid
+run-as = _gmid
+logfile = /var/log/gmid.log
+depends-on = local.target
+depends-on = network.target
+smooth-recovery = true
```
q66, line 10: nothing is really going to depend on this so it's unnecessary

user/gmid/template.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+    self.install_sysusers(self.files_path / "sysusers.conf")
+    self.install_tmpfiles(self.files_path / "tmpfiles.conf")
+    self.install_service(self.files_path / "gmid")
+
+
+@subpackage("gmid-utils")
```
q66, line 34: gmid-progs and drop subdesc

user/gmid/files/tmpfiles.conf
```diff
@@ -0,0 +1,4 @@
+# Create gmid state directories
+
+d /var/lib/gmid 0755 _gmid _gmid -
+d /var/gemini 0755 _gmid _gmid -
```
q66, line 4: surely this should not be owned by gmid (`/var/www` is not either)

should it be created at all? literally nothing seems to reference it other than some examples

ttyyls: `/var/lib/gmid` for certs
`/var/gemini` for .gem files that get served
unlike some other gemini servers, gmid does not handle the creation/rotation of certs so owning the directories isn't necessary as no writes ever happen.

maybe both should be owned by root.

the creation of both could also be skipped, im not sure

ttyyls: i guess skipping both is the way to go since there'd still be manual setup to engage in anyway by the user, the certs will more realistically go somewhere in the /etc and the content directory could be left up to the user to setup

PR #3172, opened 2024-11-13
main/sddm: fix broken 10-breeze-theme.conf location

JamiKettunen:
`/usr/lib/sddm/sddm.conf.d/10-breeze-theme.conf/10-breeze-theme.conf` used to exist which was obviously broken, though I still wasn't able to test this successfully on an already booted install where I saw the fallback theme already once despite e.g. `dinitctl stop sddm && rm -r /var/lib/sddm/{.{cache,config,local},state.conf} && dinitctl start sddm` so this may need further confirmation it actually fixes the issue..

### DIFF
--- a/main/sddm/template.py
+++ b/main/sddm/template.py
@@ -1,6 +1,6 @@
 pkgname = "sddm"
 pkgver = "0.21.0"
-pkgrel = 2
+pkgrel = 3
 build_style = "cmake"
 configure_args = [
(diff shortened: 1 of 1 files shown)


nekopsykose:
hah, yeah that was obviously the wrong place by me

nekopsykose:
even if it doesn't fix it the change is correct

PR #3173, opened 2024-11-14
main/luanti: rename from minetest, update to 5.10.0

ttyyls:


main/luanti/template.py
```diff
@@ -59,30 +60,31 @@
...
-    return ["usr/share/minetest/builtin"]
+    return ["usr/share/luanti/builtin"]
 
 
-@subpackage("minetest-server")
+@subpackage("luanti-server")
 def _(self):
     self.subdesc = "server"
-    self.depends = [self.with_pkgver("minetest-common")]
+    self.depends = [self.with_pkgver("luanti-common")]
```
q66, line 88: server is missing a provides...

ttyyls: the install process adds symlinks mintest->luanti and minetestserver->luantiserver. those get picked up by the pos-install hooks, adding them manually seemed redundant

q66: ? that doesn't make luanti-server package provide the minetest-server name (just like every other package in this template does with your changes)

ttyyls: true

PR #3180, opened 2024-11-15
user/tiny: new package

wezm:


user/tiny/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "tiny"
+pkgver = "0.12.0"
+pkgrel = 0
+build_wrksrc = "crates/tiny"
```
nekopsykose, line 4: it ends up rebuilding it in install(), should make it manually installed below instead of automatic

using `--bin tiny` instead of wrksrc probably fixes that (do that anyway), if it doesn't then manually install it :)

wezm: Thanks for the review. Updated with a second take.

PR #3188, opened 2024-11-17
Gnome remote desktop

denisstrizhkin:


main/ffnvcodec-headers/template.py
```diff
@@ -0,0 +1,29 @@
...
+pkgrel = 0
+build_style = "makefile"
+hostmakedepends = ["pkgconf"]
+pkgdesc = (
+    "FFmpeg version of headers required to interface with Nvidias codec APIs"
+)
+maintainer = "Denis Strizhkin <strdenis02@gmail.com>"
+license = "MIT"
+url = "https://github.com/FFmpeg/nv-codec-headers"
+source = f"https://github.com/FFmpeg/nv-codec-headers/archive/refs/tags/n{pkgver}.tar.gz"
```
nekopsykose, line 12:
```suggestion
source = f"{url}/archive/refs/tags/n{pkgver}.tar.gz"
```

main/ffnvcodec-headers/template.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def pre_build(self):
+    with (
+        open(self.cwd / "include/ffnvcodec/nvEncodeAPI.h") as header,
+        open(self.cwd / "LICENSE", "w") as license,
+    ):
+        lines = header.readlines()[3:25]
+        lines = [line[2:].strip() + "\n" for line in lines]
+        license.writelines(lines)
```
nekopsykose, line 25:
```suggestion
# no tests, no licence file
options = ["!check", "!distlicense"]
```

main/ffnvcodec-headers/template.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def post_install(self):
+    self.install_license("LICENSE")
```
nekopsykose, line 29:
```suggestion
```

main/gnome-remote-desktop/template.py
```diff
@@ -0,0 +1,35 @@
...
+]
+depends = []
+checkdepends = ["dbus", "pipewire", "mutter", "libgudev-devel"]
+pkgdesc = "GNOME Remote Desktop server"
+maintainer = "Denis Strizhkin <strdenis02@gmail.com>"
+license = "GPL-2.0-or-later"
+url = "https://gitlab.gnome.org/GNOME/gnome-remote-desktop"
+source = f"{url}/-/archive/{pkgver}/{pkgname}-{pkgver}.tar.gz"
+sha256 = "bfdfc646947aaac02e7c3230386a34872f501d6ddd66484154536949f51bf1f1"
+# needs dbus
```
nekopsykose, line 34: and it has it via using dbus-run-session itself, so this comment seems wrong

main/gnome-remote-desktop/template.py
```diff
@@ -0,0 +1,35 @@
+pkgname = "gnome-remote-desktop"
+pkgver = "47.1"
+pkgrel = 0
+build_style = "meson"
+configure_args = ["-Dvnc=false", "-Dsystemd=false", "-Dtests=false"]
```
nekopsykose, line 5: libvncserver exists so this should work i think

main/gnome-remote-desktop/template.py
denisstrizhkin: checks don't run https://github.com/chimera-linux/cports/actions/runs/11884781818/job/33113464596?pr=3188#step:8:546

main/gnome-remote-desktop/template.py
denisstrizhkin: I am trying to make rdp work properly for now.

denisstrizhkin: Added vncsever dep

main/ffnvcodec-headers/template.py
```diff
@@ -0,0 +1,15 @@
+pkgname = "ffnvcodec-headers"
```
q66, line 1: nv-codec-headers

main/ffnvcodec-headers/template.py
```diff
@@ -0,0 +1,15 @@
...
+pkgver = "12.2.72.0"
+pkgrel = 0
+build_style = "makefile"
+hostmakedepends = ["pkgconf"]
+pkgdesc = (
```
q66, line 6: too long, something like "Headers for Nvidia codec APIs" says basically the same thing

main/gnome-remote-desktop/template.py
nekopsykose: i have no idea what that links to but 'no module dbus' is python-dbus

nekopsykose: looking at the errors the correct comment would be `# needs real /sys /proc for some udev things to work`

denisstrizhkin: Yes, seems there is no way to run the tests under bwrap even after I fixed python-dbus and stuff.

denisstrizhkin:
The biggest problem is that grdctl segfaults and I wasn't yet able to figure out why.

denisstrizhkin:
Also the System menu does not show up for some reason but if I run the daemon directly from cli screensharing does work.

nekopsykose:
> The biggest problem is that grdctl segfaults and I wasn't yet able to figure out why.

post backtrace

> Also the System menu does not show up for some reason but if I run the daemon directly from cli screensharing does work.

i think triallax ran into the same thing, and it's the same as the kde version- it only can start the daemon from settings via dbus call to systemd service start api, otherwise it can't and one has to do it manually or maybe have a dinit service etc

denisstrizhkin:
```
(gdb) run
The program being debugged has been started already.
Start it from the beginning? (y or n) y
Starting program: /usr/bin/grdctl status

Program received signal SIGSEGV, Segmentation fault.
0x000055555555c3cb in main (argc=2, argv=0x7fffffffe5b8) at ./build/../src/grd-ctl.c:1016
1016	  if (geteuid () == pw->pw_uid)
```

q66:
that's because the pwd pointer is null because the expected user does not exist

denisstrizhkin:
I see, it seems it wants a system user
https://gitlab.gnome.org/GNOME/gnome-remote-desktop/-/blob/master/meson_options.txt#L51

denisstrizhkin:
Should I create a user like `_gnome-remote-desktop` ?

denisstrizhkin:
Then only problem remaining then is how to start the services and why it do not show up in System Gnome menu

PR #3193, opened 2024-11-19
user/zellij: update to 0.41.2

tranzystorekk:
cc @denisstrizhkin

Re: dropped manpage, upstream has declared it outdated and doesn't want to maintain it.


denisstrizhkin:
lgtm, do we need `--release` for check though?

PR #3197, opened 2024-11-21
user/maven: new package

Natthias:
Couldn't get the update string to match other versions one directory down (for when a new major version is release). Everything else should be fine

### DIFF
--- a/user/maven/template.py
+++ b/user/maven/template.py
@@ -1,0 +1,71 @@
+pkgname = "maven"
+pkgver = "3.9.9"
+pkgrel = 0
+hostmakedepends = ["openjdk21"]
+depends = ["virtual:java-jre!openjdk21-jre"]
+pkgdesc = "Software project management and comprehension tool"
+maintainer = "natthias <natthias@proton.me>"
+license = "Apache-2.0"
+url = "https://maven.apache.org"
+source = [
+    f"https://dlcdn.apache.org/maven/maven-3/{pkgver}/source/apache-maven-{pkgver}-src.tar.gz",
+    f"https://dlcdn.apache.org/maven/maven-3/{pkgver}/binaries/apache-maven-{pkgver}-bin.tar.gz",
+]
+source_paths = [
+    ".",
+    "bootstrap",
+]
+sha256 = [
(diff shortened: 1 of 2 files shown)


user/maven/template.py
```diff
@@ -0,0 +1,72 @@
...
+        "test",
+        "-Drat.skip=true",
+        "-Dmaven.repo.local=/cbuild_cache/repository",
+    )
+
+
+def install(self):
+    self.install_files("apache-maven/out", "usr/lib/", name="maven")
+
+    self.mkdir(self.destdir / "usr/bin")
```
nekopsykose, line 67: install_dir

user/maven/template.py
```diff
@@ -0,0 +1,72 @@
...
+    "8a24c448d4ac397e6b0c019a4d7250068c02d1cdb553299e6bb71c3ccca78b2c",
+    "7a9cdf674fc1703d6382f5f330b3d110ea1b512b51f1652846d9e4e8a588d766",
+]
+
+
+def prepare(self):
+    self.do(
+        "./bootstrap/bin/mvn",
+        "org.apache.maven.plugins:maven-dependency-plugin:2.8:go-offline",
+        "-Dmaven.repo.local=/cbuild_cache/repository",
```
q66, line 29: "repository" is not a good name for the path

user/maven/template.py
```diff
@@ -0,0 +1,72 @@
+pkgname = "maven"
+pkgver = "3.9.9"
+pkgrel = 0
+hostmakedepends = ["openjdk21"]
+# FIXME: depends on unversioned providers are somewhat broken in apk
+# depends = ["virtual:java-jre!openjdk21-jre"]
```
q66, line 6: they should work fine now, as long as one of the providers has a correctly set priority

user/maven/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://dlcdn.apache.org/maven/maven-3"
+pattern = r"(\d\.\d\.\d)"
```
q66, line 2: please use a more concrete pattern, something like `r>([0-9.]+)/<` should do

PR #3198, opened 2024-11-21
cbuild: fix prepare-upgrade only updating one sha256

Gnarwhal:
If multiple sha256 strings in a template start as the same initial value (i.e. all the empty string) before prepare-upgrade is run, prepare-upgrade will only update the first source's checksum. 

Example of old behaviour:
```
sha256 = [
    "",
    "",
]
```
`./cbuild prepare-upgrade user/foo`
```
sha256 = [
    "2bdb2943db86338fdfcb452fbb6d84e7a4ad05712492dd564d4e5c84a4c8d750",
    "",
]
```
`./cbuild prepare-upgrade user/foo`
```
sha256 = [
    "2bdb2943db86338fdfcb452fbb6d84e7a4ad05712492dd564d4e5c84a4c8d750",
    "8fa3b0b66914ae2dd4ddef9a954f614c5b3eb6ac9d80ee61ae2d08e3178507ec",
]
```

### DIFF
--- a/src/cbuild/hooks/fetch/000_sources.py
+++ b/src/cbuild/hooks/fetch/000_sources.py
@@ -30,16 +30,15 @@ def make_link(dfile, cksum):
             linkpath.hardlink_to(dfile)
 
 
-def verify_cksum(dfile, cksum, pkg):
+def verify_cksum(dfile, idx, pkg):
+    cksum = pkg.sha256[idx]
     pkg.log(f"verifying sha256sums for source '{dfile.name}'... ", "")
     filesum = get_cksum(dfile, pkg)
     if cksum != filesum:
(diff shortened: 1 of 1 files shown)


q66:
i'm not quite sure if this is enough, i'll need some time to think about it

q66:
so your general idea was right but had some flaws:

1) `dfiles` does not necessarily match sources (it only contains sources that are to be downloaded from scratch) which means attempting to get the index by enumerating that array won't necessarily match the sha256 array being updated
2) you didn't account for `link_cksum` update

i implemented a better version that takes the index when populating `dfiles` (where it still matches sources) and assigns the index to each dfile, then pass that around

PR #3203, opened 2024-11-23
user/incus: Add gtar as dependency

cmspam:
So happy to see incus on Chimera!

When I gave it a try, I noticed that after a base system install, gtar was missing, and was necessary for incus to launch a container. It wasn't installed as a dependency, hence this pull request.

### DIFF
--- a/user/incus/template.py
+++ b/user/incus/template.py
@@ -27,6 +27,7 @@ depends = [
     "acl-progs",
     "attr-progs",
     "dnsmasq",
+    "gtar",
     "iptables",
     "libvirt",
     "lxc",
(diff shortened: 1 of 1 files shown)


nekopsykose:
you have to increment the pkgrel

cmspam:
> you have to increment the pkgrel

You're right. Sorry for missing that. Done.

nekopsykose:
and you have to squash that commit into the first one, so it's just one commit

cmspam:
I had a feeling that was coming. I will make a new PR instead if that's ok.

nekopsykose:
you should really learn how to use rebase or just `git commit --amend` and save yourself future headache :)

PR #3204, opened 2024-11-23
user/incus: Add gtar as dependency

cmspam:
gtar is necessary for incus to launch containers but not included as a dependency, and not installed in the base install.


nekopsykose:
that failure is not related to the changes, don't worry about it

cmspam:
Haha, thanks. I was worried for a moment that incrementing the pkgrel wasn't as simple as it appeared.

nekopsykose:
thanks!

PR #3217, opened 2024-11-27
KDE Plasma 6.2.4 + Qalculate! 5.4.0

JamiKettunen:
Screenshot_20241128_010138

### DIFF
--- a/main/bluedevil/template.py
+++ b/main/bluedevil/template.py
@@ -1,5 +1,5 @@
 pkgname = "bluedevil"
-pkgver = "6.2.3"
+pkgver = "6.2.4"
 pkgrel = 0
 build_style = "cmake"
 hostmakedepends = [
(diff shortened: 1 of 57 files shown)


sidkshatriya:
Looks like you're running Linux on Snapdragon. I'm curious -- is this screenshot on bare metal Linux or in WSL ? If bare metal, are you able to run a VM ? A VM within a VM (i.e. nested VM) ? 

Thanks so much  !

JamiKettunen:
Bare-metal. Doing any KVM stuff requires booting Linux in EL2 (or making QEMU somehow work with [the Gunyah hypervisor any non-Windows OS seems currently to run under by default](https://patchwork.kernel.org/project/linux-arm-kernel/cover/20240222-gunyah-v17-0-1e9da6763d38@quicinc.com/)) which currently breaks critical stuff like charging/battery info because kernel code paths assume the hypervisor paths of certain stuff and it initializing some hw so I've not touched it, may try later but I need to even figure out what I want to do with the live image multi-DTB situation. All my work on the laptop so far is on https://github.com/JamiKettunen/cports/tree/x1e, either way best continue discussing this on the irc/matrix instead of unrelated PR comments

PR #3219, opened 2024-11-28
main/haproxy: update to 3.1.0

ttyyls:


q66:
seems to me like the readiness notification needs reworking rather than dropping?

ttyyls:
I reworked the patch and added the required pipe-var:HAPROXY_READY_FD in the service.

PR #3224, opened 2024-11-29
main/wl-kbptr: add missing dep

ttyyls:


q66:
also while at it, make it not install the hyprland helper, that needs python anyway

PR #3231, opened 2024-11-30
 main/openimageio: update to 3.0.0.3 

z-erica:


q66:
it will do it automatically, you just forgor to bump rel

PR #3233, opened 2024-12-02
user/perl-file-mimeinfo: new package

ttyyls:
## Checklist

### DIFF
--- a/user/perl-file-basedir/template.py
+++ b/user/perl-file-basedir/template.py
@@ -1,0 +1,16 @@
+pkgname = "perl-file-basedir"
+pkgver = "0.09"
+pkgrel = 0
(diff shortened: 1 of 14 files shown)


user/perl-file-basedir/template.py
```diff
@@ -0,0 +1,16 @@
+pkgname = "perl-file-basedir"
+pkgver = "0.09"
+pkgrel = 0
+build_style = "perl_module"
+hostmakedepends = ["perl"]
+depends = [
+    "perl",
```
q66, line 7: don't need perl dep if it already depends on other perl things

ttyyls: i was being pedantic but i gues i could just drop them

PR #3241, opened 2024-12-04
user/yaru-theme: new package

matrixvk:


q66:
is there a particular value in splitting it into all these subpackages

matrixvk:
> is there a particular value in splitting it into all these subpackages

you can check https://packages.fedoraproject.org/pkgs/yaru-theme/, fedora guys splitted them that way.

q66:
that doesn't answer my question...

matrixvk:
> that doesn't answer my question...

Sorry for interpreting that in a wrong way since I'm not a native speaker of English. Maybe you mean does splitting it into various subpackages make much sense? It depends. Some users would like the whole package, while the others may just prefer a few parts of the theme. I tend to take a look at other distribution's build template first, so I used fedora's splitting method.

Or maybe you mean something else? If that's true, would you explain your question   in another way? I'm all ears to your suggestions.

user/yaru-theme/template.py
```diff
@@ -0,0 +1,69 @@
...
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "meson",
+    "pkgconf",
+    "bash",
+    "sassc",
+]
+makedepends = ["gtk+3-devel"]
+pkgdesc = "Default theme for ubuntu"
```
EarldridgeJazzedPineda, line 12:
```suggestion
pkgdesc = "Default theme for Ubuntu"
```

user/yaru-theme/template.py
```diff
@@ -0,0 +1,69 @@
...
+        "usr/share/gnome-shell",
+        "usr/share/themes/*/index.theme",
+        "usr/share/themes/*/gnome-shell",
+    ]
+
+
+@subpackage("yaru-theme-gtk2")
+def _(self):
+    self.subdesc = "GTK2"
+    return ["usr/share/themes/*/gtk-2.0"]
```
EarldridgeJazzedPineda, line 33: Chimera Linux doesn't have a GTK 2 package, so why even bother with this?

matrixvk: emm you're right

q66:
> > that doesn't answer my question...
> 
> Sorry for interpreting that in a wrong way since I'm not a native speaker of English. Maybe you mean does splitting it into various subpackages make much sense? It depends. Some users would like the whole package, while the others may just prefer a few parts of the theme. I tend to take a look at other distribution's build template first, so I used fedora's splitting method.

the only reason why splitting could make sense would be if the whole thing takes up a significant amount of space and splitting it meaningfully saves anything; if it does not, then don't split it (feel free to list installed sizes for each subpackage)

if you do split it, the main package should remain empty (and this should be enforced with `options = ["empty"]`) and the subpackages all need to have `self.install_if = [self.parent, <whatever is the condition here for the specific subpackage to be useful>]` to facilitate automatic installation

regardless of whether it's split or not, the gtk2 parts and possibly other irrelevant things should be disabled or removed

PR #3245, opened 2024-12-05
user/ignis: new package

supermikea:


user/ignis/template.py
```diff
@@ -0,0 +1,35 @@
...
+]
+depends = [
+    "gtk4-layer-shell-devel",
+    "libpulse",
+    "python-loguru",
+    "python-gobject",
+    "python-click",
+    "python-requests",
+]
+pkgdesc = (
```
q66, line 24: "GTK4 widget system for building desktop shells"

user/python-loguru/template.py
```diff
@@ -0,0 +1,28 @@
...
+    "python-wheel",
+]
+depends = ["python"]
+checkdepends = [
+    "python-pytest",
+    "python-freezegun",
+    "python-colorama",
+    "python-mypy",
+]
+pkgdesc = "Python logging made (stupidly) simple"
```
q66, line 19: Logging library for Python

q66:
i added mypy separately, you can drop whatever is related to that

user/ignis/template.py
```diff
@@ -0,0 +1,35 @@
+pkgname = "ignis"
+pkgver = "0.3"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "meson",
```
q66, line 6: sort these lists by name (ditto for others)

user/ignis/template.py
q66, line 24: reformat

user/ignis/template.py
```diff
@@ -0,0 +1,35 @@
...
+    "pkgconf",
+    "git",
+]
+makedepends = [
+    "gtk4-devel",
+    "gtk4-layer-shell-devel",
+    "libpulse-devel",
+]
+depends = [
+    "gtk4-layer-shell-devel",
```
q66, line 17: this looks wrong?

user/ignis/template.py
```diff
@@ -0,0 +1,35 @@
...
+    "libpulse",
```
q66, line 18: this is autotraced

q66: also, the build results in a `libgvc.so` being dumped into `/usr/lib` which seems wrong because that's a private library

user/ignis/template.py
```diff
@@ -0,0 +1,35 @@
...
+    "python-loguru",
```
q66, line 19: this list is missing python-cairo

user/ignis/template.py
supermikea: with autotraced you mean that it is already pulled in by other dependencies?

q66: by autotraced i mean that cbuild checks executables/shared libs for their runtime dependencies and knows about the libpulse dependency already

supermikea:
This should be it (assuming "autotraced" means that it's automatically pulled in)

q66:
this is not fully fixed, there is still `libgvc.so` in `/usr/lib` even though there shouldn't - this is an upstream issue, and changing the layer shell dependency to `gtk4-layer-shell` is the correct thing to do, but it also breaks it because the code tries to open the `.so` symlink that is in `-devel`, this is also an upstream issue

you'll have to patch around these somehow, or get upstream to fix it

q66:
stale, feel free to reopen if you want to actually fix it

PR #3249, opened 2024-12-06
main/sworkstyle: add user service

ttyyls:


main/sworkstyle/files/sworkstyle.user
```diff
@@ -0,0 +1,7 @@
+# sworkstyle user service
```
q66, line 1: we don't add these comments anymore

ttyyls: removed it

PR #3252, opened 2024-12-06
user/mise: new package

ttyyls:


user/mise/template.py
```diff
@@ -0,0 +1,27 @@
...
+    "pkgconf",
+]
+makedepends = [
+    "libgit2-devel",
+    "openssl-devel",
+    "rust-std",
+    "zstd-devel",
+]
+checkdepends = ["bash"]
+pkgdesc = "Dev tools, env vars, task runner"
```
subnut, line 16: Wouldn't "Development environment setup tool" sound better?

https://mise.jdx.dev/about.html

user/mise/template.py
```diff
@@ -0,0 +1,27 @@
...
+maintainer = "ttyyls <contact@behri.org>"
+license = "MIT"
+url = "https://mise.jdx.dev"
+source = f"https://github.com/jdx/mise/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "0763599841970d9a60630fbd97301d7a06d9ce5c61d9da657fda432f2204cfae"
+
```
subnut, line 22: Seems to build fine and pass tests with 
```python
hardening = ["vis", "cfi"]
```

ttyyls: those are ignored in the cargo build style

user/mise/template.py
ttyyls: sounds good

q66:
this needs to be changed to use system lua (`--no-default-features --features=native-tls` at least plus a makedep and maybe other stuff idk)

PR #3261, opened 2024-12-09
add monaspace font

### LINKED ISSUE #3260: add non-nerd github monaspace fonts
because nerd-fonts ones are too large and mostly duplicated things.
using nerd symbol-only fonts with fontconfig is the best solution.

BarryLhm:
Closes #3260

### DIFF
--- a/main/fonts-monaspace-otf/template.py
+++ b/main/fonts-monaspace-otf/template.py
@@ -1,0 +1,20 @@
+pkgname = "fonts-monaspace-otf"
+pkgver = "1.101"
+pkgrel = 0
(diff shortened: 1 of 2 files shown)


BarryLhm:
I will make changes to the user repo

JamiKettunen:
It's all within this same cports git repository, there was no need to create a new PR (#3262) which still is broken and includes the commit from here too. Regardless forget about it and continue with the new PR now. You really should check out https://github.com/chimera-linux/cports/blob/master/USING_GIT.md taking notes for the next time and while you fix that one.

PR #3273, opened 2024-12-12
main/haproxy: update to 3.1.2

ttyyls:
supersedes #3219


q66:
the patch looks ok but fix the indentation, it's all wrong

ttyyls:
fixed the indentation

PR #3279, opened 2024-12-13
cbuild: format

z-erica:


q66:
this is fixed in 840ed850d657389dd64f95a4ca1f9bf13ada032c

formatting it like this would just make black reformat it differently again because apparently the "drop-in alternative" is not quite drop-in

PR #3283, opened 2024-12-14
user/mise: update to 2024.12.9

ttyyls:


triallax:
.12 is out

q66:
they tag a release every day, i highly suggest not updating this all the time

ttyyls:
we should probably have an official policy for these. for the sake of reducing noise we could limit daily updated packages to a sync once per week with an exception for security fixes?

q66:
no need for that, just update it whenever you feel like it but keep in mind that these releases are created by a bot so they are not "real" releases

PR #3286, opened 2024-12-15
main/qutebrowser: update to 3.4.0

ttyyls:
- **main/qutebrowser: update to 3.4.0**
- **main/pdfjs: switch to legacy distribution**

### DIFF
--- a/main/pdfjs/template.py
+++ b/main/pdfjs/template.py
@@ -1,12 +1,12 @@
 pkgname = "pdfjs"
 pkgver = "4.9.155"
-pkgrel = 0
+pkgrel = 1
 pkgdesc = "Platform for parsing and rendering PDFs"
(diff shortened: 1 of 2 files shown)


q66:
you should explain what is the deal with the legacy distribution

ttyyls:
the latest release notes recommend the switch of pdf.js to th legacy distrib.

see: https://github.com/qutebrowser/qutebrowser/discussions/8423

which I quote
> their normal release only supports the latest Chromium version and
might break in qutebrowser on updates

the only consumer of this package is qutebrowser, and seeing how both builds get updated on every release and that the only difference is api compatibility with older chromiums, I see no reason to deviate fron the qutebrowser's upstream packaging reconmendations.

Although for the time being both distributions work

PR #3288, opened 2024-12-15
user/reptyr - new package

Fantome137:
Reptyr is a utility that allows you to reparent a running program to a new terminal


q66:
you should read https://github.com/chimera-linux/cports/blob/master/Packaging.md#correct_templates and https://github.com/chimera-linux/cports/blob/master/CONTRIBUTING.md

PR #3289, opened 2024-12-16
user/scrot: new package

theuae:
Scrot is a screenshot utility for X11


JamiKettunen:
https://github.com/chimera-linux/cports/pull/3288#issuecomment-2544211186 from before still applies

PR #3291, opened 2024-12-16
user/terraform: new package

Fantome137:
Adds terrafrom, a tool for building and updating infrastructure as code idempotently

### DIFF
--- a/user/terraform/template.py
+++ b/user/terraform/template.py
@@ -1,0 +1,20 @@
+pkgname = "terraform"
+pkgver = "1.10.2"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    f"-ldflags=-X github.com/hashicorp/terraform/version.dev=no",
+]
+makedepends = ["go"]
+pkgdesc = "Tool for building infrastructure as code"
+maintainer = "LeFantome <fanome137@proton.me>"
+license = "BUSL-1.1"
+url = "https://github.com/hashicorp/terraform"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "3be0296246a96acb5d45cb354ca32417bffd55d8086a4e3fcd71073e4295bb17"
+
+def post_extract(self):
+    self.rm("internal/cloud/*_test.go", glob=True)
+
+def post_install(self):
+    self.install_license("LICENSE")


Fantome137:
I teach a class that uses Terraform. It would be great to use Chimera Linux in it. I had not heard of OpenTofu and will check it out. For the class, I am going to have to stick to Terraform.

q66:
they are the same thing, one is a fork due to terraform becoming proprietary software

Fantome137:
I totally understand, and I prefer the MPL, but the OpenTofu FAQ says explicitly that the projects will diverge after 1.6, they have both had releases since then, and they are not on the same version number now.

The course curriculum is explicitly Terraform, Kubernetes, AWS, Docker, and Jenkins. I cannot change any of that (so I cannot use Podman either). The Linux distro is not set though. The alternative libc and userland in Chimera Linux will be perfect when explaining containers.

Perhaps I will have to maintain my own repo

user/terraform/template.py
```diff
@@ -0,0 +1,17 @@
...
+license = "BUSL-1.1"
+url = "https://github.com/hashicorp/terraform"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "3be0296246a96acb5d45cb354ca32417bffd55d8086a4e3fcd71073e4295bb17"
+# Tests cause a segfault - compiled program works
+options = ["!check"]
+
+def post_install(self):
+     self.install_license("LICENSE")
+
```
q66, line 17: spurious newline

user/terraform/template.py
```diff
@@ -0,0 +1,17 @@
...
+pkgrel = 0
+build_style = "go"
+makedepends = ["go"]
+pkgdesc = "Tool for building and updating infrastructure as code idempotently"
+maintainer = "LeFantome <fanome137@proton.me>"
+license = "BUSL-1.1"
+url = "https://github.com/hashicorp/terraform"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "3be0296246a96acb5d45cb354ca32417bffd55d8086a4e3fcd71073e4295bb17"
+# Tests cause a segfault - compiled program works
```
q66, line 12: this is not informative

also you can do this instead

```
def post_extract(self):
    # needs network/some of them segfault
    self.rm("internal/cloud/*_test.go", glob=True)
```

user/terraform/template.py
```diff
@@ -0,0 +1,17 @@
+pkgname = "terraform"
+pkgver = "1.10.2"
+pkgrel = 0
+build_style = "go"
```
q66, line 4: also needs

```
make_build_args = [
    f"-ldflags=-X github.com/hashicorp/terraform/version.dev=no",
]
```

user/terraform/template.py
```diff
@@ -0,0 +1,17 @@
...
+makedepends = ["go"]
+pkgdesc = "Tool for building and updating infrastructure as code idempotently"
```
q66, line 6: just something like `Tool for infrastructure as code` should suffice

Fantome137:
Changes applied

user/terraform/template.py
```diff
@@ -0,0 +1,20 @@
+pkgname = "terraform"
+pkgver = "1.10.2"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    f"-ldflags=-X github.com/hashicorp/terraform/version.dev=no",
+]
+makedepends = ["go"]
```
q66, line 8: this is hostmakedepends

user/terraform/template.py
```diff
@@ -0,0 +1,20 @@
...
+pkgdesc = "Tool for building infrastructure as code"
+maintainer = "LeFantome <fanome137@proton.me>"
+license = "BUSL-1.1"
+url = "https://github.com/hashicorp/terraform"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "3be0296246a96acb5d45cb354ca32417bffd55d8086a4e3fcd71073e4295bb17"
+
+def post_extract(self):
+    self.rm("internal/cloud/*_test.go", glob=True)
```
q66, line 17: add a comment why these are removed

q66:
format your template with black

q66:
no response to review, you can reopen if you like

PR #3296, opened 2024-12-17
[draft] main/linux-rpi: update to 6.12.5

q66:
This updates the rpi kernel to the 6.12 branch. Device testing is necessary (I will provide a test image later).

need testing with the HA dtses @jcgruenhage


jcgruenhage:
@q66 sorry for the late reply.

I've finally found some time to test this, works as expected.

PR #3299, opened 2024-12-17
main/vmaf: new package

Fantome137:
Added new package vmaf, a perceptual video quality assessment tool developed by Netflix

It needs to be in the main repository so that it can be used to build other main packages ( such as ffmpeg )


main/vmaf/template.py
```diff
@@ -0,0 +1,22 @@
...
+meson_dir = "libvmaf"
+hostmakedepends = ["pkgconf", "meson", "nasm", "doxygen"]
+checkdepends = ["xxd"]
+pkgdesc = "Perceptual video quality assessment tool developed by Netflix"
+maintainer = "LeFantome <fantome137@proton.me>"
+license = "BSD-2-Clause-Patent"
+url = "https://github.com/Netflix/vmaf"
+source = f"{url}/archive/v{pkgver}/libvmaf-{pkgver}.tar.gz"
+sha256 = "7178c4833639e6b989ecae73131d02f70735fdb3fc2c7d84bc36c9c3461d93b1"
+options = ["splitstatic"]
```
JamiKettunen, line 14: why? it's already a default for `-devel` which is the only place `.a` static lib files should generally go under where they get further autopkg'd to `-devel-static`

main/vmaf/template.py
```diff
@@ -0,0 +1,22 @@
+pkgname = "vmaf"
+pkgver = "3.0.0"
+pkgrel = 0
+build_style = "meson"
+meson_dir = "libvmaf"
+hostmakedepends = ["pkgconf", "meson", "nasm", "doxygen"]
```
JamiKettunen, line 6: does `doxygen` generate any worthwhile documentation for end-users? `configure_args = ["-Denable_docs=false"]` at least seems to be an option and they probably aren't too valuable if also available online in the same form

also sort the list alphabetically

main/vmaf/template.py
```diff
@@ -0,0 +1,22 @@
...
+checkdepends = ["xxd"]
+pkgdesc = "Perceptual video quality assessment tool developed by Netflix"
+maintainer = "LeFantome <fantome137@proton.me>"
+license = "BSD-2-Clause-Patent"
+url = "https://github.com/Netflix/vmaf"
+source = f"{url}/archive/v{pkgver}/libvmaf-{pkgver}.tar.gz"
+sha256 = "7178c4833639e6b989ecae73131d02f70735fdb3fc2c7d84bc36c9c3461d93b1"
+options = ["splitstatic"]
+
+def post_install(self):
```
JamiKettunen, line 16: please run a formatter as per https://github.com/chimera-linux/cports/blob/master/Packaging.md#correct_style, there'd be two new lines above each function if you did

main/vmaf/template.py
```diff
@@ -0,0 +1,22 @@
...
+     self.install_files("model", "usr/share")
+     self.install_license("LICENSE")
+
+@subpackage("vmaf-devel")
+def _(self):
+    return ["usr/include", "usr/lib/*.so", "usr/lib/*.a"]
```
JamiKettunen, line 22: why not `return self.default_devel()` like most other templates? it'd do the same

JamiKettunen (changes requested):

main/vmaf/template.py
Fantome137: I added this before I was creating the -devel package. I will remove it.

Fantome137:
Fixes applied as requested.

Fantome137 (commented):
Ready for review

main/vmaf/template.py
Fantome137: Updated

main/vmaf/template.py
```diff
@@ -0,0 +1,23 @@
...
+pkgdesc = "Perceptual video quality assessment tool developed by Netflix"
+maintainer = "LeFantome <fantome137@proton.me>"
+license = "BSD-2-Clause-Patent"
+url = "https://github.com/Netflix/vmaf"
+source = f"{url}/archive/v{pkgver}/libvmaf-{pkgver}.tar.gz"
+sha256 = "7178c4833639e6b989ecae73131d02f70735fdb3fc2c7d84bc36c9c3461d93b1"
+
+
+def post_install(self):
+    self.install_files("model", "usr/share")
```
q66, line 17: this probably shouldn't be installed? or at least not like that

main/vmaf/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "vmaf"
+pkgver = "3.0.0"
+pkgrel = 0
+build_style = "meson"
+meson_dir = "libvmaf"
```
q66, line 5: you should be using build_wrksrc for this

main/vmaf/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "vmaf"
```
q66, line 1: this should probably be called libvmaf

Fantome137: Even though it creates an executable file in /usr/bin/vmaf?

q66: hm probably not then, i thought vmaf was the whole repo while the libvmaf subdir was only a library

q66: but looking at it, in this case the program should be split off because it's pretty much the same size as the lib

so create a vmaf-libs subpackage containing self.default_libs

q66:
also the PR is missing subpackage links

main/vmaf/template.py
Fantome137: How should this be installed?

main/vmaf/template.py
q66: this is not resolved

main/vmaf/template.py
q66: definitely not in a nondescript directory in /usr/share

does the program load them in some way? from some specific path? if not, you can still install them but probably split into a subpackage

if there is a path it loads them from by default, then that should be the path, otherwise maybe something like `/usr/share/vmaf-models` or similar

q66:
no response to review, you can reopen if you like

PR #3300, opened 2024-12-17
main/ffmpeg: build with support for vmaf

Fantome137:
Modified the template to build with VMAF support.

This is needed to support tools such as ab-av1

### DIFF
--- a/main/ffmpeg/template.py
+++ b/main/ffmpeg/template.py
@@ -1,6 +1,6 @@
 pkgname = "ffmpeg"
 pkgver = "7.1"
-pkgrel = 4
+pkgrel = 5
 build_style = "configure"
 configure_args = [
     "--prefix=/usr",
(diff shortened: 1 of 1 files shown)


main/vmaf/template.py
```diff
@@ -0,0 +1,22 @@
...
+hostmakedepends = ["pkgconf", "meson", "nasm", "doxygen"]
+checkdepends = ["xxd"]
+pkgdesc = "Perceptual video quality assessment tool developed by Netflix"
+maintainer = "LeFantome <fantome137@proton.me>"
+license = "BSD-2-Clause-Patent"
+url = "https://github.com/Netflix/vmaf"
+source = f"{url}/archive/v{pkgver}/libvmaf-{pkgver}.tar.gz"
+sha256 = "7178c4833639e6b989ecae73131d02f70735fdb3fc2c7d84bc36c9c3461d93b1"
+options = ["splitstatic"]
+
```
JamiKettunen, line 15: either of the two mentioned formatters on https://github.com/chimera-linux/cports/blob/master/Packaging.md#correct_style would've added an additional newline before each function, CI would fail if they're not properly formatted

Fantome137:
Changes applied

q66:
depends on stale parent PR

PR #3310, opened 2024-12-18
user/monolith: new package

sudanchapagain:
## Description

new package

## Checklist

### DIFF
--- a/user/monolith/template.py
+++ b/user/monolith/template.py
@@ -1,0 +1,15 @@
+pkgname = "monolith"
+pkgver = "2.8.3"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo-auditable", "pkgconf"]
+makedepends = [
+    "openssl-devel",
+    "rust-std",
+]
+pkgdesc = "CLI tool for saving complete web pages as a single HTML file"
+maintainer = "Sudan <sudanchapagain@proton.me>"
+license = "CC0-1.0"
+url = "https://github.com/Y2Z/monolith"
+source = f"{url}/archive/v{pkgver}/monolith-{pkgver}.tar.gz"
+sha256 = "51769e6505d5708ac296e5d93e280c9fefa7873452d471c5106aaeb7c3667f9f"


user/monolith/template.py
```diff
@@ -0,0 +1,21 @@
...
+build_style = "cargo"
+hostmakedepends = ["pkgconf"]
+makedepends = [
+    "rust-std",
+    "cargo-auditable",
+    "openssl-devel",
+]
+
+env = {
+    "OPENSSL_NO_VENDOR": "1",
```
JamiKettunen, line 13: this is already a default for cargo build_style consumers https://github.com/chimera-linux/cports/blob/d134077/src/cbuild/util/cargo.py#L35-L36, please don't copy other distro packaging vebatem, they should be references for when things don't build and you can't figure something out quickly by yourself

also for future avoid adding unnecessary newlines around default template variables like `env` here since it breaks checking for correct template variable order

user/monolith/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "monolith"
+pkgver = "2.8.3"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["pkgconf"]
+makedepends = [
+    "rust-std",
+    "cargo-auditable",
```
JamiKettunen, line 8: `cargo-auditable` belongs as first in `hostmakedepends`, try cross-building with `./cbuild -a aarch64 pkg user/monolith` assuming non-aarch64 host, also sort `makedepends` alphabetically

sudanchapagain: > try cross-building with ./cbuild -a aarch64 pkg user/monolith assuming non-aarch64 host

it got built successfully, should i do anything else or push?

JamiKettunen: I meant testing that before moving `cargo-auditable` to `hostmakedepends` locally just to see what happened before if a cross-build was attempted. things cross-compiling isn't mandatory but the dependencies at least should be placed in the correct list anyway which (always first) testing cross-build can help with

PR #3325, opened 2024-12-21
[WIP] Tcl/tk 9.0

q66:
This does not work properly yet. It builds, but at least git-gui/gitk do not work properly (gitk semi-works after removing a version check).


q66:
this will likely remain drafted for ages so let's revisit once more things actually support tcl9 properly

PR #3326, opened 2024-12-22
main/power-profiles-daemon: fix tlp conflict

subnut:
## Description
tlp is also used independently to set battery charging limits on supported hardware.  
That functionality does not clash with power-profiles-daemon.

## Checklist

### DIFF
--- a/main/power-profiles-daemon/template.py
+++ b/main/power-profiles-daemon/template.py
@@ -1,6 +1,6 @@
 pkgname = "power-profiles-daemon"
 pkgver = "0.23"
-pkgrel = 2
+pkgrel = 3
 build_style = "meson"
 configure_args = [
     "--libexecdir=/usr/lib",  # XXX drop libexec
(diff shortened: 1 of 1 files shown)


q66:
i'm still not quite sure if the conflict is a good thing at all (because it leads to stuff like this), as far as i can tell there is nothing really preventing you from having both installed? unless they dbus-activate or something? (i guess power-profiles-daemon does dbus-activate, tlp does not)

valpackett:
There's a patch around to make ppd *use* tlp: https://github.com/Rongronggg9/power-profiles-daemon/commit/e5536af79b4f1480c06fc2960ca39076b80367d2

subnut:
Ping @q66 

> I guess we could remove the conflict completely, and let the user determine what they want to use.
>
>I prefer this.
>
> @q66 lmk if I should update this PR to make this change (or you could commit it directly, in which case this PR can be closed)

triallax:
bump, i also agree that dropping the conflict is the way to go, should we go ahead and do that?

triallax:
relevant link: https://linrunner.de/tlp/faq/ppd.html#does-power-profiles-daemon-conflict-with-tlp

PR #3328, opened 2024-12-22
user/vis: new package

apprehensions:
## Description

## Checklist


main/lua5.4-lpeg/patches/makefile.patch
```diff
@@ -0,0 +1,13 @@
+use cbuild compiler flags
```
q66, line 1: the patches directory in lua5.4-lpeg should be a symlink to the one in lua5.1-lpeg instead

main/lua5.4-lpeg/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://luarocks.org/modules/gvvaughan/lpeg"
```
q66, line 1: the update file can be a symlink too

user/libtermkey/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "libtermkey"
```
q66, line 1: we previously removed libtermkey because it's dead/deprecated upstream

apprehensions: unfortunately it is still a dependency of vis.

https://github.com/martanne/vis/issues/1071

apprehensions: what do you think about vendoring libtermkey within user/vis? 

> Besides it consists of only 3 C files which could always be included in vis directly (but I prefer not to do so).

q66: vendoring it should be fine yeah

apprehensions: attempting to vendor it has been a pain in the ass since im not familiar with this, is vendoring it truly preferred?

sewnie:
I no longer use this editor, I moved on to something better, helix.

PR #3331, opened 2024-12-23
user/vgrep: new package

ttyyls:
## Description

acts as a pager for any stream in the grep result format.  the readme emphasizes
the fzf integration and the code checks for ripgrep before falling back on
git-grep or bare grep, so both were added as deps.

## Checklist

### DIFF
--- a/user/vgrep/template.py
+++ b/user/vgrep/template.py
@@ -1,0 +1,20 @@
+pkgname = "vgrep"
+pkgver = "2.8.0"
+pkgrel = 0
+build_style = "go"
+make_build_args = [f"-ldflags=-X main.version={pkgver}"]
+hostmakedepends = ["go", "go-md2man"]
+pkgdesc = "Pager for grep output"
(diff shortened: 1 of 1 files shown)


user/vgrep/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "vgrep"
+pkgver = "2.8.0"
+pkgrel = 0
+build_style = "go"
+make_build_args = [f"-ldflags=-X main.version={pkgver}"]
+hostmakedepends = ["go", "go-md2man"]
+depends = ["fzf", "ripgrep"]
```
q66, line 7: that doesn't sound like it should depend on ripgrep

q66: i don't see an obvious dependency on fzf either

ttyyls: i explained my reasoning in the pr descriptio but i'll drop regardless

PR #3335, opened 2024-12-23
main/bubblejail: update to 0.9.4

ttyyls:
## Description

also skips completion test that will always fail in cbuild env due to a lack of
completion support for /bin/sh.

## Checklist


igo95862:
> also skips completion test that will always fail

This is a bug introduced in the recent version where list command completion always fails. I plan on releasing a hotfix soon to address this.

PR #3338, opened 2024-12-23
cbuild: add cargo update function to cargo.py

Guiglin:
## Description

Add `update` function into cargo.py to easily launch `cargo update <arg>` if needed for some Rust packages. Sometimes, due to an old configuration and a more recent rust version, it may be needed to update the dependencies of a package as recorded in the local lock file.
## Checklist


q66:
there isn't really a need for a wrapper for this

Guiglin:
Indeed... I then close the MR

PR #3346, opened 2024-12-25
user/openttd: make update check only match stable versions

breakgimme:
## Description

the original update check matched the last version in the file, regardless of what channel it was
the updated pattern only matches if the version is stable
i'm not sure whether i need to bump the pkgrel for a change like this, if yes please let me know and i'll do

## Checklist


breakgimme:
had to force push because i forgot to set the proper email in git on this machine and noticed just now

PR #3361, opened 2024-12-29
main/geoclue: added service file

doskel:
## Description

added a dinit service to start geoclue.

## Checklist


doskel:
it's occurring to me this might be useless, as geoclue is autolaunched on each call via dbus; if anyone else wants to chime in about whether or not there's a point, i'd trust that more than my vaguely incompetent knowledge of dbus x3

q66:
well since it's dbus-launched, this is indeed wrong on its own, but what you could do is make the dbus service launch the dinit service, see e.g. the patch in polkit and other places that already do it

though the actual right thing to do would be to have proper dbus activation support and then just refer the dbus service to the dinit service, but we don't have that yet

q66:
i will close this, there is no point in adding workarounds and once there is dbus activation it should be addressed properly

PR #3372, opened 2024-12-31
main/gst-plugins-bad: Remove unnecessary patch file for glibc basename

Biswa96:
## Description

glibc basename was replaced with glib g_path_get_basename function in upstream. https://gitlab.freedesktop.org/gstreamer/gstreamer/-/commit/3e319081f567b439f2c4b43ba871d609e0cf8a88

## Checklist


main/gst-plugins-bad/template.py
```diff
@@ -1,6 +1,6 @@
 pkgname = "gst-plugins-bad"
 pkgver = "1.24.10"
-pkgrel = 1
+pkgrel = 2
```
q66, line 3: not needed

Biswa96: reverted

Issue #3374, opened 2024-12-31
lxc: fix template scripts


BarryLhm:
there's incompatible tools and patterns in template scripts like system sed + gnu sed syntax and system getopt + ugetopt syntax...

q66:
that's not a specific issue

q66:
either post patches or at least some actual specific information

BarryLhm:
i'm checking it further

BarryLhm:
```140c140
< if ! options=$(getopt -o d:r:a:hl -l dist:,release:,arch:,help,list,variant:,\
---
> if ! options=$(ugetopt -o d:r:a:hl -l dist:,release:,arch:,help,list,variant:,\
416c416
< sed -i -e "/lxc.net.0/{w ${LXC_PATH}/config-network" -e "d}" "${LXC_PATH}/config"
---
> gsed -i -e "/lxc.net.0/{w ${LXC_PATH}/config-network" -e "d}" "${LXC_PATH}/config"
419c419
< sed -i -e "/lxc./{w ${LXC_PATH}/config-auto" -e "d}" "${LXC_PATH}/config"
---
> gsed -i -e "/lxc./{w ${LXC_PATH}/config-auto" -e "d}" "${LXC_PATH}/config"
474,478c474,478
<   sed -i "s#LXC_NAME#${LXC_NAME}#g" "${file}"
<   sed -i "s#LXC_PATH#${LXC_PATH}#g" "${file}"
<   sed -i "s#LXC_ROOTFS#${LXC_ROOTFS}#g" "${file}"
<   sed -i "s#LXC_TEMPLATE_CONFIG#${LXC_TEMPLATE_CONFIG}#g" "${file}"
<   sed -i "s#LXC_HOOK_DIR#${LXC_HOOK_DIR}#g" "${file}"
---
>   gsed -i "s#LXC_NAME#${LXC_NAME}#g" "${file}"
>   gsed -i "s#LXC_PATH#${LXC_PATH}#g" "${file}"
>   gsed -i "s#LXC_ROOTFS#${LXC_ROOTFS}#g" "${file}"
>   gsed -i "s#LXC_TEMPLATE_CONFIG#${LXC_TEMPLATE_CONFIG}#g" "${file}"
>   gsed -i "s#LXC_HOOK_DIR#${LXC_HOOK_DIR}#g" "${file}"
484c484
[...]

PR #3376, opened 2024-12-31
user/lxc: update to 6.0.3 and fix lxc-create template scripts

BarryLhm:
## Checklist

### DIFF
--- a/user/lxc/patches/lxc-busybox.in.patch
+++ b/user/lxc/patches/lxc-busybox.in.patch
@@ -1,0 +1,11 @@
+--- a/templates/lxc-busybox.in
++++ b/templates/lxc-busybox.in
+@@ -260,7 +260,7 @@ EOF
+   return 0
+ }
+ 
+-if ! options=$(getopt -o hp:n: -l help,rootfs:,path:,name:,mapped-uid:,mapped-gid:,busybox-path: -- "$@"); then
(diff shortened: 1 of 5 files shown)


BarryLhm:
Closes #3374

q66:
you don't need gsed, the problem is that `-i` requires an argument, so changing `sed -i '...'` to `sed -i '' '...'` should be enough

BarryLhm:
what about the getopt ones

q66:
the getopt switch is okay, nothing else you can really do

BarryLhm:
```
sed: 1: "d}
": extra characters at the end of d command
```

JamiKettunen:
the patterns are compatible with FreeBSD `sed`, literally just replace `-i` with `-i ''` and it'll work at the cost of making them incompatible with `gsed` (not that this matters to us)

BarryLhm:
>or should be anyway, specify the exact command which doesn't

source/templates/lxc-download.in
/usr/share/lxc/templates/lxc-download
in line 416, 419:
`sed -i '' -e "/lxc.net.0/{w ${LXC_PATH}/config-network" -e "d}" "${LXC_PATH}/config"`
`sed -i '' -e "/lxc./{w ${LXC_PATH}/config-auto" -e "d}" "${LXC_PATH}/config"`

still don't work

q66:
indeed, specifically the 'd}' is problematic (without the 'd' it works, but then it does not delete the lines from the original file)

i'm not sure if there is a standard equivalent, i need to look into it

BarryLhm:
@q66 was is forgotten?

PR #3391, opened 2025-01-03
user/zig-bin: new package

apprehensions:
## Description

Binary toolchain distribution of the Zig programming language.

## Checklist

It seems that packages with pre-compiled sources haven't been distributed before in Chimera, so i apologize if these sorts of packages aren't allowed or are restricted to build-from-source only, since it doesn't seem to be mentioned in the documentation.

### DIFF
--- a/user/zig-bin/template.py
+++ b/user/zig-bin/template.py
@@ -1,0 +1,19 @@
+pkgname = "zig-bin"
+pkgver = "0.13.0"
+pkgrel = 0
+archs = ["aarch64", "ppc64le", "riscv64", "x86_64"]
+pkgdesc = "Zig programming language toolchain"
+maintainer = "sewn <sewn@disroot.org>"
+license = "MIT"
+url = "https://github.com/ziglang/zig"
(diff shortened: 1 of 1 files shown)


q66:
what's the reason to package this over just building the toolchain from source

either way, we don't package binary toolchains like this

apprehensions:
> what's the reason to package this over just building the toolchain from source

Building the toolchain from source no longer works since the newest zig 0.13.0 version requires LLVM 18 while chimera only has LLVM 19. Additionally, users can simply fetch this package and get going.

> either way, we don't package binary toolchains like this

Could this be documented? Maybe for general binary packages as well and the reasoning, since otherwise, a user has to fetch the binary toolchain by hand or use `zvm`.

q66:
> Building the toolchain from source no longer works since the newest zig 0.13.0 version requires LLVM 18 while chimera only has LLVM 19.

yes but packaging binaries is not a solution to that

apprehensions:
There are no other solutions other than to manage it using `zvm` or by hand. Please document packaging pre-compiled sources so others don't repeat this mistake.

q66:
of course there is, it's "figure out a way to have multiple llvm versions in parallel" and after that, fix up the ordinary packaging

apprehensions:
Too complicated and resource-heavy :(

PR #3393, opened 2025-01-03
user/spotify-player: new package at 0.20.4

apprehensions:
## Description

A Spotify player in the terminal with full feature parity

## Checklist

- [X] I have read [Packaging.md](https://github.com/chimera-linux/cports/blob/master/Packaging.md#quality_requirements) (nit-pick: maintainer-chosen package features missing)

### DIFF
--- a/user/spotify-player/template.py
+++ b/user/spotify-player/template.py
@@ -1,0 +1,36 @@
+pkgname = "spotify-player"
+pkgver = "0.20.4"
+pkgrel = 0
+build_style = "cargo"
+make_build_args = [
+    "--no-default-features",
+    "--features",
+    "sdl-backend,streaming,media-control,image,sixel,notify",
+]
+make_check_args = [*make_build_args]
+hostmakedepends = [
+    "cargo-auditable",
+    "cmake",
+    "cbindgen",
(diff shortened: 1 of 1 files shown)


q66:
commit message should be just "user/spotify-player: new package"

also, alsa-backend should not be used, either pulseaudio-backend or something generic like gstreamer-backend (or even sdl/portaudio but not sure which is best maintained/supported) is a better pick

apprehensions:
portaudio seems to be a alsa wrapper, gstreamer is most likely going to appear on a gnome system, and sdl will appear on most systems, since it is a requirement of gstreamer and mpv, and subsequently gnome, so it seems like a safe option.

q66:
gstreamer will be on any system that has gtk installed at all for any reason, or a variety of other applications

apprehensions:
i do not have gtk4 on my system but i have sdl =), gtk3 is installed but it doesnt seem to depend on gstreamer.

q66:
well if the sdl backend is well supported that's fine too, e.g. in mpv it's discouraged because of latency and some other reasons, but for plain music playback it probably matters a whole lot less

PR #3394, opened 2025-01-04
main/neovim: update to 0.10.3

Natthias:
## Description

Works fine in my testing, fixes an annoying bug with tinymist

## Checklist


q66:
can you update tree-sitter and tree-sitter-cli while at it (tree-sitter updates need rebuilds for all things depending on tree-sitter-devel and neovim is one of them)?

PR #3423, opened 2025-01-06
user/sdl3: new package

z-erica:
note that this is a stable api/abi preview: https://github.com/libsdl-org/SDL/releases/tag/preview-3.1.6. im also packaging the current git head of sdl2-compat to experiment with whether it could be viable as a drop in replacement for the current main/sdl (i tried to override the priorities so it won't be installed over main/sdl by default)

idk what the plan is to actually name this package once there's a stable release. i guess if it becomes the new main/sdl that can have a `provides = ["sdl3"]`?


q66:
i think it should just be called sdl3, the unversioned name was never a very good idea and now is an opportunity to fix it

q66:
so far my testing shows that everything pretty much works ok with compat, so i'll gradually switch builds over to using sdl2-compat-devel and then drop sdl2 entirely

PR #3433, opened 2025-01-10
user/go-task: new package

rietbergenm:
## Description
Cool tool, much like a modernized version of Makefiles with a focus towards running commands rather than compiling software.
The naming is not very consistent though; the project itself seems to call it `task` most of the time, but in other places, `go-task` is used more often. This is of course because it is such a generic name. To avoid conflict, it seemed wise to name the package go-task. What do you think of this?

## Checklist

### DIFF
--- a/user/go-task/template.py
+++ b/user/go-task/template.py
@@ -1,0 +1,23 @@
+pkgname = "go-task"
+pkgver = "3.40.1"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    f"-ldflags=-X github.com/go-task/task/v3/internal/version.version=v{pkgver}",
+    "./cmd/task",
+]
+hostmakedepends = ["go"]
+pkgdesc = "Task runner / simpler Make alternative written in Go"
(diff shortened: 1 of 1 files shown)


user/go-task/template.py
```diff
@@ -0,0 +1,23 @@
...
+license = "MIT"
+url = "https://taskfile.dev"
+source = f"https://github.com/go-task/task/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "e80cdfa2afefa69238e5078960d50a8e703de1043740b277946629ca5f3bde85"
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+
+    self.install_files("website/docs/", "usr/share/doc", name=pkgname)
```
q66, line 18: we don't usually install those

rietbergenm: Alright, I'll remove it!

If a package does not ship man pages, but it does have other sorts types documentation, do we package that in any case, or is it only man pages?

rietbergenm:
I removed the documentation and fixed the version string embedded in the binary so `task --version` now shows the version instead of `devel` :)

PR #3434, opened 2025-01-10
main/editline: new package

rietbergenm:
## Description
Simple, small readline alternative. It is claims to be call-compatible with gnu readline, but only a fraction of the size.

What I wasn't sure of is if this should go in `main` or in `user`, as it is a really simple system library on which a lot of programs could depend (I came across it when trying to package nix for chimera btw). The software has been around for ages (1992) and the build completes cleanly with no parts skipped, so stability is not really a problem I would say.
But then on the other hand, it is a new contribution, so then it should go in `user`.
Any thoughts on this?

## Checklist

### DIFF
--- a/main/editline-devel
+++ b/main/editline-devel
@@ -1,0 +1,1 @@
+editline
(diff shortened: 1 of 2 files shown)


q66:
what's the point of this? we already have libedit

PR #3454, opened 2025-01-13
user/niri: adopt + update to 25.01

z-erica:


user/niri/patches/970.patch
```diff
@@ -0,0 +1,20 @@
+From b8825ca5fd6ba49233d6f1b11fa5ce1946bc4c30 Mon Sep 17 00:00:00 2001
+From: Erica Z <zerica@callcc.eu>
+Date: Mon, 13 Jan 2025 12:17:32 +0100
+Subject: [PATCH] change niri-session shebang to /bin/bash
+
+&> redirection is undefined in POSIX sh
```
q66, line 6: just fix the redirection instead, it's just "> foo 2>&1"

user/niri/template.py
```diff
@@ -27,24 +28,28 @@
...
 pkgdesc = "Scrollable-tiling wayland compositor"
-maintainer = "Orphaned <orphaned@chimera-linux.org>"
+maintainer = "Erica Z <zerica@callcc.eu>"
 license = "GPL-3.0-or-later"
 url = "https://github.com/YaLTeR/niri"
 source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
-sha256 = "d8854830436a87215b0bc6a60b6d43f350d927a03a2798c75f0fbda228bac8d3"
+sha256 = "86b89bcfc3fc6a8ed81f9e3f0ac7a29bd30267515efb2c19e1e0bc2ccd67b649"
+# upstream recommendation
+nostrip_files = ["usr/bin/niri"]
```
q66, line 41: ironic for the infinite strip window manager

(why is that? i don't immediately see this anywhere)

q66: it makes the size go from like 14M to >100M and the only benefit is "don't have to install niri-dbg"

PR #3458, opened 2025-01-14
user/udiskie: new package

z-erica:


user/udiskie/template.py
```diff
@@ -0,0 +1,36 @@
...
+]
+checkdepends = [
+    "python-pytest",
+]
+depends = [
+    "python-docopt",
+    "python-gobject",
+    "python-pyyaml",
+]
+checkdepends += depends
```
q66, line 19: use star expansion

user/udiskie/template.py
```diff
@@ -0,0 +1,33 @@
...
+
+def pre_check(self):
+    # requires python-keyutils, unmaintained optional dep
+    self.rm("test/test_cache.py")
+
+
+def post_install(self):
+    self.install_license("COPYING")
+    # has no matching command
+    self.uninstall("usr/share/zsh/site-functions/_udiskie-canonical_paths")
```
q66, line 33: use !lintcomp instead

user/udiskie/template.py
```diff
@@ -0,0 +1,33 @@
+pkgname = "udiskie"
+pkgver = "2.5.7"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "gettext-devel",
```
q66, line 6: should probably be just gettext

PR #3474, opened 2025-01-17
main/python-numpy: make tests depend on parent + checkdepends

z-erica:
this time for real i prommy

### DIFF
--- a/main/python-numpy/template.py
+++ b/main/python-numpy/template.py
@@ -1,6 +1,6 @@
 pkgname = "python-numpy"
 pkgver = "2.2.2"
-pkgrel = 0
+pkgrel = 1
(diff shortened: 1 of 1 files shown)


main/python-numpy/template.py
```diff
@@ -101,7 +101,10 @@ def post_install(self):
 @subpackage("python-numpy-tests")
 def _(self):
     self.subdesc = "tests"
-    self.depends += ["python"]
+    self.depends = [
+        *self.parent.checkdepends,
```
q66, line 105: this can be just checkdepends...

q66: no i mean

you can write just `checkdepends` instead of `self.parent.checkdepends`

PR #3481, opened 2025-01-17
main/playerctl: add user service, adopt

ttyyls:
## Description

quoting the manpage

```
     Playerctl also comes with a daemon called playerctld which keeps track of
     media player activity. When playerctld is running, playerctl commands
     will act on the media player with the most recent activity.  Run the
     command playerctld daemon to start the daemon.

```

on systemd based distributions the dbus activation is tied to $DISPLAY, although
I don't see why not set it at a login.target since mpris is still usable and
thus potentially controllable on non-graphical environments.

regardless, the service depends on graphical.target and dbus to remain inline
with upstream expectations see:
https://github.com/altdesktop/playerctl#playerctld-autostart-issues

## Checklist

### DIFF
--- a/main/playerctl/files/playerctld.user
+++ b/main/playerctl/files/playerctld.user
@@ -1,0 +1,5 @@
+type = process
+command = /usr/bin/playerctld
+depends-on = dbus
+depends-on = graphical.target
+log-type = buffer
--- a/main/playerctl/template.py
+++ b/main/playerctl/template.py
@@ -1,6 +1,6 @@
 pkgname = "playerctl"
 pkgver = "2.4.1"
-pkgrel = 0
+pkgrel = 1
 build_style = "meson"
 configure_args = ["-Dgtk-doc=false"]
 hostmakedepends = [
@@ -10,7 +10,7 @@ hostmakedepends = [
 ]
 makedepends = ["glib-devel"]
 pkgdesc = "MPRIS media player CLI tool"
-maintainer = "Orphaned <orphaned@chimera-linux.org>"
+maintainer = "ttyyls <contact@behri.org>"
 license = "LGPL-3.0-or-later"
 url = "https://github.com/altdesktop/playerctl"
 source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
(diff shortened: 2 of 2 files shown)


main/playerctl/files/playerctld.user
```diff
@@ -0,0 +1,5 @@
+type = process
+command = /usr/bin/playerctld
+depends-on = dbus
+depends-on = graphical.target
```
q66, line 4: you sure?

ttyyls: besides what i previously mentionned in the pr description, i tested it on my machine for a couple of days, and it indeed works.

However i still don't get why the service was supposed to be launched with dbus actvation as a result of a mpris client asking for metadata or a mpris command executed earlier yet neither trigger it for me...

As i understand dbus activation should result in a lingering process (analogous to notification daemons providing org.freedesktop.notifications) but in this case i don't see it, either way it'll be far from the first dbus activated service with a dinit service as well.

so maybe?

q66: i was talking about the graphical.target dependency, does not make sense to me

q66: and yes, if it activates with dbus, then such user service is in conflict and that needs to be resolved

ttyyls: the graphical.target can be dropped as i already said in the pr desc, as for the dbus activation that doesn't work that evades me. on systemd based distributions a service file is shipped anyway and upstream recommends manually launching it as a wokraround on xorg, that imported DISPLAY so i made the equivalent dinit server depend on graphical.target out of caution, again see https://github.com/altdesktop/playerctl?tab=readme-ov-file#playerctld-autostart-issues

last commit from upstream is 4 years ago so I wouldn't hold my breath for a fix for the dbus activation

subnut: > why the service was supposed to be launched with dbus actvation as a result of a mpris client asking for metadata or a mpris command executed earlier yet neither trigger it for me

I don't think the service was supposed to be launched by dbus activation, as there's nothing in the code that's starting the service except for the command `playerctld daemon` itself.

main/playerctl/files/playerctld.user
```diff
@@ -0,0 +1,5 @@
...
+log-type = buffer
```
subnut, line 5: Why is this necessary? AFAICT playerctld doesn't emit any meaningful logs ..

PR #3502, opened 2025-01-20
user/lshw: new package

Guiglin:
## Description

lshw: HardWare LiSter for Linux: lshw is a small tool to provide detailed information on the hardware configuration of the machine. It can report exact memory configuration, firmware version, mainboard configuration, CPU version and speed, cache configuration, bus speed, etc.

## Checklist


user/lshw/template.py
```diff
@@ -0,0 +1,72 @@
...
+]
+makedepends = [
+    "gettext-devel",
+    "gtk+3-devel",
+    "linux-headers",
+    "zlib-ng-compat-devel",
+]
+depends = [
+    "cmd:gzip!chimerautils",
+    "zlib-ng-compat",
```
q66, line 33: you don't need this

user/lshw/template.py
```diff
@@ -0,0 +1,72 @@
...
+
+@subpackage("gtk-lshw")
+def _(self):
+    self.subdesc = "GTK ui"
+    self.depends = [
+        self.parent,
+        "gdk-pixbuf",
+        "glib",
+        "gtk+3",
+        "pango",
```
q66, line 66: you don't need these either

user/lshw/template.py
```diff
@@ -0,0 +1,72 @@
...
+    "cmd:gzip!chimerautils",
+    "zlib-ng-compat",
+]
+pkgdesc = "Hardware Lister"
+maintainer = "Guilhem Baccialone <guilhem.baccialone@zaclys.net>"
+license = "GPL-2.0-only"
+url = "https://ezix.org/project/wiki/HardwareLiSter"
+source = f"https://www.ezix.org/software/files/lshw-{_pkgver}.tar.gz"
+sha256 = "06d9cf122422220e5dc94e8ea5b01816a69bb6b59368f63d7f21fff31fc6922a"
+hardening = ["vis", "cfi"]
```
q66, line 41: glib program, this is not gonna work

user/lshw/template.py
```diff
@@ -0,0 +1,72 @@
...
+# no tests available and may fail to link ./core/liblshw.a
+options = ["!check", "!linkparallel"]
```
q66, line 43: may fail to link how

user/lshw/template.py
```diff
@@ -0,0 +1,72 @@
+pkgname = "lshw"
+pkgver = "02.20"
+_pkgver = f"B.{pkgver}"
```
q66, line 3: get rid of this variable

user/lshw/template.py
```diff
@@ -0,0 +1,72 @@
...
+]
+make_install_args = [
+    "install",
+    "install-gui",
+    f"VERSION={_pkgver}",
+    "NO_VERSION_CHECK=1",
+    "ZLIB=1",
+]
+make_use_env = True
+hostmakedepends = [
```
q66, line 21: this can be one line

user/lshw/template.py
```diff
@@ -0,0 +1,72 @@
...
+    "pkgconf",
+]
+makedepends = [
+    "gettext-devel",
+    "gtk+3-devel",
+    "linux-headers",
+    "zlib-ng-compat-devel",
+]
+depends = [
+    "cmd:gzip!chimerautils",
```
q66, line 32: where is this used

Guiglin: in the `src/Makefile` of the package
```make
GZIP?=gzip -9
...
%.gz: %
	$(GZIP) -c $< > $@

File: user/lshw/template.py (hunk 6 above)
Comment (Guiglin) [reply user/lshw/template.py]: done

File: user/lshw/template.py (hunk 2 above)
Comment (Guiglin) [reply user/lshw/template.py]: removed

File: user/lshw/template.py (hunk 1 above)
Comment (Guiglin) [reply user/lshw/template.py]: removed

File: user/lshw/template.py (hunk 7 above)
Comment (q66) [reply user/lshw/template.py:29]: what does that have to do with being a runtime dependency of the package?

File: user/lshw/template.py (hunk 3 above)
Comment (Guiglin) [reply user/lshw/template.py]: Removed

File: user/lshw/template.py (hunk 4 above)
Comment (Guiglin) [reply user/lshw/template.py:34]: I got the problem once. Hard to reproduce. But I remarked that the template file of void linux has this config too so I concluded it may fail not just for me

File: user/lshw/template.py (hunk 7 above)
Comment (Guiglin) [reply user/lshw/template.py]: removed. I don't know why, when I did not use it, it failed to build but perhaps due of a misconfiguration at the beginning of development of this template. Don't make any sense to have this as a dependency, I agree

File: user/lshw/template.py (hunk 4 above)
Comment (q66) [reply user/lshw/template.py:34]: no it doesn't, it disables parallel build, not parallel linking

Comment (Guiglin) [reply user/lshw/template.py:34]: yes but it does not seem they have the option to disable parallel link only. Anyway, do you prefer I remove this?

Comment (q66) [reply user/lshw/template.py:34]: you misunderstand what this option does

if there's a race in the build system this won't fix anything because it only prevents the linker process from creating parallel threads

Comment (Guiglin) [reply user/lshw/template.py:34]: arf ok then completely useless here. I'll remove it since I fail to reproduce any parallel build issue

Comment (Guiglin) [reply user/lshw/template.py]: removed

File: user/lshw/template.py [hunk 8]
```diff
@@ -0,0 +1,59 @@
...
+    "install",
+    "install-gui",
+    f"VERSION={pkgver}",
+    "NO_VERSION_CHECK=1",
+    "ZLIB=1",
+]
+make_use_env = True
+hostmakedepends = ["gettext", "pkgconf"]
+makedepends = [
+    "gettext-devel",
```
q66, line 22: probably don't need gettext-devel

user/lshw/template.py
```diff
@@ -0,0 +1,59 @@
...
+    "linux-headers",
+    "zlib-ng-compat-devel",
+]
+pkgdesc = "Hardware Lister"
+maintainer = "Guilhem Baccialone <guilhem.baccialone@zaclys.net>"
+license = "GPL-2.0-only"
+url = "https://ezix.org/project/wiki/HardwareLiSter"
+source = f"https://www.ezix.org/software/files/lshw-B.{pkgver}.tar.gz"
+sha256 = "06d9cf122422220e5dc94e8ea5b01816a69bb6b59368f63d7f21fff31fc6922a"
+# no tests available and may fail to link ./core/liblshw.a
```
q66, line 33: the comment is misleading now

user/lshw/template.py
```diff
@@ -0,0 +1,59 @@
...
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_file(
```
q66, line 38: all these files should go to gtk-lshw and they don't currently

user/lshw/template.py
```diff
@@ -0,0 +1,59 @@
...
+        "src/gui/integration/gtk-lshw.appdata.xml", "usr/share/metainfo"
+    )
+    self.install_file(
+        "src/gui/integration/gtk-lshw.pam", "usr/lib/pam.d", name="gtk-lshw"
+    )
+
+
+@subpackage("gtk-lshw")
+def _(self):
+    self.subdesc = "GTK ui"
```
q66, line 51: `Gtk UI`

user/lshw/template.py
```diff
@@ -0,0 +1,59 @@
...
+]
+make_use_env = True
+hostmakedepends = ["gettext", "pkgconf"]
+makedepends = [
+    "gettext-devel",
+    "gtk+3-devel",
+    "linux-headers",
+    "zlib-ng-compat-devel",
+]
+pkgdesc = "Hardware Lister"
```
q66, line 27: lister should be lowercase

user/lshw/template.py
Guiglin: indeed, removed

user/lshw/template.py
Guiglin: yes, removed

user/lshw/template.py
Guiglin: added a patch to correctly install file when doing a `make install-gui` and add lines to grab those files in the subpackage

user/lshw/template.py
Guiglin: fixed

q66:
i noticed none of this actually respects the distro-set `CFLAGS`/`LDFLAGS` and so on, and compiles with just its own always

so that needs to be fixed

q66:
also commit message has a typo

Guiglin:
It failed to build without it, I got a
```
/usr/include/fortify/unistd.h:156:1: error: reference to overloaded function could not be resolved; did you mean to call it?
  156 | _FORTIFY_FN(readlink) ssize_t readlink(const char *__p,
 ```

**Comment by Guiglin** [comment]
> what's wrong with _FORTIFY_SOURCE?

btw I added a new patch that fixes the problem

**Comment by q66** [comment]
instead of undefining fortify, the source itself should be fixed, the issue is due to overloading libc functions, which should not be done

**Comment by Guiglin** [comment]
> instead of undefining fortify, the source itself should be fixed, the issue is due to overloading libc functions, which should not be done

Sorry, it is what I had done but due to a wrong rebase between my machines, the patch that fixes the problem was lost. Here it is now, sorry for the time lost

**Comment by Guiglin** [comment]
> instead of undefining fortify, the source itself should be fixed, the issue is due to overloading libc functions, which should not be done
@q66 
I fix this issue, if you can take a look

File: user/lshw/template.py [hunk 13]
```diff
@@ -0,0 +1,50 @@
...
+maintainer = "Guilhem Baccialone <guilhem.baccialone@zaclys.net>"
+license = "GPL-2.0-only"
+url = "https://ezix.org/project/wiki/HardwareLiSter"
+source = f"https://www.ezix.org/software/files/lshw-B.{pkgver}.tar.gz"
+sha256 = "06d9cf122422220e5dc94e8ea5b01816a69bb6b59368f63d7f21fff31fc6922a"
+# no tests available
+options = ["!check"]
+
+
+@subpackage("gtk-lshw")
```
q66, line 36: subpackage should be called lshw-gtk to follow convention

q66:
other than the last thing, it's probably okay

user/lshw/template.py
Guiglin: done

PR #3528, opened 2025-01-26
user/impala: new package

bjin:
## Description

[impala](https://github.com/pythops/impala) is  a utility tool to manage WiFi networks via iwd in terminal.

## Checklist

### DIFF
--- a/user/impala/template.py
+++ b/user/impala/template.py
@@ -1,0 +1,17 @@
+pkgname = "impala"
+pkgver = "0.2.4"
+pkgrel = 0
+build_style = "cargo"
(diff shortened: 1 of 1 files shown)


user/impala/template.py
```diff
@@ -0,0 +1,17 @@
+pkgname = "impala"
+pkgver = "0.2.4"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo-auditable"]
+pkgdesc = "TUI for managing wifi"
```
q66, line 6: a better description would be something like `TUI frontend for iwd`

PR #3539, opened 2025-01-30
user/mupdf: new package

ttyyls:
## Description

this series of commits tries to add the mupdf library, the zathura-backend using
it and makes said backend the default for pdf files.

## NOTES

1. the mupdf source package provides no default pkgconf definition and no default desktop
2. the .so provided by mupdf defaults to the minor.patch out of the base
package's major.minor.patch versionning, which breaks the assumptions of the so
scanning logic in cbuild which thus was stubbed out.
3. different providers don't have to clash but the symlink on them conflicts
hence the -default(s)

## TODO

- [x] ~either drop .desktop or add all icons for visuals in app launchers~ added icons
- [x] ~build against a standalone mujs dep or drop it~ dropped, a release is cut
on important updates to mujs anyway and it won't be useful beyond mupdf itself.
- [x] cleanup the mupdf comments if the workarounds are approved

## Checklist


q66:
why?

q66:
mupdf is not abi-stable and the packaging for it is a big pile of nonsense so if poppler can be used instead i see no reason to use mupdf

PR #3540, opened 2025-01-30
user/sblg: new package

ttyyls:
## Checklist

### DIFF
--- a/user/sblg/template.py
+++ b/user/sblg/template.py
@@ -1,0 +1,20 @@
+pkgname = "sblg"
+pkgver = "0.6.1"
+pkgrel = 0
+build_style = "configure"
+configure_args = ["PREFIX=/usr", "MANDIR=/usr/share/man"]
+make_cmd = "bmake"
+make_check_target = "regress"
+hostmakedepends = ["bmake"]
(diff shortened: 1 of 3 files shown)


q66:
the build does not respect LDFLAGS

user/sblg/template.py
```diff
@@ -0,0 +1,28 @@
...
+maintainer = "ttyyls <contact@behri.org>"
+license = "ISC"
+url = "https://kristaps.bsd.lv/sblg"
+source = f"https://github.com/kristapsdz/sblg/archive/refs/tags/VERSION_{pkgver.replace('.', '_')}.tar.gz"
+sha256 = "f5db8c1fed5276aa90e58eea53c3cbe5e81f123057240191a8fc86a8404627d3"
+hardening = ["vis", "cfi"]
+
+
+def init_build(self):
+    self.configure_env = {
```
q66, line 21: i don't think this is needed, these are unconditionally present in the environment in the first place

ttyyls: leftover from before the configure patch

ttyyls:
patched the configure scripts to use the environment's ldflags and enabled cfi

PR #3546, opened 2025-02-01
Fractal 10

triallax:
works perfectly fine, only thing is that grass rebuilds some stuff on `install` and i'm not sure how to fix that


q66:
maybe just manually install the binary?

PR #3547, opened 2025-02-02
main/man-pages: update to 6.10

triallax:
installs a bunch of commands now (sortman, etc.) that apparently help with maintaining manpages, not sure what we want to do with these

### DIFF
--- a/main/man-pages/template.py
+++ b/main/man-pages/template.py
@@ -1,13 +1,13 @@
 pkgname = "man-pages"
-pkgver = "6.9.1"
-pkgrel = 3
+pkgver = "6.10"
+pkgrel = 0
 hostmakedepends = ["gsed", "bash"]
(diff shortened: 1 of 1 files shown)


q66:
just nuke them, they are useless (also bash and possibly relying on stuff we don't have by default)

main/man-pages/template.py
```diff
@@ -54,6 +54,11 @@ def install(self):
         self.rm("sgetspent_r.3")
         self.rm("ulckpwdf.3")
 
+    # Useless, pull in bash and other stuff we don't want
```
q66, line 57: self.uninstall("usr/bin")

q66: hm actually nevermind that, since we need to uninstall the individual manpages too, this is fine

PR #3570, opened 2025-02-07
user/jj: update to 0.26.0

ttyyls:
## Checklist

### DIFF
--- a/user/jj/patches/git2.patch
+++ b/user/jj/patches/git2.patch
@@ -1,8 +1,41 @@
 diff --git a/Cargo.lock b/Cargo.lock
-index afb7c5a..94bc609 100644
+index 1ab9daef3..70b0911f3 100644
 --- a/Cargo.lock
 +++ b/Cargo.lock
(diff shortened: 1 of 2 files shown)


user/jj/patches/git2.patch
```diff
@@ -1,8 +1,41 @@
 diff --git a/Cargo.lock b/Cargo.lock
-index afb7c5a..94bc609 100644
+index 1ab9daef3..70b0911f3 100644
```
q66, line 2: what's the deal with this patch? it seems to add a ton of seemingly irrelevant stuff

ttyyls: just a git2 update and its dependent Cargo.lock changes

q66: no, it's not, all the stuff added here is useless noise

ttyyls: those are the exact Cargo.lock changes that happen on prepare when only the Cargo.toml definition is changed anyway so i'll drop them

ttyyls: well i was about to anyway...

PR #3600, opened 2025-02-14
user/nix: new package

rietbergenm:
## Description
It took a lot of time to port this one. I have been using this template for over a month now, and I have had no problems. Getting nix up and running is now as easy as running `apk add nix`, adding yourself to the `nix` group and enabling `nix-daemon`.
It is possible to bootstrap nix with the binary tarballs containing a prepackaged nix store from nixos.org, but I think having nix as a package in cports is of added value, because it makes the installation easier and more intergrated with the OS (no broken systemd units, etc.).

### editline & libedit
Please consider having editline in cports. While it mostly does the same as libedit (being a readline compatible line editing library), nix only supports command completion and history when built with editline as opposed to readline (which is an aftertought)  or libedit (via a patch in cports). It would of course be possible to write a patch for this, to make it uses libedit's completion & history.
What I am really asking, is how far do we want to go in patching upstream software to use libedit when they are not fully compatible or losing features.

## Checklist


user/nix/files/nix.profile.d
```diff
@@ -0,0 +1,44 @@
...
+					;;
+			esac
+			;;
+	esac
+
+	# include the system nix profile as well
+	NIX_BIN_DIRS="$NIX_PREFIX/bin:/nix/var/nix/profiles/default/bin"
+	NIX_DATA_DIRS="$NIX_PREFIX/share:/nix/var/nix/profiles/default/share"
+
+	if [ "$PREPEND" == true ]; then
```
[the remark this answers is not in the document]
rietbergenm: I see you are correct! I hadn't noticed it before, because `test(1)` from freebsd (and thus `chimerautils`) allows the use of `==` in addition to  `=`, for compatibility.

I've now updated it to use the correct syntax!

user/nix/files/nix.profile.d
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh
+
+if [ -z "$__NIX_SET_ENV_DONE" ]; then
+	. /etc/default/nix
+
+	# default paths
+	DATA_PATH="${XDG_DATA_DIRS-/usr/local/share:/usr/share}"
+	XDG_STATE_HOME="${XDG_STATE_HOME-$HOME/.local/state}"
+
+	case "$USE_XDG_BASE_DIRS" in
```
folliehiyuki, line 10: The default case below already checks `use-xdg-base-directories` config option, so why do we need this custom USE_XDG_BASE_DIRS variable? Note that `use-xdg-base-directories` is a system option. It has no effect when set via `~/.config/nix/nix.conf`, for example.

user/nix/files/nix.defaults
```diff
@@ -0,0 +1,9 @@
+# Set to true to set the location of your nix profile to ~/.local/state/nix/profile.
+# Set to false to set the location of your nix profile to ~/.nix-profile.
+# Comment out to determine the location based on the settings
+# use-xdg-base-directories in nix.conf.
+#USE_XDG_BASE_DIRS=
+
+# whether to append or prepend nix directories to $PATH
+# When prepending, nix binaries get precedence over traditional tools.
+PREPEND=false
```
folliehiyuki, line 9: Appending doesn't make sense, as the whole point of using Nix is to "override" the installed packages in the system. We should only stick with prepending PATH.

folliehiyuki:
The installation process should create basic directory structure in `/nix` for nix operations to work properly.

Ref: https://github.com/NixOS/nix/blob/1068b9657ff1c4ea629a1a46e07391d16b88ff17/scripts/install-multi-user.sh#L631

user/nix/files/nix.profile.d
folliehiyuki: I remember reading an issue that nix-build doesn't work properly, as users in `_nixbld` group still look for non-xdg directories. I've always set the option globally to avoid unnecessary headache and never changed it, so my assumption here could be wrong.

user/nix/files/nix.defaults
rietbergenm: Personally, I don't use nix like this, I use it to extend the available utilities on my system; when I need to get stuff done and I find out something is not in cports yet, I can now quickly install it with nix and write a template for it later, when I have time. For me, this makes Chimera as a beta distro usable as a daily driver.

I don't use nix to install packages that are also in cports, as I prefer native packages which are tailored to better integrate with the distro. (Why use Chimera if I don't use the package it provides?) Therefore, if I would prepend nix to $PATH and a nix program depends on `coreutils`, or some other tool would get installed that is on the host, I would be using the nix version, which I don't want. It also causes confiusion as all other programs using $PATH will now be using the nix versions of programs, which can confuse packages that were ported to expect chimerautils instead of coreutils (e.g. sed vs gsed incompatibility). This is why appending by default seemed like a good idea to me.

I think there are more people out there using nix like this, so I'd say it is be reasonable to at least provide the option for users to choose whether to prepend or append nix to path

folliehiyuki: When you install a Nix "package", only its output that you choose will be exposed to the profile directory (which can be added to PATH by you - the user). Nix packages always look for runtime dependencies inside /nix/store. Those dependencies won't be exposed to the profile directory by any means. https://github.com/NixOS/patchelf is used extensively for all Nix derivations in nixpkgs to accomplish that.

folliehiyuki:
Can you also add `NIX_REMOTE=daemon` to /etc/profile.d/nix.sh? It's the requirement to allow normal users to use the nix-daemon in multi-user mode, as stated in https://nix.dev/manual/nix/2.25/installation/multi-user.html#running-the-daemon.

user/nix/files/nix.defaults
rietbergenm: Thanks for your reply.
I am aware of this. I was however convinced that this was not the case for dependencies on external command line utilities. After looking in to it some more, it seems that they are also wrapped and not put in your profile, if not installed directly.

Still, if you would install busybox, because you like `udhcpd` nix would install all programs bundled with `busybox` including most of coreutils like `ls`, `cp`, `sed`, etc. Therefore, it would be nice to at least have the _option_ to append nix to $PATH instead of prepending.

I would be fine with changing the default to prepending, but I would like to keep the option to give users the choice (except for writing their own `profile.d`-scripts, which is always an option).

user/nix/files/nix.profile.d
rietbergenm: Ah I see! `nix config check` complains indeed.
Weirdly enough, the setting does take effect however, as installing stuff on a new unprivileged account only member of the `nix` group and with the setting set in `~/.conf/nix/nix.conf`, all nix operations put stuff in `~/.local/state/nix`. The old non-xdg locations remain unpopulated. Therefore the output of `nix config show use-xdg-base-directories` seems to be consistent with the location nix programs use:

```
$ doas useradd dummy
$ doas passwd dummy
$ doas usermod -aG nix dummy
$ su dummy -
$ cat /etc/nix/nix.conf
allowed-users = @nix
build-users-group = _nixbld
extra-experimental-features = nix-command
$ mkdir -p ~/.config/nix
$ echo "use-xdg-base-directories = true" > ~/.config/nix/nix.conf
$ nix config show use-xdg-base-directories
true
$ nix config check
[PASS] PATH contains only one nix version.
[PASS] All profiles are gcroots.
warning: ignoring the client-specified setting 'use-xdg-base-directories', because it is a restricted setting and you are not a trusted user
[PASS] Client protocol matches store protocol.
[INFO] You are not trusted by store uri: daemon
$ nix-channel --add https://nixos.org/channels/nixpkgs-unstable
$ nix-channel --update
warning: ignoring the client-specified setting 'use-xdg-base-directories', because it is a restricted setting and you are not a trusted user
unpacking 1 channels...
warning: ignoring the client-specified setting 'use-xdg-base-directories', because it is a restricted setting and you are not a trusted user
$ nix-env -iA nixpkgs.figlet
warning: ignoring the client-specified setting 'use-xdg-base-directories', because it is a restricted setting and you are not a trusted user
installing 'figlet-2.2.5'
$ tree -a
.
├── .cache
│   └── nix
│       ├── fetcher-cache-v3.sqlite
│       └── fetcher-cache-v3.sqlite-journal
├── .config
│   ├── dinit.d
│   │   └── boot.d
│   └── nix
│       └── nix.conf
├── .local
│   ├── share
│   │   └── flatpak
│   │       └── db
│   └── state
│       ├── lesshst
│       ├── nix
│       │   ├── channels
│       │   ├── defexpr
│       │   │   ├── channels -> /home/dummy/.local/state/nix/profiles/channels
│       │   │   └── channels_root -> /nix/var/nix/profiles/per-user/root/channels
│       │   ├── profile -> /home/dummy/.local/state/nix/profiles/profile
│       │   └── profiles
│       │       ├── channels -> channels-1-link
│       │       ├── channels-1-link -> /nix/store/vyjj39vi7qmyx5bkb9hzxff1rklirfwa-user-environment
│       │       ├── profile -> profile-3-link
│       │       ├── profile-1-link -> /nix/store/bd2ric5lxmk0p9mzjj9iwfqybyg9w0dg-user-environment
│       │       ├── profile-2-link -> /nix/store/57904rflsfh46y7fxi62rzc7sidc9vmq-user-environment
│       │       └── profile-3-link -> /nix/store/bd2ric5lxmk0p9mzjj9iwfqybyg9w0dg-user-environment
│       └── wireplumber
│           └── stream-properties
└── .sh_history
````

It is weird though that `nix config check` complains when it clearly works... Also, all operations involving the nix-daemon, print warnings about the client setting being ignored, but then the nix programs that run as the user (so not nix-daemon) seem to honor the setting and therefore put the nix profile in the right location.

Concluding, I think the output of `nix config show` can be used as it corresponds with the place your nix profile ends up. It does however seem like something is not quite right here, but that would be an upstream issue with nix.

I agree that the setting should best be set in the global config file, but parsing that specifcally and not using `nix config show`  seems like more work than it is worth to me.

**Comment by rietbergenm** [comment]
> Can you also add `NIX_REMOTE=daemon` to /etc/profile.d/nix.sh? It's the requirement to allow normal users to use the nix-daemon in multi-user mode, as stated in https://nix.dev/manual/nix/2.25/installation/multi-user.html#running-the-daemon.

This is a good point.
Before, we were relying on the default value of `store` in **nix.conf(5)** (which `NIX_DAEMON` overwrites). The default value is [auto](https://nix.dev/manual/nix/2.25/store/types/#store-url-format), which determines what type of store we are dealing with based on a few factors. Now, in most cases this will work fine, but setting it explicitly would be better, as we assume we are dealing with a local daemon store.

I don't like having unnecessary variables in my environment, so I would rather set this via the equivalent `store = daemon` in nix.conf.

I'll push this in a moment.

**Comment by rietbergenm** [comment]
updated mdbook & toml11 (both had an update in the last 2 days)

File: user/nix/files/nix.defaults (hunk 3 above)
Comment (folliehiyuki) [reply user/nix/files/nix.defaults]: Sounds okay to me. You should keep it as appending by default, as it's your (the package's maintainer) use case. It's configurable so I can always change the value.

File: user/nix/files/nix.profile.d [hunk 4]
```diff
@@ -0,0 +1,37 @@
...
+
+if [ -z "$__NIX_SET_ENV_DONE" ]; then
+	. /etc/default/nix
+
+	# default paths
+	DATA_PATH="${XDG_DATA_DIRS-/usr/local/share:/usr/share}"
+	XDG_STATE_HOME="${XDG_STATE_HOME-$HOME/.local/state}"
+
+	# use nix command to parse settings so we don't have to
+	case `nix --extra-experimental-features nix-command config show use-xdg-base-directories` in
```
folliehiyuki, line 11: This is just nit picking. Nix sometimes unexpectedly goes to the Internet to update substitutes. It doesn't hurt being extra cautious here.

```suggestion
	case "$(nix --offline --experimental-features nix-command config show use-xdg-base-directories)" in
```

user/nix/files/nix.profile.d
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
+if [ -z "$__NIX_SET_ENV_DONE" ]; then
+	. /etc/default/nix
+
+	# default paths
+	DATA_PATH="${XDG_DATA_DIRS-/usr/local/share:/usr/share}"
+	XDG_STATE_HOME="${XDG_STATE_HOME-$HOME/.local/state}"
```
folliehiyuki, line 8: Another nit pick. Shouldn't the expansion usually be `:-` instead of `-`, so the variable is checked for both its existence and whether its value is null?

rietbergenm:
This is ready for review btw, do it at your convenience

user/editline/template.py
```diff
@@ -0,0 +1,24 @@
+pkgname = "editline"
+pkgver = "1.17.1"
+pkgrel = 0
+build_style = "configure"
```
q66, line 4: this should be gnu_configure and then you can drop the whole thing with cross, configure_args and other nonsense

user/mdbook/template.py
```diff
@@ -0,0 +1,15 @@
+pkgname = "mdbook"
```
q66, line 1: is this needed for the manpages or what

if not, then don't package this and skip docs, we don't usually package docs

user/nix/files/nix-daemon.dinit
```diff
@@ -0,0 +1,4 @@
+type = process
```
q66, line 1: assuming this continuously logs any realistic amount of stuff buffer is not a way to go

also should respect correct syntax for `=` vs `:`

user/nix/files/nix-daemon.sysusers.conf
```diff
@@ -0,0 +1,68 @@
+# system users used by nix-daemon for process isolation
+# It needs one user per build, so more users means more simultaneous builds.
+g _nixbld
+u _nixbld1  -:_nixbld "nix build user" /var/empty
```
q66, line 4: this looks really bad

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
+pkgname = "nix"
+pkgver = "2.26.2"
+pkgrel = 0
+build_style = "meson"
+configure_args = [
+    # Use /usr/lib instead of /usr/libexec
```
q66, line 6: drop all this needless verbosity

this is just

```
# add 2 and 2 together
2 + 2
```

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+    "--libexecdir=/usr/lib",
+    # /nix/var is the default, but cbuild passes --localstatedir=/var.
```
q66, line 8: like all of this is useless noise

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+    # a.k.a. nix only supports shared libraries.
+    "-Ddefault_library=shared",
+    # We want man pages.
+    "-Ddoc-gen=true",
+    # We don't need the perl bindings (and don't have the dependencies to build it).
+    "-Dbindings=false",
+    # We don't do test as it requires network, so don't build them.
+    "-Dunit-tests=false",
+]
+hostmakedepends = [
```
q66, line 33: sort

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+    "doxygen",
+    "mdbook",
+    # Preprocessor for mdbook that checks for broken links
+    # The docs include a lot of things, but we only package the man pages.
+    # man pages don't support links, so we don't care if they are dead or not.
+    # "mdbook-linkcheck",
+    "rsync",
+    "jq",
+]
+makedepends = [
```
q66, line 50: sort

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+    "brotli-devel",
+    "nlohmann-json",
+    "curl-devel",
+    "libseccomp-devel",
+    "sqlite-devel",
+    "gc-devel",
+    "lowdown-devel",
+    "libgit2-devel",
+    "toml11",
+    # Use editline as nix recommends, but readline or libedit also work.
```
q66, line 64: noise

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+    "editline-devel",
+]
+# We don't actually use these at the moment, because checks are disabled.
```
q66, line 67: noise

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+    "rapidcheck",  # not in cports yet
+    "gtest-devel",
+]
+pkgdesc = "Purely functional package manager"
+maintainer = "Mathijs Rietbergen <mathijs.rietbergen@proton.me>"
+license = "LGPL-2.1-or-later"
+url = "https://github.com/NixOS/nix"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "283ac42ec602e335ffd9c42721f35f1a20e8a7eb629ec5d9cc604a2fc2489ec0"
+# With `int` enabled, all nix tools constantly fail with "Illegal instruction".
```
q66, line 78: too verbose

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+hardening = ["!int"]
+# Checks require network to download from nixpkgs, but network is not available
```
q66, line 80: too verbose

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+# in check phase.
+# Cross compiling fails as meson cannot find CMake when cross compiling.
+# This seems like a generic issue, unrelated to nix...
+options = ["!check", "!cross"]
+
+
+def post_install(self):
+    # Remove systemd files.
```
q66, line 88: ditto

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+    self.uninstall("usr/lib/systemd")
+
+    # Creating /nix/var/nix/daemon-socket/socket is unneeded as nix-daemon can
```
q66, line 91: verbose

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+    # do it on its own, but the systemd unit files require the socket to be
+    # present before starting (why? I have no clue).
+    # That is why it exists upstream. We don't need this by any means.
+    self.uninstall("usr/lib/tmpfiles.d")
+
+    # We have our own, so we don't use these.
```
q66, line 97: you put all this verbosity in the comments but then don't state important stuff like *why* custom profile is shipped

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+    # Install the files that set up the environment, service and config.
+    self.install_sysusers(
+        self.files_path / "nix-daemon.sysusers.conf", name="nix-daemon"
+    )
+    self.install_sysusers(self.files_path / "nix.sysusers.conf", name="nix")
+    self.install_service(
+        self.files_path / "nix-daemon.dinit", name="nix-daemon"
+    )
+    self.install_file(self.files_path / "nix.conf", "etc/nix")
+    self.install_file(
```
q66, line 109: profile.d files go in `/usr/lib/profile.d`

we also don't install `/etc` files in general if possible

user/toml11/template.py
```diff
@@ -0,0 +1,21 @@
...
+pkgver = "4.4.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DTOML11_BUILD_TESTS=ON",
+    "-DTOML11_BUILD_TOML_TESTS=ON",
+]
+hostmakedepends = ["cmake", "ninja"]
+makedepends = ["doctest", "nlohmann-json"]
+pkgdesc = "Feature-rich TOML language library for C++11/14/17/20"
```
q66, line 11: TOML library for C++

user/toml11/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "toml11"
+pkgver = "4.4.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DTOML11_BUILD_TESTS=ON",
+    "-DTOML11_BUILD_TOML_TESTS=ON",
+]
+hostmakedepends = ["cmake", "ninja"]
+makedepends = ["doctest", "nlohmann-json"]
```
q66, line 10: afaict doctest is checkdepends

user/toml11/template.py
```diff
@@ -0,0 +1,21 @@
...
+pkgdesc = "Feature-rich TOML language library for C++11/14/17/20"
+maintainer = "Mathijs Rietbergen <mathijs.rietbergen@proton.me>"
+license = "MIT"
+url = "https://github.com/ToruNiina/toml11"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "815bfe6792aa11a13a133b86e7f0f45edc5d71eb78f5fb6686c49c7f792b9049"
+hardening = ["!int"]  # when enabled, test_parse_integer fails...
```
q66, line 17: put the comment above

user/nix/template.py
```diff
@@ -0,0 +1,126 @@
...
+    # /nix/var is the default, but cbuild passes --localstatedir=/var.
+    # usually this is correct, but we need nix stuff to be in /nix/var.
+    "--localstatedir=/nix/var/",
+    # profile-dir defaults to etc/profile.d which becomes
+    # /usr/etc/profile.d with the prefix /usr. This is wrong,
+    # we want these system files to go in /etc/profile.d.
+    # Note that we now remove them anyways, because we have our own,
+    # but it still felt cleaner to leave it like this.
+    "-Dnix:profile-dir=/etc/profile.d",
+    # Upstream nix prefers editline. Readline is an afterthought, so completion
```
q66, line 17: if it doesn't use libedit then don't add a patch for it

user/nix/files/nix.sysusers.conf
```diff
@@ -0,0 +1,2 @@
+# allow these users to use nix in a multi-user install
+g nix
```
q66, line 2: we don't do system users/groups without underscore

q66:
there is probably more, i'll comment that once you've addressed all this otherwise it becomes hard to keep track of for me and probably for you as well

user/nix/files/nix-daemon.sysusers.conf
q66: (i don't know what to actually do about it but i suspect it will need properly looking into)

user/mdbook/template.py
rietbergenm: it is used to build the man pages

user/nix/files/nix-daemon.sysusers.conf
rietbergenm: Every distro I used, seems to just add 32 nixbld users... There is an experimental feature `auto-allocate-uids`  (it has been around  since at least 2020), but I don't know how stable it is, as the main `nix` command and flakes are also "experimental features".

This feature would allow nix to allocate its own uids to use for building as seperate users, instead of using the nixbld* groups.

see <https://nix.dev/manual/nix/2.28/command-ref/conf-file#conf-auto-allocate-uids>

rietbergenm:
Sorry for the delay, I've been quite busy lately.

Most of the issues should be resolved now (except for the build user problem; don't really know what to do about this).

Regarding maintaining this: I am happy to maintain this package and its dependencies, but nix's upstream release cycle is quite aggressive; a new release every few weeks with patch releases in between is a bit much for me.

Now I think it would be best to package the latest _stable_ nix release (aka the one in use by NixOS). This way we have a more stable and better supported version of  nix that receives bug fixes and still has a new feature release every 6 months.
Any opinions on this?

ItsHarper:
> Now I think it would be best to package the latest _stable_ nix release

I'm nobody, but I'd love to see this land, and only packaging stable releases to keep the maintenance burden low definitely makes sense.

ItsHarper:
Definitely makes sense, Nix really seems to want to do its own thing. At some point I'll probably cherry-pick this into my `cports` repo.

ItsHarper:
> image


Actually, [lowdown is used to generate the manpages, not mdbook](https://github.com/NixOS/nix/blob/5caebab63ae54295a0b8a3ef2900d1adcf2e8512/doc/manual/source/development/documentation.md#markdown-conventions). Even without the documentation, we can know that with certainty, because `mdbook` doesn't have a manpage generator built-in (you have to use `mdbook-man`).

PR #3622, opened 2025-02-18
user/clight: new package

tribals:
## Description

Add [`clight`](https://github.com/FedeDP/Clight) user-session daemon package & dependencies. It controls screen brightness, gamma & keyboard backlight in consistent way, across any DE, WM, or even in framebuffer terminal.

## Checklist

- [x] I will take responsibility for my template and keep it up to date (until I'll stop using Chimera)

See also:

- https://github.com/FedeDP/Clight/issues/312

### DIFF
--- a/user/clight/template.py
+++ b/user/clight/template.py
@@ -1,0 +1,36 @@
+pkgname = "clight"
+pkgver = "4.11"
+pkgrel = 0
+build_style = "cmake"
+hostmakedepends = [
+    "cmake",
+    "dbus-devel",
+    "elogind-devel",
+    "gsl-devel",
+    "libconfig-devel",
+    "libmodule-devel",
+    "linux-headers",
+    "ninja",
+    "pkgconf",
+    "popt-devel",
+]
+makedepends = [
+    "bash-completion",
+    "fish-shell",
+]
+depends = [
+    "clightd",
+]
+install_if = [
+    "clightd>=5",
+]
+pkgdesc = "User daemon to adjust screen backlight based on ambient brightness"
+maintainer = "Anthony <w732qq@gmail.com>"
+license = "GPL-3.0-or-later"
+url = "https://github.com/FedeDP/Clight"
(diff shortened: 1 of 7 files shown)


user/clight/template.py
```diff
@@ -0,0 +1,36 @@
...
+    "elogind-devel",
+    "gsl-devel",
+    "libconfig-devel",
+    "libmodule-devel",
+    "linux-headers",
+    "ninja",
+    "pkgconf",
+    "popt-devel",
+]
+makedepends = [
```
q66, line 17: small lists like this can and should be one line

user/clight/template.py
```diff
@@ -0,0 +1,36 @@
+pkgname = "clight"
+pkgver = "4.11"
+pkgrel = 0
+build_style = "cmake"
+hostmakedepends = [
```
q66, line 5: but also like half of these are actually makedepends

user/clight/template.py
tribals: I did run `ruff format` (on whole repo) - whatever it did is here, on PR.

user/clight/template.py
```diff
@@ -0,0 +1,36 @@
...
+    "popt-devel",
+]
+makedepends = [
+    "bash-completion",
+    "fish-shell",
+]
+depends = [
+    "clightd",
+]
+install_if = [
```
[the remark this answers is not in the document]
tribals: There is no point to use `clight` session daemon without `clightd` system daemon. It will not work at all.

user/clight/template.py
tribals: I didn't grasp the difference between those yet. Hope you can explain it.

user/clight/template.py
tribals: Didn't understood which list do you mean - makedepends or hostmakedepends? Please make a comment at least two or three lines *below* the line you're commenting. In any case, that's just more readable as it follows generic template.

PR #3637, opened 2025-02-20
user/poweralertd: move from main, add user service

ttyyls:
## Checklist

### DIFF
--- a/user/poweralertd/files/poweralertd.user
+++ b/user/poweralertd/files/poweralertd.user
@@ -1,0 +1,4 @@
+type = process
+command = /usr/bin/poweralertd
+log-type = buffer
+depends-on: dbus
--- a/user/poweralertd/patches/tangle.patch
+++ b/user/poweralertd/patches/tangle.patch
@@ -1,0 +1,31 @@
+diff --git a/dbus.h b/dbus.h
+index e9a1fbe..839d087 100644
+--- a/dbus.h
++++ b/dbus.h
+@@ -9,4 +9,8 @@
+ #include <systemd/sd-bus.h>
+ #endif
+ 
++#if defined(HAVE_TANGLE)
++#include <tangle/sd-bus.h>
++#endif
++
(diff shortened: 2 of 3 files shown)


main/poweralertd/template.py
```diff
@@ -1,6 +1,6 @@
 pkgname = "poweralertd"
```
q66, line 1: probably switch it from elogind to tangle

and also move it to user

main/poweralertd/files/poweralertd.user
```diff
@@ -0,0 +1,5 @@
+type = process
+command = /usr/bin/poweralertd
+depends-on = dbus
```
q66, line 3: service syntax needs modernizing

ttyyls: in what sense?

q66: https://davmac.org/projects/dinit/man-pages-html/dinit-service.5.html

the canonical syntax for dependencies is `depends-on: ...`, not `depends-on = ...` and so on

user/poweralertd/files/poweralertd.user
```diff
@@ -0,0 +1,5 @@
+type = process
+command = /usr/bin/poweralertd
+depends-on: dbus
+depends-on: graphical.target
+log-type = buffer
```
q66, line 5: anything with = should go before things with :

user/poweralertd/files/poweralertd.user
```diff
@@ -0,0 +1,5 @@
+type = process
+command = /usr/bin/poweralertd
+depends-on: dbus
+depends-on: graphical.target
```
q66, line 4: is that needed

ttyyls: i assumed so since it emits notifications, but it works fine without it by my own testing so i dropped it

PR #3640, opened 2025-02-20
user/tlstunnel: new package

apprehensions:
## Description

A TLS reverse proxy.

## Checklist

#### Notes
* Should Makefile be preferred for simple installation? (eg. just copy and generate documentation from scdoc)
* For services, should there be a default configuration example in /etc?

### DIFF
--- a/user/tlstunnel/files/tmpfiles.conf
+++ b/user/tlstunnel/files/tmpfiles.conf
@@ -1,0 +1,4 @@
+# Create tlstunnel data & configuration directory
+
+d /var/lib/tlstunnel 0750 _tlstunnel _tlstunnel -
+d /etc/tlstunnel 0755 root root -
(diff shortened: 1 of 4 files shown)


q66:
this can use the go build style

user/tlstunnel/files/tmpfiles.conf
```diff
@@ -0,0 +1,3 @@
+# Create tlstunnel data directory
+
+d /var/lib/tlstunnel 0750 _tlstunnel _tlstunnel -
```
q66, line 3: the tmpfiles should also create `/etc/tlstunnel`

q66:
also codeberg is the canonical location

user/tlstunnel/files/tmpfiles.conf
apprehensions: as a convenience for the user to easily add a configuration file? okay.

PR #3642, opened 2025-02-20
user/mg: new package

haesbaert:
## Description

mg is the micro emacs clone.

There's an integer overflow that triggers on carriage return, I've disabled the overflow checks for now.
Been using this software for decades.
When I find some time I'll try to fix it upstream.

## Checklist

### DIFF
--- a/user/mg/template.py
+++ b/user/mg/template.py
@@ -1,0 +1,15 @@
+pkgname = "mg"
+pkgver = "3.7"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_gen = []
+make_dir = "."
+make_build_args = ["V=1"]
+makedepends = ["ncurses-devel"]
+pkgdesc = "Micro (GNU) Emacs-like text editor"
+maintainer = "Christiano Haesbaert <haesbaert@haesbaert.org>"
+license = "custom:none"
+url = "https://github.com/troglobit/mg"
+source = f"{url}/releases/download/v{pkgver}/mg-{pkgver}.tar.gz"
+sha256 = "05101360d2194392da0950e8b6f18d067d8af0fd2f572461ba4d4e7b4ccbb4c1"
+hardening = ["vis", "cfi", "!int"]


user/mg/template.py
```diff
@@ -0,0 +1,15 @@
...
+make_dir = "."
+make_build_args = ["V=1"]
+makedepends = ["ncurses-devel"]
+pkgdesc = "Micro (GNU) Emacs-like text editor"
+maintainer = "Christiano Haesbaert <haesbaert@haesbaert.org>"
+license = "custom:none"
+url = "https://github.com/troglobit/mg"
+source = f"{url}/releases/download/v{pkgver}/mg-{pkgver}.tar.gz"
+sha256 = "05101360d2194392da0950e8b6f18d067d8af0fd2f572461ba4d4e7b4ccbb4c1"
+hardening = ["vis", "cfi", "!int"]
```
q66, line 15: this should be easy to figure out, just add `sanruntime` into `options` and reproduce the overflow and it should tell you the exact line number and reason

user/mg/template.py
```diff
@@ -0,0 +1,15 @@
...
+pkgver = "3.7"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_gen = []
+make_dir = "."
+make_build_args = ["V=1"]
+makedepends = ["ncurses-devel"]
+pkgdesc = "Micro (GNU) Emacs-like text editor"
+maintainer = "Christiano Haesbaert <haesbaert@haesbaert.org>"
+license = "custom:none"
```
q66, line 11: this is actually Unlicense, which has an spdx name

user/mg/template.py
```diff
@@ -0,0 +1,15 @@
+pkgname = "mg"
+pkgver = "3.7"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_gen = []
+make_dir = "."
+make_build_args = ["V=1"]
```
q66, line 7: we don't usually do this

druimalban:
There are a number of competing portable versions of this. Most of the big distributions (e.g. Fedora, Debian, Alpine) use hboetes' version: https://github.com/hboetes/mg. So does NetBSD. Macports uses ibara's version: https://github.com/ibara/mg. I am not sure what FreeBSD uses.

Would it be worth clarifying the provenance of the proposed version?

q66:
this version last tagged in 2023 and the other versions seem to be preferred, we're probably better off following suit

PR #3649, opened 2025-02-22
user/firejail: new package

hbog:
## Description

Firejail is a SUID sandbox program that reduces the risk of security breaches by using Linux namespaces, seccomp-bpf and Linux capabilities.

## Checklist

### DIFF
--- a/user/firejail/files/tmpfiles.conf
+++ b/user/firejail/files/tmpfiles.conf
@@ -1,0 +1,3 @@
+# Populate /etc/firejail with default application security profiles
+
+C /etc/firejail - - - - /usr/share/firejail
--- a/user/firejail/template.py
+++ b/user/firejail/template.py
@@ -1,0 +1,27 @@
+pkgname = "firejail"
+pkgver = "0.9.72_git20250221"
+pkgrel = 0
+_commit = "733f9a9c487358ba73b06186ed348f086d8ac839"
+build_style = "gnu_configure"
+configure_args = ["--sysconfdir=/usr/share"]
+configure_gen = []
+make_dir = "."
+make_check_target = "test"
+makedepends = [
+    "gawk",
+    "base-kernel-devel"
+]
+pkgdesc = "Firejail SUID sandboxing tool"
+maintainer = "hbog <herwig.bogaert@posteo.net>"
+license = "GPL-2.0-only"
+url = "https://github.com/netblue30/firejail"
+source = f"{url}/archive/{_commit}.zip"
+sha256 = "2a05497ac0dc2053b5e3aca2ddf5c32ab5176f60be102498bc849b3042225bb3"
+file_modes = {
+    "usr/bin/firejail": ("root", "root", 0o4755)
+}
+hardening = ["vis", "cfi"]
+
+
+def post_install(self):
+    self.install_tmpfiles(self.files_path / "tmpfiles.conf")


user/firejail/template.py
```diff
@@ -0,0 +1,27 @@
...
+pkgver = "0.9.72_git20250221"
+pkgrel = 0
+_commit = "733f9a9c487358ba73b06186ed348f086d8ac839"
+build_style = "gnu_configure"
+configure_args = ["--sysconfdir=/usr/share"]
+configure_gen = []
+make_dir = "."
+make_check_target = "test"
+makedepends = [
+    "gawk",
```
q66, line 11: gawk is almost surely hostmakedepends, and what's the base-kernel-devel for?

hbog: base-kernel-devel pulled in linux-headers. Build needs `seccomp.h` (required) and `landlock.h` (optional for landlock support).  scaled this dependency down to linux-headers.  The tests need bash (which was pulled in via base-kernel-devel). Added this as `checkdepends`.

q66:
please follow the commit message requirements properly

don't add new commits, use amend feature or squash afterwards

user/firejail/files/tmpfiles.conf
```diff
@@ -0,0 +1,3 @@
+# Populate /etc/firejail with default application security profiles
+
+C /etc/firejail - - - - /usr/share/firejail
```
q66, line 3: this is probably bad because it will never get update once populated once

user/firejail/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "firejail"
+pkgver = "0.9.72_git20250221"
+pkgrel = 0
+_commit = "733f9a9c487358ba73b06186ed348f086d8ac839"
+build_style = "gnu_configure"
+configure_args = ["--sysconfdir=/usr/share"]
```
q66, line 6: this looks wrong

user/firejail/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "firejail"
+pkgver = "0.9.72_git20250221"
+pkgrel = 0
+_commit = "733f9a9c487358ba73b06186ed348f086d8ac839"
```
q66, line 4: we don't package git software

user/firejail/template.py
```diff
@@ -0,0 +1,23 @@
...
+build_style = "gnu_configure"
+configure_args = ["--sysconfdir=/usr/share"]
+configure_gen = []
+make_dir = "."
+make_check_target = "test"
+hostmakedepends = ["gawk", "linux-headers"]
```
q66, line 10: linux-headers is makedepends

user/firejail/template.py
```diff
@@ -0,0 +1,23 @@
+pkgname = "firejail"
+pkgver = "0.9.72_git20250221"
+pkgrel = 0
+_commit = "733f9a9c487358ba73b06186ed348f086d8ac839"
+build_style = "gnu_configure"
+configure_args = ["--sysconfdir=/usr/share"]
+configure_gen = []
```
q66, line 7: should say why

user/firejail/template.py
```diff
@@ -0,0 +1,23 @@
...
+make_dir = "."
+make_check_target = "test"
+hostmakedepends = ["gawk", "linux-headers"]
+checkdepends = ["bash"]
+pkgdesc = "Firejail SUID sandboxing tool"
```
q66, line 12: just `SUID sandboxing tool`

user/firejail/files/tmpfiles.conf
hbog: Yes you're right. I looked further into this and it is even worse. Built as I did, Firejail will never look at these files in /etc/. There is no need for /etc/firejail because profiles can be overridden in ~/.config/firejail/.   I will remove this. It makes no sense.

user/firejail/template.py
hbog: By default, the profiles are put in /etc/profiles, but I understood that the package manager should not own files in /etc/. Please advice on where to put them (/usr/lib/firejail/profiles?).

user/firejail/template.py
hbog: The latest release/tag dates from 2023 but the repo is active. Please advise.

user/firejail/template.py
hbog: check!

user/firejail/template.py
hbog: check!

user/firejail/files/tmpfiles.conf
hbog: There seems to be no simple and elegant way to put the profiles in another directory than /etc/firejail (and thus avoiding that the  package owns files in /etc).

user/firejail/template.py
hbog: In the mean while, new versions hev been tagged with versions number by the author. There is no need anymore to select a non-version-tagged commit.

user/firejail/template.py
hbog: There seems to be no simple and elegant way to put the profiles in another directory than /etc/firejail (and thus avoiding that the  package owns files in /etc).  I left the in /etc.

user/firejail/template.py
hbog: There seems to be no simple and elegant way to put the profiles in another directory than /etc/firejail (and thus avoiding that the  package owns files in /etc).

PR #3661, opened 2025-02-25
user/emacs-*: move from main, adopt, update to 30.1

RoastVeg:
## Description

Native compilation is now enabled by default as of 30.1.

Clang is still used for compilation of emacs itself, libgccjit is only used for native compilation of emacs lisp code at both build and run time.

## Checklist


q66:
we shouldn't really be building "normal" gcc with `--enable-host-shared` so i don't think this is the way to go

q66:
my suggestion is "just update it and don't bother with gccjit or anything else at all for now"

RoastVeg:
> my suggestion is "just update it and don't bother with gccjit or anything else at all for now"

I won't be adopting this without native compilation, as I wouldn't be using it personally. I can either build libgccjit as a separate template or close this PR, your choice.

PR #3662, opened 2025-02-25
main/gtk+3: rename update-icon-cache trigger

shawnbon206:
## Description

I noticed after installing packages the icon was not showing up in Gnome Apps Menu, turns out the cache was not updating when adding or removing packages.

I poked around in cports and found the problem, the trigger did not get renamed when some time ago the package got renamed.

## Checklist


q66:
ah, good catch...

PR #3667, opened 2025-02-26
postgresql15: new package

apprehensions:
## Description

Sophisticated object-relational DBMS, version 15.x

## Checklist

Thanks to the work of mia.

Changing from PostgresQL15 to 16 is untested, but still unnecessary for software that relies on PostgresQL15.

### DIFF
--- a/user/postgresql-postgresql15-default
+++ b/user/postgresql-postgresql15-default
@@ -1,0 +1,1 @@
+postgresql15
--- a/user/postgresql-postgresql15-oid2name-default
+++ b/user/postgresql-postgresql15-oid2name-default
@@ -1,0 +1,1 @@
+postgresql15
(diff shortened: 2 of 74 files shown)


q66:
i don't see the point, postgresql16 was the first version added, if you're doing a new deployment you should use the newest version, if you're migrating something to chimera you can afford to upgrade the database

there is *no* software in the repository that needs postgresql 15

apprehensions:
if you'd like to reject this PR in favor of waiting some weeks (feb 2025 is expected for scaleway) for the server and scaleway to support postgresql16, then that's fine.

q66:
i don't think i'm very interested in adding new software that only supports postgres that's two major versions behind either, and i don't think what scaleway supports is very important for the project, and given how complicated postgres packaging is, yeah i don't think this is worth it

PR #3668, opened 2025-02-26
user/soju: new package

sewnie:
## Description

User-friendly IRC bouncer

## Checklist


q66:
this can be made to use the regular go build style, also service should follow modern style and respect `:` vs `=`

q66:
also it will need reworking to not place a packaged configfile in /etc by default, not yet sure how

q66:
probably a wrapper script should be used for the service that does something like

```
if [ -r /etc/soju/config ]; then
    ... run soju normally ...
else
   ... run soju with -config pointed to a system default ...
fi
```

user/soju/template.py
```diff
@@ -0,0 +1,41 @@
+pkgname = "soju"
+pkgver = "0.8.2"
+_commit = "5ee7ed56f7ed4ad0ccba13d944b5610fa46de352"
+pkgrel = 0
+build_style = "makefile"
+make_build_args = ["GOFLAGS=-buildmode=pie -tags=libsqlite3"]
+hostmakedepends = ["go", "scdoc", "libcap-progs"]
+makedepends = ["sqlite-devel"]
+pkgdesc = "User-friendly IRC bouncer"
```
q66, line 9: we don't use subjective adjectives for pkgdescs, should be `IRC bouncer`

`

apprehensions: Yes you do.

https://github.com/chimera-linux/cports/tree/master/user/fd#L11
> `pkgdesc = "Simple, fast and user-friendly alternative to find"`

https://github.com/chimera-linux/cports/blob/master/user/pyonji/template.py#L6
> `pkgdesc = "Easy-to-use tool to send e-mail patches"`

https://github.com/chimera-linux/cports/blob/master/user/python-pyusb/template.py#L11
> `pkgdesc = "Easy USB access for Python"`

apprehensions:
> also it will need reworking to not place a packaged configfile in /etc by default, not yet sure how

why?

apprehensions:
added snippets from [ttyyls](https://github.com/ttyyls)'s soju PR

breakgimme:
i have a question - why is a commit used rather than just the release archive? it seems to be way later than 0.8.2

ayushnix:
We can probably add `capabilities = ^cap_net_bind_service` to the service file and skip `file_xattrs` and `file_modes`?

user/soju/files/tmpfiles.conf
```diff
@@ -0,0 +1,6 @@
+# Create soju data, runtime, configuration directories
+
+d /var/lib/soju 0750 _soju _soju -
+d /var/lib/soju/logs 0750 _soju _soju -
+d /run/soju 0750 _soju _soju -
```
q66, line 5: /run is an ephemeral directory and tmpfiles is a poor mechanism for handling those

this should be prepared by a service (dependency of `soju`)

the others are on the filesystem and are fine

user/soju/template.py
```diff
@@ -0,0 +1,54 @@
...
+
+    self.install_man("doc/soju.1")
+    self.install_man("doc/sojuctl.1")
+    self.install_tmpfiles(self.files_path / "tmpfiles.conf")
+    self.install_sysusers(self.files_path / "sysusers.conf")
+    self.install_service(self.files_path / "soju")
+    self.install_license("LICENSE")
+
+
+@subpackage("soju-utils")
```
q66, line 48: i fail to see the purpose of having a subpackage for this

PR #3669, opened 2025-02-26
font-iosevka-ttf: new package

sewnie:
## Description

Versatile typeface for code, from code.

## Checklist

### DIFF
--- a/user/font-iosevka-aile-ttf
+++ b/user/font-iosevka-aile-ttf
@@ -1,0 +1,1 @@
+font-iosevka-ttf
--- a/user/font-iosevka-curly-slab-ttf
+++ b/user/font-iosevka-curly-slab-ttf
@@ -1,0 +1,1 @@
+font-iosevka-ttf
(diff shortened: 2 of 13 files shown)


q66:
i dunno what to make of this

a few things:

1) the package names don't match convention, they should have the `-ttf` suffixes
2) the main package should not be empty, likely it should contain the default variant
3) the fixed variants aren't packaged?
4) don't get too clever with the sources appending shenanigans, just expand it
5) the subpackage names are all over the place; why not follow the upstream naming

apprehensions:
> the package names don't match convention, they should have the -ttf suffixes

nerd fonts do not have -ttf suffixes

> the fixed variants aren't packaged?

intentional, terminal will suffice

> don't get too clever with the sources appending shenanigans, just expand it

not sure how or what you mean by "just expand it", you want me to write the literal package names?

> the subpackage names are all over the place; why not follow the upstream naming

ok then, the subpackage names will be generated from the list of upstream variants

sewnie:
waiting a year here reminds me of how long i wait for void packages

q66:
i'd prefer you not making derisive comments while you see 1) that i'm going through things 2) how much i have piled up

that's just mean and besides you have no entitlement to my attention

PR #3681, opened 2025-02-28
user/dust: update to 1.1.2

ttyyls:
## Checklist

### DIFF
--- a/user/dust/template.py
+++ b/user/dust/template.py
@@ -1,6 +1,6 @@
 pkgname = "dust"
-pkgver = "1.1.1"
-pkgrel = 1
+pkgver = "1.1.2"
+pkgrel = 0
 build_style = "cargo"
 hostmakedepends = ["cargo-auditable"]
 makedepends = ["rust-std"]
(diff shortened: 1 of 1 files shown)


user/dust/template.py
```diff
@@ -1,14 +1,14 @@
...
 makedepends = ["rust-std"]
 pkgdesc = "Simplified du -h"
 license = "Apache-2.0"
 url = "https://github.com/bootandy/dust"
 source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
-sha256 = "98cae3e4b32514e51fcc1ed07fdbe6929d4b80942925348cc6e57b308d9c4cb0"
+sha256 = "31da99483ee6110d43ed5e7c56a59f40f33b389e45d09d91fca022b42d442040"
 
 
 if self.profile().wordsize == 32:
```
triallax, line 14: can be dropped now, https://github.com/bootandy/dust/commit/489d9ada440d61a780fa1a657f6bf7515a373f4a

triallax (changes requested):

user/dust/template.py
ttyyls: I noticed the change to portable atomics but I can't test it

triallax: well, i confirmed that it cross builds just fine (as opposed to the previous version which i also confirmed to fail) but if you really want to test it you could try in qemu or something

PR #3689, opened 2025-03-01
main/chimerautils: fix license name

sertonix:
I only wanted to check how chimerautils is packaged and not find typos :sob:

### DIFF
--- a/main/chimerautils/template.py
+++ b/main/chimerautils/template.py
@@ -25,7 +25,7 @@ provides = [
     self.with_pkgver("musl-rpmatch"),
 ]
 pkgdesc = "Chimera Linux userland"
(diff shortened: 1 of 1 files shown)


sertonix:
Maybe a bit too pedantic but I think `BSD-2-Clause AND BSD-3-Clause AND BSD-4-Clause AND ISC AND Beerware` would be the correct license.

sertonix:
Thanks for the clarification!

sertonix:
Should probably changed in the meson.build too: https://github.com/chimera-linux/chimerautils/blob/ef5c5f288a4fc1947788fd0218ee16ca6ca0b6e2/meson.build#L11

PR #3699, opened 2025-03-04
user/cloc: new package

ttyyls:
## Checklist

### DIFF
--- a/user/perl-algorithm-diff/template.py
+++ b/user/perl-algorithm-diff/template.py
@@ -1,0 +1,13 @@
+pkgname = "perl-algorithm-diff"
+pkgver = "1.201"
+pkgrel = 0
+build_style = "perl_module"
+hostmakedepends = ["perl"]
+makedepends = ["perl"]
+checkdepends = ["perl"]
+depends = ["perl"]
+pkgdesc = "Compute differences between files/lists"
+license = "Artistic-1.0-Perl"
+url = "https://metacpan.org/pod/Algorithm::Diff"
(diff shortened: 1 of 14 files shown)


user/perl-algorithm-diff/template.py
```diff
@@ -0,0 +1,13 @@
+pkgname = "perl-algorithm-diff"
+pkgver = "1.201"
+pkgrel = 0
+build_style = "perl_module"
+hostmakedepends = ["perl"]
+makedepends = ["perl"]
+checkdepends = ["perl"]
+depends = ["perl"]
+pkgdesc = "Compute `intelligent' differences between two files/lists"
```
q66, line 9: weird quoting, just `Compute differences between files/lists` is better

user/perl-algorithm-diff/update.py
```diff
@@ -0,0 +1,14 @@
+pkgname = "Algorithm-Diff"
```
q66, line 1: this whole update check is just weird

ttyyls: the reasoning behind it is that i ran into a weird edge case where the mathemtical diff between the numbers is not the same as a propper version diff.
example:
1.201 should be considered to 1.1902 but it's the other way around if u compare the numbers, padding with 0 on the right is an easy fix and there are precedents for it in other update checks

q66: i'd just put `1.1*` in ignore

ttyyls: suboptimal but im applying it anyway, maybe the update-check logic should be reworked as a whole at some point

PR #3701, opened 2025-03-04
user/picocom: add new package

Guiglin:
## Description

As its name suggests, picocom is a minimal dumb-terminal emulation program. It is, in principle, very much like minicom, only it is "pico" instead of "mini".

## Checklist


q66:
we have a serial client in base system btw

Guiglin:
Ah yes, python-pyserial?

q66:
no, `cu` and `tip` which are different frontends to the same (`cu` lets you specify the serial parameters on CLI, e.g. `cu -l /dev/ttyS0 -115200`, while `tip` lets you have pre-defined serial connections in `/etc/remote` and then you can just say `tip uart0` or `tip -115200 uart0`)

i'm fine adding this as well but just saying you don't need a third party program at all

Guiglin:
Ho I did not know those pieces of software. I won't use `picocom` anymore and I close this MR. Thank you :)

q66:
they are BSD-originated programs and a part of chimerautils, so they are not present in most distributions

i find them to be the most well-rounded serial clients though, people usually install picocom on linux because there isn't any other plain serial client

Guiglin:
That's why I did not know them indeed. I need to learn how to use it now, especially keyboard cmds ^^

PR #3704, opened 2025-03-06
user/sl: new package

cyuria:
## Description

From the project's README:
> SL (Steam Locomotive) runs across your terminal when you type "sl" as you meant to type "ls". It's just a joke command, and not useful at all.

https://github.com/eyJhb/sl

## Checklist

### DIFF
--- a/user/sl/template.py
+++ b/user/sl/template.py
@@ -1,0 +1,20 @@
+pkgname = "sl"
+pkgver = "5.05"
+pkgrel = 0
+build_style = "makefile"
+makedepends = ["ncurses-devel"]
+pkgdesc = "Steam Locomotive"
+license = "SL"
+url = "https://github.com/eyJhb/sl"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "6c941b526e3d01be7f91a3af4ae20a89d1e5d66b3b2d804c80123b1b1be96384"
+hardening = ["vis", "cfi"]
+# No test suite
+options = ["!check"]
+
+
+def install(self):
+    self.install_bin("sl")
+    self.install_man("sl.1")
+    self.install_man("sl.1.ja", "sl", 1, lang="ja")
+    self.install_license("LICENSE")


q66:
please format your template as Packaging.md says

user/sl/template.py
```diff
@@ -0,0 +1,21 @@
+pkgname = "sl"
+pkgver = "5.05"
+pkgrel = 0
+build_style = "makefile"
+make_install_args = [ "PREFIX=/usr" ]
+hostmakedepends = []
```
q66, line 6: these lines don't need to exist

user/sl/template.py
```diff
@@ -0,0 +1,21 @@
...
+makedepends = [ "ncurses-devel" ]
+depends = [ "ncurses" ]
```
q66, line 8: this line is wrong and should be removed

cyuria: Thanks for catching that. There is a dependency on `libncursesw.so.6`, so I've changed the dependency to `ncurses-libs`, which provides that file.

user/sl/template.py
cyuria: Thank you, I copied most of the template from other ones, and grepping through just now I noticed there was a `PREFIX=/usr` in the `make_install_args` of quite a few existing packages (also an empty `hostmakedepends` in the `tzdb` package, but that gets conditionally modified for cross compilation so I'm not sure if that's necessary or not).

cyuria:
> please format your template as Packaging.md says

I'm unable to find where my template differs from the style requirements in Packaging.md, could you possibly point me in the right direction? (If this is in regards to the license, https://spdx.org/licenses/SL.html, and if it's instead to do with the package description I'm not sure what else to put, however I could take a sentence from the project's readme if you think that's better)

user/sl/template.py
q66: that's still wrong, you don't need to manually set library dependencies

q66:
it's right in the checklist for Packaging.md, "correct style" is the first subsection

user/sl/template.py
cyuria: sorry about that, clearly I missed quite a few of the important bits of Packaging.md. It should all be resolved now

cyuria:
Thanks for dealing with my infallible ability to skip over significant portions of important documentation. It should hopefully all be good now :shrug:

q66:
the license file states `i.e., the above copyright notice has to appear in all copies.`; that means you should probably `self.install_license("LICENSE")` (the spdx data contain no way to properly check this so the lint rule is not exhaustive for minor licenses like this)

cyuria:
> the license file states i.e., the above copyright notice has to appear in all copies.; that means you should probably self.install_license("LICENSE")

fixed

PR #3759, opened 2025-03-15
main/ddclient: update to 4.0.0

Gnarwhal:
Not sure if this is a concern, but ddclient now looks for `/etc/ddclient/ddclient.conf` by default instead of `/etc/ddclient.conf` which might be confusing. It is possible to maintain the original behaviour but then it will be out of sync with documentation which might also be confusing. I opted for the new default behaviour, but if you have strong feelings the other way I will change it. 

## Checklist


main/ddclient/files/ddclient
```diff
@@ -2,5 +2,5 @@
 
 type = process
 command = /usr/bin/ddclient -foreground
-depends-on = network.target
+depends-on: network.target
```
q66, line 5: all fields with = assignments should go before those with :

q66:
also move to user/

also config file should ideally not be shipped in /etc; perhaps the service can invoke through a wrapper like

```
if [ -e /etc/ddclient/ddclient.conf ]; then
    exec /usr/bin/ddclient -foreground
else
    exec /usr/bin/ddclient -foreground -file /usr/share/ddclient/ddclient.conf
fi
```

if that makes sense at all (does the program have builtin defaults that make sense? if so, no config could be shipped by default, with perhaps the default as an example)

user/ddclient/template.py
```diff
@@ -16,8 +16,10 @@
...
     f"https://github.com/ddclient/ddclient/archive/refs/tags/v{pkgver}.tar.gz"
 )
-sha256 = "243cd832abd3cdd2b49903e1b5ed7f450e2d9c4c0eaf8ce4fe692c244d3afd77"
+sha256 = "4b37c99ac0011102d7db62f1ece7ff899b06df3d4b172e312703931a3c593c93"
 
 
 def post_install(self):
     self.install_service(self.files_path / "ddclient")
+    self.install_file("build/ddclient.conf", "usr/share/examples/ddclient")
+    self.rm(self.destdir / "etc/ddclient", recursive=True)
```
q66, line 25: self.uninstall

PR #3796, opened 2025-03-23
main/vte: split example terminal applications into seperate subpackages

breakgimme:
## Description

after the latest update vte started including desktop files for the example terminal apps which makes them show up if anything depending on vte is installed
given that these example apps are useless for the vast majority of people there's no reason for them to be included with the library

## Checklist


breakgimme:
the update check was broken because it was pulling in updates for the fast_float library, fixed it by specifying the url in update.py

q66:
there's no point in splitting both of them i think, just do one `vte-demos`

and the dependencies on `vte-gtkN` are not really necessary

breakgimme:
> and the dependencies on vte-gtkN are not really necessary

right, for whatever reason i forgot that cbuild handled dependencies automatically..

breakgimme:
sorry for the triple force push, kinda tired rn

breakgimme:
removed the update check fix due to it being done in 43ae2eb14e5576ae4757c12658f5e706bfcb3dcf

PR #3801, opened 2025-03-24
main/i2pd: new package

kul-sudo:
## Description
I2P Router written in C++

## Checklist


user/i2pd/files/i2pd
```diff
@@ -0,0 +1,9 @@
+type = process
```
q66, line 1: ditto as in the tor PR

kul-sudo: Done.

kul-sudo:
I think it is now ready to be merged.

user/i2pd/files/i2pd
```diff
@@ -0,0 +1,9 @@
...
+command = /usr/bin/i2pd --service --certsdir /usr/share/i2pd/certificates --conf /usr/share/i2pd/i2pd.conf --tunconf /usr/share/i2pd/tunnels.conf
```
q66, line 2: it's not correct to force config files to go to `/usr`

`/usr` should merely contain a default (if such default is needed at all) that the administrator can override with a file in `/etc`

i.e. it shouldn't be necessary for an `/etc` file to exist, but all user configuration should be done there

user/i2pd/files/i2pd
```diff
@@ -0,0 +1,9 @@
...
+pid-file = /var/lib/i2pd/i2pd.pid
```
q66, line 3: pid files don't apply to process services

user/i2pd/files/i2pd
```diff
@@ -0,0 +1,9 @@
...
+working-dir = /var/lib/i2pd
+restart = no
```
q66, line 5: why?

user/i2pd/files/i2pd
q66: same applies to the others btw

kul-sudo:
I think it's in the optimal state right now.

PR #3809, opened 2025-03-25
user/protontricks: update to 1.12.1

z-erica:


triallax:
hm, why the arch restriction?

q66:
given they already depend on wine there is no point

q66:
i'll just fix the ci scripts later

PR #3813, opened 2025-03-26
user/tor: new package

kul-sudo:
## Description
Anonymizing overlay network.

## Checklist

### DIFF
--- a/user/tor/files/tor
+++ b/user/tor/files/tor
@@ -1,0 +1,6 @@
+type = process
+command = /usr/bin/tor
+working-dir = /var/lib/tor
+depends-on = network.target
+depends-on = time-sync.target
+run-as = _tor
--- a/user/tor/template.py
+++ b/user/tor/template.py
@@ -1,0 +1,29 @@
+pkgname = "tor"
+pkgver = "0.4.8.16"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = [
+    "pkgconf",
+    "xz-devel",
+    "asciidoc",
+    "zstd-devel",
+    "automake",
+    "zlib-ng-compat-devel",
+    "openssl3-devel",
+    "libevent-devel",
+]
+pkgdesc = "Anonymizing overlay network"
+license = "BSD-3-Clause"
+url = "https://gitlab.com/torproject/tor"
+source = f"{url}/-/archive/tor-{pkgver}/tor-tor-{pkgver}.tar.gz"
+sha256 = "2964e2c3c7ebc9534595e59e29938e5446334510fdb912009cd877258898bfeb"
+# requires shellcheck
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_file("src/config/torrc.sample.in", "etc/tor", name="torrc")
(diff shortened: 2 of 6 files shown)


kul-sudo:
Related to #3814.

user/tor/template.py
```diff
@@ -0,0 +1,29 @@
...
+license = "BSD-3-Clause"
+url = "https://gitlab.com/torproject/tor"
+source = f"{url}/-/archive/tor-{pkgver}/tor-tor-{pkgver}.tar.gz"
+sha256 = "2964e2c3c7ebc9534595e59e29938e5446334510fdb912009cd877258898bfeb"
+# requires shellcheck
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_file("src/config/torrc.sample.in", "etc/tor", name="torrc")
```
q66, line 25: we don't put sample files in /etc

user/tor/files/tor
```diff
@@ -0,0 +1,6 @@
+type = process
```
q66, line 1: service needs reworking to properly respect : vs = fields

user/tor/template.py
```diff
@@ -0,0 +1,29 @@
+pkgname = "tor"
+pkgver = "0.4.8.16"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = [
```
q66, line 5: many of these are makedepends, not hostmakedepends

user/tor/template.py
```diff
@@ -0,0 +1,29 @@
...
+    "zlib-ng-compat-devel",
+    "openssl3-devel",
+    "libevent-devel",
+]
+pkgdesc = "Anonymizing overlay network"
+license = "BSD-3-Clause"
+url = "https://gitlab.com/torproject/tor"
+source = f"{url}/-/archive/tor-{pkgver}/tor-tor-{pkgver}.tar.gz"
+sha256 = "2964e2c3c7ebc9534595e59e29938e5446334510fdb912009cd877258898bfeb"
+# requires shellcheck
```
Gnarwhal, line 20: It looks like shellcheck might be an optional dependency for checks. Adding `bash` to checkdepends and something like [this patch](https://raw.monodon.me/fix-test.patch) get tests to pass for me. Although admittedly I don't really understand the root cause of the test failing :/

user/tor/template.py
kul-sudo: I'm afraid you didn't notice, but I do `name="torrc"`.
Also, `autofs` does this, for example:
```python
# install sample configuration
self.install_file(
    "samples/autofs.conf.default", "etc/autofs", name="autofs.conf"
)
```

kul-sudo: Done.

user/tor/files/tor
kul-sudo: Done (if I got what you meant).

user/tor/template.py
kul-sudo: OK, tor seems to create the sample on its own. I need to localize where exactly and patch it.

PR #3826, opened 2025-03-27
main/mpv: update to 0.40.0

triallax:


triallax:
seemingly found a mesa bug that surfaces with this update, will hold link to it once i've reported it

drafting for now

PR #3828, opened 2025-03-27
user/pounce: new package

ayushnix:
## Description

A new package for pounce, an IRC bouncer.

## Checklist

### DIFF
--- a/user/pounce/template.py
+++ b/user/pounce/template.py
@@ -1,0 +1,27 @@
+pkgname = "pounce"
+pkgver = "3.1"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--enable-notify",
+    "--enable-palaver",
+]
+configure_gen = []
+make_dir = "."
+make_build_target = "all"
+hostmakedepends = ["autoconf", "pkgconf"]
+makedepends = ["curl-devel", "libretls-devel", "sqlite-devel"]
+pkgdesc = "TLS-only multi-client IRC bouncer"
+license = "GPL-3.0-or-later"
+url = "https://git.causal.agency/pounce"
+source = f"{url}/snapshot/{pkgname}-{pkgver}.tar.gz"
+sha256 = "97f245556b1cc940553fca18f4d7d82692e6c11a30f612415e5e391e5d96604e"
+hardening = ["vis", "cfi"]
+# no tests
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+    self.install_man("QUIRKS.7", name="pounce-quirks")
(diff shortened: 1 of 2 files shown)


user/pounce/template.py
```diff
@@ -0,0 +1,27 @@
+pkgname = "pounce"
+pkgver = "3.1"
+pkgrel = 0
+build_style = "configure"
```
q66, line 4: gnu_configure

user/pounce/template.py
```diff
@@ -0,0 +1,27 @@
...
+configure_args = [
+    "--prefix=/usr",
```
q66, line 6: and then drop prefix/mandir

user/pounce/template.py
```diff
@@ -0,0 +1,27 @@
...
+    "--mandir=/usr/share/man",
+    "--enable-notify",
+    "--enable-palaver",
+]
+make_build_target = "all"
+hostmakedepends = ["pkgconf"]
+makedepends = ["libretls-devel", "curl-devel", "sqlite-devel"]
```
q66, line 13: sort

user/pounce/template.py
ayushnix: Done.

I'm curious, why use `gnu_configure` over `configure` in this case? I was mostly using the [catgirl](https://github.com/chimera-linux/cports/blob/master/main/catgirl/template.py) package as a reference. Does that package need an update? When would one want to use `configure` over `gnu_configure`?

user/pounce/template.py
ayushnix: Done.

user/pounce/template.py
ayushnix: Done.

user/pounce/template.py
subnut: I guess if `gnu_configure` works, it's preferred over the bundled `configure`
as the automake/autoconf files are infinitely more readable than the
`configure` file that's produced by them.

ayushnix: @subnut The `configure` script in catgirl is also manually written and is almost identical to the one in pounce.

ayushnix: oh, and using `gnu_configure` also produces this output during the `configure` build phase, I assume these warnings can be ignored?

```
warning: unsupported option --sysconfdir=/etc
warning: unsupported option --sbindir=/usr/bin
warning: unsupported option --libdir=/usr/lib
warning: unsupported option --infodir=/usr/share/info
warning: unsupported option --localstatedir=/var
warning: unsupported option --build=x86_64-chimera-linux-musl
warning: unsupported option --host=x86_64-chimera-linux-musl
```

PR #3829, opened 2025-03-28
user/python-dbus-fast: update to 2.43.0

Gnarwhal:
## Checklist


user/python-dbus-fast/patches/remove_codspeed.patch
```diff
@@ -1,6 +1,6 @@
-diff --color -ruN python-dbus-fast-2.39.3/tests/benchmarks/test_marshall.py python-dbus-fast-2.39.3-patched/tests/benchmarks/test_marshall.py
```
triallax, line 1: might be better served using `self.rm` instead of this patch

PR #3842, opened 2025-03-30
user/pdfgrep: new package

tmahlburg:
## Description

This is a commandline utility to search text in PDF files.

## Checklist

- [ x ] I have read [CONTRIBUTING.md](https://github.com/chimera-linux/cports/blob/master/CONTRIBUTING.md)

- [ x ] I have read [Packaging.md](https://github.com/chimera-linux/cports/blob/master/Packaging.md#quality_requirements)
- [ x ] I have built and tested my changes on my machine

- [ x ] I will take responsibility for my template and keep it up to date

### DIFF
--- a/user/pdfgrep/template.py
+++ b/user/pdfgrep/template.py
@@ -1,0 +1,12 @@
+pkgname = "pdfgrep"
+pkgver = "2.2.0"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = ["automake", "pkgconf"]
+makedepends = ["libgcrypt-devel", "pcre2-devel", "poppler-devel"]
+pkgdesc = "Tool to search text in PDF files"
+license = "GPL-2.0-or-later"
+url = "https://pdfgrep.org"
+source = f"https://pdfgrep.org/download/pdfgrep-{pkgver}.tar.gz"
+sha256 = "0661e531e4c0ef097959aa1c9773796585db39c72c84a02ff87d2c3637c620cb"
+hardening = ["vis", "cfi"]


user/pdfgrep/template.py
```diff
@@ -0,0 +1,22 @@
...
+pkgrel = 0
+build_style = "configure"
+configure_args = [
+    "--prefix=/usr/"
+]
+hostmakedepends = [
+    "pkgconf",
+    "poppler-devel",
+    "libgcrypt-devel",
+    "pcre2-devel"
```
subnut, line 12: Sort this list

user/pdfgrep/template.py
```diff
@@ -0,0 +1,22 @@
...
+]
+pkgdesc = "Tool to search text in PDF files"
+license = "GPL-2.0-or-later	"
```
subnut, line 15:
```suggestion
license = "GPL-2.0-or-later"
```
Remove superfluous whitespace

user/pdfgrep/template.py
```diff
@@ -0,0 +1,22 @@
+pkgname = "pdfgrep"
+pkgver = "2.2.0"
+pkgrel = 0
+build_style = "configure"
```
subnut, line 4:
```suggestion
build_style = "gnu_configure"
```

subnut (changes requested):
Please run an auto-formatter (e.g. black) on the template before submitting it.

See https://github.com/chimera-linux/cports/blob/master/Packaging.md#correct-style

user/pdfgrep/template.py
tmahlburg: Just out of interest: Is there a specific reason, why `gnu_configure` is preferred over just `configure`? It works fine both ways on my end, I'm just curious.

tmahlburg:
Thanks for the feedback, I made the changes.

user/pdfgrep/template.py
```diff
@@ -0,0 +1,13 @@
+pkgname = "pdfgrep"
+pkgver = "2.2.0"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = ["--prefix=/usr/"]
```
subnut, line 5:
```suggestion
```
This is no longer needed, as gnu_configure automatically sets `--prefix=/usr`

user/pdfgrep/template.py
subnut: Also, configure scripts that are produced using GNU autotools have some common
configuration options. The gnu_configure build style anticipates those, and
sets environment variables and configuration options accordingly.

Case in hand, the `--prefix=/usr` option needs to be specified for almost all
configure scripts generated using autotools, and therefore the gnu_configure
build style sets that automatically, as seen [here](https://github.com/chimera-linux/cports/blob/master/src/cbuild/util/gnu_configure.py#L83-L102)

tmahlburg: I see, thanks for the explanation!

user/pdfgrep/template.py
tmahlburg: Done!

user/pdfgrep/template.py
```diff
@@ -0,0 +1,12 @@
+pkgname = "pdfgrep"
+pkgver = "2.2.0"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = ["autoconf", "automake", "pkgconf"]
```
q66, line 5: specifying autoconf is redundant

PR #3851, opened 2025-04-02
user/bupstash: new package

tjheeta:
## Description

Bupstash is an encrypted backup tool. 

## Checklist


user/bupstash/patches/libc-crate-update.patch
```diff
@@ -0,0 +1,28 @@
+From b5ac70513d29a00bf46f4bac82ff5a874bc96ef6 Mon Sep 17 00:00:00 2001
```
q66, line 1: this doesn't say why (but is too old anyway for e.g. loongarch64 builds), and also is unnecessary as a patch

look at `pre_prepare` in e.g. `user/shotman`

q66:
appears to be dead, also no review response

PR #3878, opened 2025-04-07
use env var from https://docs.helix-editor.com/building-from-source.html

zacharyburnett:
## Description

when attempting `hx --grammar build`, the following error occurs:
```
error occurred in cc-rs: unknown target `x86_64-chimera-linux-musl`.

NOTE: `cc-rs` only supports a fixed set of targets when not in a build script.
- If adding a new target, you will need to fork of `cc-rs` until the target
  has landed on nightly and the auto-generated list has been updated. See also
  the `rustc` dev guide on adding a new target:
  https://rustc-dev-guide.rust-lang.org/building/new-target.html
- If using a custom target, prefer to upstream it to `rustc` if possible,
  otherwise open an issue with `cc-rs`:
  https://github.com/rust-lang/cc-rs/issues/new
```

[The docs say to use the following env var when building](https://docs.helix-editor.com/building-from-source.html):
```shell
RUSTFLAGS="-C target-feature=-crt-static"
```

## Checklist

- [ ] I have read [CONTRIBUTING.md](https://github.com/chimera-linux/cports/blob/master/CONTRIBUTING.md)

- [ ] I have read [Packaging.md](https://github.com/chimera-linux/cports/blob/master/Packaging.md#quality_requirements)


zacharyburnett:
oof, the package builds on my machine but `hx --grammar build` still doesn't work, I think I need to put the env var in a different place

q66:
i'm not sure what you are trying to achieve here but this is completely unrelated and irrelevant, all chimera targets are shared crt by default

what you are running into is this https://github.com/rust-lang/cc-rs/issues/1407 but that's not something a chance in helix build will fix

zacharyburnett:
ok, as a workaround in case anyone else finds this, I built `helix` from source, ran `hx --grammar fetch && hx --grammar build`, and then copied the `runtime/` dir to `~/.config/helix/runtime/`, and the grammars are now functioning.

q66:
it'll probably fix itself when everything updates their lockfiles to cc-v1.2.17 or newer

PR #3961, opened 2025-04-12
main/dino: update to 0.5.0

triallax:
cross fails with <code>error: Package `gee-0.8' not found in specified Vala API directories or GObject-Introspection GIR directories</code>, not sure if that's important to resolve


q66:
not important at all, no (anything with g-i is unlikely to crossbuild)

PR #3989, opened 2025-04-15
main/tcllib: remove

z-erica:
this was only used for the fossil test suite, which we don't even intend to run anymore


main/base-removed-dbg/template.py
```diff
@@ -145,6 +145,7 @@
     "qalc",
     "rfkill",
     "sdl",
+    "tcllib",
```
triallax, line 148: doesn't this need a relbump

q66: that part just shouldn't be there because that's not what the package is for

PR #4030, opened 2025-04-19
user/forgejo: fix settings override

z-erica:


user/forgejo/files/forgejo
```diff
@@ -1,10 +1,8 @@
-# forgejo service
-
 type = process
 command = /usr/bin/forgejo web
 logfile = /var/log/forgejo.log
-depends-on = network.target
-depends-on = local.target
+depends-on: network.target
```
q66, line 4: this needs sorting (: fields after =)

PR #4087, opened 2025-04-26
user/libimobiledevice: move from main, build from git head

z-erica:
last release was from 2020 and is essentially nonfunctional with newer devices. unsure if this is acceptable but wfm locally


q66:
this has revdeps in main so it can't be moved and i don't feel like tracking git just because upstream is unable to tag a thing so feel free to nag them instead

PR #4108, opened 2025-04-28
user/bfs: new package

ayushnix:
## Description

A new package called bfs which does bread-first search when used as a chimerautils find and findutils replacement.

## Checklist


user/bfs/template.py
```diff
@@ -0,0 +1,28 @@
...
+pkgdesc = "Breadth-first version of the UNIX find command"
+license = "0BSD"
+url = "https://github.com/tavianator/bfs"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "446a0a1a5bcbf8d026aab2b0f70f3d99c08e5fe18d3c564a8b7d9acde0792112"
+hardening = ["cfi", "vis"]
+
+
+def post_install(self):
+    for f in (self.cwd / "docs").glob("*.md"):
```
q66, line 26: we don't install docs

ayushnix: The doc files are picked up automatically by the `bfs-doc` subpackage. Should I not include those files in the `-doc` subpackage?

ayushnix: changes made after discussion in IRC/matrix chat room

PR #4116, opened 2025-04-29
user/docker-engine: new package

breakgimme:
## Description

this pr adds the "classic" docker daemon, for times where podman or plain containerd doesn't suffice
both rootful and rootless (given that /etc/subuid and /etc/subguid are set up) setups work

## Checklist

### DIFF
--- a/user/docker-engine/files/docker
+++ b/user/docker-engine/files/docker
@@ -1,0 +1,5 @@
+type = process
+command = /usr/bin/dockerd
+logfile = /var/log/docker.log
+depends-on: containerd
+options: signal-process-only
--- a/user/docker-engine-rootless
+++ b/user/docker-engine-rootless
@@ -1,0 +1,1 @@
+docker-engine
--- a/user/docker-engine/files/docker.user
+++ b/user/docker-engine/files/docker.user
@@ -1,0 +1,3 @@
+type = process
+command = /usr/bin/dockerd-rootless
+depends-on: dbus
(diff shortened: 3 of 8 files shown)


q66:
are there *any* cases where podman does not suffice

user/docker-engine/files/docker
```diff
@@ -0,0 +1,15 @@
+type = process
```
q66, line 1: neither of these services follow usual proper conventions for formatting and whatnot

q66: like what is all the spurious newlines, the pointless restart stuff that no other service specifies, etc

breakgimme:
> are there _any_ cases where podman does not suffice

i've had issues with the container restarting mechanism, where sometimes they just didn't start after the server rebooted
i solved this personally by moving to raw containerd + nerdctl but this opened up another issue of not having a docker compatible socket so i can check for container updates

user/docker-engine/files/docker
breakgimme: this is my first time writing a dinit service that's not for my personal use, so excuse the mess with newlines
the restarts were taken from the systemd equivalents, i'll get rid of them if they're not needed
i'll fix these issues tomorrow

breakgimme: hope the changes i've made now make the services formatted well enough

breakgimme:
ugh they are in the process of releasing an update *right now*
but rc shouldn't be in the update check though, will fix with push to 29.5.0

PR #4122, opened 2025-04-30
user/libzmq: new package

rbgarga:
## Description

Introduce zeromq in preparation to submit saltstack

## Checklist


user/zeromq/template.py
```diff
@@ -0,0 +1,30 @@
...
+pkgrel = 0
+build_style = "gnu_configure"
+configure_env = {
+    "LIBUNWIND_CFLAGS": "-I/usr/include/libunwind-nongnu",
+    "LIBUNWIND_LDFLAGS": "-lunwind-nongnu",
+}
+hostmakedepends = [
+    "asciidoc",
+    "automake",
+    "libtool",
```
q66, line 12: can you try if using slibtool works

user/zeromq/template.py
```diff
@@ -0,0 +1,30 @@
...
+    "pkgconf",
+    "xmlto",
+]
+makedepends = [
+    "libbsd-devel",
+    "libsodium-devel",
+    "libunwind-nongnu-devel",
```
q66, line 19: does this really need that

user/zeromq/template.py
```diff
@@ -0,0 +1,30 @@
+pkgname = "zeromq"
```
q66, line 1: should probably be called libzmq

q66:
also there is no need to put the version in the commit message

user/zeromq/template.py
rbgarga: nope, it was my mistake

PR #4125, opened 2025-04-30
user/salt-tls: new package (3006.10)

rbgarga:


rbgarga:
This PR requires #4121 and #4122 to be merged first

q66:
this is all squashed together and therefore impossible to review...

user/python-croniter/template.py
```diff
@@ -0,0 +1,20 @@
...
+pkgver = "6.0.0"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+]
+depends = ["python-dateutil", "python-pytz"]
+checkdepends = ["python-dateutil", "python-pytest", "python-pytz"]
```
q66: `checkdepends = ["python-pytest", *depends]`

user/python-looseversion/template.py
```diff
@@ -0,0 +1,19 @@
...
+]
+checkdepends = [
+    "python-pytest",
+]
+pkgdesc = "Python extension for comparing version strings"
+license = "PSF-2.0"
+url = "https://pypi.org/project/looseversion"
+source = f"$(PYPI_SITE)/l/looseversion/looseversion-{pkgver}.tar.gz"
+sha256 = "ebde65f3f6bb9531a81016c6fef3eb95a61181adc47b7f949e9c0ea47911669e"
+# couldn't make it to work
```
q66: probably a better comment should go here; in this case the issue seems to be that it doesn't take a regular pytest invocation; and running `tests.py` is currently too annoying without better abstractions in cbuild, so let's not worry about it for now, just put a better comment

user/python-pycurl/template.py
```diff
@@ -0,0 +1,21 @@
...
+makedepends = ["curl-devel", "openssl3-devel", "python-devel"]
+checkdepends = ["python-flask", "python-flaky", "python-pytest"]
+pkgdesc = "Python3 interface to libcurl"
+license = "LGPL-2.0-or-later AND MIT"
+url = "http://pycurl.io"
+source = f"$(PYPI_SITE)/p/pycurl/pycurl-{pkgver}.tar.gz"
+sha256 = "2b73e66b22719ea48ac08a93fc88e57ef36d46d03cb09d972063c9aa86bb74e6"
+
+
+def post_install(self):
```
q66, line 19: only one license is necessary (the MIT)

user/python-pyzmq/template.py
```diff
@@ -0,0 +1,29 @@
...
+    "libzmq-devel",
+    "python-devel",
+]
+checkdepends = ["python-pytest"]
+pkgdesc = "Python bindings for ZeroMQ"
+license = "BSD-3-Clause"
+url = "https://pypi.org/project/pyzmq"
+source = f"$(PYPI_SITE)/p/pyzmq/pyzmq-{pkgver}.tar.gz"
+sha256 = "b1f08eeb9ce1510e6939b6e5dcd46a17765e2333daae78ecf4606808442e52cf"
+# couldn't make it work
```
q66: again a better comment would be nice

user/python-tornado/template.py
```diff
@@ -0,0 +1,65 @@
+pkgname = "python-tornado"
+pkgver = "6.5.1"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-installer",
+    "python-setuptools",
+]
+depends = [
```
q66: vast majority of these are unnecessary; the requirements.txt which you presumably used here is misleading, most of these dependencies are not dependencies of tornado and they are listed as dependencies of things like sphinx, black, etc.

i think tornado itself doesn't really have dependencies?

user/salt-lts/files/salt-api
```diff
@@ -0,0 +1,6 @@
+type = process
```
q66, line 1: please use correct style for all of these services (i.e. proper usage of `=` vs `:` and so on)

user/salt-lts/template.py
```diff
@@ -0,0 +1,56 @@
...
+        "install",
+        "--prefix=/usr",
+        f"--root={self.chroot_destdir}",
+    )
+
+
+def post_install(self):
+    self.install_tmpfiles(self.files_path / "tmpfiles.conf")
+    self.install_dir("etc/salt")
+    for f in ["cloud", "master", "minion", "proxy"]:
```
q66, line 51: all of these conf files are purely comments with their default contents

we don't install files in `/etc` like that anymore

user/salt-lts/update.py
```diff
@@ -0,0 +1 @@
+pattern = r"3006\.\d+"
```
q66: won't this mean update check will skip the next lts version?

q66:
a39cf46ae087e51803a2f8d0a8feadaf59b75d21

PR #4135, opened 2025-05-02
user/riff: new package

jcgruenhage:
## Description

## Checklist

### DIFF
--- a/user/riff/template.py
+++ b/user/riff/template.py
@@ -1,0 +1,15 @@
+pkgname = "riff"
+pkgver = "3.3.10"
+pkgrel = 0
+build_style = "cargo"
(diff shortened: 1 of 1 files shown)


user/riff/template.py
```diff
@@ -0,0 +1,15 @@
+pkgname = "riff"
+pkgver = "3.3.10"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo-auditable"]
+depends = ["rust-std"]
```
q66, line 6: i don't think this is depends

jcgruenhage: sorry, this should be makedepends, right?

PR #4140, opened 2025-05-02
user/gitoxide: update to 0.44.0

jcgruenhage:


user/gitoxide/template.py
```diff
@@ -13,13 +13,12 @@
     "curl-devel",
     "rust-std",
     "sqlite-devel",
-    "zlib-ng-compat-devel",
```
jcgruenhage, line 16: They started using `zlib-rs` in https://github.com/GitoxideLabs/gitoxide/pull/1963, meaning this is being replaced with a pure rust implementation

PR #4152, opened 2025-05-04
user/apache-arrow: new package

a16bitsysop:
## Description

## Checklist

### DIFF
--- a/user/apache-arrow/template.py
+++ b/user/apache-arrow/template.py
@@ -1,0 +1,105 @@
+pkgname = "apache-arrow"
+pkgver = "20.0.0"
+pkgrel = 0
+_arrowsha = "d2a13712303498963395318a4eb42872e66aead7"
+_parquetsha = "18d17540097fca7c40be3d42c167e6bfad90763c"
+build_wrksrc = "cpp"
+build_style = "cmake"
+configure_args = [
+    "-DARROW_BUILD_EXAMPLES=OFF",
+    "-DARROW_BUILD_STATIC=OFF",
+    "-DARROW_BUILD_TESTS=ON",
+    "-DARROW_DEPENDENCY_SOURCE=SYSTEM",
+    "-DARROW_COMPUTE=ON",
+    "-DARROW_CSV=ON",
+    "-DARROW_DATASET=ON",
+    "-DARROW_FILESYSTEM=ON",
+    "-DARROW_FLIGHT=ON",
+    "-DARROW_HDFS=ON",
+    "-DARROW_JSON=ON",
+    "-DARROW_JSON=ON",
+    "-DARROW_MIMALLOC=OFF",
+    "-DARROW_ORC=OFF",
+    "-DARROW_PARQUET=ON",
+    "-DARROW_SIMD_LEVEL=NONE",
+    "-DARROW_TENSORFLOW=ON",
+    "-DARROW_USE_GLOG=ON",
+    "-DARROW_USE_LLD=ON",
+    "-DARROW_WITH_BROTLI=ON",
+    "-DARROW_WITH_BZ2=ON",
+    "-DARROW_WITH_LZ4=ON",
+    "-DARROW_WITH_MUSL=ON",
+    "-DARROW_WITH_SNAPPY=ON",
+    "-DARROW_WITH_ZLIB=ON",
+    "-DARROW_WITH_ZSTD=ON",
+    "-DPARQUET_REQUIRE_ENCRYPTION=ON",
+    "-DLZ4_ROOT=/usr",
+]
+make_check_args = [
+    "-E",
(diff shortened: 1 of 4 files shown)


user/apache-arrow/template.py
```diff
@@ -0,0 +1,130 @@
...
+    "zlib-ng-compat-devel",
+    "zstd-devel",
+]
+checkdepends = ["bash"]
+pkgdesc = "Multi-language toolbox for fast data interchange and processing"
+license = "Apache-2.0"
+url = "https://arrow.apache.org"
+source = [
+    f"https://archive.apache.org/dist/arrow/arrow-{pkgver}/apache-arrow-{pkgver}.tar.gz",
+    f"https://github.com/microsoft/mimalloc/archive/v{_mimalloc_ver}.tar.gz",
```
q66, line 77: why? just disable it

user/apache-arrow/template.py
```diff
@@ -0,0 +1,130 @@
...
+    "-DARROW_WITH_ZSTD=ON",
+    "-DPARQUET_REQUIRE_ENCRYPTION=ON",
+    "-DLZ4_ROOT=/usr",
+]
+make_check_args = [
+    "-E",
+    "arrow-array-test|arrow-compute-scalar-temporal-test|arrow-utility-test|arrow-csv-test|arrow-ipc-json-simple-test|arrow-dataset-file-parquet-test",
+]
+make_check_env = {
+    "ARROW_TEST_DATA": f"/builddir/apache-arrow-{pkgver}/testing/data",
```
q66, line 44: there is no guarantee these kind of paths will remain stable

self.chroot_cwd and do it from an appropriate `init_` hook

user/apache-arrow/template.py
```diff
@@ -0,0 +1,130 @@
...
+    "pkgconf",
+]
+makedepends = [
+    "boost-devel",
+    "brotli-devel",
+    "glog-devel",
+    "grpc-devel",
+    "gtest-devel",
+    "libevent-devel",
+    "lz4",
```
q66, line 59: that sounds like a hostdep?

q66: (if at all)

user/apache-arrow/template.py
```diff
@@ -0,0 +1,130 @@
...
+    case _:
+        configure_args += ["-DARROW_RUNTIME_SIMD_LEVEL=NONE"]
+
+
+@subpackage("apache-arrow-devel")
+def _(self):
+    return self.default_devel()
+
+
+@subpackage("libarrow")
```
q66, line 108: subpackages that are not prefixed with the parent package name are not allowed

unless there is a good reason to split them all out, `apache-arrow-libs`

q66:
given the can of worms grpc is, i suggest you submit that separately for separate review

user/apache-arrow/template.py
a16bitsysop: Okay will try that

user/apache-arrow/template.py
a16bitsysop: Ah it should be removed, added when I was trying to get it to find lz4. It needs path set later on to find it.

a16bitsysop:
Okay will split out into another pull request

a16bitsysop:
left grpc in temporarily so it still builds, I can't get init_check setting variables to work and trying to use warn to output the path gives `AttributeError: 'Logger' object has no attribute 'warn'`

Left in latest commit to show

user/apache-arrow/template.py
```diff
@@ -0,0 +1,104 @@
...
+    case "aarch64" | "x86_64":
+        configure_args += ["-DARROW_RUNTIME_SIMD_LEVEL=MAX"]
+        makedepends += ["xsimd"]
+    case _:
+        configure_args += ["-DARROW_RUNTIME_SIMD_LEVEL=NONE"]
+
+
+def init_check(self):
+    self.logger.warn(f"{self.chroot_cwd}/arrow-testing/data")
+    make_check_env = {
```
q66, line 96:
```
    self.make_check_env["ARROW_TEST_DATA"] = ...
    self.make_check_env["PARQUET_TEST_DATA"] = ...
```

q66: also, `= str(self.chroot_cwd / "arrow-testing/data")` and so on

q66:
i tried updating this but can't get it to support any less than sse4.2 compile-time baseline (it fails to build because it expects `xsimd::batch<signed char, xsimd::sse4_2>::size` to be at least 16 at compile time which is only true when the whole thing is build with `-msse4.2`, which we can't do, even though it seems to have some form of runtime dispatch)

feel free to figure it out and update it if you're still interested in it, i'll close it for now

other than that i have some minor nits with the template but i can fix them myself when it comes down to it

PR #4168, opened 2025-05-09
user/aspell-fr: new package

ttyyls:
## Checklist

### DIFF
--- a/user/aspell-fr/template.py
+++ b/user/aspell-fr/template.py
@@ -1,0 +1,20 @@
+pkgname = "aspell-fr"
+pkgver = "0.50.3"
+pkgrel = 0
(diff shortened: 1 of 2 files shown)


user/aspell-fr/template.py
```diff
@@ -0,0 +1,20 @@
+pkgname = "aspell-fr"
+pkgver = "0.50.3"
+_ver = "0.50-3"
```
triallax, line 3: this could be easily generated from the `pkgver`

ttyyls: i couldn't come up with a concise shorthand for it

triallax: see e.g. `main/lua5.1-libluv`
also let's move `_ver` below `pkgrel`

PR #4176, opened 2025-05-10
main/zxing-cpp: Build reader and write programs

Biswa96:
It was disabled in 100dcb6fe4e3e9e253c53f18b4273939ca20a8d8 commit.

Debian and Alpine both provide those programs.
* https://packages.debian.org/sid/amd64/zxing-cpp-tools/filelist
* https://pkgs.alpinelinux.org/contents?name=zxing&repo=community&branch=edge&arch=x86_64

### DIFF
--- a/main/zxing-cpp/template.py
+++ b/main/zxing-cpp/template.py
@@ -1,14 +1,14 @@
 pkgname = "zxing-cpp"
 pkgver = "2.3.0"
-pkgrel = 0
+pkgrel = 1
 build_style = "cmake"
 configure_args = [
     "-DBUILD_UNIT_TESTS=ON",
-    "-DBUILD_EXAMPLES=OFF",
+    "-DBUILD_EXAMPLES=ON",
     "-DBUILD_BLACKBOX_TESTS=OFF",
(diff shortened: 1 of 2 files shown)


Biswa96:
Example commands:

```
$ ZXingWriter -size 200/200 QRCode example example.png
$ ZXingReader example.png
Text:       "example"
Bytes:      65 78 61 6D 70 6C 65
Format:     QRCode
Identifier: ]Q1
Content:    Text
HasECI:     true
Position:   68x18 131x18 131x81 68x81
Rotation:   0 deg
IsMirrored: false
IsInverted: false
EC Level:   L
Version:    1
```

main/zxing-cpp/template.py
```diff
@@ -20,3 +20,8 @@
 @subpackage("zxing-cpp-devel")
 def _(self):
     return self.default_devel()
+
+
+@subpackage("zxing")
```
triallax, line 25: subpackage names should have the parent package name as a prefix (this will be enforced in the future)
```suggestion
@subpackage("zxing-cpp-progs")
```

Biswa96: Done. Thanks.

PR #4179, opened 2025-05-10
user/uv: update to 0.7.3

jcgruenhage:


user/uv/template.py
```diff
@@ -41,17 +41,14 @@ def init_build(self):
 
 
 def post_build(self):
-    for shell in ["bash", "fish", "nushell", "zsh"]:
-        with open(self.cwd / f"uv.{shell}", "w") as cf:
+    for cmd, shell in [
```
triallax, line 44: nothing strictly wrong with this, but i think a nested loop would be a bit more readable

user/uv/template.py
```diff
@@ -66,6 +63,9 @@ def check(self):
 
 def post_install(self):
     self.install_license("LICENSE-MIT")
-    for shell in ["bash", "fish", "nushell", "zsh"]:
-        self.install_completion(f"uv.{shell}", shell)
-        self.install_completion(f"uvx.{shell}", shell, "uvx")
+    for cmd, shell in [
+        (cmd, shell)
```
triallax, line 67: ditto

user/uv/template.py
jcgruenhage: fair

PR #4188, opened 2025-05-10
user/ty: new package

jcgruenhage:
`ty` is currently  in alpha pre-release state, but I wanted to play around with it locally, for which I have packaged it. I pushed it in case others also want to play around with it.


q66:
not interested in additional astral slopware at the moment

PR #3787, opened 2025-03-22
Asahi Linux support

breakgimme:
## Description

this pr adds support for asahi linux
it works, and this pr is being made on an m1 macbook air running chimera linux
it can be installed by creating an iso with base-asahi in it and booting it on an apple silicon mac after installing the bare uefi environment from the asahi installer (a chimera-specific asahi installer could also be created with relatively low effort)
some notes:
- i've included a patch for alsa-lib for a signed overflow that happened with the audio support in asahi
- mesa-asahi is not included into base-asahi due to limitations with providers, it has to be installed seperately
- the update-m1n1 kernel trigger in base-asahi is executed even if it's running in a chimera-live mklive environment, not sure if it's possible to fix that as i couldn't find any solution

## Checklist

Before this pull request is reviewed, certain conditions must be met.

The following must be true for all changes:

- [x] I have read [CONTRIBUTING.md](https://github.com/chimera-linux/cports/blob/master/CONTRIBUTING.md)

The following must be true for template/package changes:

- [x] I have read [Packaging.md](https://github.com/chimera-linux/cports/blob/master/Packaging.md#quality_requirements)
- [x] I have built and tested my changes on my machine

The following must be true for new package submissions:

- [x] I will take responsibility for my template and keep it up to date

PR #4203, opened 2025-05-13
Mesa 25.1

JamiKettunen:
Anyone wanting to is free to try it already, to be tested if builds on `ppc64le` (and ofc wait until at least 25.1.1 stable is out)

Also drop no-op `configure_args` for:
- `shared-glapi`: https://gitlab.freedesktop.org/mesa/mesa/-/commit/e5c76088
- `osmesa`: https://gitlab.freedesktop.org/mesa/mesa/-/commit/027ccd96
- `power8`: https://gitlab.freedesktop.org/mesa/mesa/-/commit/c4b30507

Other still functional but deprecated options (until 25.2 )
- `gallium-opencl`: https://gitlab.freedesktop.org/mesa/mesa/-/commit/a0b457ac
- `gallium-xa`: https://gitlab.freedesktop.org/mesa/mesa/-/commit/cf400997

I heard something about this release potentially allowing for dropping of `mesa-asahi` from #3787 but that's to be tested on the respective hardware etc :)


q66:
all the "transitional" packages should be dropped from base-removed-dbg, then converted to self.renames were they still exist

q66:
and yes asahi should be added to the arm list

JamiKettunen:
All done from my side; @breakgimme can you test if this now works as a replacement for [`mesa-asahi`](https://github.com/chimera-linux/cports/pull/3787/commits/eaa486004edf612883b4f80e9ba959160f985a7f) on Apple silicon?

breakgimme:
will try once I get home (in about 2 hours or so), thanks for the ping

breakgimme:
it doesn't work with the kernel version from my pr right now (will get to upgrading and pushing everything tomorrow) but after upgrading the kernel it seems to work without any issues!
Screenshot From 2025-05-22 23-51-49

main/base-removed-dbg/template.py
```diff
@@ -109,10 +103,10 @@
     "libwoff2common",
     "libwoff2dec",
     "libwoff2enc",
-    "libxatracker",
     "libxtables",
     "lidb-progs",
     "lscpu",
+    "mesa-osmesa-libs",
```
q66, line 109: drop this, it doesn't belong there

PR #4220, opened 2025-05-15
user/jj: update to 0.29.0

z-erica:


triallax:
i did a loongarch64 cross build and it seems to work, so maybe you can remove the `broken` and hopefully it works on the builder

PR #4221, opened 2025-05-16
user/widle: new package

unrealapex:
## Description

`widle(1)` is a program that runs a command when a user has idled. This functionality is useful for screenlocking with a program like `wlock(1)` .

## Checklist

### DIFF
--- a/user/widle/template.py
+++ b/user/widle/template.py
@@ -1,0 +1,23 @@
+pkgname = "widle"
+pkgver = "1.0"
+pkgrel = 0
+build_style = "makefile"
+hostmakedepends = [
+    "pkgconf",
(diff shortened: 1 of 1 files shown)


user/widle/template.py
```diff
@@ -0,0 +1,22 @@
+pkgname = "widle"
+pkgver = "1.0"
+pkgrel = 0
+build_style = "makefile"
+hostmakedepends = [
+    "pkgconf",
```
triallax, line 6: for wayland-scanner, pulled in by wayland-devel but that's in makedepends so not strictly correct
```suggestion
    "pkgconf",
    "wayland-progs"
```

PR #4224, opened 2025-05-16
cbuild: set GOCACHE under cbuild_cache

ttyyls:
## Checklist


q66:
afaict it's not safe to have a central go cache as it does not properly track changes made in cgo-imported libraries, according to the documentation

PR #4244, opened 2025-05-19
user/links: new package

unrealapex:
## Description

Links is a Lynx-like text based browser.

## Checklist

### DIFF
--- a/user/links/template.py
+++ b/user/links/template.py
@@ -1,0 +1,19 @@
+pkgname = "links"
+pkgver = "2.30"
+pkgrel = 0
+build_style = "gnu_configure"
+# TODO: configure fails with 'conditional "am__fastdepCXX" was never defined.'???
+configure_gen = []
+hostmakedepends = ["pkgconf"]
+makedepends = [
+    "libjpeg-turbo-devel",
+    "libtiff-devel",
+    "openssl3-devel",
+    "zlib-ng-compat-devel",
+]
+pkgdesc = "Lynx-like text WWW browser"
+license = "GPL-2.0-or-later"
+url = "http://links.twibright.com"
+source = f"{url}/download/links-{pkgver}.tar.gz"
+sha256 = "7f0d54f4f7d1f094c25c9cbd657f98bc998311122563b1d757c9aeb1d3423b9e"
+hardening = ["vis", "cfi"]


user/links/template.py
```diff
@@ -0,0 +1,16 @@
...
+hostmakedepends = [
+    "libjpeg-turbo-devel",
+    "libtiff-devel",
+    "openssl3-devel",
+    "zlib-ng-compat-devel",
+]
+pkgdesc = "Lynx-like text WWW browser"
+license = "GPL-2.0-or-later"
+url = "http://links.twibright.com"
+source = f"{url}/download/{pkgname}-{pkgver}.tar.gz"
```
triallax, line 15: we don't do `{pkgname}` in `source` anymore
```suggestion
source = f"{url}/download/links-{pkgver}.tar.gz"
```

user/links/template.py
```diff
@@ -0,0 +1,16 @@
...
+pkgver = "2.30"
+pkgrel = 0
+build_style = "configure"
+configure_args = ["--prefix=/usr", "--mandir=PREFIX/share/man"]
+hostmakedepends = [
+    "libjpeg-turbo-devel",
+    "libtiff-devel",
+    "openssl3-devel",
+    "zlib-ng-compat-devel",
+]
```
triallax, line 11: libraries to be built against go in `makedepends`
```suggestion
makedepends = [
    "libjpeg-turbo-devel",
    "libtiff-devel",
    "openssl3-devel",
    "zlib-ng-compat-devel",
]
```

triallax: also forgor to mention but `pkgconf` should be in `hostmakedepends` as well

user/links/template.py
```diff
@@ -0,0 +1,16 @@
+pkgname = "links"
+pkgver = "2.30"
+pkgrel = 0
+build_style = "configure"
+configure_args = ["--prefix=/usr", "--mandir=PREFIX/share/man"]
```
triallax, line 5: autotools are used here so it should be gnu_configure (also fixes the manpage installation)
```suggestion
build_style = "gnu_configure"
# TODO: configure fails with 'conditional "am__fastdepCXX" was never defined.'???
configure_gen = []
```

(would appreciate it if anyone could point out how to fix the error so we can generate configure ourselves)

triallax:
please squash

user/links/template.py
```diff
@@ -0,0 +1,18 @@
...
+    "libjpeg-turbo-devel",
+    "libtiff-devel",
+    "openssl3-devel",
+    "zlib-ng-compat-devel",
+]
+pkgdesc = "Lynx-like text WWW browser"
+license = "GPL-2.0-or-later"
+url = "http://links.twibright.com"
+source = f"{url}/download/links-{pkgver}.tar.gz"
+sha256 = "7f0d54f4f7d1f094c25c9cbd657f98bc998311122563b1d757c9aeb1d3423b9e"
```
triallax, line 18: i tested this for a little bit and it seems to be working fine, but do please let me know if it's not
```suggestion
sha256 = "7f0d54f4f7d1f094c25c9cbd657f98bc998311122563b1d757c9aeb1d3423b9e"
hardening = ["vis", "cfi"]
```

unrealapex: No issues when testing the changes you've proposed.

PR #4255, opened 2025-05-21
user/zizmor: new package

jcgruenhage:


user/zizmor/template.py
```diff
@@ -0,0 +1,33 @@
...
+    f"https://github.com/zizmorcore/zizmor/archive/refs/tags/v{pkgver}.tar.gz"
+)
+sha256 = "6f5f4da30eb7e0fa4b7558a9418b58abd7c5ab467cb2dce330c8189a00668355"
+# Generates completions using host binaries
+options = ["!cross"]
+
+
+def post_build(self):
+    for shell in ["bash", "fish", "zsh"]:
+        with open(self.cwd / f"ruff.{shell}", "w") as f:
```
triallax, line 20: `ruff.{shell}` :)

jcgruenhage:
I've added a patch that has been merged upstream which adds nushell completions so that we can already ship them.

PR #4256, opened 2025-05-22
main/firefox: update to 138.0.4

triallax:
[ci skip]


triallax:
nvm 139 is out now :D

PR #4287, opened 2025-05-26
user/prometheus: new package

nulldutra:
## Description

Prometheus is a systems and service monitoring system. It collects metrics from configured targets at given intervals, evaluates rule expressions, displays the results, and can trigger alerts when specified conditions are observed.

## Checklist

### DIFF
--- a/user/prometheus/files/prometheus
+++ b/user/prometheus/files/prometheus
@@ -1,0 +1,13 @@
+type = process
+command = /usr/bin/prometheus \
+    --config.file /etc/prometheus/prometheus.yml \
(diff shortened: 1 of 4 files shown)


user/prometheus/files/prometheus
```diff
@@ -0,0 +1,14 @@
+type = process
+command = prometheus \
+    --config.file /etc/prometheus/prometheus.yml \
+    --storage.tsdb.path /var/db/prometheus/ \
+    --web.console.templates=/etc/prometheus/consoles \
+    --web.console.libraries=/etc/prometheus/console_libraries
+
```
triallax, line 7: please drop this empty line

PR #4299, opened 2025-05-28
user/frr: new package

rbgarga:
- **user/libzmq: new package**
- **user/rtrlib: new package**
- **user/frr: new package**


user/frr/files/frr
```diff
@@ -0,0 +1,4 @@
+type = scripted
+command = /usr/lib/frr/frrinit.sh start
+stop-command = /usr/lib/frr/frrinit.sh stop
+depends-on = network.target
```
q66: `depends-on: network.target`

also needs a "proper" target dependency like local.target, network.target on its own is not sufficient

user/frr/files/frr
```diff
@@ -0,0 +1,4 @@
+type = scripted
+command = /usr/lib/frr/frrinit.sh start
```
q66: i looked at the frrinit script and it's pretty much a sysv style thing which starts a daemon with a pidfile and doesn't supervise it

so ideally this should start the daemon properly, on the foreground, in a supervised way and should not be a scripted service, should be a process service

user/frr/files/sysusers.conf
```diff
@@ -0,0 +1,2 @@
+u frr - "FRR routing suite" /nonexistent /usr/bin/nologin
```
q66: /var/empty for the homedir

also ideally this should be underscored (_frr) if it's possible to configure things that way

user/frr/files/sysusers.conf
```diff
@@ -0,0 +1,2 @@
...
+m frr frrvty
```
q66: nothing seems to create the frrvty group? also likewise should be _frrvty

user/frr/template.py
```diff
@@ -0,0 +1,86 @@
...
+    self.install_service("^/frr")
+    # devel files are only for build usage and should not be part of pkg
+    self.uninstall("usr/include")
+    self.uninstall("usr/lib/*.so", glob=True)
+    # FRR provides multiple startup scripts, lets keep only frrinit.sh
+    self.uninstall("usr/lib/frr/frr")
+    self.uninstall("usr/lib/frr/frr-reload")
+
+
+@subpackage("frr-pythontools")
```
q66: any reason to split this at all?

user/frr/template.py
```diff
@@ -0,0 +1,86 @@
...
+def _(self):
+    self.subdesc = "python tools"
+    self.depends = [self.parent]
+    return ["usr/lib/frr/*.py"]
+
+
+@subpackage("frr-rpki")
```
q66: likewise, any reason to split it?

rbgarga: It's not desired to all users to have it installed so I believe it should stay this way

user/frr/files/frr
```diff
@@ -0,0 +1,22 @@
+type = internal
+depends-ms = frr-daemon@babeld
```
q66: use correct service style, `depends-ms:` etc

user/frr/files/frr
```diff
@@ -0,0 +1,22 @@
...
+depends-ms = frr-daemon@pathd
+depends-ms = frr-daemon@pbrd
+depends-ms = frr-daemon@pim6d
+depends-ms = frr-daemon@pimd
+depends-ms = frr-daemon@ripd
+depends-ms = frr-daemon@ripngd
+depends-ms = frr-daemon@ssd
+depends-ms = frr-daemon@staticd
+depends-ms = frr-daemon@vrrpd
+before = network.target
```
q66: as previously mentioned this does not belong here, belongs in the other services

user/frr/files/frr-daemon
```diff
@@ -0,0 +1,6 @@
+type = process
+command = /usr/lib/frr/$1 -A 127.0.0.1
+depends-on = frr-zebra
```
q66: style again

(make sure order stays good, `:` fields come after `=` fields)

user/frr/files/frr-daemon
```diff
@@ -0,0 +1,6 @@
...
+depends-on = local.target
+logfile = /var/log/frr/$1.log
+restart = true
```
q66: these are superfluous

user/frr/files/tmpfiles.conf
```diff
@@ -0,0 +1,6 @@
+d /run/frr              0750 _frr _frr -
+Z /usr/share/frr/conf/* 0640 _frr _frr
```
q66: tmpfiles should not alter modes of stuff in `/usr`

nor it's particularly desirable for it to, all `/usr` can be world readable

user/frr/template.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def post_install(self):
+    for f in [
+        "daemons",
+        "frr.conf",
+        "vtysh.conf",
+        "support_bundle_commands.conf",
+    ]:
+        self.install_file(f"tools/etc/frr/{f}", "usr/share/frr/conf")
```
q66: i would use `usr/share/etc/frr`

user/frr/template.py
```diff
@@ -0,0 +1,82 @@
...
+    self.install_file(
```
q66, line 57: need to think about this one because this affects the iproute2 suite

PR #4300, opened 2025-05-28
user/snort: new package

rbgarga:


q66:
this is all squashed together thus impossible to review

user/daq/template.py
```diff
@@ -0,0 +1,19 @@
+pkgname = "daq"
+pkgver = "3.0.20"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = ["automake", "libtool", "pkgconf"]
+makedepends = ["libpcap-devel"]
+checkdepends = ["cmocka-devel"]
+pkgdesc = "Data Acquisition library for packet I/O"
+license = "GPL-2.0-only"
```
q66, line 9: i believe this is `GPL-2.0-or-later`

user/snort/template.py
```diff
@@ -0,0 +1,35 @@
+pkgname = "snort"
+pkgver = "3.9.1.0"
+pkgrel = 0
+build_style = "cmake"
+# "-DCMAKE_POLICY_VERSION_MINIMUM=3.5",
```
q66: is this a leftover?

user/snort/template.py
```diff
@@ -0,0 +1,35 @@
...
+    "libpcap-devel",
+    "libtirpc-devel",
+    "luajit-devel",
+    "numactl-devel",
+    "openssl3-devel",
+    "pcre2-devel",
+    "util-linux-uuid-devel",
+    "xz-devel",
+]
+depends = ["bash"]
```
q66, line 24: where is the bash dependency? i don't see it

user/libdnet/patches/disable_tests.patch
```diff
@@ -0,0 +1,19 @@
+commit 2322f48
+Author: Renato Botelho do Couto <renato@netgate.com>
+Date:   Wed Sep 11 16:35:45 2024
+
+    Disable broken tests
```
q66: could use specifying why they are broken

user/snort/template.py
rbgarga: There is a bash script installed at `/usr/bin/appid_detector_builder.sh`

q66: hm, that seems like an auxiliary thing? i probably wouldn't add a hard bash dep just for that

rbgarga: It's a useful tool for anyone running snort.  If you say we should remove, I will remove it but my vote is against it

rbgarga: I can split it on a separate package that won't be installed by default

q66: if it's useful as in most people would be using it, let's keep the bash dep

i'm not really "saying" anything, i'm not familiar with this tooling beyond what i see in readmes etc., so it's more like i'm trying to get clarification and your opinion

PR #4309, opened 2025-05-30
user/diffutils: new package

wezm:
## Description

Adds GNU diffutils. I need GNU diff for the test suite of the product I work on for my job. Unfortunately BSD diff and GNU diff have incompatible forms for grouping in regexs `\(`, `\)` for GNU, `(` and `)` for BSD, which we use. So I can't easily avoid the need for GNU diff without breaking things for other colleagues.

## Checklist

### DIFF
--- a/user/diffutils/template.py
+++ b/user/diffutils/template.py
@@ -1,0 +1,17 @@
+pkgname = "diffutils"
+pkgver = "3.12"
+pkgrel = 0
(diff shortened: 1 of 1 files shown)


triallax:
not 100% sure about this but since we have gsed, ggrep, etc., maybe this should also be gdiffutils?

PR #4330, opened 2025-06-02
user/gamemode: new package

shumvgolove:
## Description

Add `gamemode` package. Useful for Steam Flatpak and such.

## Checklist

### DIFF
--- a/user/gamemode/files/gamemoded.user
+++ b/user/gamemode/files/gamemoded.user
@@ -1,0 +1,4 @@
+type = process
+command = /usr/bin/gamemoded
+log-type = buffer
+depends-on: dbus
--- a/user/gamemode/patches/posix.patch
+++ b/user/gamemode/patches/posix.patch
@@ -1,0 +1,20 @@
+diff --git a/data/gamemodelist b/data/gamemodelist
+index 392c369..e959433 100755
+--- a/data/gamemodelist
++++ b/data/gamemodelist
+@@ -1,4 +1,4 @@
+-#!/bin/bash
++#!/bin/sh
+ # Created by Sam Gleske
+ # Created Sat Jan  1 16:56:54 EST 2022
+ # MIT License - https://github.com/samrocketman/home
+diff --git a/data/gamemoderun b/data/gamemoderun
+index 573b3e4..beae444 100755
+--- a/data/gamemoderun
++++ b/data/gamemoderun
+@@ -1,4 +1,4 @@
+-#!/bin/bash
++#!/bin/sh
+ # Helper script to launch games with gamemode
+ 
+ GAMEMODEAUTO_NAME="libgamemodeauto.so.0"
--- a/user/gamemode/template.py
+++ b/user/gamemode/template.py
@@ -1,0 +1,39 @@
+pkgname = "gamemode"
+pkgver = "1.8.2"
+pkgrel = 0
+build_style = "meson"
+configure_args = [
+    "--libexecdir=/usr/lib",  # XXX drop libexec
+    "-Dwith-examples=false",
+    "-Dwith-sd-bus-provider=tangle",
+    "-Dwith-privileged-group=_gamemode",
+]
+hostmakedepends = [
+    "cmake",
+    "git",
+    "meson",
+    "ninja",
+    "pkgconf",
+]
+makedepends = [
+    "dbus-devel",
+    "inih-devel",
+    "linux-headers",
+    "tangle-devel",
+]
+pkgdesc = "Optimise Linux system performance on demand"
+license = "BSD-3-Clause"
+url = "https://github.com/FeralInteractive/gamemode"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "2b4a20894822caf04767af6a1601130d6b718fc30b8a77895607341b1674740f"
+
+
+def post_install(self):
+    self.install_service(self.files_path / "gamemoded.user")
+    self.install_sysusers(self.files_path / "sysusers.conf")
+    self.install_license("LICENSE.txt")
+
+
+@subpackage("gamemode-devel")
+def _(self):
+    return self.default_devel()
--- a/user/gamemode-devel
+++ b/user/gamemode-devel
@@ -1,0 +1,1 @@
+gamemode
--- a/user/gamemode/files/sysusers.conf
+++ b/user/gamemode/files/sysusers.conf
@@ -1,0 +1,1 @@
+g _gamemode -
--- a/user/gamemode/patches/dbus.patch
+++ b/user/gamemode/patches/dbus.patch
@@ -1,0 +1,10 @@
+diff --git a/data/dbus/com.feralinteractive.GameMode.service.in b/data/dbus/com.feralinteractive.GameMode.service.in
+index ff737ea..d4833f7 100644
+--- a/data/dbus/com.feralinteractive.GameMode.service.in
++++ b/data/dbus/com.feralinteractive.GameMode.service.in
+@@ -1,4 +1,4 @@
+ [D-BUS Service]
+ Name=com.feralinteractive.GameMode
+-Exec=@BINDIR@/gamemoded -d
++Exec=/usr/bin/dinitctl --quiet start gamemoded
+ SystemdService=gamemoded.service
--- a/user/gamemode/patches/tangle.patch
+++ b/user/gamemode/patches/tangle.patch
@@ -1,0 +1,107 @@
+diff --git a/daemon/gamemode-context.c b/daemon/gamemode-context.c
+index 1e8ed9d..458aaf1 100644
+--- a/daemon/gamemode-context.c
++++ b/daemon/gamemode-context.c
+@@ -49,7 +49,6 @@ POSSIBILITY OF SUCH DAMAGE.
+ #include <stdatomic.h>
+ #include <stdlib.h>
+ #include <sys/time.h>
+-#include <systemd/sd-daemon.h> /* TODO: Move usage to gamemode-dbus.c */
+ #include <unistd.h>
+ 
+ /**
+@@ -415,7 +414,7 @@ unlock:
+ static void game_mode_context_enter(GameModeContext *self)
+ {
+ 	LOG_MSG("Entering Game Mode...\n");
+-	sd_notifyf(0, "STATUS=%sGameMode is now active.%s\n", "\x1B[1;32m", "\x1B[0m");
++	// sd_notifyf(0, "STATUS=%sGameMode is now active.%s\n", "\x1B[1;32m", "\x1B[0m");
+ 
+ 	if (game_mode_set_governor(self, GAME_MODE_GOVERNOR_DESIRED) == 0) {
+ 		/* We just switched to a non-default governor.  Enable the iGPU
+@@ -456,7 +455,7 @@ static void game_mode_context_enter(GameModeContext *self)
+ static void game_mode_context_leave(GameModeContext *self)
+ {
+ 	LOG_MSG("Leaving Game Mode...\n");
+-	sd_notifyf(0, "STATUS=%sGameMode is currently deactivated.%s\n", "\x1B[1;36m", "\x1B[0m");
++	// sd_notifyf(0, "STATUS=%sGameMode is currently deactivated.%s\n", "\x1B[1;36m", "\x1B[0m");
+ 
+ 	/* Remove GPU optimisations */
+ 	game_mode_apply_gpu(self->stored_gpu);
+diff --git a/daemon/gamemode-dbus.c b/daemon/gamemode-dbus.c
+index 29f897a..121dad2 100644
+--- a/daemon/gamemode-dbus.c
++++ b/daemon/gamemode-dbus.c
+@@ -39,6 +39,8 @@ POSSIBILITY OF SUCH DAMAGE.
+ #ifdef USE_ELOGIND
+ #include <elogind/sd-bus.h>
+ #include <elogind/sd-daemon.h>
++#elif defined(USE_TANGLE)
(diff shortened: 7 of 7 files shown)


user/gamemode/files/gamemoded.user
```diff
@@ -0,0 +1,7 @@
+# gamemode user service
+
```
JamiKettunen: the comment and empty newline are old style and are being phased out over time for packages already in cports; please drop them from here already :)

user/gamemode/template.py
```diff
@@ -0,0 +1,42 @@
...
+    "git",
+    "meson",
+    "ninja",
+    "pkgconf"
+]
+makedepends = [
+    "linux-headers",
+    "elogind-devel",
+    "inih-devel",
+    "dbus-devel"
```
JamiKettunen: sort the list alphabetically

user/gamemode/files/gamemoded.user
```diff
@@ -0,0 +1,7 @@
...
+type = process
+command = /usr/bin/gamemoded
+depends-on = dbus
+depends-on = graphical.target
```
JamiKettunen: seems to work when enabled just fine with `before: login.target` instead (of `depends-on = graphical.target`) too on my end unless you have some other insight, can't see anything odd in `/var/log/turnstiled.log` following reboot

user/gamemode/patches/posix.patch
```diff
@@ -0,0 +1,20 @@
...
+ # Created by Sam Gleske
+ # Created Sat Jan  1 16:56:54 EST 2022
+ # MIT License - https://github.com/samrocketman/home
+diff --git a/data/gamemoderun b/data/gamemoderun
+index 573b3e4..beae444 100755
+--- a/data/gamemoderun
++++ b/data/gamemoderun
+@@ -1,4 +1,4 @@
+-#!/bin/bash
++#!/usr/bin/env sh
```
JamiKettunen: fwiw these probably can be upstreamed as well since it shouldn't need bash elsewhere if it works for us, perhaps you already planned to do that too

user/gamemode/template.py
```diff
@@ -0,0 +1,38 @@
+pkgname = "gamemode"
+pkgver = "1.8.2"
+pkgrel = 0
+build_style = "meson"
+configure_args = [
+    "-Dwith-examples=false",
```
JamiKettunen, line 7: this also leaves out the default `gamemode.ini` config, I suppose this is fine and expected? according to `man 8 gamemoded` it would load the installed `/usr/share/gamemode/gamemode.ini` first followed by the others

user/gamemode/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "inih-devel",
+    "dbus-devel"
+]
+pkgdesc = "Optimise Linux system performance on demand"
+license = "BSD-3-Clause"
+url = "https://github.com/FeralInteractive/gamemode"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "2b4a20894822caf04767af6a1601130d6b718fc30b8a77895607341b1674740f"
+
+def post_install(self):
```
JamiKettunen, line 31: run the file through a formatter as per https://github.com/chimera-linux/cports/blob/master/Packaging.md#quality_requirements

user/gamemode/template.py
```diff
@@ -0,0 +1,38 @@
...
+    self.install_service(self.files_path / "gamemoded.user")
+    self.install_sysusers(self.files_path / "sysusers.conf")
+    self.install_license("LICENSE.txt")
+
+@subpackage("gamemode-static")
+def _(self):
+    return self.default_static()
```
JamiKettunen: `-static` should practically never have to be added manually like this, you just missed adding a `-devel` subpackage (which is clear due to the present `/usr/lib/pkgconfig/gamemode.pc` in `gamemode` package etc) which the [shown error also hints at these days](https://github.com/chimera-linux/cports/blob/ee5e30a/src/cbuild/hooks/pkg/098_lint.py#L101-L107)
image

user/gamemode/template.py
```diff
@@ -0,0 +1,38 @@
+pkgname = "gamemode"
+pkgver = "1.8.2"
+pkgrel = 0
+build_style = "meson"
+configure_args = [
```
JamiKettunen, line 5: we'll probably want this due to another ongoing migration to get rid of `/usr/libexec`
```suggestion
configure_args = [
    "--libexecdir=/usr/lib",  # XXX drop libexec
```

user/gamemode/files/gamemoded.user
```diff
@@ -0,0 +1,7 @@
+# gamemode user service
+
+type = process
+command = /usr/bin/gamemoded
+depends-on = dbus
```
JamiKettunen: `depends-on: dbus` and afaik should be at the end of file as per modern style similar to above

user/gamemode/template.py
```diff
@@ -0,0 +1,38 @@
...
+    "-Dwith-examples=false",
+    "-Dwith-sd-bus-provider=elogind",
+    "-Dwith-systemd-user-unit=false",
+    "-Dwith-systemd-group=false",
```
JamiKettunen: declaring any `-Dwith-systemd-*` isn't needed since we aren't using `-Dwith-sd-bus-provider=systemd`; see https://github.com/FeralInteractive/gamemode/blob/1.8.2/meson.build#L108-L131 for their only usage

user/gamemode/files/gamemoded.user
shumvgolove: Yep, thanks for the suggestion. My reasoning was that `gamemode` would probably be run by the person who already has a full-fledged desktop environment :)

user/gamemode/patches/posix.patch
shumvgolove: Yep, I plan to upstream this changes so we could drop this patch on the next `gamemode` release.

user/gamemode/template.py
shumvgolove: Hm, I was looking at the [Arch Linux](https://gitlab.archlinux.org/archlinux/packaging/packages/gamemode/-/blob/main/PKGBUILD?ref_type=heads#L42) and [Alpine Linux](https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/gamemode/APKBUILD#L24) packages, they also disable examples.

I _think_ this should be fine. In my case I just copy-pasted the configuration to `$XDG_CONFIG_HOME/gamemode.ini` and changed it to my liking.

user/gamemode/template.py
shumvgolove: Thanks! I was a bit confused: should I follow a hint or an error message?

user/gamemode/files/gamemoded.user
```diff
@@ -0,0 +1,5 @@
+type = process
+command = /usr/bin/gamemoded
+log-type = buffer
+before = login.target
```
q66: is it really necessary for this to hold up login

user/gamemode/files/gamemoded.user
```diff
@@ -0,0 +1,5 @@
...
+depends-on = dbus
```
q66: `depends-on: dbus`

user/gamemode/template.py
```diff
@@ -0,0 +1,39 @@
+pkgname = "gamemode"
+pkgver = "1.8.2"
+pkgrel = 0
+build_style = "meson"
+configure_args = [
+    "--libexecdir=/usr/lib",  # XXX drop libexec
+    "-Dwith-examples=false",
+    "-Dwith-sd-bus-provider=elogind",
```
q66: this only seems to use systemd/elogind for sd-bus, so it should be possible to make it use `tangle` instead of `elogind` with minimal patching

user/gamemode/files/gamemoded.user
JamiKettunen: probably not, that's the other dependency I saw user services defining if not `graphical.target`, would just `depends-on: dbus` suffice or do you have something else in mind?

q66: it doesn't really need anything (other than dbus ofc)

user/gamemode/template.py
shumvgolove: Hm, it seems that `gamemode` also uses `<systemd/sd-daemon.h>`, but current version of `tangle` [doesn't provide that header](https://pkgs.chimera-linux.org/contents?name=tangle-devel&repo=main&branch=current&arch=x86_64), although there is a commit in https://github.com/chimera-linux/tangle/commit/10bec005fad59a2a88e8a8a5015efbbb45357b76

user/gamemode/files/gamemoded.user
shumvgolove: Sorry, I thought this is the same thing.

user/gamemode/template.py
JamiKettunen: the original error message in this case can be confusing even if technically correct, that's why I previously added the hint spelling out what the likely cause for it showing up actually is; this is the case in other places of cbuild too

user/gamemode/template.py
q66: as far as i can tell it's not necessary, it's just included for no reason

user/gamemode/patches/posix.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/data/gamemodelist b/data/gamemodelist
+index 392c369..e959433 100755
+--- a/data/gamemodelist
++++ b/data/gamemodelist
+@@ -1,4 +1,4 @@
+-#!/bin/bash
++#!/usr/bin/env sh
```
q66: `#!/bin/sh`, ditto elsewhere

user/gamemode/template.py
shumvgolove: Without providing the header, I've got several compilation errors with the same root cause:

```log
../daemon/gamemode-dbus.c:691:2: error: call to undeclared function 'sd_notifyf'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  691 |         sd_notifyf(0, "STATUS=%sGameMode is ready to be activated.%s\n", "\x1B[1;36m", "\x1B[0m");
      |         ^
```

or:

```log
../daemon/gamemoded.c:339:2: error: call to undeclared function 'sd_notify'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  339 |         sd_notify(0, "STATUS=GameMode is quitting naturally...\n");
      |         ^
```

q66: oh, just comment these out, they're useless

user/gamemode/patches/posix.patch
shumvgolove: No problem, but is there any reason to prefer this over `#!/usr/bin/env sh`, so I have better understanding?

q66: `/bin/sh` is a posix guarantee while `/usr/bin/env sh` is an antipattern and is not

shumvgolove:
Thanks for the review!

I force-pushed the requested changes.

user/gamemode/files/gamemoded.user
```diff
@@ -0,0 +1,4 @@
+type = process
```
sewnie, line 1: What's the point of a `_gamemode` user if `gamemoded` will run as a user service? Additionally, the user doesn't seem to serve any practical purpose. The CPU governor is still untouched.

sewnie:
```
ERROR: External process failed with exit code 127
ERROR: Output was: 
ERROR: Failed to update cpu governor policy
Requesting update of split_lock_mitigate to 0
Error executing command as another user: Not authorized
``

Comment (shumvgolove) [reply user/gamemode/files/gamemoded.user:1]: This is privileged group, allowing the user to change nice level of process: https://github.com/FeralInteractive/gamemode/blob/master/data/pam_limits/10-gamemode.conf.in

Comment (sewnie) [reply user/gamemode/files/gamemoded.user:1]: > What's the point of a _gamemode user if gamemoded will run as a user service? 

Again, arguably more important values such as the CPU governor and split lock mitigation are still not used.

Comment (shumvgolove) [reply user/gamemode/files/gamemoded.user:1]: This group also enables the access of changing these parameters too: https://github.com/FeralInteractive/gamemode/blob/master/data/polkit/rules.d/gamemode.rules.in#L11

Comment (sewnie) [reply user/gamemode/files/gamemoded.user:1]: You listed a Polkit rule which allows the group to give access to GameMode's utilities, except they are not distributed in the `gamemode` package, so it is pointless. I already showed you that such a rule doesn't take effect in real world usage.

Comment (shumvgolove) [reply user/gamemode/files/gamemoded.user:1]: No idea, it worked for me when my user was in `_gamemode` group when testing with:

```
gamemoded -t
```

I can't test it right now, but will check it again in a couple of days.

sewnie: > No idea, it worked for me when my user was in _gamemode group when testing with:

I did not know i have to do this. This is mentioned for example on seatd, but for this can't you simply run the gamemoded daemon as `_gamemode` or not?

shumvgolove: Not sure, but this is a similar behavior like in other distros (`Arch Linux`, `Alpine Linux`). From the [Arch Linux Wiki](https://wiki.archlinux.org/title/GameMode):

> Add yourself to the `gamemode` [user group](https://wiki.archlinux.org/title/User_group). Without it, the GameMode user daemon will not have rights to change CPU governor or the niceness of processes.

For the Chimera Linux this group is set to `_gamemode`, which follows the general guidelines.

q66:
thanks, apologies it took me so long but i had a bunch of stuff going on and i'm going through all the PRs now

PR #3700, opened 2025-03-04
user/ocaml: new package

Chameko:
## Description

Provides the main implementation for the ocaml language

## Checklist

Before this pull request is reviewed, certain conditions must be met.

The following must be true for all changes:

- [x] I have read [CONTRIBUTING.md](https://github.com/chimera-linux/cports/blob/master/CONTRIBUTING.md)

The following must be true for template/package changes:

- [x] I have read [Packaging.md](https://github.com/chimera-linux/cports/blob/master/Packaging.md#quality_requirements)
- [x] I have built and tested my changes on my machine

The following must be true for new package submissions:

- [x] I will take responsibility for my template and keep it up to date

PR #4369, opened 2025-06-10
user/chibi-scheme: new package

druimalban:
## Description

Add the Chibi Scheme system

## Checklist

### DIFF
--- a/user/chibi-scheme/template.py
+++ b/user/chibi-scheme/template.py
@@ -1,0 +1,17 @@
+pkgname = "chibi-scheme"
+pkgver = "0.11"
+pkgrel = 0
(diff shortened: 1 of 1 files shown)


druimalban:
One question about `tool_flags` - `Packaging.md` says, that we should avoid overwriting, say, ;CFLAGS; and use the default, passing the change afterwards. I couldn't work out how to do this (or if my usage of `tool_flags` doesn't do this). 

Unsetting compiler optimisations is necessary here as the test suite fails on quotient / division otherwise.

Further, unsetting the `int` hardening option is necessary. This made me feel uneasy, but it is similar to the open ocaml port: https://github.com/chimera-linux/cports/pull/3700.

(I have put comments in the template file but I can remove them if these are unhelpful.)

PR #4372, opened 2025-06-11
user/nushell: update to 0.105.1

alecStewart1:
## Description

Just keep nushell up-to-date with current releases. 0.105.0 adds the meaningful changes, 0.105.1 is a revision of the version for the coreutils crate. 

https://www.nushell.sh/blog/2025-06-10-nushell_0_105_0.html

https://www.nushell.sh/blog/2025-06-10-nushell_0_105_1.html

## Checklist

### DIFF
--- a/user/nushell/template.py
+++ b/user/nushell/template.py
@@ -1,5 +1,5 @@
 pkgname = "nushell"
-pkgver = "0.104.1"
+pkgver = "0.105.0"
 pkgrel = 0
 build_style = "cargo"
 make_check_args = [
@@ -20,7 +20,7 @@ pkgdesc = "Shell with a focus on structured data"
 license = "MIT"
 url = "https://www.nushell.sh"
 source = f"https://github.com/nushell/nushell/archive/refs/tags/{pkgver}.tar.gz"
-sha256 = "3dafca8bf892f5a2afaac1122a88a7eb7f22a0b62ef901f550173a11d5cbdf6e"
+sha256 = "2c2f4062820602a773682c5b9189d52ddc03793bc2a5d367d88f73c604f290d3"
 
 
 def post_install(self):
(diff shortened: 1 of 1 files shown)


user/nushell/template.py
```diff
@@ -1,5 +1,5 @@
 pkgname = "nushell"
-pkgver = "0.104.1"
+pkgver = "0.105.1"
 pkgrel = 0
 build_style = "cargo"
```
tranzystorekk, line 4:
```suggestion
build_style = "cargo"
make_build_args = [
    "--no-default-features",
    "--features=plugin,trash-support,sqlite,native-tls",
]
make_install_args = [*make_build_args]
```

https://www.nushell.sh/blog/2025-06-10-nushell_0_105_0.html#no-more-openssl-toc

alecStewart1: Is there any particular reason nushell _needs_ to be built with openssl? Like a security thing, or just so that it works on most if not all architectures?

tranzystorekk: likely both, but I also assume once we've opted for ossl for the package, we should keep it this way as long as it's maintainable

alecStewart1: Makes sense.

alecStewart1:
Sorry. Unfortunately, if I want to update anything on my work laptop (Mac) I can't quite because the repo has this weird thing with `networkmanager.conf` and `NetworkManager.conf`. If you clone it down on a Mac, you'll see:

```
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'main/networkmanager/files/NetworkManager.conf'
  'main/networkmanager/files/networkmanager.conf'
```

So I thought "well, I could wait until I'm home, or, since this is just a minor update, I could use Git's editor instead."

I'll close the PR and create a proper branch.

PR #4375, opened 2025-06-12
user/avahi2dns: new package

z-erica:


ItsHarper:
Thanks for PRing this!

@q66 mDNS support is something that regular users will expect to just work. I think Chimera should strongly consider including this package in the base install in the future (with appropriate configuration already set up of course).

Here's the documentation for how Alpine Linux recommends using it (the need for this is caused by our usage of musl): https://wiki.alpinelinux.org/wiki/MDNS

user/avahi2dns/files/avahi2dns
```diff
@@ -0,0 +1,5 @@
+type = process
+command = /usr/bin/avahi2dns
+depends-on: avahi-daemon
```
ItsHarper, line 3: I think you also need to specify that this package depends on avahi in a `depends` field in `template.py`

z-erica: the dependency is already tracked through the dinit service file

```
0:00:43.937 => avahi2dns-dinit-0.1.0-r0: changed dependencies:
0:00:43.937 =>   +avahi2dns=0.1.0-r0
0:00:43.937 =>   +dinit-chimera
0:00:43.937 =>   +svc:avahi-daemon
0:00:43.937 =>   +svc:dbus
0:00:43.937 =>   +svc:local.target
```

but it probably makes sense to make the base package (not `-dinit`) depend on the avahi base package anyway...

ItsHarper: I'm still learning about Chimera, but is that how that works? It lists a dependency on `svc:avahi-daemon` (the dinit service), but it doesn't list the actual `avahi` package.

z-erica: the `avahi-dinit` package both provides that service and depends on `avahi`:
```
avahi-dinit-0.8-r11 provides:
svc:avahi-daemon=0.8-r11

avahi-dinit-0.8-r11 depends on:
avahi=0.8-r11
dinit-chimera
```

(which means installing `avahi2dns-dinit` transitively pulls in `avahi`)

ItsHarper: Gotcha, that makes sense. Thanks for taking the time to explain!

Do you happen to know what happens if multiple packages say they provide a service with the same name?

z-erica: i know there's a `provider_priority` mechanism, but beyond that, not sure? i'd imagine it'd prioritize a provider that's in `/etc/world` for example

ypopovych:
Hello one more time.

I configured it on my machine, works awesome. I have on suggestion how to improve current setup. I added a configuration file to mine.

What I did:

1. Inside dinit service file I added env file and changed command line like that:
```config
command = /usr/bin/avahi2dns --domains $/DOMAINS --addr $BIND --port $PORT --timeout $TIMEOUT $/DEBUG
env-file = /etc/default/avahi2dns
```
2. And created `/etc/default/avahi2dns` file with settings:
```config
DOMAINS=home internal intranet lan local private test
BIND=localhost
PORT=5354
TIMEOUT=0
# DEBUG=--debug
``` 

This will allow to configure service. Because to work properly it shouldn't ran on port 53, because port 53 will be used by `dnsmasq` or `unbound` to serve all DNS requests and route only local ones to avahi.

I hope it makes sense.

q66:
envfiles for configuration are an antipattern so don't do that

if you want to source an /etc/default file do it with a shell script wrapper for command

q66:
or even better: just don't

anyone wanting to configure this can create an /etc/dinit.d/avahi2dns and make it contain something like so:

```
@include /usr/lib/dinit.d/avahi2dns
command += --whatever --args --you --want
```

ypopovych:
@q66 interesting idea with dinit overrides.  Thanks!

user/avahi2dns/template.py
```diff
@@ -0,0 +1,17 @@
+pkgname = "avahi2dns"
+pkgver = "0.1.0"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go"]
+makedepends = ["avahi-dinit"]
```
q66: `makedepends = ["avahi", "dinit-chimera", "dinit-dbus"]`

user/avahi2dns/template.py
```diff
@@ -0,0 +1,17 @@
...
+depends = ["avahi", "dbus"]
```
q66: drop the dbus rundep

user/avahi2dns/template.py
z-erica: done

user/avahi2dns/template.py
z-erica: also done

PR #4378, opened 2025-06-13
user/oils-for-unix: new package

z-erica:


user/oils-for-unix/template.py
```diff
@@ -0,0 +1,39 @@
...
+sha256 = "34fdb53f1dbd2b8742ef74cd3a46d87abfcda18d0d645f29762c4fddb9a2b3ac"
+# check: no tests in tarball
+options = ["!check"]
+
+
+def init_configure(self):
+    self.configure_args += [
+        "--readline",
+        f"{self.profile().sysroot}/usr",
+    ]
```
triallax: i think this can go at the top level? `self.profile().sysroot` should be available there

user/oils-for-unix/template.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def build(self):
+    self.do("_build/oils.sh", "--cxx", self.get_tool("CXX"))
+
+
+def install(self):
+    ovm_name = list(self.cwd.glob("_bin/*"))[0].name
```
triallax: minor nit
```suggestion
    ovm_name = next(self.cwd.glob("_bin/*")).name
```

z-erica: was referencing the existing `main/python-tap.py`, but sure :P

user/oils-for-unix/template.py
z-erica: so it is!

z-erica: some reference on which template fields are available at the top level exactly could be nice....

triallax: agreed, i had to check other templates to find out

PR #4382, opened 2025-06-13
 user/dolphin-emu: update to 2506a 

z-erica:


user/dolphin-emu/patches/system-libs.patch
```diff
@@ -1,12 +1,12 @@
...
-@@ -672,10 +672,10 @@
+--- a/CMakeLists.txt	2025-06-05 08:52:15.000000000 +0200
++++ b/CMakeLists.txt	2025-06-13 17:47:47.334890260 +0200
+@@ -689,10 +689,10 @@
  
  dolphin_find_optional_system_library_pkgconfig(ZSTD libzstd>=1.4.0 zstd::zstd Externals/zstd)
  
--add_subdirectory(Externals/zlib-ng)
-+dolphin_find_optional_system_library_pkgconfig(ZLIB zlib-ng ZLIB::ZLIB Externals/zlib-ng)
+-dolphin_find_optional_system_library_pkgconfig(ZLIB zlib>=1.3.1 ZLIB::ZLIB Externals/zlib-ng)
```
triallax: hm, is this still necessary? `zlib-ng-compat-devel` provides `pc:zlib` which i think should satisfy the original dep

z-erica: hmm guess not!

PR #4399, opened 2025-06-17
main/qbittorrent: make service depend on local.target

z-erica:
was giving me problems on boot
```
******** Information ********
Unable to bind to IP: ::1, port: 8090. Reason: The address is not available
To fix the error, you may need to edit the config file manually.
```


triallax:
`load-options` should also be switched to use `:` and moved to the bottom

PR #4418, opened 2025-06-22
user/krdp: new package

Biswa96:


JamiKettunen:
were you able to setup and use `Remote Desktop` from system settings with this? my look into https://invent.kde.org/plasma/krdp/-/blob/v6.4.0/src/kcm/kcmkrdpserver.cpp would suggest no

PR #4432, opened 2025-06-24
 user/fonts-tex-gyre-*-math-otf: new package 

z-erica:


user/fonts-tex-gyre-bonum-math-otf/template.py
```diff
@@ -0,0 +1,14 @@
+pkgname = "fonts-tex-gyre-bonum-math-otf"
+pkgver = "1005"
+pkgrel = 0
+pkgdesc = "Math companion for TeX Gyre Bonum"
+license = "LPPL-1.3c"
```
triallax: apparently the license is strictly speaking the GFL, which is "legally equivalent" to this one but with an extra non-legally binding clause

i don't know if it really matters that much, but maybe we specifically mention that one (ditto for the other templates too)?

```suggestion
license = "custom:GFL"
```

z-erica: hmm sure... i just figured we could get away with something that's already in our list

triallax:
hm, why was this closed? i was going to merge it

PR #4436, opened 2025-06-25
main/firefox: make it compile on ARMv7

Gelbpunkt:
## Description

Crashreporter fails to compile just like loongarch64, debug info needs to be disabled entirely to not OOM during linking.

## Checklist


Gelbpunkt:
Need to investigate why Thunderbird still OOMs, will apply the same fixes for both, so marking as draft for now

PR #4439, opened 2025-06-25
user/difftastic: update to 0.64.0

Biswa96:
Install man page also.

### DIFF
--- a/user/difftastic/template.py
+++ b/user/difftastic/template.py
@@ -1,5 +1,5 @@
 pkgname = "difftastic"
-pkgver = "0.63.0"
+pkgver = "0.64.0"
 pkgrel = 0
 build_style = "cargo"
 hostmakedepends = [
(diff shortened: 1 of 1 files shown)


Biswa96:
Do you mean this commit https://github.com/Wilfred/difftastic/commit/7628f5f51ca209cb3de752dc7ed59bdf6658e63b ? It seems to fix an issue. I am not sure if changing that will break the program.

triallax:
i've tested with the system allocator (which is mimalloc for the record) using the test files in https://github.com/Wilfred/difftastic/pull/297, and `difft` just dies with `SIGKILL` after a few seconds, so nice (vendored jemalloc works fine)

using the vendored jemalloc is not something we want however due to https://github.com/Wilfred/difftastic/issues/850, maybe you could make it link against our `jemalloc-devel` instead as it already handles this issue

triallax:
update merged in #4494, thanks!

PR #4476, opened 2025-07-02
user/hare*: update to 0.25.2

ttyyls:
## Checklist


user/hare-update/template.py
```diff
@@ -0,0 +1,16 @@
+pkgname = "hare-update"
+pkgver = "0.25.2.0"
+pkgrel = 0
+build_style = "makefile"
+make_install_args = ["LIBEXECDIR=/usr/lib"]
```
triallax:
```suggestion
make_install_args = ["LIBEXECDIR=/usr/lib"] # XXX libexecdir
```
ditto for the other templates

user/hare/template.py
```diff
@@ -32,10 +40,16 @@
...
+    make_build_args += ["HARE=hare", "HAREDOC=haredoc"]
 else:
-    make_build_args += ["HARE=.bin/hare", "HAREC=/usr/bin/harec"]
+    make_build_args += [
+        "HARE=.bin/hare",
+        "HAREC=/usr/bin/harec",
+        "HAREDOC=.bin/haredoc",
+    ]
+
+self.make_jobs = 1
```
triallax: `!parallel` in options, but why is this needed?

user/hare-update/template.py
ttyyls: why?

triallax: just a small convention, it's done pretty much everywhere else too

ttyyls: i don't see the point but i added it anyway

PR #4494, opened 2025-07-04
user/difftastic: update to 0.64.0

z-erica:


triallax:
see https://github.com/chimera-linux/cports/pull/4439, in particular https://github.com/chimera-linux/cports/pull/4439#issuecomment-3036635103

triallax:
i mean, sure, but jemalloc does seem to perform significantly better, at least on these large files (on my system it takes "only" about 730 MB as opposed to mimalloc just completely failing)

still i wouldn't be opposed to using system allocator anyway i suppose

PR #4503, opened 2025-07-05
main/coturn: update to 4.7.0

ttyyls:
## Checklist

### DIFF
--- a/main/coturn/files/tmpfiles.conf
+++ b/main/coturn/files/tmpfiles.conf
@@ -2,4 +2,4 @@
 
 d /var/lib/coturn 0755 _coturn _coturn -
 C /var/lib/coturn/turndb - - - - /usr/share/turnserver/turndb
-z /var/lib/coturn/turndb 0644 _coturn _coturn - -
+Z /var/lib/coturn/turndb 0644 _coturn _coturn - -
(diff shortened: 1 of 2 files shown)


main/coturn/files/tmpfiles.conf
```diff
@@ -2,4 +2,4 @@
 
 d /var/lib/coturn 0755 _coturn _coturn -
 C /var/lib/coturn/turndb - - - - /usr/share/turnserver/turndb
-z /var/lib/coturn/turndb 0644 _coturn _coturn - -
+Z /var/lib/coturn/turndb 0644 _coturn _coturn - -
```
triallax, line 5: just checking, what does this change do?

ttyyls: last change in  https://github.com/chimera-linux/cports/commit/aa46c11938c3d39471461f0cf988776ba8871649 introduced inconsistency where the turndb was installed and ownded effectively by root, then moved to a directory owned by root and that directory then changes owner to _coturn but that doesn't apply on turndb itself.

before the fix introduced here apk would complain on every upgrade with the following
```
sd-tools-0.99.0-r4.trigger: Executing script...
sd-tools-0.99.0-r4.trigger: Creating group '_coturn' with GID 987.
sd-tools-0.99.0-r4.trigger: Creating user '_coturn' (Coturn user) with UID 987 and GID 987.
sd-tools-0.99.0-r4.trigger: Detected unsafe path transition /var/lib/coturn (owned by 987) => /var/lib/coturn/turndb (owned by 0) during canonicalization of var/lib/coturn/turndb.
```

PR #4508, opened 2025-07-06
main/bash: update to 5.3, main/readline: update to 8.3

triallax:
rolled readline update here since bash seems to fail to build without it


q66:
leave bash alone, i'll do it myself

PR #4513, opened 2025-07-06
user/listenbrainz-mpd: update to 2.3.9

JetpackJackson:
## Checklist

### DIFF
--- a/user/listenbrainz-mpd/template.py
+++ b/user/listenbrainz-mpd/template.py
@@ -1,5 +1,5 @@
 pkgname = "listenbrainz-mpd"
-pkgver = "2.3.8"
+pkgver = "2.3.9"
 pkgrel = 0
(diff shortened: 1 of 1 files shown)


user/listenbrainz-mpd/template.py
```diff
@@ -1,5 +1,5 @@
 pkgname = "listenbrainz-mpd"
-pkgver = "2.3.8"
+pkgver = "2.3.9"
 pkgrel = 0
 build_style = "cargo"
 hostmakedepends = ["cargo", "pkgconf"]
```
triallax: why `cargo` and not `cargo-auditable`? simply an oversight?

JetpackJackson: Yeah, an oversight. Lemme make sure it builds before I push that change.

PR #4533, opened 2025-07-10
main/wine: enable on aarch64

Biswa96:


Biswa96:
I have tested wine in Raspberry Pi 4. Here is a screenshot of that with notepad++ from a ssh session.

image

main/wine/template.py
```diff
@@ -68,32 +67,49 @@
...
-            "usr/bin/wmc",
-            "usr/bin/wrc",
-            "usr/lib/wine/i386-windows/*.a",
-            "usr/lib/wine/x86_64-unix/*.a",
-            "usr/lib/wine/x86_64-windows/*.a",
-        ]
-    )
+
+    # llvm-strip/objcopy cannot handle windows .a's
+    match self.rparent.profile().arch:
```
q66: why do all this stuff? you can just change the wildcard to a looser one, like `"usr/lib/wine/*-*/*.a"`

Biswa96: ok, understood. I tried to follow the syntax in other templates.

main/wine/template.py
```diff
@@ -1,11 +1,10 @@
 pkgname = "wine"
 pkgver = "10.11"
 pkgrel = 0
-archs = ["x86_64"]
+archs = ["x86_64", "aarch64"]
```
q66: sort alphabetically

Biswa96: done. thanks.

PR #4537, opened 2025-07-12
user/gomplate: new package

ayushnix:
## Description

gomplate is a go template renderer. 

## Checklist

### DIFF
--- a/user/gomplate/template.py
+++ b/user/gomplate/template.py
@@ -1,0 +1,22 @@
+pkgname = "gomplate"
+pkgver = "4.3.2"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    "-ldflags",
+    f"-X github.com/hairyhenderson/gomplate/v4/version.Version=v{pkgver}",
+    "./cmd/gomplate",
+]
+hostmakedepends = ["go"]
+depends = ["ca-certificates"]
+pkgdesc = "Template renderer with datasources"
+license = "MIT"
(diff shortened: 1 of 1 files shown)


user/gomplate/template.py
```diff
@@ -0,0 +1,22 @@
...
+]
+hostmakedepends = ["go"]
+depends = ["ca-certificates"]
+pkgdesc = "Template renderer with datasources"
+license = "MIT"
+url = "https://github.com/hairyhenderson/gomplate"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "d15c66230d72bdc13b0155f28d391c55cac45b7fdbe1ff4a73db8ee263471a3d"
+# some checks need internet connection
+options = ["!check"]
```
triallax: i checked and it's not that many tests that need network, we could skip only those ones (see e.g. `user/glow`)

ayushnix: I had to skip a considerable amount of tests for the `check` stage to finish successfully. 😄

triallax: ok yeah wow, somehow they didn't seem like that many to me when checking the output;-;

triallax: this doesn't look very great so i think we probably go back to skipping tests, sorry for making you go through the trouble in vain :/

ayushnix: no worries, thanks for merging it

PR #4539, opened 2025-07-13
user/diff-so-fancy: new package

unrealapex:
## Description
diff-so-fancy is a diff colorizer.

## Checklist


user/diff-so-fancy/patches/use-usr-share-lib-dir.patch
```diff
@@ -0,0 +1,11 @@
+--- a/diff-so-fancy	2023-06-09 11:05:43.000000000 -0700
++++ b/diff-so-fancy	2025-07-12 14:25:40.960896422 -0700
+@@ -11,7 +11,7 @@
+ use File::Spec;                                             # For catdir
+ use File::Basename;                                         # For dirname
+ use Cwd qw(abs_path);                                       # For realpath()
+-use lib dirname(abs_path(File::Spec->catdir($0))) . "/lib"; # Add the local lib/ to @INC
++use lib "/usr/share/diff-so-fancy";
```
triallax, line 8: i don't know much about perl packaging, why put this in /usr/share and not /usr/lib?

unrealapex: I'm not certain if it's the best place, but other distros(Gentoo, Arch) use this location.

triallax:
there are also tests, let's run those in `check`

user/diff-so-fancy/template.py
```diff
@@ -0,0 +1,47 @@
...
+_bats_assert_pkgver = "2.1.0"
+source = [
+    f"{url}/archive/refs/tags/v{pkgver}.tar.gz",
+    f"{_bats_core_url}/archive/refs/tags/v{_bats_core_pkgver}.tar.gz",
+    f"{_bats_support_url}/archive/refs/tags/v{_bats_support_pkgver}.tar.gz",
+    f"{_bats_assert_url}/archive/refs/tags/v{_bats_assert_pkgver}.tar.gz",
+]
+source_paths = [
+    ".",
+    "test/bats",
```
triallax, line 23: we have bats packaged already, and we can also look at packaging bats-support and bats-assert (i actually had templates for both at some point but nuked them when i didn't need them anymore, they're trivial so i can bring them back easily though)

Issue #4548, opened 2025-07-15
windows programs that use audio input fail to start under recent wine


z-erica:
will update as i try more versions:
* `10.10` does not work
* `10.0` is safe
* `10.5` is safe
* `10.7` does not work
* `10.6` does not work
* commit `5d14c97623033aaddc73aded0b4d6427723844dd` is good
* commit `3c83127a1d8f6a9faff53460da053366065fcf7d` is bad
* commit `a6aa0108f365b1465bc594a053af84116a28ce34` is bad
* commit `1b02b055a7b2a803949129b18515e0e6aa2082f0` is good
* commit `695b2a64d6431503a8b6c773be90d9bf92673a2b` is good
* commit `c42be7921b46c5557f986fd7ca93d2abc31dcd5e` is good
* commit `78fe10140330be75c7c15c2f00add82b179a11bd` is good
* commit `f7044c4b8f3ab02ca8070cb37c7cc90e311d512f` is bad
* commit `6d73d8837420cd97fe1c844b82e3375ccd1548dd` is good

first bad commit is https://gitlab.winehq.org/wine/wine/-/commit/f7044c4b8f3ab02ca8070cb37c7cc90e311d512f

triallax:
#4551

PR #4551, opened 2025-07-16
 main/wine: revert problematic commit 

z-erica:
see #4548, reported to upstream in https://bugs.winehq.org/show_bug.cgi?id=58508


triallax:
i mean this has the bump too but sure

triallax:
rebase + undraft?

PR #4553, opened 2025-07-16
user/darkman: new package

zacharyburnett:
## Description

[`darkman`](https://gitlab.com/WhyNotHugo/darkman) runs user-defined scripts at sunrise and sunset (based on your location derived via `geoclue` or manually specified coordinates), to automatically toggle "light mode" and "dark mode"

## Checklist

### DIFF
--- a/user/darkman/files/darkman.user
+++ b/user/darkman/files/darkman.user
@@ -1,0 +1,5 @@
+type = process
+command = /usr/bin/darkman run
+smooth-recovery = yes
+log-type = buffer
+depends-on: graphical.target
--- a/user/darkman/template.py
+++ b/user/darkman/template.py
@@ -1,0 +1,61 @@
+pkgname = "darkman"
+pkgver = "2.0.1"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    f"-ldflags=-X main.Version=v{pkgver}",
+    "./cmd/darkman",
+]
+hostmakedepends = [
+    "go",
+    "scdoc",
+]
+pkgdesc = "Control dark-mode and light-mode transitions"
+license = "ISC"
+url = "https://gitlab.com/WhyNotHugo/darkman"
+source = f"{url}/-/archive/v{pkgver}/darkman-v{pkgver}.tar.gz"
+sha256 = "4d87ee5dcefcd237be43d2b3d27bea658d23ebe14b5b5951bc67942f1ec839e5"
+# builds completions with generated binary
+options = ["!cross"]
+
+
+def post_build(self):
+    with open(self.cwd / "darkman.1.scd", "rb") as scd_file:
+        with open(self.cwd / "darkman.1", "w") as man_file:
+            self.do("scdoc", input=scd_file.read(), stdout=man_file)
+
+    for shell in ["bash", "fish", "zsh"]:
+        with open(self.cwd / f"darkman.{shell}", "w") as comp_file:
+            self.do(
+                "build/darkman",
+                "completion",
+                shell,
+                stdout=comp_file,
+            )
+
+
+def post_install(self):
+    self.install_license("LICENCE")
+    self.install_files(
+        "contrib/dbus/nl.whynothugo.darkman.service",
+        "usr/share/dbus-1/services",
+    )
+    self.install_files(
+        "contrib/dbus/org.freedesktop.impl.portal.desktop.darkman.service",
+        "usr/share/dbus-1/services",
+    )
+    self.install_files(
+        "contrib/portal/darkman.portal",
+        "usr/share/xdg-desktop-portal/portals",
+    )
+    self.install_files(
+        "darkman.desktop",
+        "usr/share/applications",
+    )
+
+    self.install_service("^/darkman.user")
+
+    self.install_man("darkman.1")
+
+    for shell in ["bash", "fish", "zsh"]:
+        self.install_completion(f"darkman.{shell}", shell)


user/darkman/files/darkman.user
```diff
@@ -0,0 +1,4 @@
+type             = process
+command          = /usr/bin/darkman run
+smooth-recovery  = no
+log-type         = buffer
```
triallax: don't align these, also probably want `depends-on = graphical.target`

user/darkman/files/darkman.user
```diff
@@ -0,0 +1,4 @@
+type             = process
+command          = /usr/bin/darkman run
+smooth-recovery  = no
```
triallax: why `no`? probably should be `yes`

user/darkman/template.py
```diff
@@ -0,0 +1,71 @@
...
+make_build_args = [
+    f"-ldflags=-X main.Version=v{pkgver}",
+    "./cmd/darkman",
+]
+hostmakedepends = [
+    "scdoc",
+    "go",
+]
+pkgdesc = "🌇 control dark-mode and light-mode transitions"
+license = "0BSD"
```
triallax: > darkman is licensed under the ISC licence.

triallax: i checked and the licence is indeed ISC, though gitlab does show 0bsd for some reason:

README, BSD Zero Clause License, CHANGELOG

cc @whynothugo

user/darkman/template.py
```diff
@@ -0,0 +1,71 @@
...
+build_style = "go"
+make_build_args = [
+    f"-ldflags=-X main.Version=v{pkgver}",
+    "./cmd/darkman",
+]
+hostmakedepends = [
+    "scdoc",
+    "go",
+]
+pkgdesc = "🌇 control dark-mode and light-mode transitions"
```
triallax: No emojis in `pkgdesc` please

```suggestion
pkgdesc = "Control dark-mode and light-mode transitions"
```

user/darkman/template.py
```diff
@@ -0,0 +1,71 @@
...
+        "bash": "darkman.bash",
+        "fish": "darkman.fish",
+    }.items():
+        with open(self.cwd / comp_filename, "w") as comp_file:
+            self.do(
+                "build/darkman",
+                "completion",
+                shell,
+                stdout=comp_file,
+            )
```
triallax: `install_completion` handles installing with the correct filename already
```suggestion
    for shell in ["bash", "fish", "zsh"]:
        with open(self.cwd / f"darkman.{shell}", "w") as comp_file:
            self.do(
                "build/darkman",
                "completion",
                shell,
                stdout=comp_file,
            )
```

user/darkman/template.py
```diff
@@ -0,0 +1,71 @@
...
+    self.install_files(
+        "contrib/portal/darkman.portal",
+        "usr/share/xdg-desktop-portal/portals/",
+    )
+    self.install_files(
+        "darkman.desktop",
+        "usr/share/applications/",
+    )
+
+    self.install_service(self.files_path / "darkman.user")
```
triallax: shorthand
```suggestion
    self.install_service("^/darkman.user")
```

user/darkman/template.py
```diff
@@ -0,0 +1,71 @@
...
+
+    self.install_man("darkman.1")
+
+    for shell, comp_filename in {
+        "zsh": "_darkman.zsh",
+        "bash": "darkman.bash",
+        "fish": "darkman.fish",
+    }.items():
+        self.install_completion(comp_filename, shell)
```
triallax:
```suggestion
    for shell in ["bash", "fish", "zsh"]:
        self.install_completion(f"shotman.{shell}", shell)
```

user/darkman/template.py
```diff
@@ -0,0 +1,71 @@
...
+        "usr/share/dbus-1/services/",
+    )
+    self.install_files(
+        "contrib/portal/darkman.portal",
+        "usr/share/xdg-desktop-portal/portals/",
+    )
+    self.install_files(
+        "darkman.desktop",
+        "usr/share/applications/",
+    )
```
triallax: as a matter of style we typically don't use trailing slashes here

```suggestion
    self.install_files(
        "contrib/dbus/nl.whynothugo.darkman.service",
        "usr/share/dbus-1/services",
    )
    self.install_files(
        "contrib/dbus/org.freedesktop.impl.portal.desktop.darkman.service",
        "usr/share/dbus-1/services",
    )
    self.install_files(
        "contrib/portal/darkman.portal",
        "usr/share/xdg-desktop-portal/portals",
    )
    self.install_files(
        "darkman.desktop",
        "usr/share/applications",
    )
```

user/darkman/template.py
```diff
@@ -0,0 +1,71 @@
...
+                stdout=comp_file,
+            )
+
+
+def install(self):
+    self.install_bin("build/darkman")
+    self.install_license("LICENCE")
+
+
+def post_install(self):
```
triallax: just merge em, binary should be installed automatically i think
```suggestion
def post_install(self):
    self.install_license("LICENCE")
```

user/darkman/template.py
WhyNotHugo: The licence rendered on the sidebar by GitLab is (erroneously) auto-detected and I have no way of fixing it.

The correct licence is the ISC licence, as stated in both README.md and LICENCE.

user/darkman/template.py
```diff
@@ -0,0 +1,71 @@
...
+                "build/darkman",
+                "completion",
+                shell,
+                stdout=comp_file,
+            )
+
+
+def install(self):
+    self.install_bin("build/darkman")
+    self.install_license("LICENCE")
```
WhyNotHugo: Can you not simply use `make install`? This will also install the man page, d-bus service files and completions.

zacharyburnett: I believe using `self.install_*` lets the package manager track which files belong to `darkman`, though I may be wrong about that; @triallax, would `make install` work here as well?

zacharyburnett:
Thank you for taking the time to review @triallax @WhyNotHugo, I appreciate it! :smile:

user/darkman/template.py
triallax: > I believe using self.install_* lets the package manager track which files belong to darkman

I'm honestly not sure what you mean by this, for cbuild there is no difference between files installed via `self.install*` and files installed by the package's build system.

> would `make install` work here as well?

probably, give it a try!

zacharyburnett: Oh! Then I apologize, I think was making this more complicated than it needed to be. It appears that the only thing preventing this template from using `build_style = "makefile"` is that the `go` compiler requires network access in the `build` phase (but the network is not available after the `fetch` phase): 
```
...
0:00:02.223 => darkman-2.0.1-r0: running build...
go build -ldflags "-X main.Version=`git describe --tags --dirty 2>/dev/null || echo 0.0.0-dev`" ./cmd/darkman
scdoc < darkman.1.scd > darkman.1
go: github.com/adrg/xdg@v0.3.3: Get "https://proxy.golang.org/github.com/adrg/xdg/@v/v0.3.3.mod": dial tcp: lookup proxy.golang.org on 10.0.0.1:53: dial udp 10.0.0.1:53: connect: network is unreachable
make: *** [Makefile:12: darkman] Error 1
0:00:02.235 => A failure has occurred!
...
0:00:02.239 => Phase 'build' failed for package 'darkman'.
```

WhyNotHugo: > I believe using self.install_* lets the package manager track which files belong to darkman, though I may be wrong about that

The installation is done inside an empty directory. All files placed there are tracked as belonging to the package, regardless of how they're copied.

zacharyburnett: yeah, it seems to be much simpler than it was in my head :sweat_smile: 

thank you for being patient with me as I figure this out!

user/darkman/files/darkman.user
```diff
@@ -0,0 +1,5 @@
+type = process
+depends-on = graphical.target
+command = /usr/bin/darkman run
+smooth-recovery = yes
+log-type = buffer
```
triallax: i'm the one who suggested the initial version, so this is on me, but:

```suggestion
command = /usr/bin/darkman run
smooth-recovery = yes
log-type = buffer
depends-on: graphical.target
```

user/darkman/template.py
```diff
@@ -0,0 +1,75 @@
...
+]
+hostmakedepends = [
+    "scdoc",
+    "go",
+]
+pkgdesc = "Control dark-mode and light-mode transitions"
+license = "ISC"
+url = "https://gitlab.com/WhyNotHugo/darkman"
+source = f"{url}/-/archive/v{pkgver}/darkman-v{pkgver}.tar.gz"
+sha256 = "4d87ee5dcefcd237be43d2b3d27bea658d23ebe14b5b5951bc67942f1ec839e5"
```
triallax, line 17:
```suggestion
sha256 = "4d87ee5dcefcd237be43d2b3d27bea658d23ebe14b5b5951bc67942f1ec839e5"
# builds completions with generated binary
options = ["!cross"]
```

user/darkman/template.py
triallax: the trailing slashes are still here

user/darkman/files/darkman.user
```diff
@@ -0,0 +1,5 @@
+type = process
+command = /usr/bin/darkman run
+smooth-recovery = yes
+log-type = buffer
+depends-on: graphical.target
```
zacharyburnett, line 5:
```suggestion
depends-on = graphical.target
```

triallax: the colon was on purpose, please bring it back

user/darkman/template.py
triallax: now this whole block of code is a duplicate of the one below it

PR #4557, opened 2025-07-18
user/opam: new package

kit-ty-kate:
## Description

This packages opam, the OCaml package manager.
Our archive and build system ships with everything needed (including the compiler), so very minimal dependencies are required. Once https://github.com/chimera-linux/cports/pull/3700, it would be possible to change this package to use it instead, but in the meantime it doesn't matter.

I haven't tested this locally yet so this is a draft for now. I should be able to test it around next week.

## Checklist

### DIFF
--- a/user/opam/template.py
+++ b/user/opam/template.py
@@ -1,0 +1,18 @@
+pkgname = "opam"
+pkgver = "2.4.1"
+pkgrel = 0
+build_style = "makefile"
+make_build_target = "cold"
+make_install_target = "install"
+make_install_args = ["prefix=/usr"]
+makedepends = ["curl", "rsync", "unzip", "bubblewrap", "bash", "clang", "gmake"]
+depends = [*makedepends]
+pkgdesc = "OCaml Package Manager"
+license = "LGPL-2.1-only WITH OCaml-LGPL-linking-exception"
+url = "https://opam.ocaml.org"
+source = f"https://github.com/ocaml/opam/releases/download/{pkgver}/opam-full-{pkgver}.tar.gz"
+sha256 = "c4d053029793c714e4e7340b1157428c0f90783585fb17f35158247a640467d9"
+tools = {"ASPP": "cc -c", "AS": "cc -c"}
+hardening = ["!int"]
+# check is disabled because it requires bubblewrap and doesn't work inside the sandbox
+options = ["!cross", "!check"]


kit-ty-kate:
> as opam requires GNU patch specifically.

not anymore since 2.4.0

kit-ty-kate:
dw, i'll squash everything once i undraft this

kit-ty-kate:
I've tested it locally and it works fine now after many adjustments. This is ready to merge.

user/opam/template.py
```diff
@@ -0,0 +1,18 @@
+pkgname = "opam"
+pkgver = "2.4.1"
+pkgrel = 0
+build_style = "makefile"
```
q66, line 4: i'd prefer this to be a gnu_configure package without the makefile wrapper

user/opam/template.py
```diff
@@ -0,0 +1,18 @@
...
+make_build_target = "cold"
+make_install_target = "install"
+make_install_args = ["prefix=/usr"]
+makedepends = ["curl", "rsync", "unzip", "bubblewrap", "bash", "clang", "gmake"]
```
q66, line 8: these should be hostmakedepends; clang should not be there, presumably neither should be make

rsync is a tests-only dependency so it should be in `checkdepends` and not carried into runtime dependencies

is bash an actual dep?

kit-ty-kate: every single one of these dependencies are used at runtime (see `opam init`)

user/opam/template.py
```diff
@@ -0,0 +1,18 @@
+pkgname = "opam"
+pkgver = "2.4.1"
+pkgrel = 0
+build_style = "makefile"
+make_build_target = "cold"
```
kit-ty-kate, line 5:
```suggestion
build_style = "gnu_configure"
configure_args = [
    "--with-vendored-deps",
    "--with-mccs",
]
```

user/opam/template.py
q66: i only see rsync in `required_packages_for_cygwin`, same with make

i don't see bash specifically being required anywhere

i see `cc` and make as a recommended thing, but i don't see any reason to make those hard deps especially as different packages could provide that

kit-ty-kate: Please do try `opam init`. bash is used in the sandbox, rsync is used in for local pins and ssh support, cc and gmake are used to compile OCaml (for which opam is fairly useless without, although it is """technically""" possible to use without)

PR #4567, opened 2025-07-22
user/ydotool: new package

BarryLhm:
## Description

Generic command-line automation tool (no X!)

## Checklist

### DIFF
--- a/user/ydotool/template.py
+++ b/user/ydotool/template.py
@@ -1,0 +1,17 @@
+pkgname = "ydotool"
+pkgver = "1.0.4"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DCMAKE_POLICY_VERSION_MINIMUM=3.5"]
+hostmakedepends = ["cmake", "pkgconf", "ninja", "scdoc"]
+makedepends = ["linux-headers"]
+pkgdesc = "Generic command-line automation tool"
+license = "AGPL-3.0-or-later"
+url = "https://github.com/ReimuNotMoe/ydotool"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "ba075a43aa6ead51940e892ecffa4d0b8b40c241e4e2bc4bd9bd26b61fde23bd"
+
+
+def post_install(self):
+    self.uninstall("usr/lib/systemd/user")
+    self.install_license("LICENSE")


user/ydotool/patches/fix_cmake_min_version.patch
```diff
@@ -0,0 +1,10 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1,4 +1,4 @@
+-cmake_minimum_required(VERSION 3.4)
++cmake_minimum_required(VERSION 3.4...4.0)
```
JamiKettunen: `configure_args = ["-DCMAKE_POLICY_VERSION_MINIMUM=3.5"]` in template instead and drop the patch, that's how (all?) other packages deal with the configure failure and nowadays even `cmake` itself gives you the same hint about trying this

user/ydotool/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "ydotool"
+pkgver = "1.0.4"
+pkgrel = 0
+build_style = "cmake"
+make_cmd = "make"
+make_dir = "."
```
JamiKettunen: why declare `make_*` instead of just providing `ninja` in `hostmakedepends` like other cmake templates?

user/ydotool/template.py
```diff
@@ -0,0 +1,25 @@
...
+hostmakedepends = ["cmake", "pkgconf", "scdoc"]
+makedepends = ["linux-headers"]
+pkgdesc = "Generic command-line automation tool"
+subdesc = "(like xdotool, but not for X11)"
```
JamiKettunen: `subdesc` by definition already includes the opening/closing parenthesis but I'm not sure if its usage here is warranted really, maybe for someone else to weigh in on. I'd probably just `pkgdesc = "Generic command-line input automation tool like xdotool"`

user/ydotool/template.py
```diff
@@ -0,0 +1,25 @@
...
+license = "AGPL-3.0-or-later"
+url = "https://github.com/ReimuNotMoe/ydotool"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "ba075a43aa6ead51940e892ecffa4d0b8b40c241e4e2bc4bd9bd26b61fde23bd"
+
+
+def install(self):
+    self.install_bin("ydotool")
+    self.install_bin("ydotoold")
```
JamiKettunen: whole `install()` unneeded here even before switching from `make` to `ninja`, please only add needed stuff

user/ydotool/template.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def post_install(self):
+    self.install_man("manpage/ydotool.1")
+    self.install_man("manpage/ydotoold.8")
```
JamiKettunen: man pages are installed by default when you don't override `install()`, also following that you'd want `self.uninstall("usr/lib/systemd/user")` instead here like some other templates are doing

user/ydotool/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://api.github.com/repos/ReimuNotMoe/ydotool/git/refs/tags"
+pattern = r"refs/tags/v([0-9.]+)"
```
JamiKettunen: drop unneeded `update.py` since `./cbuild update-check user/ydotool verbose` already works, in other packages where something like this may exist it's only used if the defaults aren't already enough (e.g. project does a lot of pre-release tags etc)

user/ydotool/template.py
BarryLhm: the cmakelist defaultly generates makefile

user/ydotool/update.py
BarryLhm: omg i didnt think cbuild was so automatical

user/ydotool/template.py
JamiKettunen: that may be a `cmake` default (like when you build it manually) but that's not exactly what happens with `cbuild`; see https://github.com/chimera-linux/cports/blob/5026157/src/cbuild/build_style/cmake.py#L79 + https://github.com/chimera-linux/cports/blob/5026157/src/cbuild/build_style/cmake.py#L11 + https://github.com/chimera-linux/cports/blob/5026157/src/cbuild/util/cmake.py#L97

user/ydotool/template.py
BarryLhm: i'm afraid it'll cause misunderstand so i'll just remove subdesc

PR #4571, opened 2025-07-24
user/fnf: update to 0.3.1

unrealapex:
## Checklist

### DIFF
--- a/user/fnf/template.py
+++ b/user/fnf/template.py
@@ -1,12 +1,13 @@
 pkgname = "fnf"
-pkgver = "0.2"
-pkgrel = 1
+pkgver = "0.3.1"
+pkgrel = 0
 build_style = "makefile"
+makedepends = ["linux-headers"]
 pkgdesc = "CLI fuzzy finder"
 license = "MIT"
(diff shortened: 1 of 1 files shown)


user/fnf/template.py
```diff
@@ -1,12 +1,13 @@
 pkgname = "fnf"
-pkgver = "0.2"
+pkgver = "0.3.1"
 pkgrel = 1
 build_style = "makefile"
+hostmakedepends = ["linux-headers"]
```
triallax: this goes in `makedepends`

user/fnf/template.py
```diff
@@ -1,12 +1,13 @@
 pkgname = "fnf"
-pkgver = "0.2"
+pkgver = "0.3.1"
 pkgrel = 1
```
triallax:
```suggestion
pkgrel = 0
```
did you not use `prepare-upgrade`? it resets the pkgrel automatically
if not i recommend you do use it next time

unrealapex: I did, I thought that was unintentional.

PR #4585, opened 2025-07-29
main/linux-stable: update to 6.16.0

q66:
Draft until 6.16.1.


JamiKettunen:
6.16.1 is finally out; also a reminder about including [`CONFIG_INTERCONNECT_QCOM=m`](https://github.com/chimera-linux/cports/commit/a6efa8e) & [`CONFIG_IOMMU_DEFAULT_DMA_STRICT=y`](https://github.com/chimera-linux/cports/commit/7f8a350) for aarch64 which were reverted for now (though which this should've still applied on top cleanly when I later checked)

PR #4587, opened 2025-07-30
user/nushell: Update to 0.106.1, add core plugins as subpackages

ItsHarper:
## Description

Updates nushell to 0.106.1, and builds the [core plugins](https://www.nushell.sh/book/plugins.html#core-plugins) as subpackages. All of the subpackages are installed with the main `nushell` package by default, and include nushell scripts that will automatically [register the appropriate plugin](https://www.nushell.sh/book/plugins.html#registering-plugins).

## Checklist

### DIFF
--- a/user/nushell/template.py
+++ b/user/nushell/template.py
@@ -1,5 +1,5 @@
 pkgname = "nushell"
-pkgver = "0.105.1"
+pkgver = "0.106.1"
 pkgrel = 0
 build_style = "cargo"
 make_build_args = [
@@ -9,8 +9,6 @@ make_build_args = [
 make_install_args = [*make_build_args]
 make_check_args = [
     "--",
-    "--skip=shell::environment::env::env_shlvl_in_exec_repl",
-    "--skip=shell::environment::env::env_shlvl_in_repl",
     "--skip=shell::environment::env::path_is_a_list_in_repl",
 ]
 hostmakedepends = ["cargo-auditable", "pkgconf"]
@@ -25,8 +23,7 @@ pkgdesc = "Shell with a focus on structured data"
 license = "MIT"
 url = "https://www.nushell.sh"
 source = f"https://github.com/nushell/nushell/archive/refs/tags/{pkgver}.tar.gz"
-sha256 = "2c52ef5aef2ba1a3ae873e84bf72b52220f47c8fe47b99950b791e678a43d597"
-
+sha256 = "3e24044c354d050a850b69dc77c99cc503542c3d9d75fed0aef1c12fefdf380b"
 
 def post_install(self):
     self.install_shell("/usr/bin/nu")
--- a/user/nushell/template.py
+++ b/user/nushell/template.py
@@ -1,12 +1,15 @@
 pkgname = "nushell"
 pkgver = "0.106.1"
-pkgrel = 0
+pkgrel = 2
 build_style = "cargo"
-make_build_args = [
+_features_args = [
     "--no-default-features",
     "--features=plugin,trash-support,sqlite,native-tls",
 ]
-make_install_args = [*make_build_args]
+make_build_args = [
+    *_features_args,
+    "--workspace",
+]
 make_check_args = [
     "--",
     "--skip=shell::environment::env::path_is_a_list_in_repl",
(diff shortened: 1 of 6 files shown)


user/nushell/template.py
```diff
@@ -25,9 +26,43 @@
...
+    "polars",
+    "formats",
+    "gstat",
+    "query",
+    "inc",
+]
+
+
+def install(self):
+    self.cargo.install(args=[*_features_args])
```
jcgruenhage, line 40: Does it make sense to use `self.cargo.install` here? The binary has already been built and could be installed using `self.install_bin` directly.

user/nushell/template.py
```diff
@@ -25,9 +26,43 @@
...
+    nu_autoload_path = "usr/share/nushell/vendor/autoload"
+    self.install_dir(nu_autoload_path)
+    for _plugin in _plugins:
+        self.cargo.install(wrksrc=f"crates/nu_plugin_{_plugin}")
```
jcgruenhage, line 44: Same here

user/nushell/template.py
```diff
@@ -25,9 +26,43 @@
...
 def post_install(self):
     self.install_shell("/usr/bin/nu")
     self.install_license("LICENSE")
+
+
+def _genmod(pname):
+    @subpackage(f"nushell-plugin-{pname}")
+    def _(self):
+        self.subdesc = f"{pname} plugin"
+        self.install_if = [self.parent]
```
jcgruenhage, line 60: I don't think we should auto-install the plugins, especially since these also contain a script to auto-enable them. Btw, how does this interact with automatic nushell completion packages? https://github.com/chimera-linux/cports/blob/26df0f7b3384515b82b29aeead8495598c33cbfc/src/cbuild/core/template.py#L2326-L2331?

Do we get `nushell-plugin-{pname}-nucomp` packages with the auto-enable script?

user/nushell/template.py
ItsHarper: For context, I'm an experienced software engineer, but I'm new to Rust, cbuild, and package maintenance in general.

My thought process was that `cargo install` is probably doing something important/different, since that's what the `install` step runs by default. Since you're asking this, I'm guessing that it works that way just because `cargo` already knows which binaries it's building, so by using `cargo install`, we can save the package author from having to list out the binaries to install, at the cost of some extra build time. Am I understanding that correctly?

If my new understanding is correct, then I'll definitely make that change.

user/nushell/template.py
ItsHarper: > I don't think we should auto-install the plugins, especially since these also contain a script to auto-enable them.

Fair enough, I'm fine with changing that.

> How does this interact with automatic nushell completion packages?

It does capture them into `-nucomp` packages.

```nushell
~/cports> apk search nushell-plugin
nushell-plugin-formats-0.106.1-r2
nushell-plugin-formats-nucomp-0.106.1-r2
nushell-plugin-gstat-0.106.1-r2
nushell-plugin-gstat-nucomp-0.106.1-r2
nushell-plugin-inc-0.106.1-r2
nushell-plugin-inc-nucomp-0.106.1-r2
nushell-plugin-polars-0.106.1-r2
nushell-plugin-polars-nucomp-0.106.1-r2
nushell-plugin-query-0.106.1-r2
nushell-plugin-query-nucomp-0.106.1-r2
```

I found an [interesting discussion](https://github.com/chimera-linux/cports/pull/2767#discussion_r1730134869) from a year ago about this, where this was said:

> taking [non-completion nushell scripts for `-nucomp` subpackages] is fine, the only issue is the name might be misleading. but it is indeed shell-specific data that is taken with the shell so the contents are correct and work fine

Still, I agree it's better not to have them, and there aren't any other automatically-split subpackages that I'd have to re-create, so I'll set the `!autosplit` option.

user/nushell/template.py
jcgruenhage: `cargo install` does nothing "special", it just knows what binaries exists and installs them into some destination directory for binaries.

Usually it's not even extra build time, because cargo uses pre-existing caches. The reason I'm asking here is that here we basically already list out all the binaries *and* we introduce complexity by separating `_feature_args` out of `make_build_args`.

jcgruenhage:
@ItsHarper I have adjusted this locally, if you want to take a look, my commit is under https://github.com/jcgruenhage/cports/commit/3c2dd8fa4d57ebfb46739544cdf649ebb4ffdb0e

triallax:
merged in #4587

PR #4596, opened 2025-07-31
user/zizmor: update to 1.11.0

jcgruenhage:


user/zizmor/template.py
```diff
@@ -1,5 +1,5 @@
 pkgname = "zizmor"
-pkgver = "1.8.0"
+pkgver = "1.11.0"
 pkgrel = 0
 build_style = "cargo"
 prepare_after_patch = True
```
triallax: not needed anymore

PR #4604, opened 2025-08-03
user/tayga: new package

z-erica:
picked from https://github.com/chimera-linux/cports/pull/4473, seems useful to pr on its own

### DIFF
--- a/user/tayga/template.py
+++ b/user/tayga/template.py
@@ -1,0 +1,26 @@
+pkgname = "tayga"
+pkgver = "0.9.5"
+pkgrel = 0
+build_style = "makefile"
+makedepends = ["linux-headers"]
+checkdepends = [
+    "iproute2",
+    "python",
+]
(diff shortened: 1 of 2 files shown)


user/tayga/template.py
```diff
@@ -0,0 +1,26 @@
...
+sha256 = "d44cc1158f60623d1bcd245f811957a162092c8f5e725489438de12e5500ae49"
+# check: needs network namespaces
+options = ["!check"]
+
+
+def check(self):
+    self.do("sh", "test/fullsuite.sh")
+
+
+def install(self):
```
triallax, line 23: there's an install target, any reason we're not using it?

z-erica: not in the actual release https://github.com/apalrd/tayga/blob/0.9.5/Makefile#L31-L32

triallax: ah, right, i was looking at master

PR #4621, opened 2025-08-06
main/firefox: drop obsolete patch

WhyNotHugo:
## Description

The value returned by find_candidate() now matches RUST_TARGET on all platforms. This patch is no longer necessary.

I've corroborated that this compiles and works properly on Alpine, I'm opening a Draft PR here to trigger CI and confirm that my observations are true on other distros shipping the same patch.

## Checklist


q66:
this won't trigger ci unless you bump pkgrel

but anyway i prefer to keep the patch given it does nothing but bypass mozilla's buildsystem weirdness that they may just break at any time

PR #4642, opened 2025-08-08
main/initramfs-tools: update to 0.150

JamiKettunen:
at least for most Snapdragon X Elite laptops https://salsa.debian.org/kernel-team/initramfs-tools/-/commit/d82a1e91 was of interest. works for me after the conflict resolves but possible I missed something.

adapts `0005-general-portability-fixes-for-bsdutils-and-musl.patch` for https://salsa.debian.org/kernel-team/initramfs-tools/-/commit/9fa9ae01 etc.


q66:
since 0.151 will need additional changes i'll take care of this myself later

PR #4646, opened 2025-08-09
user/dnscrypt-proxy: don't install default config

ayushnix:
## Description

as discussed on IRC, installing a default config file in /etc using tmpfiles in the manner that I did would prevent installation of an updated config file later

## Checklist


ayushnix:
I can also add a [wrapper script](https://github.com/chimera-linux/cports/pull/3668#issuecomment-2716022375) if that sounds more appropriate. Let me know if I should @q66 @triallax.

PR #4655, opened 2025-08-11
 user/s-tui: new package 

z-erica:


user/python-urwid/template.py
```diff
@@ -0,0 +1,40 @@
...
+    *depends,
+]
+pkgdesc = "Console UI library"
+license = "LGPL-2.1-or-later"
+url = "https://urwid.org"
+source = f"$(PYPI_SITE)/u/urwid/urwid-{pkgver}.tar.gz"
+sha256 = "e7cb70ba1e7ff45779a5a57e43c57581ee7de6ceefb56c432491a4a6ce81eb78"
+
+
+def init_check(self):
```
q66: just set make_check_args at toplevel, why do this

z-erica: oh i totally had a {self.python_version} thing going on there in a previous revision haha ill move it up

PR #4662, opened 2025-08-13
user/pre-commit: new package

hatzka-nezumi:
## Description

`pre-commit` is a tool for managing shared Git pre-commit hooks.

It's worth noting that it can use several other language runtimes and compilers for the hooks themselves, but this package doesn't depend on most of them, because that would be an absurd amount of dependencies most users won't use---and the ones they do use, they will likely be installing directly anyway for the actual coding part. It does unfortunately depend on Node, despite containing no more JavaScript than it does Swift or Haskell.

## Checklist

### DIFF
--- a/user/python-cfgv/template.py
+++ b/user/python-cfgv/template.py
@@ -1,0 +1,24 @@
+pkgname = "python-cfgv"
+pkgver = "3.4.0"
+pkgrel = 0
+build_style = "python_pep517"
+hostmakedepends = [
+    "python-build",
+    "python-setuptools",
+]
+checkdepends = [
+    "python-installer",
+    "python-pytest",
+]
+depends = [
+    "python",
+]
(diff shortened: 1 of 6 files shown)


user/python-cfgv/template.py
```diff
@@ -0,0 +1,29 @@
...
+    "python",
+]
+pkgdesc = "Validate configuration and produce human readable error messages"
+license = "MIT"
+url = "https://github.com/asottile/cfgv"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "a8a4e4ad5618b6e65af563b14f46ccb1e45f5725c3b733d0ba15bb2a3b5bf9fd"
+
+
+def build(self):
```
q66: it's fine to use the pep517 build-style with setuptools things instead of doing stuff manually

PR #4673, opened 2025-08-16
user/bettercap: new package

breakgimme:
## Description

bettercap is a Swiss Army knife for WiFi, Bluetooth Low Energy, wireless HID hijacking, CAN-bus and IPv4 and IPv6 networks reconnaissance and MITM attacks.

## Checklist

### DIFF
--- a/user/bettercap/template.py
+++ b/user/bettercap/template.py
@@ -1,0 +1,14 @@
+pkgname = "bettercap"
+pkgver = "2.41.4"
+pkgrel = 0
+build_style = "go"
+hostmakedepends = ["go", "pkgconf"]
+makedepends = ["libnetfilter_queue-devel", "libpcap-devel", "libusb-devel"]
+pkgdesc = "Swiss Army knife for network attacks"
+license = "GPL-3.0-only"
+url = "https://bettercap.org"
+source = (
+    f"https://github.com/bettercap/bettercap/archive/refs/tags/v{pkgver}.tar.gz"
+)
+sha256 = "384848630f594fadd48e80406f4cf8ceccfe3f32dd9182f7e18c20240e74a5fd"
+env = {"CGO_ENABLED": "1"}


user/bettercap/template.py
```diff
@@ -0,0 +1,21 @@
...
+pkgdesc = "Swiss Army knife for network attacks"
+license = "GPL-3.0-only"
+url = "https://bettercap.org"
+source = (
+    f"https://github.com/bettercap/bettercap/archive/refs/tags/v{pkgver}.tar.gz"
+)
+sha256 = "c00a489110a01b799796bfc5701bbaea882e0a1aa675d16ce2aba25bd0d71ad1"
+
+
+# make sure cgo is enabled while cross-compiling
```
q66: this looks wrong

just `env = {"CGO_ENABLED": "1"}` in main block should be enough

breakgimme: > just `env = {"CGO_ENABLED": "1"}` in main block should be enough

it's enough as long as it's not cross-compiled, some other packages also have to use this hack and the only package that doesn't (user/emptty) doesn't seem to cross-compile

q66: you're confusing something

the only reason they need to apply that stuff is because they're makefile templates and not go templates

a go template already has the env otherwise set up

```
q66@chimera: /home/q66/cports$ ./cbuild -a aarch64 pkg user/bettercap
...
0:00:32.383 => bettercap-2.41.4-r0: generating packages...
  apk: bettercap-2.41.4-r0.apk in user
  apk: bettercap-dbg-2.41.4-r0.apk in user/debug
```

```
q66@chimera: /home/q66/cports$ git diff
diff --git a/user/bettercap/template.py b/user/bettercap/template.py
index 1987b18ce..b3e1a7500 100644
--- a/user/bettercap/template.py
+++ b/user/bettercap/template.py
@@ -11,11 +11,4 @@ source = (
     f"https://github.com/bettercap/bettercap/archive/refs/tags/v{pkgver}.tar.gz"
 )
 sha256 = "384848630f594fadd48e80406f4cf8ceccfe3f32dd9182f7e18c20240e74a5fd"
-
-
-# make sure cgo is enabled while cross-compiling
-def init_build(self):
-    from cbuild.util import golang
-
-    self.env["CGO_ENABLED"] = "1"
-    self.env.update(golang.get_go_env(self))
+env = {"CGO_ENABLED": "1"}
```

breakgimme: you're absolutely right, i'm sorry for wasting your time with this
i was confusing env and make_env

q66:
lgtm now

PR #4683, opened 2025-08-21
user/fonts-monaspace: move from user/fonts-monaspace-otf, update to 1.300

BarryLhm:
## Description

install all variants(except web .woff2) fonts.

uses scripted installaion and subpkg generation to simplify variants control

## Checklist


quadraplat:
just choosing not to package like 90% of the font is wrong. do it like arch or alpine does it if package size is an issue

BarryLhm:
changed to scripted subpkg generation like fonts-noto

BarryLhm:
@q66 how should i name those subpkgs?

PR #4686, opened 2025-08-23
user/cicada: new package

JustSoup312:
## Description

Adds the `cicada` package. Cicada is a bash-like shell that is meant for interactive use. Due to this, it does not include most POSIX shell scripting logic and is not supposed to be a replacement for `/bin/sh`.

## Checklist

### DIFF
--- a/user/cicada/template.py
+++ b/user/cicada/template.py
@@ -1,0 +1,24 @@
+pkgname = "cicada"
+pkgver = "1.1.2"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo-auditable"]
+makedepends = [
+    "rust-std",
+    "sqlite-devel",
+]
+checkdepends = [
+    "less",
+]
+pkgdesc = "Bash-like Unix shell"
+license = "MIT"
+url = "https://github.com/mitnk/cicada"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "fa98e4b15d62ce0343d69a9f51d737fe02b8200f5903c14f5131dec6c7a54656"
+# Needs exact matching coreutils command output
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+    self.install_shell("/usr/bin/cicada")
--- a/user/cicada/patches/unbash-and-release.patch
+++ b/user/cicada/patches/unbash-and-release.patch
@@ -1,0 +1,166 @@
+From https://github.com/mitnk/cicada/pull/55
+
+From 8718bb77fcbeb5570d49d63970bd129f94f95554 Mon Sep 17 00:00:00 2001
+From: Aster Boese <asterboese@mailbox.org>
+Date: Sat, 23 Aug 2025 14:36:11 -0400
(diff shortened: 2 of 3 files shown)


user/cicada/template.py
```diff
@@ -0,0 +1,33 @@
...
+sha256 = "fa98e4b15d62ce0343d69a9f51d737fe02b8200f5903c14f5131dec6c7a54656"
+hardening = ["vis", "cfi", "int"]
+# Needs exact matching coreutils command output
+options = ["!check"]
+
+
+def check(self):
+    self.cargo.check()
+    arch_mode = self.profile().arch + "-chimera-linux-musl"
+    self.do("sh", "-c", f"MODE={arch_mode}/release ./tests/test_scripts.sh")
```
q66: why the whole thing with sh -c

user/cicada/template.py
```diff
@@ -0,0 +1,33 @@
...
+license = "MIT"
+url = "https://github.com/mitnk/cicada"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "fa98e4b15d62ce0343d69a9f51d737fe02b8200f5903c14f5131dec6c7a54656"
+hardening = ["vis", "cfi", "int"]
+# Needs exact matching coreutils command output
+options = ["!check"]
+
+
+def check(self):
```
q66: if you're disabling check what is the point of this

user/cicada/template.py
```diff
@@ -0,0 +1,33 @@
+pkgname = "cicada"
+pkgver = "1.1.2"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = [
```
q66: `hostmakedepends = ["cargo-auditable"]`, same with the others

user/cicada/template.py
```diff
@@ -0,0 +1,33 @@
...
+]
+checkdepends = [
+    "less",
+]
+pkgdesc = "Bash-like Unix shell"
+license = "MIT"
+url = "https://github.com/mitnk/cicada"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "fa98e4b15d62ce0343d69a9f51d737fe02b8200f5903c14f5131dec6c7a54656"
+hardening = ["vis", "cfi", "int"]
```
q66: drop this line, it doesn't do anything

user/cicada/template.py
JustSoup312: That's how the tests are done using the makefile. The tests are done using a shell script that execs cicada and tests its output against diffs.

user/cicada/template.py
JustSoup312: Oh, I didn't know that check was removed it check is disabled. Will do.

user/cicada/template.py
q66: `self.do("./tests/test_scripts.sh", env={"MODE": f"{arch_mode}/release"})`

JustSoup312: Thanks. I'll note that for if tests are ever fixed in a way that I can just patch the program.

PR #4718, opened 2025-08-28
main/vectorscan: update to 5.4.12

rbgarga:


rbgarga:
benchmarks are failing to link and I couldn't fix it

q66:
this still does a bunch of -march/-mtune passing, which should be eliminated (not the way the original patch worked, but by preventing the buildsystem from doing so)

PR #4722, opened 2025-08-28
user/dua-cli: new package

jcgruenhage:
## Description

TUI disk usage analyzer

## Checklist

### DIFF
--- a/user/dua-cli/template.py
+++ b/user/dua-cli/template.py
@@ -1,0 +1,35 @@
+pkgname = "dua-cli"
+pkgver = "2.34.0"
+pkgrel = 0
+build_style = "cargo"
+make_check_args = [
+    "--",
+    "--skip=interactive::app::tests::unit::it_can_handle_ending_traversal_reaching_top_but_skipping_levels",
+    "--skip=interactive::app::tests::unit::it_can_handle_ending_traversal_without_reaching_the_top",
+]
+hostmakedepends = ["cargo-auditable"]
+makedepends = ["rust-std"]
+pkgdesc = "TUI disk usage analyzer"
+license = "MIT"
+url = "https://github.com/Byron/dua-cli"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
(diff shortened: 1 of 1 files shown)


user/dua-cli/template.py
```diff
@@ -0,0 +1,33 @@
+pkgname = "dua-cli"
```
wezm, line 1: Since the binary is `dua` and I don't believe there's a need to distinguish between a non-CLI version, perhaps the package should be `dua`? I see packages for other platforms are split on whether to call it `dua` or `dua-cli`.

user/dua-cli/template.py
```diff
@@ -0,0 +1,33 @@
...
+url = "https://github.com/Byron/dua-cli"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "8a3495d2ec0c9bee961e512c79b11c9945797a3d4f979b4ec63fd50d73d80c94"
+
+
+def post_build(self):
+    for shell in ["bash", "fish", "zsh"]:
+        with open(self.cwd / f"dua.{shell}", "w") as outf:
+            self.do(
+                f"./target/{self.profile().triplet}/release/dua",
```
wezm, line 25: Package will need `options = ["!cross"]` due the need to run the binary.

user/dua-cli/template.py
jcgruenhage: I chose this name because of this part in our docs:

https://github.com/chimera-linux/cports/blob/7fd54f3a46e256c1ef9befdadce009d154606711/Packaging.md?plain=1#L643-L646

The name of the upstream repo is dua-cli, and the name of the crates.io package is also dua-cli.

I'm fine with renaming it though, if you think `dua` is more fitting. I agree, but thought our docs disagreed.

user/dua-cli/template.py
jcgruenhage: Good catch, thanks :)

user/dua-cli/template.py
wezm: With that quote and the crate name being dua-cli I think you were right all along.

PR #4723, opened 2025-08-29
main/kea: update to 3.0.1

rbgarga:
Starting on Kea 2.6.3, it enforces the directory defined at compile time to be the only option to be used for sockets.  From changelog:

```
    3. To address potential denial-of-service and spoofing attacks, Kea no
    longer creates sockets in `/tmp`, which is a world-writable directory on
    most systems. Instead, it uses the
    ``[kea-installation-dir]/var/run/kea`` directory for sockets. The new
    default configurations have this change applied so new installations are
    secured, but please alter your configuration if upgrading.
```

The default runstatedir is /var/run but we are using /run.  It usually is not a big deal since /var/run is a symlink to /run but in this case we need to use what is defined at compile-time.  Change it to /run as the simplest solution.

### DIFF
--- a/main/kea/template.py
+++ b/main/kea/template.py
@@ -1,14 +1,11 @@
 pkgname = "kea"
-pkgver = "2.6.3"
+pkgver = "3.0.1"
 pkgrel = 0
-build_style = "gnu_configure"
-configure_args = [
-    "--disable-static",
-    "--enable-shell",
-]
+build_style = "meson"
+configure_args = ["-Drunstatedir=run"]
 hostmakedepends = [
-    "automake",
-    "slibtool",
+    "meson",
+    "pkgconf",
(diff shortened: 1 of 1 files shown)


main/kea/template.py
```diff
@@ -1,14 +1,11 @@
...
-pkgver = "2.6.3"
+pkgver = "3.0.1"
 pkgrel = 0
-build_style = "gnu_configure"
-configure_args = [
-    "--disable-static",
-    "--enable-shell",
-]
+build_style = "meson"
+configure_args = ["-Drunstatedir=run"]
```
q66, line 5: is the missing slash here intended?

q66: ah, yes, indeed it is

PR #4726, opened 2025-08-30
user/cli11: new package

zacharyburnett:
## Description

> [CLI11](https://github.com/CLIUtils/CLI11) is a command line parser for C++11 and beyond that provides a rich feature set with a simple and intuitive interface.

## Checklist

### DIFF
--- a/user/cli11/template.py
+++ b/user/cli11/template.py
@@ -1,0 +1,27 @@
+pkgname = "cli11"
+pkgver = "2.6.1"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DCLI11_SINGLE_FILE=OFF",  # Build the CLI11.hpp file from the sources. Requires Python (version 3 or 2.7).
+    "-DCLI11_PRECOMPILED=OFF",  # generate a precompiled static library instead of header-only
+    "-DCLI11_SINGLE_FILE_TESTS=OFF",  # Run the tests on the generated single file version as well
+    "-DCLI11_BUILD_DOCS=ON",  # build CLI11 documentation and book
+    "-DCLI11_BUILD_EXAMPLES=OFF",  # Build the example programs.
+    "-DCLI11_BUILD_EXAMPLES_JSON=OFF",  # Build some additional example using json libraries
+    "-DCLI11_INSTALL=ON",  # install CLI11 to the install folder during the install process
+    "-DCLI11_FORCE_LIBCXX=OFF",  # use libc++ instead of libstdc++ if building with clang on linux
+    "-DCLI11_CUDA_TESTS=OFF",  # build the tests with NVCC
+    "-DCLI11_BUILD_TESTS=ON",  # Build the tests.
+]
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+checkdepends = ["catch2-devel"]
+pkgdesc = "Command line parser for C++11 and beyond"
+license = "BSD-3-Clause"
+url = "https://github.com/CLIUtils/CLI11"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "377691f3fac2b340f12a2f79f523c780564578ba3d6eaf5238e9f35895d5ba95"
+
+
+def post_install(self):
+    self.install_license("LICENSE")


user/cli11-devel/template.py
```diff
@@ -0,0 +1,27 @@
+pkgname = "cli11-devel"
+pkgver = "2.5.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DCLI11_SINGLE_FILE=OFF",  # Build the CLI11.hpp file from the sources. Requires Python (version 3 or 2.7).
+    "-DCLI11_PRECOMPILED=ON",  # generate a precompiled static library instead of header-only
```
zacharyburnett, line 7: setting `CLI11_PRECOMPILED=ON` allows creation of a `cli11-devel-static` subpackage

user/cli11/template.py
```diff
@@ -0,0 +1,32 @@
+pkgname = "cli11"
+pkgver = "2.5.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DCLI11_SINGLE_FILE=OFF",  # Build the CLI11.hpp file from the sources. Requires Python (version 3 or 2.7).
+    "-DCLI11_PRECOMPILED=ON",  # generate a precompiled static library instead of header-only
```
zacharyburnett, line 7: setting `CLI11_PRECOMPILED=ON` allows creation of a `cli11-devel-static` subpackage

user/cli11/template.py
```diff
@@ -0,0 +1,27 @@
+pkgname = "cli11"
+pkgver = "2.5.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DCLI11_SINGLE_FILE=OFF",  # Build the CLI11.hpp file from the sources. Requires Python (version 3 or 2.7).
+    "-DCLI11_PRECOMPILED=OFF",  # generate a precompiled static library instead of header-only
```
zacharyburnett, line 7: I tried setting each of these options to `ON` and adding `cli11-devel` and `cli11-devel-static` subpackages respectively, but kept getting the following error:
```
0:00:49.387 => cli11-2.5.0-r0: empty non-meta packages must be marked as such
0:00:49.388 => cli11-2.5.0-r0: ERROR: package lint failed
```

z-erica:
this can't be merged as long as the “update to 2.6.1” commit isn't squashed with the “new package” commit

user/cli11/template.py
zacharyburnett: we can figure out how to package `-devel` and `-static` later; for now, the current build config works

zacharyburnett:
> this can't be merged as long as the “update to 2.6.1” commit isn't squashed with the “new package” commit

done, thanks!

user/cli11/template.py
```diff
@@ -0,0 +1,27 @@
...
+    "-DCLI11_BUILD_EXAMPLES_JSON=OFF",  # Build some additional example using json libraries
+    "-DCLI11_INSTALL=ON",  # install CLI11 to the install folder during the install process
+    "-DCLI11_FORCE_LIBCXX=OFF",  # use libc++ instead of libstdc++ if building with clang on linux
+    "-DCLI11_CUDA_TESTS=OFF",  # build the tests with NVCC
+    "-DCLI11_BUILD_TESTS=ON",  # Build the tests.
+]
+hostmakedepends = ["boost", "cmake", "doxygen", "ninja", "pkgconf", "python"]
+makedepends = ["catch2-devel"]
+pkgdesc = "Command line parser for C++11 and beyond"
+license = "custom:cli11"
```
z-erica: why custom? project seems to be under `BSD-3-Clause` https://github.com/CLIUtils/CLI11/tree/v2.6.1?tab=readme-ov-file#license

user/cli11/template.py
z-erica: seems fine? looks like arch and fedora both keep it as a header only library, and i think here it's preferable to a static library without a shared counterpart (that's why the base package ends up empty here, usually it'd contain the .so)

user/cli11/template.py
```diff
@@ -0,0 +1,27 @@
...
+    "-DCLI11_SINGLE_FILE_TESTS=OFF",  # Run the tests on the generated single file version as well
+    "-DCLI11_BUILD_DOCS=ON",  # build CLI11 documentation and book
+    "-DCLI11_BUILD_EXAMPLES=OFF",  # Build the example programs.
+    "-DCLI11_BUILD_EXAMPLES_JSON=OFF",  # Build some additional example using json libraries
+    "-DCLI11_INSTALL=ON",  # install CLI11 to the install folder during the install process
+    "-DCLI11_FORCE_LIBCXX=OFF",  # use libc++ instead of libstdc++ if building with clang on linux
+    "-DCLI11_CUDA_TESTS=OFF",  # build the tests with NVCC
+    "-DCLI11_BUILD_TESTS=ON",  # Build the tests.
+]
+hostmakedepends = ["boost", "cmake", "doxygen", "ninja", "pkgconf", "python"]
```
z-erica: some of these look unnecessary to me:
- `boost` isn't even found according to the build log (`-- Boost not found, not adding boost tests`), i think because it's looking specifically for a version that's older than what we have? https://github.com/CLIUtils/CLI11/blob/v2.6.1/tests/CMakeLists.txt#L26 plus it seems to only be used for testing? which makes it a `checkdepends` if anything at all
- `doxygen` is found, however it also says `missing components: dot`. considering the `-doc` subpackage only ends up containing the license file i don't think it's doing anything
- `python` is only required for a single include file build, which is disabled above (and by default...)

user/cli11/template.py
```diff
@@ -0,0 +1,27 @@
...
+makedepends = ["catch2-devel"]
```
z-erica: should be `checkdepends`

z-erica: or... not necessarily!! guess it downloads it unconditionally

zacharyburnett:
Thanks for reviewing! Sorry there were some extra packages!

PR #4749, opened 2025-09-07
user/haproxy-dataplaneapi: new package

ayushnix:
## Description

The HAProxy DataPlane API. This is useful starting from version 3.2.x because it supports writing certificates to disk when using the native ACME feature of HAProxy. Without this package, HAProxy will keep certificates in memory and one has to manually dump those certs on disk.

I haven't added a dinit service file for this package because the program expects `/etc/haproxy/dataplaneapi.yaml` and we can't ship files in `/etc`. Should I use a wrapper script as mentioned in the soju PR or do something else? Let me know.

## Checklist

### DIFF
--- a/user/haproxy-dataplaneapi/template.py
+++ b/user/haproxy-dataplaneapi/template.py
@@ -1,0 +1,17 @@
+pkgname = "haproxy-dataplaneapi"
+pkgver = "3.3.5"
+pkgrel = 0
+build_style = "go"
+make_build_args = [
+    "-ldflags="
+    + f" -X main.GitTag={pkgver}"
+    + " -X main.GitCommit="
+    + " -X main.GitDirty= ",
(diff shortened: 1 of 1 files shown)


q66:
please update and add a service file

handling of /etc is up to you, i can't quite figure out from the docs how much the yaml is needed - if it needs the yaml to run and doesn't have good defaults and it is possible to provide a yaml that has good defaults, you can install it the location, else we can leave setting it up to the user

ayushnix:
On second thoughts, I'll prefer closing this for now until I figure out how to dump the certs on disk using dataplane. It's helpful for formatting the `haproxy.cfg` file and some declarative updates to the config but I'm not sure I want to maintain this package until I can figure out how to dump certs on disk.

Apologies for the noise. 😓

q66:
ok, sure

PR #4760, opened 2025-09-10
user/salt-lts: update to 3006.16

rbgarga:


rbgarga:
> .16 is out, also please use https url

done

PR #4767, opened 2025-09-10
user/vnstat: new package

ayushnix:
## Description

add [vnstat](https://humdi.net/vnstat), a network traffic and bandwidth monitor

seems similar to nload

## Checklist

### DIFF
--- a/user/vnstat/files/vnstatd
+++ b/user/vnstat/files/vnstatd
@@ -1,0 +1,6 @@
+type = process
+command = /usr/bin/vnstatd -n
+run-as = _vnstatd
+log-type = buffer
+depends-on: local.target
+depends-on: network.target
--- a/user/vnstat/files/sysusers.conf
+++ b/user/vnstat/files/sysusers.conf
@@ -1,0 +1,3 @@
+# Create vnstatd system user
+
+u _vnstatd - "vnstatd user" /var/lib/vnstat /usr/bin/nologin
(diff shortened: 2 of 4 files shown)


q66:
i don't get how this is supposed to be used given there is no service and the sysuser is not used by anything

ayushnix:
missed the service file when I raised the PR, apologies

should be fine now

user/vnstat/files/vnstatd
```diff
@@ -0,0 +1,7 @@
+type = process
+command = /usr/bin/vnstatd -n
+restart = on-failure
+smooth-recovery = true
+run-as = _vnstatd
+log-type = buffer
+depends-on: network.target
```
q66, line 6: network.target is never used alone

user/vnstat/files/vnstatd
```diff
@@ -0,0 +1,7 @@
+type = process
+command = /usr/bin/vnstatd -n
+restart = on-failure
+smooth-recovery = true
```
q66: this is mostly pointless boilerplate

q66: we should probably just build dinit with `--default-auto-restart=on-failure`, it makes much more sense

user/vnstat/files/vnstatd
ayushnix: added `local.taget`

user/vnstat/files/vnstatd
ayushnix: > this is mostly pointless boilerplate

removed restart and recovery vars

PR #4799, opened 2025-09-25
user/atop: new package

Laikulo:
## Description

This PR adds a `atop` to the `user` repo.  
Atop is (my personal favorite) top variant.

### Packaging notes:
* Upstream supports "sysv", "systemd" and "generic" install modes. "Generic" is used here.
* With the default hardening (specifically `int`), the application dies on startup. I've not yet run down exactly where in the application this is occurring (the notes in PACKAGING were really useful to understand what was happening)
* Upstream does not support `make check` or equivalent.

## Checklist

### DIFF
--- a/user/atop/template.py
+++ b/user/atop/template.py
@@ -1,0 +1,39 @@
+pkgname = "atop"
+pkgver = "2.13.0"
+pkgrel = 0
+build_style = "makefile"
+make_build_args = [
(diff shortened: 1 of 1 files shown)


Laikulo:
Thanks for taking the time to talk and assist in IRC last night. I was def. a bit out of my depth and in somewhat strange territory.

PR #4833, opened 2025-10-01
main/eartag: update to 1.0.2

sudanchapagain:
## Description

## Checklist


sudanchapagain:
i think eartag should be moved to user. the package was never properly packaged and has not been updated too for quite a while.

i tried to package all it's dependencies (aio libs) but i gave up because its like 4+ packages till now and i dont even know how much more needs to be packaged.

the core functionality works. so updating in this state is fine ig but the whole package needs reconsideration.

PR #4834, opened 2025-10-02
user/fish-shell: update to 4.1.1

ayushnix:
## Description

updated fish to 4.1.1

The release type 'None' fails to build with 4.1.1 and also prints a warning. The gettext dependency is now needed only for msgfmt and fish has otherwise switched to its own gettext implementation. Additional details can be found in the 4.1.0 release notes.

https://github.com/fish-shell/fish-shell/releases/tag/4.1.0

## Checklist


ayushnix:
The tests succeed on my RPi4, not sure why the test failed in the arm64 container.

PR #4841, opened 2025-10-07
[CI EXPERIMENT] relbump numpy, no other changes

z-erica:
see https://github.com/chimera-linux/cports/pull/4210#issuecomment-3375304014


q66:
Screenshot From 2025-10-16 09-05-22

passes fine on builder so i guess disregard CI

PR #4885, opened 2025-10-16
[CI SKIP] user/wasmtime: update to 37.0.2

subnut:
Now that there's the pulley backend, wasmtime _should_ technically work on architectures which cranelift doesn't support yet. **Further testing needed.**

---

Known to work
- x86_64

Should work (hasn't been tested yet)
- aarch64
- riscv64

Needs testing
- loongarch64
- ppc64le
- ppc64
- ppc


subnut:
nothing requires wasmtime 37.x yet.. I originally intended this for tree-sitter, but that only needs wasmtime 31.x

PR #4914, opened 2025-10-26
user/prs: new package

jcgruenhage:
## Description

Adds `prs` with the three currently implemented backends: GnuPG, gpgme and rpgpie.

## Checklist

### DIFF
--- a/user/prs-gnupg-bin/template.py
+++ b/user/prs-gnupg-bin/template.py
@@ -1,0 +1,53 @@
+# Keep prs-gnupg-bin, prs-gpgme and prs-rpgpie in sync
+pkgname = "prs-gnupg-bin"
+pkgver = "0.5.7"
+pkgrel = 0
+build_wrksrc = "cli"
+build_style = "cargo"
+prepare_after_patch = True
+make_build_args = [
+    "--no-default-features",
+    "--features="
+    + ",".join(
+        [
+            "alias",
+            "backend-gnupg-bin",
+            "clipboard",
+            "notify",
+            "select-fzf-bin",
+            "select-skim",
+            "tomb",
+            "totp",
+        ]
+    ),
+]
+hostmakedepends = ["cargo-auditable"]
+makedepends = ["rust-std"]
+depends = ["gnupg"]
+pkgdesc = "Password manager CLI inspired by pass"
+subdesc = "GnuPG binary backend"
+license = "GPL-3.0-only"
+url = "https://timvisee.com/projects/prs"
+source = (
+    f"https://gitlab.com/timvisee/prs/-/archive/v{pkgver}/prs-v{pkgver}.tar.gz"
+)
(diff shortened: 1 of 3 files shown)


user/tessen/patches/0001-add-prs-support.patch
```diff
@@ -0,0 +1,113 @@
...
++set -x
+ 
+ # GLOBAL VARIABLES
+ declare _PASS_BACKEND _DMENU_BACKEND _TSN_ACTION _TSN_CONFIG
+@@ -102,6 +102,11 @@ get_pass_data() {
+     if [[ ${#passdata[@]} -eq 0 ]]; then
+       _die "the selected file is empty"
+     fi
++  elif [[ $_PASS_BACKEND == "prs" ]]; then
++    mapfile -t passdata < <(prs show "$_TSN_PASSFILE" 2> /dev/null)
```
timvisee: You may use `--quiet` to hide the header and prevent writing anything to stderr:

```suggestion
+    mapfile -t passdata < <(prs show --quiet "$_TSN_PASSFILE")
```

If the application is not an interactive CLI this may be useful too:

```suggestion
+    mapfile -t passdata < <(prs show --quiet --no-interact "$_TSN_PASSFILE")
```

(untested)

jcgruenhage: I have considered this in https://tangled.org/@jcg.re/tessen/commit/9657b646ae1fddf455759ba110c98edb353d523c, thanks for the feedback!

Although, just FYI: For me locally `--no-interact` wouldn't work, it'd still show a fuzzy search window

timvisee: Thanks!

> Although, just FYI: For me locally --no-interact wouldn't work, it'd still show a fuzzy search window

I see. I consider that a bug. Thanks for the report.

q66:
i still don't get what's the purpose behind having 3 backends

it makes this template nasty

q66:
i dug up the old irc logs...

imo the best path is to build a `prs` package with a default backend that is the most useful and then for the other backends, create separate templates that should be matched to the main one (e.g. `prs-rpgpie`) - they will conflict with the main `prs` package (due to providing the same command) but that is fine as i don't think it's an issue to expect the user to install only one

so please do that, let's have 3 regular, clean templates instead of this

jcgruenhage:
The three builds within one template dedupe quite a few things and also mean that most of the build can be cached and we only have the completions / man pages once. If three templates are preferable, I can also split this into three templates with comments to keep them in sync.

Thanks for spending the time digging this up!

q66:
the completions/man pages don't take much and i prefer 3 well readable templates over one that's convoluted and results in an empty base package for no reason

jcgruenhage:
Makes sense, splitting it, will update the PR once the three templates are ready.

q66:
lgtm

jcgruenhage:
Thanks! :)

PR #4922, opened 2025-11-03
user/fish-shell: update to 4.1.2

sugar700:
## Description

First of all, this removes the gettext patch. Fish shell doesn't use gettext anymore (albeit gettext is still needed at build time), see https://fishshell.com/docs/current/relnotes.html#changelog-4-1-gettext. This also means the `fish-shell-locale` package is gone, with all locales being stored directly within the executable (it is possible to disable localization, but that requires disabling cargo feature, making it impractical for `-locale` split).

The check for lintpixmaps was deleted too, as a shortcut icon was removed: fish-shell/fish-shell#11104.

As for removing `/etc/fish`, that could have been done for a long time, since fish-shell/fish-shell#2799.

## Checklist


ayushnix:
A PR has already been raised for [fish-shell 4.1.2](https://github.com/chimera-linux/cports/pull/4834).

q66:
this one looks better because it does not needlessly mess with the cmake changes

ayushnix:
Hmm, strange. I do remember facing build failures without changing the release profile at the time but now it builds successfully without me touching the `none-profile.patch` file.

PR #4925, opened 2025-11-03
user/cockatrice: new package

afiw0:


user/cockatrice/template.py
```diff
@@ -0,0 +1,30 @@
...
+depends = [
+    "protobuf",
+    "qt6-qtbase",
+    "qt6-qtimageformats",
+    "qt6-qtmultimedia",
+    "qt6-qtsvg",
+    "qt6-qttools",
+    "qt6-qtwebsockets",
+    "zlib-ng-compat",
+]
```
sugar700: You probably don't need to declare dependencies explicitly, considering [automatic dependencies](https://github.com/chimera-linux/cports/blob/master/Packaging.md#automatic_deps) system.

ghost: Thanks! I removed the `depends`, and they all got correctly picked up.

q66: okay, but why make it a separate commit if you're just fixing a thing that was wrong in the original commit in the first place

ghost: q66: Sorry. Is it correct now?

PR #4947, opened 2025-11-09
main/shadow: update gnome login keyring with passwd

Gnarwhal:
## Description

`passwd` does not currently update the gnome login keyring

## Checklist


q66:
i'm reluctant to add this, also this will break badly if gnome-keyring is not present on the system and will lock you out of the tools

q66:
yeah i don't think the low level tools need this

PR #4971, opened 2025-11-26
user/tokei: update to 14.0.0

jcgruenhage:


jcgruenhage:
Given how many PRs are open upstream I am not very hopeful. If the git2 fix is merged I'll look into more upstream updates, but until then I'll not bother.

wezm:
Yeah fair enough 👍

PR #5007, opened 2025-12-11
cbuild: use GIT_CONFIG_{GLOBAL,SYSTEM} to ignore global and system gitconfigs

### LINKED ISSUE #5427: Using `HOME` in gitconfig breaks cbuild
### Description
The conditional cbuild uses to see if the cports directory is a Git repo *always* returns false if the user includes a path in their gitconfig that uses `HOME`.  cbuild removing `HOME` from the environment causes paths that use it to fail to expand in the gitconfig and causes Git to error.

### Workarounds
It appears `HOME` is undefined so cbuild doesn't use the user's gitconfig. This can be fixed by setting Git's `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` environment variables to `/dev/null`

Having the user expand `HOME` in paths that reference it manually.

### Error messages
```shell
$ ./cbuild
You have to run cbuild from a git clone
$ # stderr from the command cbuild ran
$ env -u HOME git rev-parse --is-inside-work-tree
error: could not expand include path '~/.config/git/credentials'
fatal: bad config line 2 in file /home/user/.config/git/config
```

### How to Reproduce
1. Setup gitconfig

~/.config/git/config
```gitconfig
[include]
         path = ~/.config/git/credentials
[init]
        defaultBranch = main
```
~/.config/git/credentials
```gitconfig
[user]
	name = Alice
	email = alice@example.com
```

2. Run `cbuild`
 ```
$ cd cports
$ ./cbuild
```
3. See error

unrealapex:
Suggested-by: pj

## Description

This patch ignores system and global gitconfigs using `GIT_CONFIG_{GLOBAL,SYSTEM}` instead of unsetting `HOME`. Unsetting `HOME` would cause cbuild's Git repository check to error when the user includes a path in their gitconfig that uses `HOME`.

## Checklist

- [ ] I have read [Packaging.md](https://github.com/chimera-linux/cports/blob/master/Packaging.md#quality_requirements)

Closes  #5427

### DIFF
--- a/src/cbuild/core/git.py
+++ b/src/cbuild/core/git.py
@@ -1,14 +1,16 @@
 # silly wrapper around git so we can ignore ~/.gitconfig as needed
 
+import os
 import subprocess
 
 
 def call(args, gitconfig=False, foreground=False, cwd=None):
-    if gitconfig:
-        bcmd = ["git"]
-    else:
-        # still use the rest of the environment
-        bcmd = ["env", "-u", "HOME", "--", "git"]
+    if not gitconfig:
+        env = os.environ
+        env["GIT_CONFIG_GLOBAL"] = "/dev/null"
+        env["GIT_CONFIG_SYSTEM"] = "/dev/null"
+
+    bcmd = ["git"]
 
     ret = subprocess.run(bcmd + args, capture_output=not foreground, cwd=cwd)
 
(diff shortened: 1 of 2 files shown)


q66:
we can't do this because these env vars are a too new of a feature and cbuild may run on some pretty old hosts (eg the version on the riscv64 builder can't satisfy this)

q66:
should probably work just as well

unrealapex:
That makes sense, this will do!

PR #5026, opened 2025-12-17
user/fht-compositor: new package

aniruddhgutta:
## Description

fht-compositor is a dynamic tiling Wayland compositor based on the [Smithay](https://github.com/smithay) compositor library. It has a layout model inspired by popular X11 window managers such as [DWM](https://dwm.suckless.org/), [AwesomeWM](https://awesomewm.org/) and [XMonad](https://xmonad.org/). Each connected output has 9 workspaces, each workspace managing a bunch of windows.

## Checklist

### DIFF
--- a/user/fht-compositor/template.py
+++ b/user/fht-compositor/template.py
@@ -1,0 +1,40 @@
+pkgname = "fht-compositor"
+pkgver = "25.10.1"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = [
+    "cargo-auditable",
+    "pkgconf",
+]
+makedepends = [
+    "libdisplay-info-devel",
+    "libinput-devel",
+    "libseat-devel",
+    "libxkbcommon-devel",
+    "mesa-devel",
+    "pipewire-devel",
+    "pixman-devel",
+    "rust-std",
+    "udev-devel",
+]
+depends = ["so:libEGL.so.1!mesa-egl-libs", "xwayland-satellite"]
+pkgdesc = "Dynamic tiling Wayland compositor"
+license = "GPL-3.0-only"
+url = "https://github.com/nferhat/fht-compositor"
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "6ed6816a8e003b2d0b1675475bd73f5da14e5c6bd255573f7e7c154acd9adb46"
+hardening = ["vis", "cfi"]
+
+
+def install(self):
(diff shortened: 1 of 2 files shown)


user/fht-compositor/template.py
```diff
@@ -0,0 +1,34 @@
...
+source = f"{url}/archive/refs/tags/{pkgver}.tar.gz"
+sha256 = "6ed6816a8e003b2d0b1675475bd73f5da14e5c6bd255573f7e7c154acd9adb46"
+
+
+def install(self):
+    self.install_bin(f"target/{self.profile().triplet}/release/fht-compositor")
+    self.install_license("LICENSE")
+    self.install_file("res/systemd/fht-compositor.desktop", "usr/share/wayland-sessions")
+    self.install_file("res/fht-compositor.portal", "usr/share/xdg-desktop-portal/portals")
+    self.install_file("res/fht-compositor-portals.conf", "usr/share/xdg-desktop-portal")
```
subnut: did you run `ruff format` ?  
https://github.com/chimera-linux/cports/blob/master/Packaging.md#correct-style

user/fht-compositor/template.py
```diff
@@ -0,0 +1,34 @@
+pkgname = "fht-compositor"
+pkgver = "25.10.1"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = [
+    "cargo-auditable",
+    "pkgconf",
+    "wayland-protocols",
```
subnut: shouldn't this be in `makedepends`?

user/fht-compositor/template.py
aniruddhgutta: ooo no i didn't mb, i'll fix it

user/fht-compositor/template.py
aniruddhgutta: niri's template puts them in `hostmakedepends` so i assumed the same, should i change it?
wayland-protocols provides XML files for code generators, not libraries to link against, so `hostmakedepends` is correct

subnut: I meant `wayland-protocols` specifically. That seems to be a makedepends AFAICT.

aniruddhgutta: Well I just realised both `wayland-protocols` and `wayland-devel` weren't even needed as dependencies. I've removed both, it should be correct now no?

q66:
the version of pipewire-rs they use fails to build with current toolchain and there is no easy way to patch the update in

feel free to reopen if they cut a new release, lgtm otherwise

will also need the same broken= for 32bit like niri

PR #5043, opened 2025-12-24
 user/opensc: new package 

z-erica:
## Description
Smart card tools and middleware.

## Checklist


user/opensc/template.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@subpackage("opensc-devel")
+def _(self):
+    # exclude symlinks to non-library shared objects (onepin-opensc-pkcs11.so@ -> opensc-pkcs11.so)
+    return [
+        "usr/lib/*.a",
+        "usr/lib/lib*.so",
+        "usr/lib/pkgconfig",
+    ]
```
z-erica, line 42: this warns:
> WARNING: .so symlinks should be in the -devel package

however i believe the symlink does belong in the base package (not a library, but a loadable module)

user/opensc/template.py
```diff
@@ -0,0 +1,43 @@
...
+url = "https://github.com/OpenSC/OpenSC"
+source = f"{url}/releases/download/{pkgver}/opensc-{pkgver}.tar.gz"
+sha256 = "f16291a031d86e570394762e9f35eaf2fcbc2337a49910f3feae42d54e1688cb"
+
+
+def post_install(self):
+    # remove bash completions for commands we don't install
+    for tool in ["npa-tool"]:
+        # guard against the command being installed in the future
+        assert not (self.destdir / "usr/bin" / tool).exists()
```
z-erica: i believe this executable requires OpenPACE, which we don't have packaged. this guard is for the case that's eventually added to makedepends, in which case the completion should be installed

z-erica: this is prolly better so nm

PR #5050, opened 2025-12-26
user/minicom: new package

Laikulo:
## Description

Adds minicom, a text-mode serial terminal emulator that can translate VT102 and ANSI to whatever $TERM and terminfo say it should be.

Note that if there are no serial device specified, and /dev/modem does not exist, the application will exit with no text and rc=1. I believe this to be intended upstream behavior. 

The default may be changed by either system-wide or per-user configuration.

## Checklist

(Tested specifically with a USB Serial adapter, specified with -D/dev/serial/by-id/....)

### DIFF
--- a/user/minicom/template.py
+++ b/user/minicom/template.py
@@ -1,0 +1,18 @@
+pkgname = "minicom"
+pkgver = "2.10"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = [
+    "automake",
+    "gettext-devel",
+    "pkgconf",
+]
+makedepends = [
+    "linux-headers",
+    "ncurses-devel",
+]
+pkgdesc = "Serial communication program"
+license = "GPL-2.0-or-later"
+url = "https://salsa.debian.org/minicom-team/minicom"
+source = f"{url}/-/archive/{pkgver}/minicom-{pkgver}.tar.gz"
+sha256 = "66ff82661c3cc49ab2e447f8a070ec1a64ba71d64219906d80a49da284a5d43e"


user/minicom/template.py
```diff
@@ -0,0 +1,19 @@
...
+makedepends = [
+    "autoconf",
+    "automake",
+    "clang",
+    "gettext-devel",
+    "gm4",
+    "linux-headers",
+    "ncurses-devel",
+    "pkgconf",
+]
```
wezm, line 13:
```suggestion
hostmakedepends = [
    "automake",
    "pkgconf",
    "gettext-devel",
]
makedepends = [
    "linux-headers",
    "ncurses-devel",
]
```

- clang is present by default
- automake pulls in autoconf and gm4
- binaries that run on the host need to be in `hostmakedepends`

user/minicom/template.py
```diff
@@ -0,0 +1,19 @@
...
+pkgdesc = "Friendly serial communication program"
```
wezm:
```suggestion
pkgdesc = "Serial communication program"
```

user/minicom/template.py
Laikulo: Agree on this, I'll toss a commit once I've validated.

user/minicom/template.py
Laikulo: This text is taken verbatim from the manpage's description, unsure about this.

user/minicom/template.py
Laikulo: Done. Validated that I can cross compile, but I cant get a riscv64 vm to behave enough to verify properly.

Laikulo: Update: after fiddling with the qemu config, I was able to install and verify that a cross amd64>riscv64 works as expected.

user/minicom/template.py
Laikulo: In light of that, would you still like this modified?

q66: it doesn't matter at all what the manpage's description says

Laikulo:
> still not sure what's the point given the base system has a better serial console builtin, but whatever i suppose...

Do tell, my weak `apk search`-fu failed me when I needed one. Wound up falling back to `screen`. I tried to get picocom up first, but it had issues with the way it was trying to do custom bauds.

(I've not yet figured out how to get it to print descriptions in search results, even when searching descriptions, so all I get is a list of packages with no context)

user/minicom/template.py
Laikulo: Applied

PR #5051, opened 2025-12-26
user/stow: new package

angelwood-xyz:
## Description
GNU Stow is a Symlink farm manager/Dotfiles manager written in perl. 

There was a previous [pull request](https://github.com/chimera-linux/cports/pull/4794) to add this package, but it needed to patch the build system to ignore warnings. Due to some upstream changes patching is no-longer required!

There is also 2 trivial dependencies included in this pull request(they are only needed for the unit testing of the build): `perl-test-output` and `perl-capture-tiny`

## Checklist

### DIFF
--- a/user/perl-capture-tiny/template.py
+++ b/user/perl-capture-tiny/template.py
@@ -1,0 +1,13 @@
+pkgname = "perl-capture-tiny"
+pkgver = "0.50"
+pkgrel = 0
+build_style = "perl_module"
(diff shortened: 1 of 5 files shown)


wezm (commented):
We already have `user/xstow`, do we need this version too?

angelwood-xyz:
Last time I used `xstow` it was missing some core features. But it has been quite a few years so I should re-investigate it as a option

PR #5066, opened 2026-01-01
main/[shadow, util-linux]: use sulogin from util-linux

firasuke:
## Description

This PR enables `sulogin` from util-linux as it is more widely used (and I think it is the expected version based on [this comment](https://github.com/chimera-linux/dinit/blob/master/doc/linux/services/recovery#L2)), and removes shadow's version of `sulogin`.

## Checklist

### DIFF
--- a/main/shadow/template.py
+++ b/main/shadow/template.py
@@ -56,10 +56,6 @@ def pre_install(self):
 def post_install(self):
     self.uninstall("usr/sbin")
 
-    # install sulogin which is noinst
-    self.install_bin("src/sulogin")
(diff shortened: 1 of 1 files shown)


q66:
i have no plans to switch to further util-linux utilities

also you haven't tested this

q66:
not sure why you're pointing to that comment considering chimera does not use dinit's sample service files either

firasuke:
> not sure why you're pointing to that comment considering chimera does not use dinit's sample service files either

Oh my bad then, I thought you intended to use the version from `util-linux`. Closing..

PR #5067, opened 2026-01-02
user/fish-shell: update to 4.4.0

ayushnix:
## Description

updated fish-shell to 4.4.0

The none-profile.patch is the same as before but created with
git-format-patch.

cross-compilation doesn't really work because a native binary is used to
build man pages and docs during build, partially fixed by defining
Rust_CARGO_TARGET introduced in the following commit

https://github.com/fish-shell/fish-shell/commit/205d80c75aa8fb2e11dbf477e2f377ca2b98ace2

python-sphinx is now needed to build man pages and documentation but it
doesn't work when cross compiling fish as mentioned earlier

https://github.com/fish-shell/fish-shell/commit/135fc73191dc8d7d39ead030f033e071ca05dc23

## Checklist

### DIFF
--- a/user/fish-shell/patches/none-profile.patch
+++ b/user/fish-shell/patches/none-profile.patch
@@ -1,13 +1,23 @@
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -54,8 +54,8 @@ function(CREATE_TARGET target)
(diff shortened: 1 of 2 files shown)


ayushnix:
Hmm, looks like pre-build docs were removed from 4.3.0. I'll push a fix soon.

PR #5078, opened 2026-01-06
user/transmission: update license

ile6695:
Transmission can be redistributed and/or modified under the terms of the GNU GPLv2 (http://www.gnu.org/licenses/license-list.html#GPLv2), the GNU GPLv3 (http://www.gnu.org/licenses/license-list.html#GNUGPLv3), or any future license endorsed by Mnemosyne LLC.

In addition, linking to and/or using OpenSSL is allowed.

## Description

One line change that is more of a documentation issue.

## Checklist

### DIFF
--- a/user/transmission/template.py
+++ b/user/transmission/template.py
@@ -39,7 +39,7 @@ makedepends = [
     "qt6-qttools-devel",
 ]
 pkgdesc = "BitTorrent client"
-license = "GPL-2.0-or-later"
+license = "GPL-3.0-only"
 url = "https://github.com/transmission/transmission"
 source = f"{url}/releases/download/{pkgver}/transmission-{pkgver}.tar.xz"
 sha256 = "2a38fe6d8a23991680b691c277a335f8875bdeca2b97c6b26b598bc9c7b0c45f"


ile6695:
If there ever is GPLv4, the `GPL-2.0-or-later` would allow its usage, while the project only says GPLv2 or GPLv3 are allowed.

q66:
then use an spdx license expression that actually matches reality

q66:
we have SPDX license expressions to be properly able to express multi-licensing

q66:
yes, you clarified that after i asked about it

the PR description says what the licensing is, but initially to me, our current license string seemed to be the more accurate one given there is no GPLv4 (and won't be for a while)

the distro doesn't choose the license, it declares the available licenses in the packaging metadata, and we have expressions to allow this to be described, so `GPL-2.0-only OR GPL-3.0-only` would've been better than both (probably still not entirely accurate due to the openssl exception, but that's probably being overly pedantic, especially considering we only ship openssl version with the apache license now)

ile6695:
I'll admit that the openssl could make the spdx more complicated, but stating that it "does not match reality" instead of saying what is wrong with it is just too much for me.

PR #5106, opened 2026-01-16
user/clifm: new package

listentolist:
## Description

This PR adds a new package: user/clifm, a command line terminal file manager.

## Checklist

### DIFF
--- a/user/clifm/template.py
+++ b/user/clifm/template.py
@@ -1,0 +1,18 @@
+pkgname = "clifm"
+pkgver = "1.27.1"
+pkgrel = 0
+build_style = "makefile"
+makedepends = [
+    "acl-devel",
+    "file-devel",
+    "libcap-devel",
+    "linux-headers",
+    "readline-devel",
+]
+pkgdesc = "Shell-like, command line terminal file manager"
+license = "GPL-2.0-or-later"
+url = "https://github.com/leo-arch/clifm"
+source = f"{url}/archive/v{pkgver}.tar.gz"
+sha256 = "a35cd1ccbb83f1261c3c5b14b5b4733cf0555be68579b3cb19fa8b36076a5339"
+# no tests
+options = ["!check"]


user/clifm/template.py
```diff
@@ -0,0 +1,19 @@
...
+pkgrel = 0
+build_style = "makefile"
+makedepends = [
+    "acl-devel",
+    "file-devel",
+    "libcap-devel",
+    "linux-headers",
+    "readline-devel",
+]
+depends = ["cmd:cp!chimerautils", "cmd:mv!chimerautils", "cmd:rm!chimerautils"]
```
q66: this is just unnecessary, there's never going to be a system without cp/rm/mv

user/clifm/template.py
```diff
@@ -0,0 +1,19 @@
+pkgname = "clifm"
+pkgver = "1.27.1"
+pkgrel = 0
+build_style = "makefile"
+makedepends = [
+    "acl-devel",
+    "file-devel",
+    "libcap-devel",
+    "linux-headers",
+    "readline-devel",
```
q66, line 10: unless it breaks, we prefer using libedit (you can use `libedit-readline-devel` to avoid patching)

q66: though in this case it likely breaks because i see a lot of niche readline features used

user/clifm/template.py
listentolist: Okay, I removed it.

user/clifm/template.py
listentolist: Then I would just leave the readline-devel make dependency as it is?

q66: dunno, test it with libedit and see?

listentolist: It does not build with libedit-readline-devel as it complains about several undeclared functions and macros. It seems that clifm depends heavily on readline. Would you prefer not having it added then?

q66: just leave it as is then

PR #5120, opened 2026-01-24
user/mpdscribble: new package

aniruddhgutta:
## Description

A Music Player Daemon (MPD) client which submits information about tracks being played to a scrobbler (e.g. last.fm).

## Checklist


q66:
i improved the services (added system service + dependency on mpd is wrong because mpd may be running somewhere remote; we still want to start this after mpd though, so `after` is the correct type) and some other minor adjustments

PR #5132, opened 2026-01-27
user/uv: update 0.9.28

voidisnull:
## Description

Refreshed the patch to align with upstream

## Checklist

### DIFF
--- a/user/uv/template.py
+++ b/user/uv/template.py
@@ -1,5 +1,5 @@
 pkgname = "uv"
-pkgver = "0.9.17"
+pkgver = "0.9.28"
 pkgrel = 0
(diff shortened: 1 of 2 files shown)


user/uv/template.py
```diff
@@ -36,7 +36,7 @@ def prepare(self):
         "--package",
         "libc",
         "--precise",
-        "0.2.174",
+        "0.2.177",
         allow_network=True,
     )
```
wezm: This can be removed. The lock file has 0.2.178 in it now.

voidisnull: I thought that but nvm I have updated the PR to remove the cargo update part

PR #5140, opened 2026-01-31
user/cagebreak: new package

couldbeanemail:
## Description

cagebreak is a Wayland compositor in the spirit of ratpoison on X.
i have been using this package for a while now and haven't ran into non-self-caused issues.
Upstream does have some bugs, and the tests don't really work in a cbuild build root.
I spent a bit of time trying to get the checks to work. That would require:
a) gsed dep, patching tests to use that or fixing it upsream.
b) patching some bad paths in test/arguments ~> build-arguments{/,}
    pretty sure this is not some gnu behavior but just a bug in the tests.
c) patching missing XDG_RUNTIME_DIR=$(pwd)
a through c could be fixed upstream but its hardly worth the effort since
d) tests still won't work because missing drm backend in bldroot
   so wlroots explodes, and well i think all hope is lost to test this in cbuild.
this is just for testing the basic test suite.
thats also why no hardening opts since no tests.

## Checklist

### DIFF
--- a/user/cagebreak/patches/00_fix_etc.patch
+++ b/user/cagebreak/patches/00_fix_etc.patch
@@ -1,0 +1,45 @@
+diff --git a/cagebreak.c b/cagebreak.c
+index 878d5b9..b513d74 100644
+--- a/cagebreak.c
++++ b/cagebreak.c
+@@ -682,6 +682,14 @@ main(int argc, char *argv[]) {
+ 			conf_ret = set_configuration(&server, default_conf);
+ 		}
+ 
++		// Fallback Chimera
++		if(conf_ret == 1) {
++			char *default_conf = "/usr/share/etc/cagebreak/config";
++			wlr_log(WLR_INFO, "Loading fallback distro config: \"%s\"",
++			        default_conf);
++			conf_ret = set_configuration(&server, default_conf);
++		}
++
+ 		if(conf_ret != 0 || !server.running) {
+ 			ret = 1;
+ 			goto end;
+diff --git a/man/cagebreak-config.5.md b/man/cagebreak-config.5.md
+index f6de22a..d3500db 100644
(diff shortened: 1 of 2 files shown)


couldbeanemail:
i forgot, i wanted to also patch the man pages to reflect patches to fallback config file in /usr/share/etc/cagebreak, will fix and then set as a non draft.

user/cagebreak/patches/00_fix_etc
jcgruenhage, line 1: File should have a `.patch` extension

jcgruenhage (changes requested):

user/cagebreak/patches/00_fix_etc.patch
```diff
@@ -0,0 +1,45 @@
...
+ 		if(conf_ret != 0 || !server.running) {
+ 			ret = 1;
+ 			goto end;
+diff --git a/man/cagebreak-config.5.md b/man/cagebreak-config.5.md
+index f6de22a..d3500db 100644
+--- a/man/cagebreak-config.5.md
++++ b/man/cagebreak-config.5.md
+@@ -8,6 +8,8 @@ cagebreak-config(5) "Version 3.1.0" "Cagebreak Manual"
+ 
+ *\$XDG_CONFIG_PATH/cagebreak/config*
```
jcgruenhage, line 35: That's not related to your patch, but https://specifications.freedesktop.org/basedir/latest/ does not know `XDG_CONFIG_PATH`. Am I missing something or is this wrong upstream?

couldbeanemail:
Ah I missed this. This is wrong upstream in their man page but not in
their code. All references to _PATH i found with ddg seemed to just be
typos.

jcgruenhage (approved):
I think this is good as is now. @wezm, wdyt?

couldbeanemail:
no longer using chimera

PR #5156, opened 2026-02-02
user/erlang: update to 28.3.1

wezm:
## Description

**Note:** I'm not super experienced with Erlang. I updated the package, installed it, and confirmed that it could run a Hello World Gleam program. If anyone has more experience, additional testing would be appreciated.

**Update:** I did further testing with a less trival application once I had packaged `rebar3`, and that also worked.

## Checklist

### DIFF
--- a/user/erlang/template.py
+++ b/user/erlang/template.py
@@ -1,6 +1,6 @@
 # TODO: split devel, maybe the libs too? (may not be worth it)
 pkgname = "erlang"
-pkgver = "28.1.1"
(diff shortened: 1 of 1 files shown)


wezm:
Confirmed the crash is due to overflow hardening:

https://github.com/erlang/otp/blob/OTP-28.3.1/erts/emulator/beam/erl_process.c#L9750

I've updated the template to disable it, as I don't see a clear way to address it from quickly looking at the code.

PR #5157, opened 2026-02-02
user/rebar3: new package

wezm:
## Description

`rebar3` is an official build tool for erlang that deals with dependencies, compilation and artifacts. It's also invoked by the gleam (`user/gleam`) compiler when targeting erlang, and using erlang dependencies.

## Checklist

### DIFF
--- a/user/rebar3/template.py
+++ b/user/rebar3/template.py
@@ -1,0 +1,34 @@
+pkgname = "rebar3"
+pkgver = "3.27.0"
+pkgrel = 0
+hostmakedepends = ["erlang"]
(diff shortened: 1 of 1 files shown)


user/rebar3/template.py
```diff
@@ -0,0 +1,34 @@
+pkgname = "rebar3"
+pkgver = "3.26.0"
+pkgrel = 0
+makedepends = ["erlang"]
```
q66: hostmakedepends probably?

wezm: Yes, you're right. I've fixed it.

PR #5174, opened 2026-02-03
user/prismlauncher: update to 10.0.2, move from main

natthias:
## Description

This is based on #4330, as prismlauncher has introduced a dependency on gamemode.
Also adds toml++, another new dependency of prismlauncher.

## Checklist


q66:
i added gamemode, i think we can clean this up now

q66:
looks like the main thing is to update it? the rest looks pretty good

natthias:
Updated to 11.0.2. Minecraft 26 does not work, but it didn't work for me on previous versions either.

PR #5197, opened 2026-02-17
packaging: update link to list of sanitizers

wezm:
The list was moved in https://github.com/llvm/llvm-project/commit/2622e6bfa076c60d4556c066245895e2766a7285


wezm:
Good idea, I've updated the link.

PR #5199, opened 2026-02-19
main/gnupg: apply freepg patchset

jcgruenhage:
## Description

We've had a HACK-revert-rfc4880bis-default.patch for a while, and by now there's
a better patchset available, [FreePG](https://freepg.org/).

Multiple distros carry patches from this patchset, and I recognize the people
behind this. I think shipping this patchset is reasonable, a lot more reasonable
than shipping unpatched GnuPG.

Alternatives to consider:
 - cherry-pick what patches we apply from FreePG instead of taking all of them.
   Most distros opt for this, but I think taking the whole thing is better,
   basically everything in there looks reasonable to me at a glance. If we'd
   like to only partially pick patches, I'm fine with trimming it down, just
   tell me to do so then.
 - have unpatched upstream GnuPG and FreePG as two different conflicting
   packages. I'd rather not.

## Checklist


q66:
i ended up taking a different approach because i don't like bundling the whole patchset into our tree, but this is now done

PR #5220, opened 2026-02-27
user/iamb: new package

couldbeanemail:
## Description

iamb is a decent matrix client. it's a tui application offering vim like keybindings and workflows.

It does have very slow moving releases. This has been somewhat unfortunate at times bc when authenticated media(? i think that's what it was called) was introduced, it took like i think a year, until the current release was there that would fix this. (the fix was faster, just that afaik it's policy to strongly discourage using non release versions like some 'random' git ref) . just as a heads up that this could happen again potentially?
It is also not a fully feature complete matrix client - as in you cannot interact with all features of the matrix protocol. it does most of what you would wanna do. 

these are the only concerns i could see with this. to me (and for me) this is the sanest matrix client still.

## Checklist

### DIFF
--- a/user/iamb/patches/00_man_share_etc.patch
+++ b/user/iamb/patches/00_man_share_etc.patch
@@ -1,0 +1,13 @@
+diff --git a/docs/iamb.5 b/docs/iamb.5
+index 8357f96..cdfee31 100644
+--- a/docs/iamb.5
++++ b/docs/iamb.5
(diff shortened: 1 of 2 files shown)


q66:
if the client does not read files from `/etc`, what sense does it make to put the example there? that's just confusing

files in `/usr/share/etc` are those that are meant to actually live in `/etc`

couldbeanemail:
no longer using chimera

PR #5307, opened 2026-03-14
main/geoclue: enable demo agent

ghost:
change source tarball from .bz2 to .gz
remove `demo-agent=false` flag

## Description
i am using darkman to auto change dark mode. i set location provider to geoclue, but it never works. after searching and debugging, i found out that it needs demo agent. so where is `/usr/libexec/geoclue-2.0/demos/agent`, this is what the commit did. and it tested successfully on my machine.

## Checklist

### DIFF
--- a/main/geoclue/template.py
+++ b/main/geoclue/template.py
@@ -1,13 +1,12 @@
 pkgname = "geoclue"
 pkgver = "2.8.0"
-pkgrel = 1
+pkgrel = 2
 build_style = "meson"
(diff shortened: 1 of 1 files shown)


wezm:
I would suggest changing it back. There's no need for consistency. Other packages typically use gzip tarballs because that's all that's available. There are plenty of others that use a smaller format, like xz when that's available.

yiyitiantian:
done.

PR #5314, opened 2026-03-15
user/cowsay: new package

cyberia-ng:
## Description

Cowsay is a well-known humorous program that generates an ASCII art cow with a speech bubble containing user input.

## Checklist

### DIFF
--- a/user/cowsay/template.py
+++ b/user/cowsay/template.py
@@ -1,0 +1,22 @@
+pkgname = "cowsay"
+pkgver = "3.8.4"
+pkgrel = 0
+build_style = "makefile"
+make_install_args = [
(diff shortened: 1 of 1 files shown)


q66:
not sure what these destdir shenanigans were supposed to be but i fixed it for you

cyberia-ng:
I think it was cruft copied over from the template for signify. Thanks.

q66:
oh, it's unnecessary in signify too, i just missed it

i dropped it there too, it's overall wrong

cyberia-ng:
Fair enough. I seem to remember it installing to the wrong place without it, but if it works now it works :shrug:

q66:
the builtin destdir is exactly the same, except by building it manually you're exposing internal paths that are an implementation detail

PR #5315, opened 2026-03-15
user/signify: new package

cyberia-ng:
## Description

OpenBSD tool to sign and verify signatures on files

## Checklist

### DIFF
--- a/user/signify/template.py
+++ b/user/signify/template.py
@@ -1,0 +1,19 @@
+pkgname = "signify"
+pkgver = "33"
+pkgrel = 0
+build_style = "makefile"
+make_install_args = [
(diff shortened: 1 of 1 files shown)


cyberia-ng:
OK so, it seems the CI fails because supposedly a newer version is available -- this is tagged as `v33` in the git repository -- however, the v33 tag is not released as a tarball, so if I set `pkgver = 33` in the `template.py` it fails. Is there a way to skip the newer version check in CI for now? (I have filed an issue upstream to get the v33 release fixed.)

cyberia-ng:
Development moved to Codeberg, where v33 does have a release tarball.

PR #5319, opened 2026-03-16
user/simdutf: new package

alex-s168:
## Description
simdutf is a library used by many other projects. I'm interested in it for Ladybird.

Doing a `grep simdutf` shows that we have a few packages that vendor simdutf, and we also have a few patches for those, but I hope they aren't required in latest simdutf anymore...

## Checklist

### DIFF
--- a/user/simdutf/template.py
+++ b/user/simdutf/template.py
@@ -1,0 +1,20 @@
+pkgname = "simdutf"
+pkgver = "8.2.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DBUILD_SHARED_LIBS=ON"]
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+pkgdesc = "Library of SIMD-accelerated Unicode routines"
+license = "MIT"
+url = "https://github.com/simdutf/simdutf"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "033a91b1d7d1cb818c1eff49e61faaa1b64a3a530d59ef9efef0195e56bda8b1"
+
+
+def post_install(self):
+    self.install_license("LICENSE-MIT")
+
+
+@subpackage("simdutf-devel")
+def _(self):
+    return self.default_devel()
--- a/user/simdutf-devel
+++ b/user/simdutf-devel
@@ -1,0 +1,1 @@
+simdutf
--- a/user/simdutf/patches/loongarch_include_hwcap.patch
+++ b/user/simdutf/patches/loongarch_include_hwcap.patch
@@ -1,0 +1,10 @@
+--- a/include/simdutf/internal/isadetection.h
++++ b/include/simdutf/internal/isadetection.h
+@@ -72,6 +72,7 @@
+
+ #if defined(__loongarch__) && defined(__linux__)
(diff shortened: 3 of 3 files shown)


user/simdutf/template.py
```diff
@@ -0,0 +1,25 @@
+pkgname = "simdutf"
+pkgver = "8.2.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = ["-DSIMDUTF_TESTS=OFF", "-DBUILD_SHARED_LIBS=ON"]
```
q66: tests disabled for no reason

user/simdutf/template.py
```diff
@@ -0,0 +1,25 @@
...
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
+pkgdesc = "Text processing at billions of characters per second"
```
q66: pkgdesc should be neutral and not an advertising line

user/simdutf/template.py
```diff
@@ -0,0 +1,25 @@
...
+license = "MIT"
+url = "https://github.com/simdutf/simdutf"
+source = (
```
q66: {url}

user/simdutf/template.py
```diff
@@ -0,0 +1,25 @@
...
+    f"https://github.com/simdutf/simdutf/archive/refs/tags/v{pkgver}.tar.gz"
+)
+sha256 = "033a91b1d7d1cb818c1eff49e61faaa1b64a3a530d59ef9efef0195e56bda8b1"
+# breaks shared library
+hardening = ["!vis"]
```
q66: that's the default

user/simdutf/template.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def post_install(self):
+    self.install_license("LICENSE-MIT")
+
+
+@subpackage("simdutf-devel")
+def _(self):
+    self.depends += [self.parent]
```
q66: not needed

user/simdutf/template.py
alex-s168: There are 114 tests which take a while to compile...

user/simdutf/template.py
alex-s168: "Fast Unicode and Base64 parsing library"?

user/simdutf/template.py
q66: and?

user/simdutf/template.py
alex-s168: There are lots of templates in main with `hardening = ["!vis"]`

q66: there's 25 of them that do that alone (1% of main) and it's just as unnecessary in those?

user/simdutf/template.py
alex-s168: Okay true. Even though upstream probably runs tests, it's better to also run them for chimera's specific setup

q66: the packaging manual specifically mandates that unit tests be enabled unless there is a really good reason and taking extra 1 minute to build (on my laptop) is not one...

user/simdutf/template.py
q66: "Library of SIMD-accelerated Unicode routines"

PR #5378, opened 2026-03-28
user/anyrun: new package

zacharyburnett:
## Description

Follow-on to https://github.com/chimera-linux/cports/pull/2817

Anyrun is a krunner-like runner for wayland that I use often
https://github.com/anyrun-org/anyrun

Anyrun itself now requires `anyrun-provider` as a go-between with plugins

## Checklist

### DIFF
--- a/user/anyrun/template.py
+++ b/user/anyrun/template.py
@@ -1,0 +1,31 @@
+pkgname = "anyrun"
+pkgver = "25.12.0"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo-auditable", "pkgconf"]
+makedepends = [
+    "cairo-devel",
+    "gdk-pixbuf-devel",
+    "glib-devel",
+    "gtk4-devel",
+    "gtk4-layer-shell-devel",
+    "pango-devel",
+]
+depends = ["anyrun-provider"]
+pkgdesc = "Wayland native, highly customizable runner"
+license = "GPL-3.0-or-later"
+url = "https://github.com/anyrun-org/anyrun"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "4213a2f65fd6139829128d3c7a7f4b54fec3181f8d549e212021341dd10c3d50"
+
+
+def install(self):
+    self.install_bin(f"target/{self.profile().triplet}/release/{pkgname}")
+    self.install_license("LICENSE")
+
+    self.install_file(
+        "target/*/release/*.so",
+        "usr/lib/anyrun/plugins/",
+        glob=True,
+    )
+    self.install_file("examples/config.ron", "etc/anyrun")
--- a/user/anyrun-provider/template.py
+++ b/user/anyrun-provider/template.py
@@ -1,0 +1,16 @@
+pkgname = "anyrun-provider"
+pkgver = "25.12.0"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo-auditable", "pkgconf"]
+makedepends = []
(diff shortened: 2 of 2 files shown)


user/anyrun/template.py
```diff
@@ -0,0 +1,27 @@
...
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "4213a2f65fd6139829128d3c7a7f4b54fec3181f8d549e212021341dd10c3d50"
+
+
+def install(self):
+    self.install_bin(f"target/{self.profile().triplet}/release/{pkgname}")
+    self.install_license("LICENSE")
+
+    self.install_file("target/release/*.so", "etc/anyrun/plugins/", glob=True)
+    self.install_file("examples/config.ron", "etc/anyrun/config.ron")
```
zacharyburnett: the anyrun installation instructions suggest making a `~/.config/anyrun/` directory and putting compiled plugins there, but I figured the user can copy them from `/etc/`. I know PACKAGING.md says to prefer `/usr/share/` but when I tried putting the `*.so`s there the checks failed with "ELF file in /usr/share/"

zacharyburnett: nevermind, that's what `/usr/lib/` is for :facepalm:

user/anyrun/template.py
```diff
@@ -0,0 +1,29 @@
...
+sha256 = "4213a2f65fd6139829128d3c7a7f4b54fec3181f8d549e212021341dd10c3d50"
+
+
+def install(self):
+    self.install_bin(f"target/{self.profile().triplet}/release/{pkgname}")
+    self.install_license("LICENSE")
+
+    self.install_file(
+        "target/release/*.so", "usr/lib/anyrun/plugins/", glob=True
+    )
```
zacharyburnett, line 30: only the `libanyrun_macros.so` plugin is copied here:
```shell
zrb@jupiter ~> ls /usr/lib/anyrun/plugins/
libanyrun_macros.so*
```

however, the build process _should_ be building all of these plugins:
```
zrb@jupiter ~/r/b/anyrun (s|master master@origin)> ls target/release/*.so
target/release/libactions.so*       target/release/librandr.so*
target/release/libanyrun_macros.so* target/release/librink.so*
target/release/libapplications.so*  target/release/libshell.so*
target/release/libdictionary.so*    target/release/libstdin.so*
target/release/libkidex.so*         target/release/libsymbols.so*
target/release/libniri_focus.so*    target/release/libtranslate.so*
target/release/libnix_run.so*       target/release/libwebsearch.so*
```

Maybe I'm using the glob incorrectly? I'll play around with this

zacharyburnett: oh, from running `self.do("ls", "target/release/")` locally, only the `libanyrun_macros.so` plugin is actually built in the first place with `cbuild`:
```shell
0:04:28.319 => anyrun-25.12.0-r0: running install...
.cargo-lock		deps			libanyrun_macros.d
.fingerprint		examples		libanyrun_macros.so
build			incremental
```

zacharyburnett: alright well, I can't figure out how to get the default plugins (see https://github.com/chimera-linux/cports/pull/5378#discussion_r3006540957) to build with `cbuild`; they build just fine and are placed in `target/release/` on my machine, but they just don't show up when running with `cbuild pkg`

Other than that however, the core `anyrun` and `anyrun-provider` build just fine and work; without the default plugins, they just don't show those plugins' results.

zacharyburnett: figured it out; they were in `target/x86_64-chimera-linux-musl/release/*.so`

PR #5409, opened 2026-04-05
main/php8.3: fix memory corruption on aarch64 and ppc64

wezm:
## Description

Running a PHP application on a ROCK64 (`aarch64`) I was seeing the `php-fpm` workers aborting as follows:

```
[04-Apr-2026 20:02:22] WARNING: [pool www] child 6775 said into stderr: "zend_mm_heap corrupted"
[04-Apr-2026 20:02:42] WARNING: [pool www] child 6775 exited on signal 6 (SIGABRT) after 62.634257 seconds from start
```

I found https://github.com/php/php-src/commit/93d32eae2751089ed0d255ba2cf37d5c8a7353e5 in the PHP repo which addresses an issue identified on aarch64 with LTO. Applying it seems to have resolved it. I can at least now access pages that previously aborted. I have not tested the ppc change, only pulled it in from the upstream commit.

Some tests were also failing prior to this change:

- Bug #74796: TLS encryption fails behind HTTP proxy [ext/openssl/tests/bug74796.phpt]
- sni_server [ext/openssl/tests/sni_server.phpt]
- sni_server with separate pk and cert [ext/openssl/tests/sni_server_key_cert.phpt]

The two sni ones are also disabled on Alpine, so they've been added to the skip list. A patch has been pulled in from upstream that addresses bug74796.phpt.

The clobbered registers patch seems to have resolved the tests that were being skipped on aarch64 as these now pass for me, so these have been re-enabled. It's possible that the tests being skipped on ppc64le also pass now, but I don't have an easy way to test this—binfmt-misc builds of this package are super slow.

## Checklist

### DIFF
--- a/main/libtool/template.py
+++ b/main/libtool/template.py
@@ -1,6 +1,6 @@
 pkgname = "libtool"
 pkgver = "2.5.4"
-pkgrel = 3
+pkgrel = 4
 build_style = "gnu_configure"
 configure_gen = []
 hostmakedepends = [
--- a/main/php8.3/patches/fix-clobbered-registers.patch
+++ b/main/php8.3/patches/fix-clobbered-registers.patch
@@ -1,0 +1,24 @@
+this is https://github.com/php/php-src/commit/93d32eae2751089ed0d255ba2cf37d5c8a7353e5
+backported to 8.3.
+--- a/Zend/zend_multiply.h	2026-04-05 00:24:14.250604523 +1000
++++ b/Zend/zend_multiply.h	2026-04-05 00:28:54.242121268 +1000
+@@ -267,7 +267,8 @@
+ 		: "=&r"(res), "=&r"(m_overflow)
+ 		: "r"(nmemb),
+ 		  "r"(size),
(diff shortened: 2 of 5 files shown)


wezm:
Unrelated to this change one test is failing not too, marking draft until I can sort that out.

The failed update check seems to have been a transient error:

```
./cbuild -v update-check main/php8.3
Checking for updates: php8.3=8.3.30
Found update.py, using overrides...
Adding 'https://www.php.net' for version check...
Adding 'https://www.php.net/distributions/' for version check...
Fetching 'https://www.php.net' for version checks...
Fetching 'https://www.php.net/distributions/' for version checks...
Checking found version: 8.3.26
Checking found version: 8.3.27
Checking found version: 8.3.28
Checking found version: 8.3.29
Checking found version: 8.3.30
```

wezm:
The update check was working locally, but failing when run in CI, so I updated it to use a URL less likely to be blocked, the releases RSS feed. However, that failed in CI too. Not sure what else to try.

**Edit 3 weeks later:** seems to be working now

PR #5413, opened 2026-04-09
user/fonts-maple-mono: new package

Cmooon-dev:
## Description

This adds the MapleMono font int the ligature (default) version. There are additional versions  (No-Ligature,Normal-Ligature,Normal-No-Ligature) and different formats (CN, NerdFont, etc.) which could also be added as subpackages later as well. For now, this just serves as baseline.

## Checklist

### DIFF
--- a/user/fonts-maple-mono/template.py
+++ b/user/fonts-maple-mono/template.py
@@ -1,0 +1,13 @@
+pkgname = "fonts-maple-mono"
+pkgver = "7.9"
+pkgrel = 0
+pkgdesc = "Monospace font with round corners and ligatures"
+license = "OFL-1.1"
+url = "https://font.subf.dev"
+source = f"https://github.com/subframe7536/maple-font/releases/download/v{pkgver}/MapleMono-OTF.zip"
+sha256 = "ecf47b851ae4001b00564399511af8dc9615339d3ae9ded54e8547d6d1ad3da1"
+
+
+def install(self):
+    self.install_file("*.otf", "usr/share/fonts/maple-mono", glob=True)
+    self.install_license("LICENSE.txt")
--- a/user/fonts-maple-mono/template.py
+++ b/user/fonts-maple-mono/template.py
@@ -4,10 +4,32 @@ pkgrel = 0
 pkgdesc = "Monospace font with round corners and ligatures"
 license = "OFL-1.1"
 url = "https://font.subf.dev"
-source = f"https://github.com/subframe7536/maple-font/releases/download/v{pkgver}/MapleMono-OTF.zip"
-sha256 = "ecf47b851ae4001b00564399511af8dc9615339d3ae9ded54e8547d6d1ad3da1"
+source = [
+    f"https://github.com/subframe7536/maple-font/releases/download/v{pkgver}/MapleMono-OTF.zip",
+    f"https://github.com/subframe7536/maple-font/releases/download/v{pkgver}/MapleMono-NF-unhinted.zip",
+]
+source_paths = ["maple-otf", "maple-nf"]
+sha256 = [
+    "ecf47b851ae4001b00564399511af8dc9615339d3ae9ded54e8547d6d1ad3da1",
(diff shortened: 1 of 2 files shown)


Cmooon-dev:
Changed the commit message to comply with CONTRIBUTING.md

user/fonts-maple-mono/template.py
```diff
@@ -0,0 +1,16 @@
...
+url = "https://font.subf.dev"
+source = f"https://github.com/subframe7536/maple-font/releases/download/v{pkgver}/MapleMono-OTF.zip"
+sha256 = "ecf47b851ae4001b00564399511af8dc9615339d3ae9ded54e8547d6d1ad3da1"
+
+
+def install(self):
+    self.install_file("*.otf", "usr/share/fonts/maple-mono", glob=True)
+
+
+def post_install(self):
```
wezm: No need for a post install, when you're defining `install` (then can be combined).

user/fonts-maple-mono/template.py
```diff
@@ -0,0 +1,16 @@
+pkgname = "fonts-maple-mono"
+pkgver = "7.9"
+pkgrel = 0
+pkgdesc = "Open source monospace font focused on smoothing your coding flow"
```
wezm: `pkgdesc` should concisely describe what it is to someone that doesn't know about it already. (open source is redundant as most packages in cports are open source).

```suggestion
pkgdesc = "Rounded monospace font"
```

wezm (changes requested):

user/fonts-maple-mono/template.py
Cmooon-dev: Ah I see. Applied in new push :)

user/fonts-maple-mono/template.py
Cmooon-dev: Alright, I changed the description. Now also includes mention of ligatures to be even more specific (since some do not like them)

alecStewart1:
Not up to me, but I've noticed quite a few font packages have a `-otf` or `-ttf` suffix. Should there be two packages for the Maple Mono font; one for OTF and one for TTF?

user/fonts-maple-mono/template.py
```diff
@@ -0,0 +1,35 @@
...
+pkgver = "7.9"
+pkgrel = 0
+pkgdesc = "Monospace font with round corners and ligatures"
+license = "OFL-1.1"
+url = "https://font.subf.dev"
+source = [
+    f"https://github.com/subframe7536/maple-font/releases/download/v{pkgver}/MapleMono-OTF.zip",
+    f"https://github.com/subframe7536/maple-font/releases/download/v{pkgver}/MapleMono-NF-unhinted.zip",
+]
+source_paths = ["maple-otf", "maple-nf"]
```
aniruddhgutta, line 11: args in `source_paths` when >1 must be in separate lines

user/fonts-maple-mono/template.py
```diff
@@ -0,0 +1,35 @@
...
+        "maple-otf/*.otf", "usr/share/fonts/maple-mono", glob=True
+    )
+    self.install_file(
+        "maple-nf/*.ttf", "usr/share/fonts/maple-mono-nf", glob=True
+    )
+    self.install_license("maple-otf/LICENSE.txt")
+    self.install_license("maple-nf/LICENSE.txt", pkgname="fonts-maple-mono-nf")
+
+
+@subpackage("fonts-maple-mono-nf")
```
aniruddhgutta, line 29: Not 100% sure, but I believe nerd fonts must be added to the `fonts-nerd` template instead of being packaged separately...

aniruddhgutta (changes requested):

q66:
thanks @aniruddhgutta, but i'm clearing the remaining PR queue without review currently, so i don't need @Cmooon-dev to address anything - i just do all necessary changes myself

regular review will be done once all of the backlog is gone

user/fonts-maple-mono/template.py
q66: this is incorrect

PR #5416, opened 2026-04-09
user/mbpfan: new package

mbecher:
## Description

Added mbpfan to user. This is a tiny daemon to manage the fan on Macbooks. Tested by me only on x86_64.

## Checklist

- [ x] I have read [CONTRIBUTING.md](https://github.com/chimera-linux/cports/blob/master/CONTRIBUTING.md)
- [ x] I acknowledge that overtly not following the above or the below will result in my pull request getting closed

- [ x] I have read [Packaging.md](https://github.com/chimera-linux/cports/blob/master/Packaging.md#quality_requirements)
- [ x] I have built and tested my changes on my machine

- [x ] I will take responsibility for my template and keep it up to date

### DIFF
--- a/user/mbpfan/files/mbpfan
+++ b/user/mbpfan/files/mbpfan
@@ -1,0 +1,4 @@
+type = process
+command = /usr/bin/mbpfan -f
+smooth-recovery = true
+depends-on: local.target
--- a/user/mbpfan/patches/upstream-musl.patch
+++ b/user/mbpfan/patches/upstream-musl.patch
@@ -1,0 +1,474 @@
+diff -Nur a/Makefile b/Makefile
+--- a/Makefile	2023-04-02 10:04:13.000000000 +0200
++++ b/Makefile	2026-04-09 05:53:14.787504240 +0200
+@@ -19,7 +19,7 @@
+ INCLUDES =
+ LIBS = -lm
+ LIBPATH =
+-CFLAGS += $(COPT) -g $(INCLUDES) -Wall -Wextra -Wno-unused-function -std=c99 -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=500
++CFLAGS += $(COPT) -g $(INCLUDES) -Wall -Wextra -Wno-unused-function -Wno-unused-parameter -std=c99 -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=500 -D_DEFAULT_SOURCE
+ LDFLAGS += $(LIBPATH) -g
+ 
+ OBJS := $(patsubst %.$(C),%.$(OBJ),$(wildcard $(SOURCE_PATH)*.$(C)))
+@@ -51,23 +51,15 @@
+ tests: all
+ 	./bin/mbpfan-tests
+
+ uninstall:
+-	rm /usr/sbin/mbpfan
++	rm /usr/bin/mbpfan
+ 	rm /etc/mbpfan.conf
+-	rm /lib/modules-load.d/mbpfan.depend.conf
+-	rm /lib/systemd/system/mbpfan.service
+ 	rm /usr/share/man/man8/mbpfan.8.gz
+-	rm -rf /usr/share/doc/mbpfan
+ 
+ install: all
+-	install -d $(DESTDIR)/usr/sbin
++	install -d $(DESTDIR)/usr/bin
+ 	install -d $(DESTDIR)/etc
+-	install -d $(DESTDIR)/lib/systemd/system
+-	install -d $(DESTDIR)/usr/share/doc/mbpfan
+-	install -d $(DESTDIR)/lib/modules-load.d
+-	install $(BIN) $(DESTDIR)/usr/sbin
++	install $(BIN) $(DESTDIR)/usr/bin
+ 	install -m644 $(CONF) $(DESTDIR)/etc
+-	install -m644 $(DEPEND_MODULE) $(DESTDIR)/lib/modules-load.d
+-	install -m644 $(DOC) $(DESTDIR)/usr/share/doc/mbpfan
+ 	install -d $(DESTDIR)/usr/share/man/man8
+ 	install -m644 $(MAN) $(DESTDIR)/usr/share/man/man8
+ 	@echo ""
+diff -Nur a/src/daemon.c b/src/daemon.c
+--- a/src/daemon.c	2023-04-02 10:04:13.000000000 +0200
++++ b/src/daemon.c	2026-04-09 04:27:42.551585353 +0200
+@@ -22,10 +22,9 @@
+ #include <stdio.h>
+ #include <stdlib.h>
+ #include <stdbool.h>
+-#include <fcntl.h>
+ #include <errno.h>
+ #include <syslog.h>
+-#include <string.h>
++#include <sys/syslog.h>
+ #include <assert.h>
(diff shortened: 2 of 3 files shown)


user/mbpfan/patches/upstream-musl.patch
```diff
@@ -0,0 +1,525 @@
...
++    -readability-braces-around-statements,
++    -readability-else-after-return,
++    -readability-function-cognitive-complexity,
++    -readability-identifier-length,
++    -readability-inconsistent-declaration-parameter-name,
++    -readability-isolate-declaration,
++    -readability-magic-numbers,
++    -readability-math-missing-parentheses,
++    -readability-suspicious-call-argument,
++'
```
wezm: Not sure we need clang-tidy changes in the patch.

user/mbpfan/template.py
```diff
@@ -0,0 +1,18 @@
+pkgname = "mbpfan"
+pkgver = "2.4.0"
+pkgrel = 0
+build_style = "makefile"
+makedepends = ["gmake", "dinit-chimera", "linux-headers"]
+pkgdesc = "Daemon to control fan speed on MacBook/MacBook Pro"
+subdesc = "The kernel modules applesmc and coretemp are required to be loaded"
```
wezm: I don't think this is an appropriate use of `subdesc`.

user/mbpfan/template.py
```diff
@@ -0,0 +1,18 @@
...
+license = "GPL-3.0-or-later"
+url = "https://github.com/linux-on-mac/mbpfan"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "e1cdcffaba52be215ae40a8568949190866d500d6ae2a1e96b71ab5372f3580b"
+hardening = ["vis", "cfi"]
+# didn't want to, they make no sense here either
```
wezm: Probably enough to say that they require a MacBook to run.

user/mbpfan/files/mbpfan
```diff
@@ -0,0 +1,4 @@
+type = process
+command = /usr/bin/mbpfan -f
+restart = true
+waits-for: local.target
```
wezm: We typically use `depends-on: local.target`

user/mbpfan/files/mbpfan
```diff
@@ -0,0 +1,4 @@
+type = process
+command = /usr/bin/mbpfan -f
+restart = true
```
wezm: This is the default (can be omitted). You may also want to add `smooth-recovery = true`.

user/mbpfan/template.py
```diff
@@ -0,0 +1,18 @@
...
+options = ["!check"]
+
+
+def post_install(self):
+    self.install_service(self.files_path / "mbpfan")
```
wezm, line 24: You can use the shorthand:

```suggestion
    self.install_service("^/mbpfan")
```

user/mbpfan/patches/upstream-musl.patch
mbecher: Yes, this is in fact not tidy at all. Sorry, I didn't care about them, since they won't do no harm. I'll remove them.

user/mbpfan/template.py
mbecher: I may just remove the subdesc. Have been looking for a way to tell, that these modules are required to be loaded. Chances are, they will already be loaded anyway and it is in the manpage, too.

user/mbpfan/template.py
mbecher: And even more precise. I'll change that.

user/mbpfan/files/mbpfan
mbecher: Ok, depends-on works, too.

user/mbpfan/files/mbpfan
mbecher: Sounds good to me.

user/mbpfan/template.py
mbecher: Nice, I didn't know about that. I'm still new to this and there is a lot to learn currently. I'll do my best.

mbecher:
I apologize for being new to this github workflow. I have made the changes and added another commit. Is that what I'm supposed to do? It looks like the commit shows up over here, but I'm not sure ...

wezm:
> I have made the changes and added another commit. Is that what I'm supposed to do?

You want all the changes in one commit. Now that you have two commits you can do this with `rebase.  See https://github.com/chimera-linux/cports/blob/master/USING_GIT.md#interactive-rebase

You will need to force push to push the updated commit, see section following the one linked above for that.

mbecher:
Thank you! I hope it is all well now.

q66:
your patch does not apply...

```
Patching file src/settings.c using Plan A...
Hunk #1 failed at 34.
1 out of 1 hunks failed while patching src/settings.c
Hmm...  The next patch looks like a unified diff to me...
The text leading up to this was:
--------------------------
|diff -Nur a/src/settings.h b/src/settings.h
|--- a/src/settings.h	2023-04-02 10:04:13.000000000 +0200
|+++ b/src/settings.h	2026-04-06 17:41:13.604192384 +0200
--------------------------
Patching file src/settings.h using Plan A...
Hunk #1 failed at 31.
1 out of 1 hunks failed while patching src/settings.h
```

mbecher:
I'm not really sure what happened here. This has been running for months on my laptop. Anyway, I reviewed the patch and updated it. I also set "etcfiles", which seems to be new. Still not sure whether this should be x86_64 only ...

q66:
probably? does it work on apple silicon stuff? if so also aarch64

q66:
the patch still fails to apply too

i found this is due to settings.c and settings.h containing windows line endings, and the patch gets its line endings normalized when i download a patch of this PR

i suggest providing a pure unix line ending patch, and normalizing the two files:

```
def post_extract(self):
    self.do("dos2unix", "src/settings.h")
    self.do("dos2unix", "src/settings.c")
```

(with dos2unix in hostmakedepends)

user/mbpfan/template.py
```diff
@@ -0,0 +1,17 @@
+pkgname = "mbpfan"
+pkgver = "2.4.0"
+pkgrel = 0
+build_style = "makefile"
+makedepends = ["gmake", "dinit-chimera", "linux-headers"]
```
q66: gmake is superfluous (and wrong, because that's a host dependency)

user/mbpfan/template.py
```diff
@@ -0,0 +1,17 @@
...
+url = "https://github.com/linux-on-mac/mbpfan"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "e1cdcffaba52be215ae40a8568949190866d500d6ae2a1e96b71ab5372f3580b"
+hardening = ["vis", "cfi"]
+# tests require a MacBook to run
+options = ["etcfiles", "!check"]
+
+
+def post_install(self):
+    self.install_service("^/mbpfan")
```
q66: drop the ^/ and use self.files_path, you received incorrect advice

the shorthand was a testing feature

mbecher:
I still don't understand why this did work without problems on my computer, but I do see how much cleaner your approach is. I have added your suggested changes and also set arch to x86_64. I don't have a silicon mac, but this has clearly been made for Intel macs.

PR #5429, opened 2026-04-17
user/links: remove unnecessary dependencies

### LINKED ISSUE #4837: graphical links / unnecessary makedepends in the links template
hi there,

i want to use links with graphics support on chimera. i don't think there is a lot of demand to use links with graphics support though. people that want it can just have their own packages downstream and it would be fine to just leave it at that imo. in that case: the `makedepends` of the template contains `libtiff` and `libjpeg-turbo` unnecessarily. (or at least i am pretty sure, what i didn't test was stuff like sixel and i have no clue if links does that when in a terminal and without being compiled with graphics support, would test that then. i mean it builds and i can open wikipedia normally)

i would however be happy to see graphics support upstream (bc i am a links fan). i don't think it would be a good idea if the template would just be updated to be links compiled with graphics support as one then needs to also depend on `libpng`, `librsvg`, `libtiff` and `libx11`. forcing people who most likely use links in their terminal emulators, to install these would be bad imo. 

since one then needs to have separate binaries, i think the two options are: a second links package (i would assume nobody likes that). or a subpackage and then duplicating the build, building one binary with graphics support and one without, and having the return of the subpackage be the one with graphics support overwriting the non-graphical one.
(if links is compiled with graphics support and libraries aren't available it will not run even in non graphical mode, thus one can not compile it with graphics support, have the dependencies as an 'optional' subpackage).

i am happy to write a pr removing the `makedepends` or one for a subpackage. is there interest in any of the two?

take care :)

unrealapex:
remove dependencies for graphics mode which links is not compiled with

## Description

Removes images dependencies used for Links' graphics mode because Links is not compiled with graphics mode enabled.

Closes #4837

## Checklist

### DIFF
--- a/user/links/template.py
+++ b/user/links/template.py
@@ -6,8 +6,6 @@ build_style = "gnu_configure"
 configure_gen = []
 hostmakedepends = ["pkgconf"]
 makedepends = [
-    "libjpeg-turbo-devel",
-    "libtiff-devel",
     "openssl3-devel",
     "zlib-ng-compat-devel",
 ]


user/links/template.py
```diff
@@ -1,13 +1,11 @@
 pkgname = "links"
 pkgver = "2.30"
-pkgrel = 0
+pkgrel = 1
```
wezm: No need to rebuild it. They are unused `makedepends` that the final package doesn't end up depending on:

Depends (6):

- so:libc.so
- so:libcrypto.so.3
- so:liblzma.so.5
- so:libssl.so.3
- so:libz.so.1
- so:libzstd.so.1

wezm (changes requested):

PR #5437, opened 2026-04-18
[TODO, needs go update] main/tailscale: update to 1.96.4

gigirassy:
## Checklist

- [] I have built and tested my changes on my machine


gigirassy:
Looks like Go needs to be updated to the latest version, how would I go about this without having to update the entirety of go (which is a big endeavour)? Do I just need to patch go.mod?

wezm:
> Looks like Go needs to be updated to the latest version, how would I go about this without having to update the entirety of go (which is a big endeavour)?

I think it's best to just wait until Go is updated.

gigirassy:
Good plan. I'll wait until it's updated before I dedraft this.

PR #5456, opened 2026-04-19
user/got: update to 0.124

gigirassy:
## Checklist

### DIFF
--- a/user/got/template.py
+++ b/user/got/template.py
@@ -1,5 +1,5 @@
 pkgname = "got"
-pkgver = "0.120"
+pkgver = "0.124"
 pkgrel = 0
 build_style = "gnu_configure"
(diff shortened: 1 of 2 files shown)


q66:
please fix update-check

gigirassy:
> please fix update-check

That's weird. The previous update check should find 0.124. I tried another, more complicated version of doing that, but it also comes up with no matches. Maybe gameoftrees might be limiting github runner traffic.

I just put ``ignore = True`` for now temporarily.

gigirassy:
> > please fix update-check
> 
> Maybe gameoftrees might be limiting github runner traffic.
> 
This might be the case as the pull is stalling out, even when I put ignore = True to the check.

PR #5483, opened 2026-04-23
[WIP] main/libxml2: update to 2.15.3

gigirassy:
Fixes the following cve. Python bindings now require doxygen, and aren't enabled by default. 

~~The fact they now need doxygen makes them also recursive (which breaks Chimera's packaging), so I disabled them.~~

Update: That was silly. I was so eepy.

https://access.redhat.com/security/cve/cve-2026-6732

## Checklist


q66:
you can't just remove something that other things depend on?

q66:
and according to CI this doesn't even build, have you tested it

gigirassy:
Oops sorry, I'm still working on it. Sorry for my silliness and bad decision to remove the bindings. I was so eepy yesterday, that was not a good idea.

I just forgot to mark this as an incomplete draft.

@q66 Do you know how to build libxml2's python bindings since Doxygen is recursive? It won't let me build the python-bindings if I put doxygen as a dependency. Sorry for my silly question.

q66:
the bindings are deprecated, so getting rid of them is okay, but you can't do it while things still use it

there were two users in the repo, virt-manager and itstool

virt-manager had an alternative implementation added upstream that uses python stdlib xml.etree, so i backported that earlier this morning

itstool is kinda dead with no real changes since 2021 but used by a bunch of stuff

there is a PR that will likely never get merged that switches to lxml but it would need lots of careful testing

gigirassy:
Thank you so much for the time and for the helpful information.

gigirassy:
Looks like this PR is a thing, neat:

https://github.com/itstool/itstool/pull/57

I am not sure whether it will be merged, as the project is a bit inactive.

PR #5485, opened 2026-04-24
user/tea: add shell completions and man page

### LINKED ISSUE #5428: Install Tea Man page
`user/tea` comes with a manual page, however, it is only created with the command `tea man`. Tea's template should create the manual page from `tea` and install it.

Biswa96:
## Description

-  user/tea: add shell completions and man page
-  user/tea: fix version reporting
    Previously, the version was shown as following.

    $ tea --version
    Version: development    golang: 1.26.0

- Fixes #5428 

## Checklist

### DIFF
--- a/user/tea/template.py
+++ b/user/tea/template.py
@@ -1,6 +1,6 @@
 pkgname = "tea"
 pkgver = "0.14.0"
-pkgrel = 0
+pkgrel = 1
 build_style = "go"
 make_build_args = [
     "-ldflags",
@@ -13,7 +13,21 @@ license = "MIT"
 url = "https://gitea.com/gitea/tea"
 source = f"{url}/archive/v{pkgver}.tar.gz"
 sha256 = "f509de217ac0e57491ffdab2750516e8c505780881529ee703b9d0c86cc652a3"
+# generates completions and manpage with host binary
+options = ["!cross"]
+
+
+def post_build(self):
+    for shell in ["bash", "fish", "zsh"]:
+        with open(self.cwd / f"tea.{shell}", "w") as o:
+            self.do("build/tea", "completion", shell, stdout=o)
+
+    with open(self.cwd / "tea.1", "w") as o:
+        self.do("build/tea", "man", stdout=o)
 
 
 def post_install(self):
+    for shell in ["bash", "fish", "zsh"]:
+        self.install_completion(f"tea.{shell}", shell)
+    self.install_man("tea.1")
     self.install_license("LICENSE")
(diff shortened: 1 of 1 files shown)


user/tea/template.py
```diff
@@ -13,7 +13,21 @@
...
 source = f"{url}/archive/v{pkgver}.tar.gz"
 sha256 = "f509de217ac0e57491ffdab2750516e8c505780881529ee703b9d0c86cc652a3"
+# generates completions and manpage with host binary
+options = ["!cross"]
+
+
+def post_build(self):
+    for completion in ["bash", "fish", "zsh"]:
+        with open(f"{self.cwd}/tea.{completion}", "w") as o:
+            self.do("build/tea", "completion", completion, stdout=o)
```
wezm: This is minor, but for consistency with other packages I'd suggest this:

```suggestion
    for shell in ["bash", "fish", "zsh"]:
        with open(self.cwd / f"tea.{shell}", "w") as o:
            self.do("build/tea", "completion", shell, stdout=o)
```

Same for the loop in `post_install`.

user/tea/template.py
```diff
@@ -13,7 +13,21 @@
...
+
+    with open(f"{self.cwd}/tea.8", "w") as o:
```
wezm: man section 8 is system commands. I think section 1 would make more sense.

Biswa96: I agree but the generated man page has 8 in title heading. so, I choose tea.8.

```
$ tea man | head -n 2
.nh
.TH tea 8
```

user/tea/template.py
Biswa96: Done. Thanks.

user/tea/template.py
wezm: I think that's a bug. They call `ToMan`

https://gitea.com/gitea/tea/src/commit/892905d482df5095d35dd8c70e79573fb60b19e4/cmd/man.go#L32

which defaults to section 8 in the library that actually generates the man page:

https://github.com/urfave/cli-docs/blob/72b87d11c482c43243050f3efbc1ed2225efa162/docs.go#L161-L164

They should call `ToManWithSection` instead:

https://github.com/urfave/cli-docs/blob/72b87d11c482c43243050f3efbc1ed2225efa162/docs.go#L150

Biswa96: ok, I have moved the man page to section 1.

wezm: Great. I've opened a PR to fix it upstream as well: https://gitea.com/gitea/tea/pulls/969

PR #5490, opened 2026-04-26
user/niri: update to 26.04

pastthepixels:
## Description

NIRI BLUR !!!!! anyway. ahem. this just bumps the niri version to 26.04 and renames the git repository (it moved to niri-wm/niri)

## Checklist

### DIFF
--- a/user/niri/template.py
+++ b/user/niri/template.py
@@ -1,5 +1,5 @@
 pkgname = "niri"
-pkgver = "25.11"
+pkgver = "26.04"
 pkgrel = 0
 build_style = "cargo"
 make_build_args = [
@@ -31,9 +31,9 @@ depends = ["so:libEGL.so.1!mesa-egl-libs", "xwayland-satellite"]
 checkdepends = ["xkeyboard-config"]
 pkgdesc = "Scrollable-tiling wayland compositor"
 license = "GPL-3.0-or-later"
-url = "https://github.com/YaLTeR/niri"
+url = "https://github.com/niri-wm/niri"
 source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
-sha256 = "9a9a58dbe12e065776cc80424f22c89489f2662e881152ceae46e68bb8677d8c"
+sha256 = "134c602d8e0d53413a52d6cd58f9ce7e79a07d03288ee0a51ba1abd5db1b1ad9"
 # cross: generates completions using host binary
 options = ["!cross"]
 


silverdyne:
Apologies if this isn't the right place for this, but I think this could also be a good opportunity to update the package to handle the various dinit facilities that have been included upstream, such as dinit services files (marked as TODO in the template) and the `niri-session` script (currently unused here) that now supports dinit.

aniruddhgutta:
> Apologies if this isn't the right place for this, but I think this could also be a good opportunity to update the package to handle the various dinit facilities that have been included upstream, such as dinit services files (marked as TODO in the template) and the `niri-session` script (currently unused here) that now supports dinit.

I agree, and I've started [work](https://github.com/aniruddhgutta/cports/tree/niri) on handling them. For now I've only packaged the services, but they don't work yet. I'll have to figure out why but here's the error -

```
[2m2026-04-27T16:50:05.813717Z[0m [32m INFO[0m [2mniri[0m[2m:[0m starting version 26.04 (unknown commit)
[2m2026-04-27T16:50:05.834551Z[0m [34mDEBUG[0m [2mniri_config[0m[2m:[0m loaded config from "/home/vye/.config/niri/config.kdl"

thread 'main' (7759) panicked at src/main.rs:186:6:
called `Result::unwrap()` on an `Err` value: error initializing the TTY backend

Caused by:
    0: Error creating a session. This might mean that you're trying to run niri on a TTY that is already busy, for example if you're running this inside tmux that had been originally started on a different TTY
    1: Failed to open session: Function not implemented (os error 38)
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

sidenote: i know the services have to be installed in dinit.d/user, ill fix the template once i figure out how to fix this issue

pastthepixels:
Oh, I didn't read that in the release notes! It would be really nice to get that new script/dinit thingies working, but I feel it would be best as a separate commit/PR

user/niri/template.py
zacharyburnett, line 45: are these still necessary?

pastthepixels: Honestly I've seen a few things that need to be updated, but given (a) according to the guidelines commits should do minimal work, and this just literally updates niri (b) it builds and everything works (c) oh my goodness I do not have the time to maintain the build script I just want to update niri, I think I'll leave the PR as-is unless there's a breaking change

aniruddhgutta: I understand you don't want to work on handling dinit services yet, but they've been left as "TODO" for a year and it would make sense to work on them right now.

If you'd like to, you can close your pr and I'll raise one with my template to discuss these changes.

aniruddhgutta:
> Apologies if this isn't the right place for this, but I think this could also be a good opportunity to update the package to handle the various dinit facilities that have been included upstream, such as dinit services files (marked as TODO in the template) and the `niri-session` script (currently unused here) that now supports dinit.

Looks like upstream fixed the `niri-session` script and updated the `dinit` services to use `notify-fd` too, I believe niri expects its users to use it solely via `niri-session`, do correct me if I'm wrong.

If I'm right however, then this [pull request](https://github.com/chimera-linux/cports/pull/5494) should be ready to merge.

zacharyburnett:
> Looks like upstream fixed the `niri-session` script and updated the `dinit` services to use `notify-fd` too, I believe niri expects its users to use it solely via `niri-session`, do correct me if I'm wrong.
> 
> If I'm right however, then this [pull request](https://github.com/chimera-linux/cports/pull/5494) should be ready to merge.

I suggest that, if we want to reduce the amount of changes per PR, we should merge this one (#5490) and then you can rebase your PR #5494) on top of that change while  incrementing `pkgrel=1`

unless y'all think the dinit services and usage of niri session are important enough to include in the version update?

aniruddhgutta:
> I suggest that, if we want to reduce the amount of changes per PR, we should merge this one (#5490) and then you can rebase your PR #5494) on top of that change while incrementing `pkgrel=1`
> 
> unless y'all think the dinit services and usage of niri session are important enough to include in the version update?

Yes they are important, especially considering how they were fixed in this release

There would be no point in me rebasing my pr on this since wtv this pr does is already done in my pr

zacharyburnett:
can we at least have version `26.04`? We don't have to enable the `dinit` services if we just run with `niri-session`

q66:
i pushed the update yesterday already, what more do you want

zacharyburnett:
> i pushed the update yesterday already, what more do you want

oh I apologize! I was looking at the closed pull requests and didn't think to check the commits to main, sorry!

q66:
the commit is linked on this page just above your first post...

PR #5502, opened 2026-05-02
user/slopcheck: new package

gigirassy:
## Description

Slopcheck is a simple CLI Rust tool to check whether a Git repo has "AI"/LLM-generated code, shows signs of it, or whether its dependencies do. If it finds any, it will state the amount of times models have generated code in commits, as well as whether it has files known to be related to LLMs. It does this by looking for contributors sharing the same names as popular chatbots, or LLM-related files.

It currently has no license or numbered releases, so it might be a bit weird among Chimera's other packages. I find it very helpful though. Hope that's okay.

## Checklist


q66:
ditto as in your other PR

PR #5520, opened 2026-05-04
user/mustache: new package

devopnem:
## Description

Adds `mustache`, a header-only C++ implementation of the Mustache template format.
This package is needed as a build dependency for `libkiwix`.

## Checklist

### DIFF
--- a/user/mustache/template.py
+++ b/user/mustache/template.py
@@ -1,0 +1,15 @@
+pkgname = "mustache"
+pkgver = "4.1"
+pkgrel = 0
+pkgdesc = "Mustache text templates for modern C++"
+license = "BSL-1.0"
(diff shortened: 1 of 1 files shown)


q66:
not interested in packaging things like this if not used by any package, try again if you end up with anything actually useful

also missing checklist

devopnem:
> not interested in packaging things like this if not used by any package, try again if you end up with anything actually useful
> 
> also missing checklist

this is used by kiwix-desktop. i hace checked the cheklist

q66:
yes and i'm saying that unless you have the packaging for a useful application actually ready, i'm not interested in adding this on its own

you can just bundle it into one PR with something else

PR #5545, opened 2026-05-10
user/wlr-which-key: new package

0xF32:
## Description

`wlr-which-key` is a configurable tool to show keymaps onto the screen for wlroots-based compositors. It is very fast and minimal.

## Checklist

### DIFF
--- a/user/wlr-which-key/template.py
+++ b/user/wlr-which-key/template.py
@@ -1,0 +1,19 @@
+pkgname = "wlr-which-key"
+pkgver = "1.3.0"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = [
+    "cargo-auditable",
+    "pkgconf",
+]
+makedepends = [
+    "glib-devel",
+    "libxkbcommon-devel",
+    "pango-devel",
+    "rust-std",
+]
+pkgdesc = "Keymap manager for wlroots-based compositors"
+license = "GPL-3.0-or-later"
+url = "https://github.com/MaxVerevkin/wlr-which-key"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "50fc06f60e67dc678ffe7bb167d662910d98085e185d2030c09ebd5793bf2794"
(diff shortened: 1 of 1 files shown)


user/wlr-which-key/template.py
```diff
@@ -0,0 +1,18 @@
...
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = [
+    "cargo-auditable",
+    "pkgconf",
+]
+makedepends = [
+    "glib-devel",
+    "libxkbcommon-devel",
+    "pango-devel",
```
q66, line 12: missing rust-std

0xF32: It appeared to build fine without it, however it is probably more idiomatic to include.

0xF32:
I doubt anyone will use wlr-which-key if it is merged to be honest. I thought I would personally use chimera as a desktop OS, but I am too used to nixos. I think chimera could be a decent server OS though.

If cports ever makes it possible to use cbuild templates as loose files or in the git source of a project as a build helper/tool, like nix, without having to maintain a local copy of cports itself, I, and others, would probably prefer cbuild over nix, because it has all of the useful features of nix, as well as being easy to understand, make, and maintain,  not requiring all of the bloat and convoluted complexity that nix does, and allowing musl and dinit to be used.

Apologies for the delay and large amount of unrelated text.

q66:
don't speak for others, also cbuild is for maintenance of the repo and not a tool for building your local junk, if you don't use it i'm closing it

0xF32:
Hello,

I would like to request this to be reopened and merged, mainly because past me was stupid and inconsiderate and I do not like leaving things unfinished.

My intent was not to speak for others, so I will avoid implicitly doing that in the future.

I have tried going back to nix for a while, but have ultimately decided to come back to cports, this time with the intent of actually contributing, instead of hiding in a local repo, because if I have stuff built locally I may as well make it worthy to be in cports.

Sorry for wasting your time,
0xF32

PR #5624, opened 2026-05-22
user/awesome: add 4.3

User-web698:
New package: awesome 4.3

Highly configurable, next gen framework window manager for X.
Homepage: https://awesomewm.org

Notes:
- -fcommon required due to duplicate symbol definitions in headers,
  incompatible with musl's strict linker (same fix as xbps-src)
- AWESOME_IGNORE_LGI=1 set to bypass build-time lgi check, known
  incompatibility between lgi 0.9.2 and Awesome 4.3 that does not
  affect runtime behavior

Tested on x86_64 Chimera Linux.

### DIFF
--- a/user/awesome/patches/fix-cmake.patch
+++ b/user/awesome/patches/fix-cmake.patch
@@ -1,0 +1,12 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 9d3d3b4..8456fba 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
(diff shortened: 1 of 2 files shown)


User-web698:
https://github.com/lgi-devs/lgi/issues/362

q66:
i don't see how this is supposed to work at all without https://github.com/awesomeWM/awesome/pull/4022

q66:
also you're missing the PR checklist and not following it

q66:
no response, closing

PR #5638, opened 2026-05-25
user/quickshell: new package

jl4c:
## Description

Quickshell is a toolkit for building status bars, widgets, lockscreens, and other desktop components using QtQuick. It can be used alongside your wayland compositor or window manager to build a complete desktop environment.

Adds quickshell plus its three new dependencies:

- cli11 — command line parser for C++11
- libdwarf — library to access DWARF debugging information
- cpptrace — stacktrace library for C++11 and newer (used by quickshell's crash handler; depends on libdwarf)

cpptrace and quickshell use the system libunwind, which quickshell's crash handler requires.

## Checklist

### DIFF
--- a/user/libdwarf/update.py
+++ b/user/libdwarf/update.py
@@ -1,0 +1,4 @@
+# template url points to prevanders.net, which pollutes the check with
+# legacy date tags (YYYYMMDD) and an old 0.x series. Restrict to GitHub tags.
+url = "https://github.com/davea42/libdwarf-code/tags"
+pattern = r"/tag/v([\d.]+)"
--- a/user/cli11/template.py
+++ b/user/cli11/template.py
@@ -1,0 +1,18 @@
+pkgname = "cli11"
+pkgver = "2.6.2"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DCLI11_BUILD_EXAMPLES=OFF",
+    "-DCLI11_BUILD_TESTS=OFF",
+]
+hostmakedepends = ["cmake", "ninja", "pkgconf"]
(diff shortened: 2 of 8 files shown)


user/libdwarf/update.py
```diff
@@ -0,0 +1,4 @@
+# template url points to prevanders.net, which pollutes the check with
+# legacy date tags (YYYYMMDD) and an old 0.x series. Restrict to GitHub tags.
+url = "https://github.com/davea42/libdwarf-code/tags"
+pattern = r"/tag/v([\d.]+)"
```
subnut, line 4: instead of this, wouldn't just specifying `ignore = ....` suffice?

jl4c: The issue is that the template url points to `prevanders.net`, which the default check scrapes alongside the GitHub tags, pulling in legacy `YYYYMMDD` date tags and the old 0.x series as false positives. Pointing url at the GitHub tags page fixes the source of the noise rather than filtering it, and still detects real new releases.

It produces a clean result:
```
$ ./cbuild update-check user/libdwarf -v
Checking for updates: libdwarf=2.3.1
Found update.py, using overrides...
Adding 'https://github.com/davea42/libdwarf-code/tags' for version check...
Fetching 'https://github.com/davea42/libdwarf-code/tags' for version checks...
Checking found version: 2.0.0
Checking found version: 2.1.0
Checking found version: 2.2.0
Checking found version: 2.3.0
Checking found version: 2.3.1
```
That said, if `ignore` is the preferred convention for these cases, happy to switch.

colin-heffernan:
Working with no issues on my machines.

PR #5639, opened 2026-05-26
user/constrict: new package

not-fiore:
## Description

constrict is a GNOME application that easily allows to compress a video down to a specified size.

## Checklist

### DIFF
--- a/user/constrict/template.py
+++ b/user/constrict/template.py
@@ -1,0 +1,21 @@
+pkgname = "constrict"
+pkgver = "26.4"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "blueprint-compiler",
+    "desktop-file-utils",
+    "gettext",
+    "meson",
+    "pkgconf",
+]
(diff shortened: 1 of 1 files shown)


not-fiore:
actually wondering if this should be called `gnome-constrict` instead, since it is a gnome application. if anyone seeing this has input on it. it seems to me that maybe it should stay like this since the gnome apps that are called `gnome-*` usually have a generic name, while this one does not.

user/constrict/template.py
```diff
@@ -0,0 +1,21 @@
...
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = [
+    "blueprint-compiler",
+    "desktop-file-utils",
+    "gettext",
+    "meson",
+    "pkgconf",
+]
+makedepends = [
```
q66, line 12: not sure what you're expecting this to do considering it's a python program, did you just put this there arbitrarily

the paint thing had the same problem so keep this mind and read meson output

Issue #5644, opened 2026-05-28
documentation: package conflict resolution


mark-z-r:
There is almost no documentation or mentions of how to deal with commands provided by multiple packages

The one that sparked this issue in particular was mandoc's and groff's 'cmd:soelim' 
mandoc's and groff's soelim might be compatible, but at the end of the day, they are not 1 to 1, 
so it might cause some issues.

A solution I found was the "replaces" option. the documentation for the "replaces" option however
mentions its primary purpose as something else

another solution would be too split the packages to create (mandoc-soelim) and (groff-soelim)
and have a install_if rule as well as cmd:soelim dependecy

q66:
there isn't anything to document, because there isn't any single guideline

sometimes packages are supposed to conflict, sometimes renaming the binaries works, etc

PR #5645, opened 2026-05-28
user/groff: New package

mark-z-r:
draft because of soelim conflict, see issue #5644

## Description

groff typsetting system. added bc nethack requires nroff it to build its docs.
groff does not vendor nroff, it would be more accurate to say it provides gnroff, and also provides a nroff symlink to grnoff

## Checklist


q66:
we don't need groff, nethack can be built with `USE_MANDOC=1` and so can just about anything else

PR #5648, opened 2026-05-30
user/ibus-hangul: new package

RangHo:
## Description

`ibus-hangul` provides Hangul input method for IBus engine.

## Checklist

### DIFF
--- a/user/ibus-hangul/template.py
+++ b/user/ibus-hangul/template.py
@@ -1,0 +1,31 @@
+pkgname = "ibus-hangul"
+pkgver = "1.5.5"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--libexecdir=/usr/libexec",  # TODO switch libexec
+]
+hostmakedepends = [
+    "automake",
+    "gettext-devel",
+    "libtool",
+    "pkgconf",
+    "python",
+    "swig",
+]
+makedepends = [
+    "gtk+3-devel",
+    "ibus-devel",
+    "libhangul-devel",
+]
+depends = [
+    "ibus",
+    "python-gobject",
+]
+pkgdesc = "Korean input engine for IBus"
+license = "GPL-2.0-or-later"
+url = "https://github.com/libhangul/ibus-hangul"
+source = f"{url}/releases/download/{pkgver}/{pkgname}-{pkgver}.tar.xz"
+sha256 = "a5aac88286cd18960229860e3e1a778978a7aeaa484ad9acfa48284b87fdc3bb"
+# tests require interactive environment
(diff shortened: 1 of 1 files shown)


user/ibus-hangul/template.py
```diff
@@ -0,0 +1,33 @@
+pkgname = "ibus-hangul"
+pkgver = "1.5.5"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--libexecdir=/usr/libexec",  # TODO switch libexec
```
q66, line 6: why?

user/ibus-hangul/template.py
```diff
@@ -0,0 +1,33 @@
...
+    "swig",
+]
+makedepends = [
+    "gtk+3-devel",
+    "ibus-devel",
+    "libhangul-devel",
+]
+depends = [
+    "ibus",
+    "librsvg",
```
q66: does it?

user/ibus-hangul/template.py
```diff
@@ -0,0 +1,33 @@
+pkgname = "ibus-hangul"
+pkgver = "1.5.5"
+pkgrel = 0
+build_style = "gnu_configure"
+configure_args = [
+    "--libexecdir=/usr/libexec",  # TODO switch libexec
+    "--with-python=/usr/bin/python3",
```
q66: any reason to specify this explicitly?

user/ibus-hangul/template.py
RangHo: I've checked other ibus-related packages, and it seems that they specify the libexecdir to be /usr/libexec with the same comment on it. I am aware that there is an issue to eliminate libexec directory altogether, but I figured it is best to avoid any potential issue until ibus gets its own directory in /usr/lib.

user/ibus-hangul/template.py
RangHo: This one is an artifact left over from migrating the same Void Linux package to Chimera Linux. I'm not particularly sure why they have this there, but I can test the package without librsvg dependency and remove it.

user/ibus-hangul/template.py
RangHo: I've followed what `ibus-libpinyin` does in this case. The configure script does specify the minimum python version to be 2.5 (https://github.com/libhangul/ibus-hangul/blob/9ee4419ebca803a4f7f2d9c31550abb2c8017de0/configure.ac#L91), though.

RangHo:
I've removed the unneeded dependency on `librsvg` and redundant specification of Python path. The `/usr/libexec` situation remains to be addressed, however.

PR #5657, opened 2026-06-04
user/emmylua-analyzer-rust: new package

colin-heffernan:
## Description

`emmylua-analyzer-rust` is a language server, linter, and documentation generator for Lua.

## Checklist

### DIFF
--- a/user/emmylua-analyzer-rust/template.py
+++ b/user/emmylua-analyzer-rust/template.py
@@ -1,0 +1,22 @@
+pkgname = "emmylua-analyzer-rust"
+pkgver = "0.23.2"
+pkgrel = 0
+build_style = "cargo"
+hostmakedepends = ["cargo-auditable", "openssl3-devel", "pkgconf"]
+makedepends = ["rust-std"]
+pkgdesc = "Lua language server, formatter, linter, and doc generator"
+license = "MIT"
+url = "https://github.com/EmmyLuaLs/emmylua-analyzer-rust"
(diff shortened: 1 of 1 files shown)


colin-heffernan:
The same repo provides a linter and documentation generator as well, both of which I also plan on packaging. I could separate each binary into its own package or combine all three into this one. Which would work better?

EDIT: Typo.

q66:
please follow CONTRIBUTING.md

natthias:
Why not just package all the utils and call this package emmylua-analyzer-rust?

colin-heffernan:
That's certainly possible, but since the executables don't depend on each other and can be installed separately via Cargo, I figured individual packages would work. I can absolutely force-update the branch to bundle them, though.

q66:
if they build together they should install together

natthias:
there's also luafmt

colin-heffernan:
Submitting a PR to add `emmylua_formatter` to their docs, particularly their `README.md` and `CONTRIBUTING.md`, because I didn't see that `luafmt` existed until I went digging. Should I also include `schema_to_emmylua`? It's another util meant for users (as opposed to `edit_version`, `schema_json_gen`, and `std_i18n`, which seem to be for internal development), but I rarely see it packaged.

natthias:
that can go both ways i feel, i only noticed luafmt because it was a release asset, which schema_json_gen is not. But i see no reason not to package it if its a user facing util

PR #5662, opened 2026-06-06
user/neochat: new package

alois31:
Before it can be added, two dependencies need to be fixed (one in the packaging and one upstream), which is done here.

## Checklist


q66:
what's with the random rebuilds? why is that necessary?

q66:
ah, their soname is the full version...

anyway you have a too long commit message title there

alois31:
CI complained about cmark being out of date, but now it complains about prismlauncher and I fear this would turn into whack a mole. What would you is the best way forward here?

q66:
also, this is not the right fix in cmark

the right fix is to drop the pointless `-libs` subpackage, `cmark` has 12 KiB installed size and no extra dependencies

add a `renames = ["cmark-libs"]` to the main block and drop the subpackage and that's all

```
--- a/main/cmark/template.py
+++ b/main/cmark/template.py
@@ -1,5 +1,5 @@
 pkgname = "cmark"
-pkgver = "0.31.1"
+pkgver = "0.31.2"
 pkgrel = 0
 build_style = "cmake"
@@ -9,11 +9,12 @@ hostmakedepends = [
     "pkgconf",
 ]
 checkdepends = ["python"]
+renames = ["cmark-libs"]
 pkgdesc = "C implementation of the CommonMark markdown specification"
 license = "BSD-2-Clause"
 url = "https://github.com/commonmark/cmark"
-sha256 = "3da93db5469c30588cfeb283d9d62edfc6ded9eb0edc10a4f5bbfb7d722ea802"
+sha256 = "f9bc5ca38bcb0b727f0056100fac4d743e768872e3bacec7746de28f5700d697"
 # defaults to Release which sets this, and the tests crash in an assert without it..
 tool_flags = {"CFLAGS": ["-DNDEBUG"], "CXXFLAGS": ["-DNDEBUG"]}
 
@@ -25,8 +26,3 @@ def post_install(self):
 @subpackage("cmark-devel")
 def _(self):
     return self.default_devel()
-
-
-@subpackage("cmark-libs")
-def _(self):
-    return self.default_libs()
```

alois31:
All moles seem to have been lowered now. I'm not sure I like it yet, as testing has been quite superficial (neochat of course was more thorough, and I clicked around a bit in mkvtoolnix-gui and evolution and at least it doesn't seem to crash, only build-tested the other rebuilds). Please let me know if there is a better way.

Regarding cmark, is my understanding correct that the fix was wrong not on principle, but instead because the package is so tiny the extra complexity isn't worth it? I do not want to make such a mistake again.

q66:
generally we only split packages if leaving it unsplit would mean significant extra dependencies and/or size and most users don't require the split package

(anyway i am testing this now with some minor local changes and will push it after)

PR #5673, opened 2026-06-13
user/opam: new package

RangHo:
## Description

opam is the package manager for OCaml programming language.

Some notes regarding dependencies:
- `bash` is required to detect the isolation method. You can force it to work without, but then `bubblewrap` is useless, and you'd have to encounter big error message, so...
- `curl`, `libarchive-progs`, and `unzip` are used to download and extract OCaml sources, and are listed as hard dependencies.
- Since it compiles OCaml for every switch, `clang` and `gmake` are pretty much requirements, but _technically_ one can use opam without, so they are not listed as hard dependencies.

Supersedes #4557.

## Checklist

### DIFF
--- a/user/opam/template.py
+++ b/user/opam/template.py
@@ -1,5 +1,5 @@
 pkgname = "opam"
-pkgver = "2.5.1"
+pkgver = "2.5.2"
 pkgrel = 0
 build_style = "gnu_configure"
 configure_args = [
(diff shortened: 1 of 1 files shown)


ifreund (approved):
Rebased and bumped the version, LGTM!

q66:
i'd like to investigate this a little bit more when it comes to stuff like dependencies, but it will definitely be looked at in the coming days

ifreund:
> i'd like to investigate this a little bit more when it comes to stuff like dependencies, but it will definitely be looked at in the coming days

Alright, I'll leave this to you then. I did use it to successfully fetch dependencies and build https://codeberg.org/river/tinyrwm/pulls/15 FWIW

(Though I did need to explicitly set CC=clang to get opam to build the uring package's vendored liburing copy, bleh)

PR #5696, opened 2026-06-19
main/zsh: update to 5.9.2

flukeeey:
## Description

Dropped patches:

- clang-15.patch: no longer relevant, fixed upstream
- pcre2-{001,002,mdd}.patch: PCRE2 support merged upstream

Tests no longer failing for me (x86_64):

- Test/D07multibyte.ztst
- Test/Y03arguments.ztst

New tests failing:

- Test/A03quoting.ztst
- Test/B03print.ztst 
- Test/D04parameter.ztst

I haven't attempted to patch these yet.

## Checklist


q66:
so did you figure out why those tests failed

flukeeey:
Not yet. Should get chance to delve into them tomorrow.

flukeeey:
There are currently 3 tests failing:

- D04parameter
- B03print
- A03quoting

```
## D04parameter
## expects:
# -128
# -129
# +57216
# +57217
a=( 0x80 0x81 )
/usr/bin/printf "%s\n" ${(#)a} |
while read x; do echo $(( #x )); done

## B03print
## expects:
# -f0
# +dff0
printf '%x\n' $(printf '"\xf0')

## A03quoting
## expects:
#  16#4D
#  16#42
#  16#53
# -16#DC
# +16#DFDC
chars=$(print -r $'BS\\MBS\M-\\')
for (( i = 1; i <= $#chars; i++ )); do
  char=$chars[$i]
  print $(( [#16] #char ))
done
```

Output when run on my host install:

```
128
129
f0
16#42
16#53
16#5C
16#4D
16#42
16#53
16#DC
```

I extracted the tests to a script (above) on my host machine, and all three pass okay. The tests also pass fine if I `./cbuild chroot` in, and run the above script using the built zsh. I assume there's a problem with the Makefile or how the tests are being called, but I don't see anything immediately wrong.

I thought it might be related to https://www.zsh.org/mla/workers/2021/msg00805.html, but I tried unsetting all mentioned variables to no avail, setting them to 'C' also doesn't work (via `make_check_wrapper`).

I'm not super interested in spending too much time on this, tbh. The reproducer passes fine on my actual host machine with the built zsh, so I'd be happy just ignoring these tests for now?

q66:
it's fine to ignore them if things work okay, i just wanted to see the run to make sure it's not anything bad, this is sufficient, thanks

subnut:
the aforementioned tests fail when LC_CTYPE (or LANG) is set to **C**  
zsh-5.9.2/Test/ztst.zsh (line 29-30) sets `LANG=C` and unsets all LC_\* variables

i presume its passing outside the testsuite because you have LANG or LC_ALL or LC_CTYPE set to some variation of \*.utf8 or \*.UTF-8

i've been looking into it for some time, and the issue seems to be centered around how musl and glibc handle `wctomb`

**EDIT:** If line 30 of ztst.zsh is changed to `export LANG=C.UTF-8` then A03quoting and B03print pass.  
I do not know what to do with this information...

main/zsh/template.py
subnut: Should be removed.  
Test/ztst.zsh unsets everything anyways (see comment above)

main/zsh/template.py
```diff
@@ -32,7 +32,7 @@
 license = "MIT AND GPL-3.0-or-later"
 url = "https://www.zsh.org"
 source = f"{url}/pub/zsh-{pkgver}.tar.xz"
-sha256 = "9b8d1ecedd5b5e81fbf1918e876752a7dd948e05c1a0dba10ab863842d45acd5"
+sha256 = "36fa734374b44783582cec09bcd67822e2f992c779ec1624ab5596df078d2f81"
 # FIXME int: test failures
 hardening = ["!int"]
```
subnut: Is this still necessary? Tests seem to pass...

main/zsh/template.py
flukeeey: This indeed isn't needed (verified with an `export -p` in one of the tests).

main/zsh/template.py
flukeeey: Doesn't appear to be needed, added vis+cfi hardening too and so far all good.

flukeeey:
> the aforementioned tests fail when LC_CTYPE (or LANG) is set to **C** zsh-5.9.2/Test/ztst.zsh (line 29-30) sets `LANG=C` and unsets all LC_* variables
> 
> i presume its passing outside the testsuite because you have LANG or LC_ALL or LC_CTYPE set to some variation of *.utf8 or *.UTF-8
> 
> i've been looking into it for some time, and the issue seems to be centered around how musl and glibc handle `wctomb`
> 
> **EDIT:** If line 30 of ztst.zsh is changed to `export LANG=C.UTF-8` then A03quoting and B03print pass. I do not know what to do with this information...

Good info indeed, but yeah, for the sake of ignoring a couple of tests and manually verifying on my own machine, I don't feel too pressed to spend more time on this, tbh.

main/zsh/template.py
```diff
@@ -1,6 +1,6 @@
 pkgname = "zsh"
-pkgver = "5.9"
-pkgrel = 3
+pkgver = "5.9.2"
+pkgrel = 1
```
subnut, line 3: `pkgrel` should be zero

PR #5700, opened 2026-06-19
user/perl-ipc-run: update to 20260402.0

flukeeey:
## Checklist


q66:
fails to fetch

flukeeey:
Works fine if I set https://cpan.metacpan.org/authors/id/T/TO/TODDR/IPC-Run-20260402.0.tar.gz
as the source when building locally.

Shall I switch from using `$(CPAN_SITE)`?

PR #5705, opened 2026-06-21
main/gstreamer: update to 1.28.4

alois31:
The update to 1.28.2 has introduced a regression where in some combinations of
media player and codec, video would be completely corrupted. (Reproduced bad
combinations are GNOME Video Player (showtime) with DVCAM or AV1 codec.) This
issue seems to be fixed in 1.28.4.

## Checklist


q66:
this misses some components (grep around for version number)

alois31:
Thank you for noticing, actually I have prepared the change using this exact same method but forgot to commit gst-libav.

PR #5709, opened 2026-06-21
main/wlsunset: add user service

ghost:
## Description

This PR adds a dinit user service for wlsunset. Since dinit doesn't have a predefined directory for service configurations, like openrc's `conf.d`, I've chosen `$HOME/.config/wlsunset.env` as the `env-file` path. There is one problem with this approach. If the `env-file` doesn't exist, the dinit service will fail to start, so the user should have an empty `env-file` even if they're not interested in passing any options. For wlsunset, I think most users wants to set the latitude and longitude, so I guess it should be fine.

```
~$ dinitctl start wlsunset
dinitctl: error loading service (or dependency of service).
dinitctl: try 'dinit-check <service-name>' or check log for more information.
~$ doas tail -n1 /var/log/turnstiled.log
dinit: Could not load service wlsunset: could not load environment file: No such
file or directory
~$ dinit-check -n wlsunset
Checking service: wlsunset...
Service 'wlsunset': could not load environment file: No such file or directory
Checking service: graphical.target...
Checking service: graphical.monitor...
Checking service: login.target...
Performing secondary checks...
Secondary checks complete.
One or more errors/warnings issued.
```

There is a similar user service in [alpine](https://git.alpinelinux.org/aports/tree/community/wlsunset/wlsunset.initd), which assumes the latitude and longitude as mandatory arguments. I tried that to solve the empty file problem, but it fails the manual time use cases.

```
~$ wlsunset -l 0 -L 0 -s 18:00
latitude and longitude are not valid in manual time mode
```

## Checklist

### DIFF
--- a/main/wlsunset/files/wlsunset.user
+++ b/main/wlsunset/files/wlsunset.user
@@ -1,0 +1,5 @@
+type = process
+command = /usr/bin/wlsunset
+smooth-recovery = true
+log-type = buffer
+depends-on: graphical.target
--- a/main/wlsunset/template.py
+++ b/main/wlsunset/template.py
@@ -1,9 +1,14 @@
 pkgname = "wlsunset"
 pkgver = "0.4.0"
-pkgrel = 1
+pkgrel = 2
 build_style = "meson"
 hostmakedepends = ["meson", "pkgconf", "scdoc", "wayland-progs"]
-makedepends = ["wayland-devel", "wayland-protocols"]
+makedepends = [
+    "dinit-chimera",
+    "turnstile",
+    "wayland-devel",
+    "wayland-protocols",
+]
 pkgdesc = "Day/night gamma adjustments for Wayland"
 license = "MIT"
 url = "https://git.sr.ht/~kennylevinsen/wlsunset"
@@ -15,3 +20,4 @@ hardening = ["!int", "vis", "cfi"]
 
 def post_install(self):
     self.install_license("LICENSE")
+    self.install_service(self.files_path / "wlsunset.user")


q66:
abusing env-files for this is really bad and we don't do it anywhere

there are two ways to do it properly:

1) either use a wrapper script (see eg incus)

or

since any configuration file thing for this is purely homegrown, maybe just use a command with no arguments

then one can create a `~/.config/dinit.d/wlsunset` and put something like this in there

```
@include /usr/lib/dinit.d/user/wlsunset
command += -foo -bar
```

we could even come up with some convention for naming services that are not meant to be used as is

ghost:
I saw your [commit](https://github.com/chimera-linux/cports/commit/29bc77797463d434276ff95676adf293bcd63310) on tailscale. It looks like you faced the same issue that I described in the PR description. So for now, I will remove the `env-file` and command arguments.

On alternatives, I like the second approach more than the wrapper script. It's simple and straightforward. But then the problem is, how will a user know that they can extend it? I'm not really sure whether the naming convention will help. There are no required arguments for wlsunset, so a user can execute it without extending it, and there are no ways to prevent a user from running a non-extended service. Also, every service is extensible with the `@include` meta command, so we can't precisely differentiate between extensible and non-extensible services. For example, I could even extend the foot user service to add more arguments even if it doesn't state it explicitly. Maybe can we add some examples about this dinit feature in the documentation? Let me know what you think.

ghost:
I thought on this a bit. I don't think anyone will run the wlsunset service without arguments, so we can define it as a fragment. From [dinit-service(5)](https://davmac.org/projects/dinit/man-pages-html/dinit-service.5.html),

> A "fragment" may be the service description file itself, or it may be a file included via `@include` or similar.

The fragment is a file with an `.fragment` extension stored in the service description directories. It is a partially defined service with an undefined execution state. It could execute successfully or fail depending on how complete the service description is. In our case, it will execute successfully as wlsunset has no required arguments.

```
@include /usr/lib/dinit.d/user/wlsunset.fragment

command += -l <lat> -L <long>
```

Now to inform the users, we can add a "Service customization" section (or something similar) in the [4.2 Service management](https://chimera-linux.org/docs/configuration/services) document to explain about the `@include` meta command and the `.fragment` files.

Let me know your suggestions. We can also pick a different naming convention if the above one is unsatisfactory. I picked it because anyone can infer its meaning, provided that they read the man page.

ghost:
I reread what I wrote earlier and realized, "Why am I putting a partially defined service in a service description directory when it will most likely fail to start?." I don't think we need a special naming convention like `<service>.fragment`, instead we need to store them out of the service description directories. Maybe directories like `/usr/share/dinit.d` and `/usr/share/dinit.d/user` might be a good start, but they could be reserved by the upstream for future use. In that case, we could have a different directory in `/usr/share` or a subdirectory in `/usr/share/dinit.d`. This is how it will look in practice.

```
@include /usr/share/dinit.d/user/wlsunset

command += -l <lat> -L <long>
```

Let me know your thoughts on this one.

q66:
i kinda came to the conclusion that this doesn't really need a pre-shipped user service, considering it doesn't have a reasonable configuration file support or anything

the amount of boilerplate that you add for your custom-configuration service is nearly the same it takes to write one from scratch

just drop a file in .config/dinit.d with whatever contents you want

PR #5719, opened 2026-06-23
main/curl: stop disabling optimization

alois31:
In commit b68f9dabac85fae5715b683a15ea1c5f1e361e25, the default compiler optimization was disabled. As no explicit reason was given, this was likely either a temporary hack or a mistake (such as a belief that it just disables extra optimization flags added by curl). Remove the offending flag to restore optimization, which does not seem to break anything.

## Checklist

### DIFF
--- a/main/curl/template.py
+++ b/main/curl/template.py
@@ -1,9 +1,8 @@
 pkgname = "curl"
 pkgver = "8.20.0"
-pkgrel = 1
+pkgrel = 2
 build_style = "gnu_configure"
 configure_args = [
-    "--disable-optimize",
     "--enable-ares",
     "--enable-httpsrr",
     "--enable-ipv6",


q66:
hm? no, this is not wrong

we generally disable project-provided optimization flags, they get in the way and aren't necessary

alois31:
Maybe I worded it wrong, but the way it works is that only `--enable-optimization` passes curl-provided optimization flags, passing nothing keeps the default ones, and `--disable-optimization` passes explicit `-O0` at the end.

q66:
oh, yes, you seem to be right

fixed properly in f59ffd80216625d615e54706f773aac88d4565eb

alois31:
No worries, and thanks for the quick and more thorough fix! So much code that in the end makes people's life harder :(

q66:
i honestly wonder what was the reasoning behind this rather than just putting whatever CFLAGS you provide at the end like any reasonable build system does, but at least they had the decency to make this separate macro invocations so if you're regenerating autoconf you can patch it out easily

PR #5723, opened 2026-06-24
user/raylib: new package

clifflocked:
## Description

raylib is a simple C library for making games.

## Checklist


natthias:
Please reread Contributing.md, the commit message should be `user/raylib: new package`, and subsequent commits should be squashed.

Also is there any particular reason we need the statically linked library as well?

user/raylib/template.py
```diff
@@ -0,0 +1,55 @@
...
+    cmake.build(self, build_dir="build-static")
+
+
+def post_install(self):
+    self.install_file("build-static/raylib/libraylib.a", "usr/lib")
+
+
+@subpackage("raylib-devel")
+def _(self):
+    return [
```
natthias: use `return self.default_devel()` instead

clifflocked: done

clifflocked:
> Also is there any particular reason we need the statically linked library as well?

Because it is the default option when compiling raylib without any flags, people might expect a static library to be present on the system.

natthias:
Looking at the build system, we're also going to want `DUSE_EXTERNAL_GLFW` instead of `DGLFW_BUILD_WAYLAND`.

> Because it is the default option when compiling raylib without any flags, people might expect a static library to be present on the system.

I don't really see this as an issue. By that logic, we should never deviate from upstreams default options. Also, I could not find any other distributions, which package the static library.

clifflocked:
> I don't really see this as an issue. By that logic, we should never deviate from upstreams default options. Also, I could not find any other distributions, which package the static library.

Fair enough, seems it isn't common practice.

PR #5746, opened 2026-06-29
user/ntpd-rs: new package

flukeeey:
## Description

NTP server and client with NTS support, written in Rust.

Sponsored by NLnet/ISRG/Sovereign Tech Fund, and looks like Ubuntu will
be switching to it.

Development seems more active than chrony and it already has NTPv5
support (experimental?).

I haven't packaged the ntp-metrics-exporter binary and service, as I
don't need or want Prometheus/OpenMetrics metrics. Happy to include it
if there's demand for it, though.

## Checklist

### DIFF
--- a/user/ntpd-rs/files/ntp-daemon
+++ b/user/ntpd-rs/files/ntp-daemon
@@ -1,0 +1,8 @@
+type = process
+command = /usr/bin/ntp-daemon
+run-as = _ntp-daemon
+logfile = /var/log/ntp-daemon.log
+capabilities = ^cap_sys_time
+smooth-recovery = true
+depends-on: network.target
+depends-on: local.target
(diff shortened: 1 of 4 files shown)


q66:
this has a bunch of issues (eg tmpfiles being a poor mechanism for handling ephemeral service state, no readiness notification for time sync target and hogging a generic service name among others) but a project by terminal RIIR pushers does not spark joy in me therefore i can't be bothered to point them out at this time

sorry

use chrony it works fine

q66:
PRs such as https://github.com/pendulum-project/ntpd-rs/pull/2278 also indicate to me that this is going to be an absolute long term maintenance nightmare ("vendor openssl when musl is used" etc nonsense which we'd have to roll back because rust people are seemingly unable to comprehend that musl is not "that thing you use for static linking")

and i already have plenty of pull requests in the backlog to wade through for all sorts of new packages so i really can't afford to pile on more stuff that exists seemingly only to annoy me

flukeeey:
No worries at all, this is your rodeo. :)

Honestly, I mostly started looking into ntpd-rs because it has the most
mature NTPv5 implementation from what I could see, which brings some
pretty nice protocol cleanups.

But as you said, chrony works just fine and is pretty much the reference 
anyways (afaik), and I can keep on using ntpd-rs via my local cports, if
desired. All good!

Issue #5694, opened 2026-06-19
[SOLVED] blender-5.1.1-r1 fails to launch on wayland without xwayland


Monochrotec:
Dear whoever is reading this

blender when launched from the terminal gives the following error
'''
00:00.062  ghost.system     | Failed to initialize display for back-end(s): ['X11']
                            |   'X11': unable to open a display!
                            |
00:00.062  ghost.system     | ERROR Unable to initialize GHOST, exiting!
Error: Not freed memory blocks: 4638, total unfreed memory 1.207977 MB
'''
the blender wiki states that blender supports both x11 and wayland and defaults to wayland when possible
i started looking for a solution on the internet and then i found something on the debian mailing list
https://lists.debian.org/debian-multimedia/2023/03/msg00109.html
it might have something to do with the flag WITH_GHOST_WAYLAND.
i might try compiling blender from source once i figure out how and see if the problem persists and if its a packaging problem 
or a problem with blender itself.

Sincerely,
me.

Monochrotec:
i didn't compile from source but tested blender 5.1.2 from flathub and it launches on wayland without any issues so I'm guessing that its a packaging problem but it could also be a bug in blender 5.1.1 that got patched in 5.1.2

PR #5768, opened 2026-07-05
main/blender: update to 5.1.2 and fixed wayland support

Monochrotec:
## Description

This pull request updates blender from 5.1.1 to 5.1.2. it also fixes the wayland support problem that i pointed out here: #5694 . the problem was due to missing dependencies i believe. also I'm not sure if 'libxkbcommon' should be in the depends section or 'libxkbcommon-devel' in 'makedepends' is enough. and I tested blender on my machine and can confirm that it works.

## Checklist

- [x] I have read [Packaging.md](https://github.com/chimera-linux/cports/blob/master/Packaging.md#quality_requirements) (not fully. i felt like it was important to mention that)

### DIFF
--- a/main/blender/template.py
+++ b/main/blender/template.py
@@ -1,6 +1,6 @@
 pkgname = "blender"
-pkgver = "5.1.1"
-pkgrel = 1
+pkgver = "5.1.2"
+pkgrel = 0
 build_style = "cmake"
 configure_args = [
     "-DCMAKE_BUILD_TYPE=Release",
@@ -50,6 +50,7 @@ makedepends = [
     "libsndfile-devel",
     "libtiff-devel",
     "libwebp-devel",
+    "libxkbcommon-devel",
     "llvm-devel",
     "onetbb-devel",
     "openal-soft-devel",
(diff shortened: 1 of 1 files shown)


Monochrotec:
this is my first time trying to contribute to a project so I'm sorry if there is any problems with this pull request

q66:
the makedepends changes are enough, don't need to change `depends` nor `configure_args`

also please use correct commit message as per CONTRIBUTING.md (not the PR title, but the git commit message)

Monochrotec:
thanks. I modified it and removed those parts. do i need to recompile and test it again or is it enough?

q66:
no, but squash your commits

q66:
in the final commit message you might also want a `Fixes https://github.com/chimera-linux/cports/issues/5694` in the body of the commit message

ie there should be one commit with the message

```
main/blender: update to 5.1.2, enable wayland support

Fixes https://github.com/chimera-linux/cports/issues/5694
```

q66:
thanks, since you built and tested this i don't think we need to run the CI for hours so let's merge it now

PR #5808, opened 2026-07-11
user/nut-ups: new package

Gwenio:
## Description

Adds package for Network UPS Tools (NUT).

Technically the package is ready, but I have marked it as a draft to ask for an opinion on a couple of things:

- The package name
- The system user's name

I called it 'nut-ups' because if I was searching for a package containing these tools I would start with `apk list | grep nut` or `apk list | grep ups`.

The system user is '_nut', I would have called it '_ups' if Arch did not call it 'nut' (and therefore that is what anyone looking at the Arch wiki would expect).

## Checklist


q66:
https://gts.chimera-linux.org/@chimera/statuses/01KWARHE1BXBMXB8WPXK0BBM1B try at another time

q66:
also please don't claim "Technically the package is ready", you don't know that and this is up to the reviewer to decide

this template has several major issues like adjusting mode on files installed by apk outside of apk (which breaks audit) among others, so it will not be reviewed at this time as per the above

PR #5815, opened 2026-07-14
[RFC] main/sddm: switch to wayland by default

q66:
The greeter is kinda wonky when it comes to session selection. However, it is wonky regardless even in x11 mode, just slightly less. Use kwin for wayland compositor because we already bring in KDE dependencies anyway through plasma-workspace, and with kwin we get a nice virtual keyboard and stuff, without significant new deps

testing appreciated

cc @JamiKettunen

### DIFF
--- a/main/sddm/files/sddm.config
+++ b/main/sddm/files/sddm.config
@@ -11,8 +11,8 @@ User=
 
 [General]
 # Which display server should be used.
-# Valid values are: x11, x11-user, wayland. Wayland support is experimental
-DisplayServer=x11
+# Valid values are: x11, x11-user, wayland (considered experimental)
+DisplayServer=wayland
 
 # Comma-separated list of environment variables to be set
 GreeterEnvironment=
@@ -93,7 +93,7 @@ ReuseSession=true
 
 [Wayland]
 # Path of the Wayland compositor to execute when starting the greeter
-CompositorCommand=weston --shell=kiosk
+CompositorCommand=kwin_wayland --drm --no-lockscreen --no-global-shortcuts --locale1 --inputmethod plasma-keyboard
 
 # Enable Qt's automatic high-DPI scaling
 EnableHiDPI=true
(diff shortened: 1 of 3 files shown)


TheOneDoc:
[GMKtec NucBox M5 Plus Mini PC AMD RyzenTM 7 5825U](https://www.gmktec.com/products/amd-ryzen-7-5825u-mini-pc-nucbox-m5-plus)

I did no session selection testing as I only have Wayland KDE installed.

alois31:
It works for my needs, but I couldn't get the virtual keyboard to actually show up either on the main machine or in a mostly clean testing VM. Another thing that might be surprising to some users is that this change may make the X11 option go away, the `xserver-xorg-core` in the `install_if` of `plasma-desktop-x11-meta` was only pulled in by the `sddm` -> `xserver-xorg-input-libinput` -> `xserver-abi-input~24` dependency chain.

TheOneDoc:
> I couldn't get the virtual keyboard to actually show up either on the main machine or in a mostly clean testing VM. 

Can confirm. Doesn't show up on any of my test systems.

> Another thing that might be surprising to some users is that this change may make the X11 option go away, the `xserver-xorg-core` in the `install_if` of `plasma-desktop-x11-meta` was only pulled in by the `sddm` -> `xserver-xorg-input-libinput` -> `xserver-abi-input~24` dependency chain.

I see this as intentional X11 KDE should only install if XOrg is explicitly installed as well. The default is clean Wayland.

q66:
shows up fine here on my touchscreen laptop: https://i.imgur.com/egbEmNR.jpeg

only appears upon touch events

q66:
another thing is the keyboard layout indicator no longer working but that's expected (there is no wayland keyboard layout backend in sddm, it's only stubs)

otoh with this change it follows locale1 at very least, so it can be set/changed with localectl

q66:
~~another thing i want to do is make the initial VT more runtime-based instead of hardcoding it at 7 (perhaps even make it follow `ACTIVE_CONSOLES` via a helper script, so that it can query the first VT that has no gettys on it), but i will do that before merging this~~ done

PR #5823, opened 2026-07-18
lvm2-dm: fix libexecdir override, actually landing in /usr/lib

oldake:
lvm2's own `configure` script defines a separate `--with-libexecdir` AC_ARG_WITH that unconditionally overrides the standard `--libexecdir` autoconf flag if not passed explicitly (falls back to `${prefix}/libexec` otherwise). This meant `lvresize_fs_helper` (and friends) landed in `/usr/libexec` instead of `/usr/lib`, which fails cports lint on Chimera (no `/usr/libexec` split). Passing `--with-libexecdir=/usr/lib` explicitly fixes it. Verified with a clean isolated `cbuild pkg -f -t` rebuild.

### DIFF
--- a/main/lvm2/template.py
+++ b/main/lvm2/template.py
@@ -1,9 +1,13 @@
 pkgname = "lvm2"
 pkgver = "2.03.33"
-pkgrel = 1
+pkgrel = 2
 build_style = "gnu_configure"
 configure_args = [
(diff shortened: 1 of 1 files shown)


q66:
i explicitly didn't want to touch lvm2 until i've updated it so this is intended

PR #5824, opened 2026-07-18
pipewire: explicitly disable libsystemd feature

oldake:
`--auto-features=enabled` promotes every meson `type: 'feature', value: 'auto'` option to required, including `libsystemd` (separate from the already-correctly-configured `logind-provider=libelogind`). Chimera doesn't ship `libsystemd.pc`, so this makes the meson configure step hard-fail. Passing `-Dlibsystemd=disabled` explicitly fixes it since elogind already provides the logind integration. Verified with a clean isolated `cbuild pkg -f -t` rebuild.

### DIFF
--- a/main/pipewire/template.py
+++ b/main/pipewire/template.py
@@ -1,12 +1,17 @@
 pkgname = "pipewire"
 pkgver = "1.6.7"
-pkgrel = 0
+pkgrel = 1
 build_style = "meson"
 configure_args = [
     "--auto-features=enabled",
     "-Ddocs=enabled",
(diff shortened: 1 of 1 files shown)


q66:
no? elogind-devel has libsystemd.pc and pipewire builds fine

```
$ apk info --who-owns /usr/lib/pkgconfig/libsystemd.pc 
/usr/lib/pkgconfig/libsystemd.pc is owned by elogind-devel-255.17-r1
```

q66:
also these changes smell of LLM and if that's the case i ask you to cease

PR #5825, opened 2026-07-18
perl-authen-sasl: fetch from permanent metacpan path

oldake:
`$(CPAN_SITE)` (`www.cpan.org/modules/by-module`) only mirrors the latest release per module and 403s on older pinned versions once upstream ships a newer one (this module has since moved on to 2.2000). Switching to the permanent author-keyed metacpan path (`cpan.metacpan.org/authors/id/.../Authen-SASL-<ver>.tar.gz`) fixes the fetch permanently, matching the pattern already used by several other `perl-*` templates in this repo. Verified with a clean isolated build.

### DIFF
--- a/main/perl-authen-sasl/template.py
+++ b/main/perl-authen-sasl/template.py
@@ -1,6 +1,6 @@
 pkgname = "perl-authen-sasl"
 pkgver = "2.1900"
-pkgrel = 0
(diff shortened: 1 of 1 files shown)


q66:
i have no interest in doing this for every perl module

PR #5826, opened 2026-07-18
perl-crypt-urandom: fetch from permanent metacpan path

oldake:
Same issue as perl-authen-sasl: `$(CPAN_SITE)` only mirrors the latest release per module and 403s on this pinned older version now that upstream has shipped 0.55. Switching to the permanent author-keyed metacpan path fixes the fetch permanently. Verified with a clean isolated build.


q66:
ditto

PR #5828, opened 2026-07-19
 main/easyeffects: update to 8.2.7 

z-erica:


main/easyeffects/template.py
```diff
@@ -1,51 +1,60 @@
...
     "speexdsp-devel",
     "turnstile",
+    "webrtc-audio-processing-devel",
     "zita-convolver-devel",
 ]
-# most plugins are from here and it can crash without them (and at least prints
-# 9 million warnings), so just always pull it
-depends = ["dinit-dbus", "lsp-plugins-lv2"]
+depends = [
+    "dinit-dbus",
```
q66: i think we don't really need this anymore, as cbuild traces the service dependencies anyway...

q66:
i think the user service needs some tweaking, because:

```
0:04:32.870 => easyeffects-dinit-8.2.7-r0: WARNING: service 'easyeffects' has a dependency field with '='
0:04:32.870 => easyeffects-dinit-8.2.7-r0: WARNING: service 'easyeffects' has a dependency field with '='
```

the `--help` also says:

```
  --gapplication-service                  Deprecated. Use --service-mode
                                          instead.
```

so it needs some more changes...

z-erica:
mb! also needed to add the `--hide-window` flag alongside `--service-mode` to preserve the previous behavior

q66:
indeed, also i've been testing this and it's such a ui downgrade :(

but that's what upstream has chosen so let's get this in

PR #5856, opened 2026-07-23
main/openssh: update to 10.4_p1

flukeeey:
## Description

I think we're good?

 * sshd(8): on Linux systems with the seccomp sandbox enabled,
   failures to enable SECCOMP or NO_NEW_PRIVS are now fatal.
   Previously sshd(8) would log the error but continue operation,
   to support systems that lacked these features. Now systems that
   lack these should instead disable the sandbox at configure time.

https://www.openssh.org/releasenotes.html

Tested ssh, ssh-agent, and sshd either way. Looks fine.

## Checklist

### DIFF
--- a/main/openssh/template.py
+++ b/main/openssh/template.py
@@ -1,6 +1,6 @@
 pkgname = "openssh"
-pkgver = "10.3_p1"
-pkgrel = 1
+pkgver = "10.4_p1"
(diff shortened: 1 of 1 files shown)


q66:
i'll handle this, i have some things to do

PR #5918, opened 2026-07-24
main/hwloc: update + split libs

z-erica:
this allows packages like onetbb to link to libhwloc.so.15 without needlessly pulling in in cairo or libx11

## Checklist


z-erica:
alternatively, maybe -progs is better?

q66:
yeah i think i prefer splitting progs, since it's primarily a library...

q66:
you can use default_progs + put usr/share as extra in it, i think that should do

q66:
i was testing this and it turned out i had done the work already so i just pushed it, it was a little bit more than i thought

q66:
oops i forgor to push the update, sorry

PR #5922, opened 2026-07-25
user/joshuto: new package

gigirassy:
## Description

Joshuto is an aesthetically pleasing Ranger-inspired terminal file manager.

Note that the version of libgit2 Chimera uses doesn't seem to be compatible, so it fails to build with it.

The library bundled in the application's build, as a result, needs to be used instead.

The binary also uses itself to generate completions.

## Checklist

### DIFF
--- a/user/joshuto/template.py
+++ b/user/joshuto/template.py
@@ -1,0 +1,33 @@
+pkgname = "joshuto"
+pkgver = "0.9.9"
+pkgrel = 0
+build_style = "cargo"
+# needs bundled libgit version
+make_env = {"LIBGIT2_NO_VENDOR": "0"}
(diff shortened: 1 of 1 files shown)


user/joshuto/template.py
```diff
@@ -0,0 +1,33 @@
+pkgname = "joshuto"
+pkgver = "0.9.9"
+pkgrel = 0
+build_style = "cargo"
+# needs bundled libgit version
+make_env = {"LIBGIT2_NO_VENDOR": "0"}
```
q66, line 6: i don't want this

q66:
it's technically possible to upgrade libgit2 versions but https://gts.chimera-linux.org/@chimera/statuses/01KWARHE1BXBMXB8WPXK0BBM1B

also AI commits

PR #5923, opened 2026-07-25
user/nxbelld: new package

gigirassy:
## Description

This is a tiny, simple daemon for X that will do an action instead of playing the speaker if someone doesn't like their computer's default speaker beep, or if the computer does not have a PC speaker.

## Checklist


q66:
this is just entirely dead

use https://docs.pipewire.org/page_module_x11_bell.html

PR #5957, opened 2026-07-29
user/nushell: update to 0.114.1

alecStewart1:
## Description

Nushell has updated a few times.

A couple of things that might be interesting:

~~1. `http-parser` is needed now. I could probably find where it's used if needed.~~ <- **EDIT** don't know how I thought this was the case
1. `nu-plugin` has a hard requirement of the `local-socket` feature, it will not build if it's not enabled.
2. Other features (LSP server, system clipboard) I explicitly listed as features built in this package, more so to make it clear to anyone that might wonder/ask in the future.

For anyone using Nushell, there's some deprecations you should probably be aware of:

https://www.nushell.sh/blog/2026-05-23-nushell_v0_113_0.html#deprecations
https://www.nushell.sh/blog/2026-05-23-nushell_v0_113_0.html#other-changes
https://www.nushell.sh/blog/2026-07-04-nushell_v0_114_0.html#deprecations
https://www.nushell.sh/blog/2026-07-04-nushell_v0_114_0.html#other-changes

## Checklist

### DIFF
--- a/user/nushell/template.py
+++ b/user/nushell/template.py
@@ -1,10 +1,10 @@
 pkgname = "nushell"
-pkgver = "0.112.2"
+pkgver = "0.114.1"
 pkgrel = 0
 build_style = "cargo"
 make_build_args = [
     "--no-default-features",
-    "--features=plugin,trash-support,sqlite,native-tls,network",
+    "--features=plugin,trash-support,lsp,local-socket,system-clipboard,sqlite,native-tls,network",
     "--workspace",
 ]
 make_check_args = [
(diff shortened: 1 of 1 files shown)


alecStewart1:
```
0:00:23.777 => nushell-0.114.1-r0: ERROR: target dependency 'http-parser' does not exist
```

Wh-...what did I build with locally then? It should be `http-parser-devel` anyway. I'll fix that.

q66:
https://github.com/chimera-linux/cports/commit/58d15343589b2aecb23088fb65263fab8c04300b there is no http-parser

q66:
i also don't see anywhere any indication that it needs that

alecStewart1:
Huh...well, I'll fix that then.

q66:
what exactly is the point of making a separate commit for a fix for an issue that the same PR introduced

you should read CONTRIBUTING.md again

alecStewart1:
My bad. It's one commit now.

PR #5965, opened 2026-08-01
user/android-tools: update to 37.0.0

Biswa96:
## Description

* Remove go and perl which are not required to build this.
* Remove ppc64le patch which is fixed in upstream.
* Remove test source file from cmake. It fails to build due to sys/cdefs.h include.

## Checklist


Biswa96:
p.s. It was my mistake that I left the test file in cmake. musl-bsd-headers can be removed in next release.

PR #5966, opened 2026-08-01
user/apparmor: new package

natthias:
## Description

Adds apparmor and libapparmor. Apparmor and libapparmor are in seperate templates, as i found it to be pretty jank to deal with different build styles in the same template.
There are still some considerations to be made, such as the dinit service, which when restarted, briefly disables all profiles. However, in my testing it works just fine, although this is only on x86_64 as that is the only hardware I have available.

I know reviewer bandwidth is low currently, but I saw no good reason to just sit on this template in private, so I'll just pick this up when there is more review capacity.

## Checklist

### DIFF
--- a/user/apparmor-extra-profiles
+++ b/user/apparmor-extra-profiles
@@ -1,0 +1,1 @@
+apparmor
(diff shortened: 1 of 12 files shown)


natthias:
It might actually to be better to build them all together after all, as USE_SYSTEM is used to toggle both building against system libapparmor, but also using host installed versions of the utils provided by apparmor.

This makes testing jank, as we would have to carry a patch that splits the USE_SYSTEM into two variables, one for libs and one for programs, which could be completely avoided by building together.

PR #5971, opened 2026-08-03
Packaging: add guidance for when to install licenses

cyberia-ng:
## Description

Add a section to `Packaging.md` describing when a license should be installed with a package.

## Checklist

### DIFF
--- a/Packaging.md
+++ b/Packaging.md
@@ -16,6 +16,7 @@ you should not rely on them or expect them to be stable.
   * [Writing Correct Templates](#correct_templates)
     * [Handling /etc](#handling_etc)
     * [Hardening Templates](#template_hardening)
+    * [Licensing](#licensing)
 * [Build Phases](#phases)
 * [Package Naming](#naming)
   * [Bootstrap Packages](#bootstrap_packages)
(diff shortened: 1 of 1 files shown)


q66:
this is not correct

there are various licenses that do not require installation (including the gpl) and can get away purely with spdx markers, the base-files docs package ships some of them, particularly the copyleft ones, but whether a license must be installed is not necessarily linked to that

cyberia-ng:
On IRC, you said
```
<cyberia> under what circumstances should one install the license in post_install in a cbuild template?
<q66> when the license wants it
<q66> there's no other answer to that
[...]
<q66> the license text for select licenses that don't explicitly want you to include a specific copyright clause (eg apache, gpl* except agpl) are installed with base-files already
```

From this, I understood that when a license requires distribution with the binaries it covers, it should be included in a cbuild template **except** when its inclusion in base-files-docs is sufficient. This is the process I described in the documentation update.

If this is not the case, can you clarify your answer to my question?

cyberia-ng:
In general, I think the documentation is lacking in guidance as to how to know when to install the license in the template and when not to. My intention is to make this decision easier for packagers.

q66:
you can't know that without reading the license and there is no spdx metadata declaring this

i don't think much can be done other than "if it's a common license, don't install it unless cbuild errors (it does that for common known licenses), if it's a niche license, then install it"

cyberia-ng:
> you can't know that without reading the license and there is no spdx metadata declaring this

OK? Then packagers can read the license? Nowhere do I say that you should be able to make the decision based on SPDX metadata alone.

I invite you to read the documentation that I wrote again. I think perhaps you have misread it as "install the license unless it is in this list". That is not what it says -- it says "If a software license requires that the license is distributed with the software, ...", which you correctly point out requires reading the license (or at least knowing what it says).

q66:
i read it and the whole "the exception is when ..." thing is entirely off the mark and you misinterpreted my comment on irc in a way that wasn't intended

cyberia-ng:
In that case, as I have asked, please can you clarify what you meant on IRC?

q66:
i don't think i'm at all deserving of you getting pissy at me and i don't appreciate that at all, i haven't done anything to you

i mentioned the base-files thing as a "just fyi" thing but it's by no means exhaustive and mostly only covers a few mostly copyleft licenses that are of interest

cyberia-ng:
I'm sorry, "obtuse" does not mean what I thought it meant.

I am trying to help here -- and reduce your workload in having to correct templates which make the wrong decision about installing the license. I still do not understand the correct process for making this decision, if it is (as you say) not what I wrote in the documentation.

Given that I can't contribute a good documentation update if I don't know the correct procedure, I will close this PR.

q66:
the guideline is what I wrote above, for commonplace licenses you can just defer to cbuild because it'll error for most common licenses needing installation

anything else needs reading

q66:
regardless i don't think this needs documenting now, ideally cbuild should catch enough that you shouldn't think about it very much, where it doesn't it should be extended

PR #5974, opened 2026-08-05
main/tailscale: update to 1.102.2

z-erica:
also point update.py to the actual releases

## Checklist

### DIFF
--- a/main/tailscale/update.py
+++ b/main/tailscale/update.py
@@ -1,1 +1,3 @@
-ignore = ["1.96.0", "1.96.1"]
+# tags are unreliable as they may randomly be RCs
+url = "https://tailscale.com/changelog"
+pattern = r">Tailscale\ v([\d.]+)<"
--- a/main/tailscale/template.py
+++ b/main/tailscale/template.py
@@ -1,5 +1,5 @@
 pkgname = "tailscale"
-pkgver = "1.102.0"
+pkgver = "1.102.2"
 pkgrel = 0
 build_style = "go"
 make_build_args = [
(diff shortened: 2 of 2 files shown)


main/tailscale/update.py
```diff
@@ -1 +1,2 @@
-ignore = ["1.96.0", "1.96.1"]
+url = "https://github.com/tailscale/tailscale/releases"
+pattern = r"/tailscale/tailscale/tree/v([0-9.]+)\""
```
q66: i don't understand what is the purpose of this part

z-erica: it's to match the releases listed in https://tailscale.com/changelog. any other tagged version seem not to even have a changelog attached, so im uncertain theyre intended to be packaged at all

looking a little more into it, the changelog entry for v1.92.1 says, for example, that
> Note: 1.92.0 was a [release candidate](https://tailscale.com/kb/1168/versions#release-candidate-track) intended for testing only.

q66: i see, so some tags are not meant to be used

sounds to me like maybe using the changelog page itself would be a better fit?

```
# tags are unreliable as they may randomly be RCs
url = "https://tailscale.com/changelog"
pattern = r">Tailscale\ v([\d.]+)<"
```

z-erica: fair enough! lets do that

q66: after changing that from this listing i see that 1.102.0 wasn't one either, and this goes for a lot of the .0 releases, but seemingly not all of them...

main/tailscale/update.py
```diff
@@ -1 +1,3 @@
-ignore = ["1.96.0", "1.96.1"]
+# tags are unreliable as they may randomly be RCs
+url = "https://tailscale.com/changelog"
+pattern = r"Tailscale\ v([\d.]+)<"
```
q66: add a > before the Tailscale else it matches some version numbers with . like:

```
Checking found version: 1.84.0
Checking found version: 1.84.0.
```

(due to pattern match within the text)

PR #5996, opened 2026-08-07
main/wf-shell: update to 0.11.0, move gtk4-layer-shell to main, wayfire and wf-config updated to 0.11.0 and temporary patch deployed for wf-shell to compile against clang

gurkulsahoo08:
## Description

Update wf-shell to 0.11.0.

This required moving gtk4-layer-shell from user to main and
updating it to 1.3.0, to satisfy wf-shell's build dependencies.

I have updated wayfire and wf-config to 0.11.0 also on recommendation of maintainers.

I have deployed a patch for wf-shell as it would not compile against clang.

## Checklist

### DIFF
--- a/main/wf-shell/template.py
+++ b/main/wf-shell/template.py
@@ -48,7 +48,6 @@ sha256 = [
     "1999dcab22ccf159b56e670476c0f6eab309ac2f69d16ad3c117824f17201313",
     "0163f8e7250d46a18905b04e966f3a4c849a3afb810ca1e862cb685f8a92bc2e",
 ]
-
 options = ["etcfiles"]
 
 
--- a/main/wf-shell/template.py
+++ b/main/wf-shell/template.py
@@ -27,7 +27,7 @@ makedepends = [
 pkgdesc = "Desktop shell for Wayfire"
 license = "MIT"
 url = "https://wayfire.org"
-_wfjson = "a85b53d8a45565b876465bdf7482776eb01cc54a"
+_wfjson = "70039e13cdeaebd8ec498ed30bf5ab91c2e313ec"
 _wlogout = "1cffc6fd0be5d8127f394b25eb3c82e71fbaabe1"
 _gvc = "5f9768a2eac29c1ed56f1fbb449a77a3523683b6"
 source = [
@@ -44,7 +44,7 @@ source_paths = [
 ]
 sha256 = [
     "328b1a01c5fc63ebc6fa79152bafdcd2e1525c8bb3caf5b49361ff2792dd57f4",
-    "91cec007677cbf83e84aae0abd0cc4436dbef142fd9bede01b6df289b4df3e06",
+    "749d9f649f82570d860c0742c11ff83ff1b108ca7a8ec02c5c3e5f439084fafa",
     "1999dcab22ccf159b56e670476c0f6eab309ac2f69d16ad3c117824f17201313",
     "0163f8e7250d46a18905b04e966f3a4c849a3afb810ca1e862cb685f8a92bc2e",
 ]
@@ -54,8 +54,4 @@ options = ["etcfiles"]
 
 def post_install(self):
     self.install_license("LICENSE")
-    (self.destdir / "usr/lib/pam.d").mkdir(parents=True, exist_ok=True)
-    (self.destdir / "etc/pam.d/wf-locker").rename(
-        self.destdir / "usr/lib/pam.d/wf-locker"
-    )
-    (self.destdir / "etc/pam.d").rmdir()
+    self.rename("etc/pam.d", "usr/lib/pam.d", relative=False)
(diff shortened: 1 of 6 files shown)


main/wf-shell/template.py
```diff
@@ -11,19 +11,51 @@
...
+    "yyjson-devel",
 ]
 pkgdesc = "Desktop shell for Wayfire"
 license = "MIT"
 url = "https://wayfire.org"
-source = f"https://github.com/WayfireWM/wf-shell/releases/download/v{pkgver}/wf-shell-{pkgver}.tar.xz"
-sha256 = "49a7fc861849051a3be5de353e3d7442a37170c990a3ffd8d83b67a369edca93"
+_wfjson = "a85b53d8a45565b876465bdf7482776eb01cc54a"
+_wlogout = "1cffc6fd0be5d8127f394b25eb3c82e71fbaabe1"
+_gvc = "5f9768a2eac29c1ed56f1fbb449a77a3523683b6"
```
q66: these should be the same as for the wf-shell tag and at least wfjson isn't

also making separate variables for them is pointless

main/wf-shell/template.py
```diff
@@ -1,5 +1,5 @@
 pkgname = "wf-shell"
-pkgver = "0.10.0"
```
q66, line 2: this shouldn't be updated without also updating wayfire

main/wf-shell/template.py
```diff
@@ -11,19 +11,51 @@
...
+options = ["etcfiles"]
 
 
 def post_install(self):
     self.install_license("LICENSE")
+    (self.destdir / "usr/lib/pam.d").mkdir(parents=True, exist_ok=True)
+    (self.destdir / "etc/pam.d/wf-locker").rename(
+        self.destdir / "usr/lib/pam.d/wf-locker"
+    )
+    (self.destdir / "etc/pam.d").rmdir()
```
q66: `self.rename("etc/pam.d", "usr/lib/pam.d", relative=False)`

instead of whatever this is

main/wf-shell/template.py
```diff
@@ -11,19 +11,51 @@
...
+    "subprojects/wayland-logout",
+    "subprojects/gvc",
+]
+sha256 = [
+    "328b1a01c5fc63ebc6fa79152bafdcd2e1525c8bb3caf5b49361ff2792dd57f4",
+    "91cec007677cbf83e84aae0abd0cc4436dbef142fd9bede01b6df289b4df3e06",
+    "1999dcab22ccf159b56e670476c0f6eab309ac2f69d16ad3c117824f17201313",
+    "0163f8e7250d46a18905b04e966f3a4c849a3afb810ca1e862cb685f8a92bc2e",
+]
+
```
q66: spurious newline

main/wf-shell/template.py
gurkulsahoo08: wfjson commit is different, I am going to fix this, but I don't see the troubles with other variables and how they are listed as I followed the layout one other package main/libei/template.py already uses, diff sub modules with their separate hash, also the new ver 0.11.0 of wf shell fails to build without these as the author of wf shell did a explicit large overhaul migration to gtk 4.0 and added like 10 new features in a single release

main/wf-shell/template.py
gurkulsahoo08: I will also update wayfire and resubmit despite during testing wf-shell successfully binded along with wayfire 0.10.0

gurkulsahoo08: I have found a problem with version sync of wf-shell 0.11.0 and updated wf-config 0.11.0 and wayfire 0.11.0 which I updated locally using ./cbuild. Turns out wf-shell 0.11.0 has a real incompatibility that only surfaces when built against wf-config 0.11.0 together.

While testing the full stack (wf-config, wayfire, wf-shell all at 0.11.0), I hit a hard compile failure in src/locker/timedrevealer.cpp

`error: statement requires expression of integer type ('WfOption<int>' invalid)
    switch (hide_animation)`

I digged deeper to find the root cause, first trying with clang strict santizer flags off, which also failed, but I found that `base_option_wrapper_t's` implicit conversion operator in wayfire's `option-wrapper.hpp` uses a `decltype(auto)` return type, which Clang doesn't accept as a valid contextual conversion for `switch` (GCC apparently does, or so I presume because arch maintainers successfully build it and it is in their repo and I checked their pkgbuild and it contained no patches, pure default build). 

Apparently the fix is a one-liner and I've verified it locally
`switch (hide_animation.value())`
instead of relying on the implicit conversion. I am going to file a issue in wf-shell repo asking if they want a PR for this one line fix or maintainer will fix it. Till then I would like to ask for the cports — carry a local patch until it's merged upstream, if maintainers accept it.

gurkulsahoo08: I have sent upstream PR fix to wf-shell maintainers, just a one line change in `src/locker/timedrevealer.cpp` from` switch (hide_animation)` to  switch `(hide_animation.value())`, hoping for a quick merge

gurkulsahoo08: I have a applied a local patch and now everything builds perfectly

PR #6001, opened 2026-08-07
user/moby: update to 29.7.2

breakgimme:
## Checklist

### DIFF
--- a/user/moby/template.py
+++ b/user/moby/template.py
@@ -1,8 +1,8 @@
 pkgname = "moby"
-pkgver = "29.6.2"
+pkgver = "29.7.2"
 pkgrel = 0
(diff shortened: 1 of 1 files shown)


breakgimme:
update check failing seems to be related to cd718a027f3900446db9b0e35459ca76557c81ce, as reverting it makes the update check work again for me

q66:
oh yeah, things that rely on matching patterns from xml to override the default regex pattern to match a subset of tags will break

i fixed the pattern

q66:
i don't think there is a lot of things that do this because usually the default pattern works fine though

breakgimme:
thanks, will rebase and force push

PR #6009, opened 2026-08-09
user/libratbag: only restart on failure

z-erica:
this matches the behavior or the upstream systemd unit (`Restart=on-abort`). otherwise it keeps pointlessly restarting when it exits on idle

## Checklist

### DIFF
--- a/user/libratbag/files/ratbagd
+++ b/user/libratbag/files/ratbagd
@@ -1,4 +1,5 @@
 type = process
 command = /usr/bin/ratbagd
+restart = on-failure
 depends-on: dbus
 depends-on: local.target
--- a/user/libratbag/template.py
+++ b/user/libratbag/template.py
@@ -1,6 +1,6 @@
 pkgname = "libratbag"
 pkgver = "0.18"
-pkgrel = 0
+pkgrel = 1
 build_style = "meson"
 configure_args = [
     "-Dsystemd=false",


q66:
i don't think we should be adding on-failure in places because that should kinda be the default behavior of dinit and we should switch to it

i have been considering it but i found that the official way to do it via configure has undesirable side effects so https://github.com/davmac314/dinit/issues/564 needs addressing first

z-erica:
fair enough! for the record, the reason for this pr specifically is that my logs were being spammed like
```
2026-08-09 08:51:36.491  koshka  notice                      dinit: service ratbagd stopped.
2026-08-09 08:51:36.502  koshka  notice                      dinit: service ratbagd started.
2026-08-09 09:12:36.493  koshka  notice                      dinit: service ratbagd stopped.
2026-08-09 09:12:36.493  koshka  notice                      dinit: service ratbagd started.
2026-08-09 09:33:36.493  koshka  notice                      dinit: service ratbagd stopped.
2026-08-09 09:33:36.493  koshka  notice                      dinit: service ratbagd started.
2026-08-09 09:54:36.493  koshka  notice                      dinit: service ratbagd stopped.
2026-08-09 09:54:36.493  koshka  notice                      dinit: service ratbagd started.
```
while the dinit-buffered output was
```
Exiting after idle
(dinit: note: service restarted)
```
and the service itself is purely dbus activated

q66:
yeah seems like it just kinda goes after a while

the current default restart behavior of dinit doesn't make sense to me

if anything we should be specifying `restart = yes` in stuff that explicitly wants this but everything else should implicitly restart on fail only...

q66:
i'll just merge this for now as an immediate fix for the issue so you don't have to wait, i'll be clearing on-failure from other service files when this is addressed in dinit anyway

PR #6010, opened 2026-08-09
add noctalia and sdbus-cpp

plavpixel:
## Description

`user/noctalia`: Noctalia desktop shell. Its purpose is to provide a easily configurable, unified Linux desktop experience without using separate bars, notification centers, settings menu, etc. Required 3 simple patches to remove GNU/libstdc++-isms and be more portable.
`user/sdbus-cpp`: Dependency for Noctalia. It is a modern dbus library which is also being used by other wayland desktop components.

## Checklist

### DIFF
--- a/user/noctalia/patches/make-sed-portable.patch
+++ b/user/noctalia/patches/make-sed-portable.patch
@@ -1,0 +1,13 @@
+diff --git a/meson.build b/meson.build
+index 21939b5..8b2415e 100644
+--- a/meson.build
++++ b/meson.build
+@@ -397,7 +397,7 @@ _wlr_ls_h = custom_target('wlr-layer-shell-unstable-v1-client-header',
+   output:  'wlr-layer-shell-unstable-v1-client-protocol.h',
+   command: [
+     'sh', '-c',
+-    '"$1" client-header "$2" "$3" && sed -i s/namespace/name_space/g "$3"',
++    '"$1" client-header "$2" "$3" && sed s/namespace/name_space/g "$3" > "$3.tmp" && mv "$3.tmp" "$3"',
+     '_', wayland_scanner.full_path(), '@INPUT@', '@OUTPUT@',
+   ],
+ )
--- a/user/noctalia/template.py
+++ b/user/noctalia/template.py
@@ -1,0 +1,48 @@
+pkgname = "noctalia"
+pkgver = "5.0.0_beta8"
+pkgrel = 0
+build_style = "meson"
+hostmakedepends = ["meson", "pkgconf"]
+makedepends = [
+    "cairo-devel",
+    "curl-devel",
+    "elogind-devel",
+    "fontconfig-devel",
+    "freetype-devel",
+    "harfbuzz-devel",
+    "jemalloc-devel",
+    "libepoxy-devel",
+    "libical-devel",
+    "libjxl-devel",
+    "libqalculate-devel",
+    "librsvg-devel",
+    "libsecret-devel",
+    "libsndfile-devel",
+    "libsodium-devel",
+    "libwebp-devel",
+    "libxkbcommon-devel",
+    "libxml2-devel",
+    "linux-pam-devel",
+    "md4c-devel",
+    "mesa-devel",
+    "nlohmann-json",
+    "pango-devel",
+    "pipewire-devel",
+    "polkit-devel",
+    "sdbus-cpp-devel",
+    "stb",
+    "tomlplusplus-devel",
+    "wayland-devel",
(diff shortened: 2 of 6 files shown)


user/noctalia/patches/make-sed-portable.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/meson.build b/meson.build
+index 21939b5..8b2415e 100644
+--- a/meson.build
++++ b/meson.build
+@@ -397,7 +397,7 @@ _wlr_ls_h = custom_target('wlr-layer-shell-unstable-v1-client-header',
+   output:  'wlr-layer-shell-unstable-v1-client-protocol.h',
+   command: [
+     'sh', '-c',
+-    '"$1" client-header "$2" "$3" && sed -i s/namespace/name_space/g "$3"',
++    '"$1" client-header "$2" "$3" && sed s/namespace/name_space/g "$3" > "$3.tmp" && mv "$3.tmp" "$3"',
```
jcgruenhage, line 10:
```suggestion
-    '"$1" client-header "$2" "$3" && sed -i s/namespace/name_space/g "$3"',
+    '"$1" client-header "$2" "$3" && sed -i '' s/namespace/name_space/g "$3"',
```

jcgruenhage:
Started using this yesterday, quite nice :)

user/sdbus-cpp/template.py
```diff
@@ -0,0 +1,17 @@
+pkgname = "sdbus-cpp"
+pkgver = "2.3.1"
+pkgrel = 1
+build_style = "cmake"
+configure_args = ["-DSDBUSCPP_BUILD_CODEGEN=ON", "-DSDBUSCPP_BUILD_TESTS=OFF"]
```
q66: we build tests

make sure they work

user/noctalia/template.py
```diff
@@ -0,0 +1,48 @@
...
+    "pipewire-devel",
+    "polkit-devel",
+    "sdbus-cpp-devel",
+    "stb",
+    "tomlplusplus-devel",
+    "wayland-devel",
+    "wayland-protocols",
+    "wireplumber-devel",
+]
+pkgdesc = "Sleek, customizable desktop shell crafted for Wayland"
```
q66: pkgdescs shouldn't contain subjective adjectives

"Desktop shell for Wayland"

user/noctalia/template.py
```diff
@@ -0,0 +1,48 @@
+pkgname = "noctalia"
+pkgver = "5.0.0_beta8"
```
q66, line 2: this is a beta version? most people use the stable 4.x versions and i heard complaints that 5.x is nowhere near as capable

user/noctalia/template.py
```diff
@@ -0,0 +1,48 @@
...
+pkgrel = 1
```
q66: pkgrel starts with 0

user/noctalia/template.py
ItsLemmy: Your information seems a bit off.

v5 is the main/leading version, v4 is deprecated and unmaintained (not packaged anymore).

v4 will probably totally break for most people on the next Qt update since a majority of our user base use  rolling distros.

There is only one feature from v4 that was not reimplemented (on purpose) in v5, it's the "framed" bar, that wrap the edge of the screens, because it was a really bad (technical) idea to start with.

Even if labelled beta, v5 is more stable than v4 ever was and covers 55% of the user base, the trend is shifting fast.   We plan to get out of beta in september.

user/noctalia/template.py
plavpixel: Same for sdbus-cpp ?

q66: Packaging.md:

pkgrel (int) The release number for the package. When changes are made to the template that require rebuilding of the package, this is is incremented by one. The initial value should be zero. When bumping to a new version, it should be reset back to zero.

plavpixel: Sorry about that. still used to void linux templates

user/noctalia/template.py
jcgruenhage, line 2: 5.0.1 has been released as the first stable release of the v5 series.

jcgruenhage (changes requested):

user/noctalia/template.py
plavpixel: Some tests have not been building. have been working with project devs to resolve it

bunkbail: v5.1.0 builds cleanly with a rebased stb and sed patch.

PR #6013, opened 2026-08-12
user/lact: new package

alliumm:
## Description

Useful for GPU undervolting and power limiting. (and overclocking i guess but who does that)
The libc crate update patch is for musl `renameat2` support.

## Checklist


alliumm:
I think this should technically work on other architectures but it's not supported and I don't have the means to test. I'll just set archs to x86_64.

PR #6064, opened 2026-08-24
user/testdisk: new package

xerydian:
## Description

This PR provides a suite of data recovery tools:

- **TestDisk** can be used to recover lost partitions, fix boot sectors, and recover files from damaged or deleted file systems. It is mainly used to recover data from hard drives, but it can also work with other storage devices such as USB drives and memory cards. It supports a wide variety of file systems, including NTFS, exFAT, FAT, ext2, ext3, and ext4.

- **PhotoRec** is designed to recover lost files, including photos, videos, and music files, from various types of storage devices. It can recover files from hard drives, memory cards, and USB drives, and it can also recover files from digital cameras and other portable devices.

- **Fidentify** checks all the files from a directory with the same signatures than PhotoRec. It’s useful to check if PhotoRec is able to recover these file formats

## Checklist

### DIFF
--- a/user/testdisk/patches/basename.patch
+++ b/user/testdisk/patches/basename.patch
@@ -1,0 +1,12 @@
+--- a/src/hdaccess.c
++++ b/src/hdaccess.c
+@@ -48,6 +48,9 @@
+ #ifdef HAVE_STRING_H
+ #include <string.h>
+ #endif
++#ifdef HAVE_LIBGEN_H
++#include <libgen.h>
++#endif
+ #ifdef HAVE_FCNTL_H
+ #include <fcntl.h> 	/* open */
+ #endif
--- a/user/testdisk/template.py
+++ b/user/testdisk/template.py
@@ -1,0 +1,17 @@
+pkgname = "testdisk"
+pkgver = "7.2"
+pkgrel = 0
+build_style = "gnu_configure"
+hostmakedepends = ["automake", "pkgconf"]
+makedepends = [
+    "e2fsprogs-devel",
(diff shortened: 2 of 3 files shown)


user/testdisk/template.py
```diff
@@ -0,0 +1,17 @@
...
+    "e2fsprogs-devel",
+    "libjpeg-turbo-devel",
+    "ncurses-devel",
+    "ntfs-3g-devel",
+    "zlib-ng-compat-devel",
+]
+pkgdesc = "Data recovery suite"
+license = "GPL-2.0-or-later"
+url = "https://www.cgsecurity.org"
+source = f"{url}/{pkgname}-{pkgver}.tar.bz2"
```
q66: we don't expand pkgname

user/testdisk/patches/basename.patch
```diff
@@ -0,0 +1,12 @@
+--- a/src/hdaccess.c
```
q66: this patch is incorrect

libgen.h provides the POSIX version of basename which may modify its argument and the expected version here is the GNU version (not provided by musl)

better to do something like `main/attr/patches/basename.patch` does

user/testdisk/update.py
```diff
@@ -0,0 +1,2 @@
+url = "https://github.com/cgsecurity/testdisk/tags"
```
q66:
```                                                                     
url = "https://github.com/cgsecurity/testdisk/info/refs?service=git-upload-pack"
pattern = r"refs/tags/v([\d.]+)"
```

no pkgname

PR #6071, opened 2026-08-25
main/*: remove redundant sysusers.conf files

wezm:
## Description

With these commits I believe the need for build-time user/group creation was removed for the respective packages:

- dbus: dc42f96722ec784ee1dd9185152d91b45b2f9218
- qemu: bf4b80acdf0d5c8175fef2846a76c232d3b71852
- wireshark: 58f6b8ebd723d20b91a5e7e073adb796b45789c2

Incidentally, I'm not sure assigning non-root users and groups in `file_modes` works any more. `make` is invoked via [chroot.enter with fakeroot=True](https://github.com/chimera-linux/cports/blob/c104aea3013e4a902db303f5ff9d7dec390c6f07/src/cbuild/core/build.py#L866) and that in turn sets [`FAKEROOTDONTTRYCHOWN=1`](https://github.com/chimera-linux/cports/blob/c104aea3013e4a902db303f5ff9d7dec390c6f07/src/cbuild/core/chroot.py#L852). So, from what I can tell [the `chown` lines put into the Makefile](https://github.com/chimera-linux/cports/blob/c104aea3013e4a902db303f5ff9d7dec390c6f07/src/cbuild/apk/generate.py#L362-L366) do nothing.

This isn't an issue in practice since there are no cases left where the user and group is non-`root`. However, the docs do still mention the functionality.

## Checklist

### DIFF
--- a/main/dbus/sysusers.conf
+++ b/main/dbus/sysusers.conf
@@ -1,1 +1,0 @@
-u dbus - "dbus user" /tmp /usr/bin/nologin
--- a/main/qemu/sysusers.conf
+++ b/main/qemu/sysusers.conf
@@ -1,1 +1,0 @@
-u _qemu - "qemu user" /tmp /usr/bin/nologin
--- a/main/wireshark/sysusers.conf
+++ b/main/wireshark/sysusers.conf
@@ -1,1 +1,0 @@
-g _wireshark -


q66:
the functionality works fine, fakeroot wraps the user/group apis and keeps its own ephemeral mappings that exist for the duration of the run

the env var prevents it from propagating back to the filesystem, but that is the point

by chowning it during the same fakeroot run as apk mkpkg, apk can retrieve the ownership to store in the adb, since there is no other way to pass it (this is also why apk needs to be dynamic during mkpkg, because you can't preload-override symbols in a static bin, but it just uses the bldroot's so it's fine)

wezm:
Thanks for the explanation. I got as far as [chown in fakeroot](https://salsa.debian.org/clint/fakeroot/-/blob/main/libfakeroot.c#L899), but overlooked the 3 lines above where the internal state is manipulated 🤦‍♂️.

PR #6112, opened 2026-08-31
user/lidm: New package

mic0wav:
## Description

new package: [LiDM](https://github.com/javalsai/lidm)

LiDM is a really light unix login manager made in C, and highly customizable _sadly in /etc, figuring out /usr/share/examples/ might take some time_. Whats more important for why i want this in the repos is that the only login managers are sddm and gdm, with no option for a lighter TUI login manager, and compared to other TUI login managers (ly, lemurs, etc.) LiDM has a short dependency list, is written in a pretty standard language, and from my testing just works.

I understand if this gets turned down from the repos

## Checklist

- [x] I acknowledge https://gts.chimera-linux.org/@chimera/statuses/01KWARHE1BXBMXB8WPXK0BBM1B

### DIFF
--- a/user/lidm/files/lidm
+++ b/user/lidm/files/lidm
@@ -1,0 +1,8 @@
+type            = process
+command         = /sbin/agetty tty7 linux-c -n -l /usr/local/bin/lidm -o 7
+restart         = true
+depends-on      = login.target
+termsignal      = HUP
+smooth-recovery = true
+inittab-id      = 7
+inittab-line    = tty7
--- a/user/lidm/template.py
+++ b/user/lidm/template.py
@@ -1,0 +1,21 @@
+pkgname = "lidm"
+pkgver = "2.0.2"
+pkgrel = 0
+build_style = "makefile"
+makedepends = [
+    "dinit-chimera",
+    "linux-headers",
+    "linux-pam-devel",
+]
+pkgdesc = "Customizable TUI display manager"
+license = "GPL-3.0-only"
+url = "https://github.com/javalsai/lidm"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "08bb013df15987b5a685c5925a41b2c9a8531e4af3c9a097c0dbb1fa8a6d8a79"
+# no test suite, files in /etc
+options = ["!check", "etcfiles"]
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+    self.install_service(self.files_path / "lidm")


natthias:
> Whats more important for why i want this in the repos is that the only login managers are sddm and gdm, with no option for a lighter TUI login manager

What about tuigreet?

user/lidm/templates.py
```diff
@@ -0,0 +1,30 @@
+pkgname = "lidm"
+pkgver = "2.0.2"
+pkgrel = 0
+build_style = "makefile"
+make_build_args = ["PREFIX=/usr"]
+make_install_args = ["PREFIX=/usr"]
+make_use_env = True
```
natthias: why is this needed? seems like it isn't

user/lidm/templates.py
```diff
@@ -0,0 +1,30 @@
...
+hostmakedepends = ["linux-pam-devel", "linux-headers"]
+makedepends = ["linux-pam-devel", "linux-headers"]
+
+options = ["etcfiles", "!check"]
```
natthias: Should go after sha256, and you should provide a reason

user/lidm/templates.py
```diff
@@ -0,0 +1,30 @@
...
+pkgdesc = "Customizable TUI display manager"
+license = "GPL-3.0-only"
+url = "https://github.com/javalsai/lidm"
+source = f"{url}/archive/refs/tags/v{pkgver}.tar.gz"
+sha256 = "08bb013df15987b5a685c5925a41b2c9a8531e4af3c9a097c0dbb1fa8a6d8a79"
+
+
+def post_install(self):
+    self.install_license("LICENSE")
+    self.do(
```
natthias: Use self.install_service. and don't seperate self.do invocations with a semicolon, thats a gross hack and is difficult to read

q66: it just doesn't work period, this just creates a bunch of weirdly named directories and doesn't install any service file (which, if it did, it would get installed in a wrong path and the lint would fail, and also the upstream service file doesn't conform to requirements)

mic0wav: okay sorry im gonna get to work fixing these, sorry this is my first time packaging

user/lidm/templates.py
mic0wav: the `linux-pam-devel` dependency is needed for the pam library, used for user authentication. `linux-headers` is also needed for a library. the option `etcfiles` is there because lidm has a config file that is in /etc/lidm.ini, `!check` i added because of help in the matrix chat:

> doesnt seem like the Makefile has a check target defined? you can probably skip the check phase with `options = ["!check"]`

sorry i am bad at communicating is this what you asked ?

mic0wav:
> > Whats more important for why i want this in the repos is that the only login managers are sddm and gdm, with no option for a lighter TUI login manager
> 
> What about tuigreet?

oh #@!# sorry i did not know that was in the repos, i only searched for displaymanager and loginmanager and similar terms (`*dm*` and `*lm*`)

mic0wav:
im having some problems with git but here is the update:

```
pkgname = "lidm"
pkgver = "2.0.2"
pkgrel = 0
build_style = "makefile"
makedepends = [
    "dinit-chimera",
    "linux-headers",
    "linux-pam-devel",
]
pkgdesc = "Customizable TUI display manager"
license = "GPL-3.0-only"
url = "https://github.com/javalsai/lidm"
sha256 = "08bb013df15987b5a685c5925a41b2c9a8531e4af3c9a097c0dbb1fa8a6d8a79"
# no test suite, files in /etc
options = ["!check", "etcfiles"]


def post_install(self):
    self.install_service(self.files_path / "lidm")
```

and a new `files/lidm` for the dinit service:

```
type            = process
command         = /sbin/agetty tty7 linux-c -n -l /usr/local/bin/lidm -o 7
restart         = true
depends-on      = login.target
termsignal      = HUP
smooth-recovery = true
inittab-id      = 7
inittab-line    = tty7
```

PR #6120, opened 2026-09-01
user/*: add Hyprland and dependencies

milchinskiy:
## Description

Add Hyprland (v0.56.2) and its required dependencies.
the PR is large mainly because Hyprland has a fairly long chain of separately packaged dependencies, each commit introduces one self-contained package.

Most individual packages use conventional Meson/CMake packaging without unusual integration work, so the number of commits mostly reflects the dependency strcuture rather than complexity within individual packages. All of them are fairly conventional, so I expect future updates to be  reasonably straightforward.
I suppose I have to mention that Alpine and FreeBSD ports were used partially as references for dependency/build-definition cross-checking, obviously chimera specific decisions were still validated against Chimera's own packaging rules and local builds. Currently scope is deliberately limited to Hyprland and the libraries/tools needed for it; additional Hypr* ecosystem applications are planned as separate follow-up packages.

The whole stack was built locally and tested in a fresh VM. The generated local APK repository was added to the qemu VM and Hyprland was installed using "doas apk add hyprland", allowing apk to reslve the dependency chain as expected. Hyprland starts successfully in the VM and locally, including XWayland support.

## Checklist

### DIFF
--- a/user/aquamarine-devel
+++ b/user/aquamarine-devel
@@ -1,0 +1,1 @@
+aquamarine
--- a/user/aquamarine/template.py
+++ b/user/aquamarine/template.py
@@ -1,0 +1,45 @@
+pkgname = "aquamarine"
+pkgver = "0.14.0"
+pkgrel = 0
+build_style = "cmake"
+configure_args = [
+    "-DOPENGL_opengl_LIBRARY=/usr/lib/libGL.so",
+]
+make_check_args = [
+    "-E",
+    "simpleWindow",
+]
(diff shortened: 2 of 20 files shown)


q66:
given we've been pretty vocal about not wanting hyprland what makes you think this will get accepted this time

absolutely not

milchinskiy:
I wasn't aware of the previous hyprland PRs, otherwise I obviously would not have spent the effort preparing this. I see, thanks for clarifying

PR #6125, opened 2026-09-02
user/emacs*: update to 31.1

colin-heffernan:
## Description

This bumps the various Emacs builds to 31.1. One notable change is the deprecation of its native `ctags` in favor of `universal-ctags`.

## Checklist


q66:
just disable the test that calls into gcc

user/emacs-pgtk/template.py
```diff
@@ -18,9 +18,11 @@
     " %tramp-tests.el"  # TODO: fails mysteriously
     " %shr-tests.el"  # TODO: zoom-image times out
     " %process-tests.el"  # TODO: times out
+    " %package-vc-tests.el"  # TODO: hangs
 ]
 hostmakedepends = [
     "automake",
+    "ctags",
```
alecStewart1: Is this meant to be in the `hostmakedepends` for the other versions?

alecStewart1 (changes requested):

user/emacs-pgtk/template.py
colin-heffernan: Presumably yes, but the others built fine without it, so it may not be necessary. I know I got an error on pgtk without it, but I don't see why any one version would depend on it differently, given it's not related to UI. Will test shortly and update the PR accordingly.

colin-heffernan: Whatever caused me to add `ctags` to `hostmakedepends` isn't reproducible anymore, so I removed it.

PR #6132, opened 2026-09-04
user/kyanite: new package

cyrneko:
## Description

This packages Kyanite, a text-editor made with GTK4.

> [!CAUTION]
> this PR is **not yet ready to merge**, upstream still requires a version to be tagged. I am submitting it here so that once a version is tagged, the template can be updated and the PR be marked as ready. The template works, I run it daily, but I do not submit it as ready for that reason.

## Checklist

### DIFF
--- a/user/kyanite/template.py
+++ b/user/kyanite/template.py
@@ -1,0 +1,51 @@
+pkgname = "kyanite"
+pkgver = "0.0.1"
+pkgrel = 0
+build_style = "cargo"
+configure_args = [
+    "-Dbugreport_url=https://github.com/chimera-linux/cports/issues",
+]
+hostmakedepends = ["blueprint-compiler"]
+makedepends = [
+    "cargo",
+    "gtk4-devel",
+    "gtksourceview-devel",
+    "libadwaita-devel",
+    "libpanel-devel",
+    "pkgconf",
+    "vte-gtk4-devel",
+]
+pkgdesc = "Kyanite text editor"
+license = "GPL-3.0-or-later"
+url = "https://codeberg.org/pastthepixels/kyanite"
+source = "https://codeberg.org/pastthepixels/kyanite/archive/8e27198895ca5b7574fdf5176267c87c1cc1849e.tar.gz"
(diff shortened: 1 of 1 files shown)


q66:
the moratorium on new packages that you explicitly acknowledge in the checklist is still in place yet everyone is consistently ignoring it

at very least you could not submit unfinished work...

cyrneko:
sorry! I read "if it's not a really simple template as is" and figured this would be okay, because there's not much special involved here

though the template itself is finished, I just don't think packages without a version should be merged judging from cbuild's warning about the versioning scheme

q66:
and how exactly do you determine this is "finished"? this template has several deviations from common packaging conventions and things that don't make sense (what's the configure_args supposed to mean? cargo never uses that? what's with the hardening thing? no other template does that and most of it doesn't have any effect? do you know about post_install? etc)

the packager doesn't determine if something is finished unless they have the experience to be able to judge it and even then it comes down solely to a reviewer

cyrneko:
idk! I'm just trying to learn how this works, I tried following cbuild's output[^1], took a look at a few other GTK4 apps and saw what they were doing[^2], and I do know about `post_install` but figured it wasn't what I was looking for when comparing it to other templates[^3]

really I wanted to take this as an exercise so that I could perhaps improve and maintain a few more packages in the future, the only other change I've made in cports was just a version bump but that didn't really require me to understand the build system at all

[^1]: which gave me a notice about hardening options, which to me sounded like "these are things you could still add"; it wasn't 100% clear if cbuild did anything with the options becuase there was neither acknowledgement nor an error about them
[^2]: which is where the configure_args comes from, I figured it'd be a more universal thing for GNOME applications
[^3]: I looked at helix off the top of my head, knowing it's also written in rust, to try and understand where which helper functions are best used in practice, namely `cargo.install()`, and is why I didn't use `post_install` because I took said package as reference for how to package a rust application, which also uses `install`

q66:
the reasons aren't important/meaningful for this, particularly because i'm not asking you to justify anything

the point is that when i ask people to not throw stuff at me because i want to clear things up and this is suffocating, assuming something is "complete" is the wrong way to go when that turns out not to be the case

PR #6136, opened 2026-09-05
main/networkmanager: update to 1.58.1

flukeeey:
## Description

- upstream: smallish fixes
- added conf.d, dnsmasq.d to tmpfiles as these are common customisation
  paths that are recommended for user drop-in configs
- switched to readline because libedit was echoing passwords in
  plaintext

Let me know if any of these tweaks should be reverted.

## Checklist


main/networkmanager/template.py
```diff
@@ -33,7 +33,7 @@
     "-Dintrospection=true",
     "-Ddocs=true",
     "-Dcrypto=nss",
-    "-Dreadline=libedit",
+    "-Dreadline=libreadline",
```
q66, line 36: add a comment why

PR #6151, opened 2026-09-08
user/just-lsp: new package

xerydian:
## Description

**just-lsp** implements the Language Server Protocol for [just](https://github.com/casey/just), the command runner. 
This provides editor features for justfiles (completions, hover information, definitions, ...)

## Checklist


user/just-lsp/patches/cargo-checksum.patch
```diff
@@ -0,0 +1,4 @@
+--- /dev/null
++++ b/vendor/tree-sitter-just/.cargo-checksum.json
+@@ -0,0 +1 @@
++{"files":{},"package":null}
```
q66: this seems wrong

the "vendor" directory already exists in the tree and is used for a different purpose

you can drop this patch and instead add something like

```
def prepare(self):
    # "vendor" dir already exists for a different purpose
    self.cargo.vendor(["vendored-crates"])
```

to resolve the conflict by using a different path

xerydian: Oh, that makes way more sense than what i had. Thank you

